PR #17, opened 2018-09-10
Update hyperopt_tuner.py

xuehui1991:


src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
```diff
@@ -51,7 +51,7 @@ class OptimizeMode(Enum):
 def json2space(in_x, name=ROOT):
     '''
     Change json to search space in hyperopt.
-	'''
+    '''
```
QuanluZhang: please check the whole file. Also, sometimes you use single quote and sometimes you use double quote.

xuehui1991: fixed

PR #25, opened 2018-09-10
Update Python dependencies

liuzhe-lz:


src/sdk/pynni/setup.py
```diff
@@ -32,14 +32,12 @@ def read(fname):
 
     python_requires = '>=3.5',
     install_requires = [
+        'hyperopt',
         'json_tricks',
         'numpy',
         'pymc3',
```
QuanluZhang: also remove pymc3 here?

PR #29, opened 2018-09-11
add basic auto-gbdt example

xuehui1991:


examples/trials/auto-gbdt/config.yml
```diff
@@ -0,0 +1,20 @@
+authorName: default
+experimentName: example_auto-gbdt
+trialConcurrency: 1
+maxExecDuration: 10h
+maxTrialNum: 1
+#choice: local, remote
+trainingServicePlatform: local
+searchSpacePath: ~/nni/examples/trials/auto-gbdt/search_space.json
```
QuanluZhang: for your convenience, nni has supported relative path, you can simply use ./search_space.json :)

examples/trials/auto-gbdt/config.yml
```diff
@@ -0,0 +1,20 @@
+authorName: default
+experimentName: example_auto-gbdt
+trialConcurrency: 1
+maxExecDuration: 10h
+maxTrialNum: 1
```
QuanluZhang: could we set maxTrialNum 5 or 10, because some users complain that it is a little confusing that the experiment stopped after one trial job is completed.

PR #31, opened 2018-09-11
add ~/.local/bin to PATH when installing

Crysple:
add ~/.local/bin to PATH when installing

### DIFF
--- a/src/nni_manager/common/utils.ts
+++ b/src/nni_manager/common/utils.ts
@@ -154,6 +154,10 @@ function parseArg(names: string[]): string {
 function getMsgDispatcherCommand(tuner: any, assessor: any): string {
     let command: string = `python3 -m nni --tuner_class_name ${tuner.className}`;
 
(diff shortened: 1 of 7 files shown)


src/nni_manager/common/utils.ts
```diff
@@ -154,6 +154,10 @@ function parseArg(names: string[]): string {
 function getMsgDispatcherCommand(tuner: any, assessor: any): string {
     let command: string = `python3 -m nni --tuner_class_name ${tuner.className}`;
 
+    if (process.env.VIRTUAL_ENV) {
+        command = path.join(process.env.VIRTUAL_ENV + 'bin/') +command;
+    }
+
     if (tuner.classArgs !== undefined) {
```
QuanluZhang: it is not good to modify code here

Crysple: I'm just still testing travis. Not ready

Crysple:
Abandoned

PR #58, opened 2018-09-13
Fix links in README.md 

narumiruna:


QuanluZhang:
Thanks a lot for pointing this out :). We have fix the break links in pr #59

PR #73, opened 2018-09-14
Bugfix and partial pylint fix for SQuAD QA model.

tansinan:
This pull request fix bugs prevent SQuAD QA model from running. It also fixes some of pylint warnings.

### DIFF
--- a/examples/trials/ga_squad/train_model.py
+++ b/examples/trials/ga_squad/train_model.py
@@ -37,7 +37,7 @@ class GAGConfig:
 
         self.dropout = 0.1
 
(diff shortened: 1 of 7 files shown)


examples/trials/ga_squad/train_model.py
```diff
@@ -32,12 +32,13 @@
 
 
 class GAGConfig:
+    """The class for model hyper-parameter configuration."""
     def __init__(self):
         self.batch_size = 128
```
xuehui1991: remove empty line

Issue #50, opened 2018-09-13
stderr path is not consistent with "trial ls" output information 


SparkSnail:

mydmdm:
same issue, there is a extra ".nni"  in my case

chicm-ms:
Fixed in master branch.

PR #74, opened 2018-09-14
Fix stderr path issue #50

chicm-ms:


QuanluZhang:
also commit this change to v0.1?

chicm-ms:
> also commit this change to v0.1?

Sure

PR #78, opened 2018-09-15
Code cleanup for SQuAD example.

tansinan:
Cleanup unused classes and functions for attention in SQuAD example.


tansinan:
A problem in default dataset config is also fixed.

Issue #85, opened 2018-09-17
the WebUI displayed with no content


hjctop1:
the starting process seems OK, but there is no content on the web.
![screenshot from 2018-09-17 12-46-55](https://user-images.githubusercontent.com/174086/45606483-0820a880-ba78-11e8-8f91-d0555b00972b.png)

ubuntu@ubuntu-Super-Server:~$ nnictl create --config nni/examples/trials/mnist-annotation/config.yml
Info: Checking experiment...
Info: Starting restful server...
Info: Checking restful server...
Info: Restful server start success!
Info: Setting local config...
Info: Success!
Info: Starting experiment...
Info: Checking web ui...
Info: Starting web ui...
Info: Starting web ui success!
Info: Web UI url: http://127.0.0.1:8080   http://192.168.123.101:8080   http://172.17.0.1:8080
Info: Start experiment success! The experiment id is ImMmO066, and the restful server post is 51188.
You can use these commands to get more information about this experiment:
         commands                       description
1. nnictl experiment show        show the information of experiments
2. nnictl trial ls               list all of trial jobs
3. nnictl stop                   stop a experiment
4. nnictl trial kill             kill a trial job by id
5. nnictl --help                 get help information about nnictl
6. nnictl webui url              get the url of web ui

[stderr.txt](https://github.com/Microsoft/nni/files/2387377/stderr.txt)

QuanluZhang:
It seems nni is not correctly installed. You can check nni python package by running `python3 -c "import nni"`.
If there is no nni, make sure which python3 you are using and nni is installed in this python.
BTW, how did you install nni?

hjctop1:
yes, there is no nni.

buntu@ubuntu-Super-Server:~/nni/experiments/jb9oZYlK/log$ which python3
/home/ubuntu/anaconda3/bin/python3

I use the install direction in the doc:
pip3 install -v --user git+https://github.com/Microsoft/nni.git@v0.1
source ~/.bashrc

I also tried:
Install NNI through source code
git clone -b v0.1 https://github.com/Microsoft/nni.git
cd nni
chmod +x install.sh
source install.sh

QuanluZhang:
Thanks @hjctop1, you are using anaconda. Actually we have not tested nni on anaconda, we are working on it though. We will inform you when anaconda is fully tested, it should be very soon.

hjctop1:
OK!
Thanks a lot.

scarlett2018:
[...]

Issue #80, opened 2018-09-15
 Run an experiment at local met problem！


wenku1992abc:
According to GetStarted.md ,as follws,

Install NNI through source code

git clone -b v0.1 https://github.com/Microsoft/nni.git
cd nni
chmod +x install.sh
source install.sh

 I had sucessfully installed NNI ,But when I  run an experiment at local ,do such commad: nnictl create --config ~/nni/examples/trials/mnist-annotation/config.yml ,It shows an error,as follws,

Info: Checking experiment...
Info: Starting restful server...
Info: Checking restful server...
Error: Restful server start failed!

Could you help me?

QuanluZhang:
1. Please make sure that port 51188 is not occupied by other processes. 
2. Then please run 'nnictl log stderr' to see the error message. 
3. You can also use 'ps aux | grep node' to check is there any running rest server process.

wenku1992abc:
1.Check port 51188 is not occupied:

2. Run 'nnictl log stderr' ,It shows,as follws:
(node:18566) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
/home/user/.local/bin/nnimanager: line 3: 18566 Terminated              /home/user/.local/nni/node/bin/node main.js $@
(node:31077) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
/home/user/.local/bin/nnimanager: line 3: 31077 Terminated              /home/user/.local/nni/node/bin/node main.js $@
(node:40883) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
/home/user/.local/bin/nnimanager: line 3: 40883 Terminated              /home/user/.local/nni/node/bin/node main.js $@
(node:870) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
/home/user/.local/bin/nnimanager: line 3:   870 Terminated              /home/user/.local/nni/node/bin/node main.js $@
[...]

PR #90, opened 2018-09-18
support conda

Crysple:
- specify which python to execute pip install
- inherit environment variables in nnimanager's spawn to fully support virtual environment


Crysple:
refer to issue: #80 #85

PR #98, opened 2018-09-19
Add trial log path

SparkSnail:


SparkSnail:
I've updated the document.

PR #112, opened 2018-09-21
Add built-in tuner test in CI

Crysple:
- Check whether each built-in tuner can be run normally in travis CI


test/test_builtin_tuner/naive_trial.py
```diff
@@ -0,0 +1,7 @@
+import nni
+
+params = nni.get_parameters()
+print('params:', params)
```
xuehui1991: shall we use print as test method?? Or use unit test to instead?

test/test_builtin_tuner/run.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python3
```
xuehui1991: I'm not sure shall we add license here....@quanlu zhang

test/test_builtin_tuner/search_space.json
```diff
@@ -0,0 +1,7 @@
+{
```
xuehui1991: shall we add all kind of search space to test their format?

Crysple:
This pr is not ready. Further modification needs to be done because of refracting exp stop #113

test/test_builtin_tuner/run.py
QuanluZhang: yes, could add license here

QuanluZhang:
please refactor this pr, since we have new experiment stopping logic.

Crysple:
Will refactor this pr

Issue #106, opened 2018-09-20
nnictl crashed if config path is not correct


goooxu:
Use command
`nnictl create --config xxx`
If the path is not correct, the nnictl tool crashed
Need a friendly tip

PR #121, opened 2018-09-25
fix config path error message

SparkSnail:


SparkSnail:
https://github.com/Microsoft/nni/issues/106

PR #125, opened 2018-09-26
Fix multiphase datastore problem

chicm-ms:


src/nni_manager/core/nniDataStore.ts
```diff
@@ -161,9 +165,8 @@ class NNIDataStore implements DataStore {
 
     private async getFinalMetricData(trialJobId: string): Promise<any> {
         const metrics: MetricDataRecord[] = await this.getMetricData(trialJobId, 'FINAL');
-        assert(metrics.length <= 1);
-        if (metrics.length === 1) {
-            return metrics[0];
+        if (metrics.length > 0) {
+            return metrics[metrics.length - 1];
```
yds05: Should we also consider to log error if metrics.length > 1 for single phase exp?

chicm-ms: Ok, added back the assertion for single phase exp.

src/nni_manager/core/nniDataStore.ts
```diff
@@ -161,11 +167,31 @@ class NNIDataStore implements DataStore {
...
             return metrics[0];
+        }
+    }
+
+    private async isMultiPhase(): Promise<boolean|undefined> {
+        if (this.multiPhase === undefined) {
+            this.multiPhase = (await this.getExperimentProfile(getExperimentId())).params.multiPhase;
+        }
+
+        if (this.multiPhase !== undefined) {
```
yds05: Optional: change to 
if (this.multiPhase) { 
}
for simplicity

src/nni_manager/core/nniDataStore.ts
```diff
@@ -161,11 +167,31 @@ class NNIDataStore implements DataStore {
...
+        const multiPhase: boolean | undefined = await this.isMultiPhase();
+
+        if (multiPhase) {
+            if (metrics.length > 0) {
+                return metrics[metrics.length - 1];
+            } else {
+                return undefined;
+            }
+        } else {
+            assert(metrics.length <= 1);
```
yds05: Log error here, rather than assert

src/nni_manager/core/nniDataStore.ts
chicm-ms: Fixed.

src/nni_manager/core/nniDataStore.ts
chicm-ms: Fixed.

PR #128, opened 2018-09-26
PAI Training Service implementation

yds05:
**1. Implement PAITrainingService
**2. Add trial-keeper python module, and modify setup.py to install the module
**3. Add PAItrainingService rest server to collect metrics from PAI container.


Makefile
```diff
@@ -92,14 +92,9 @@ build:
...
+	cd src/nni_manager/training_service_tool && python3 setup.py build
 
-# Standard installation target
-# Must be invoked after building
-.PHONY: install
-install: install-python-modules
-install: install-node-modules
-install: install-scripts
-install: install-examples
 install:
```
QuanluZhang: why removing these installs?

Makefile
```diff
@@ -109,6 +104,7 @@ install:
 .PHONY: remote-machine-install
 remote-machine-install:
 	cd src/sdk/pynni && python3 setup.py install $(PIP_MODE)
+	cd src/nni_manager/training_service_tool && python3 setup.py install $(PIP_MODE)
```
QuanluZhang: i suggest move training_service_tool into training_service, for example, src/nni_manager/training_service/tools

src/nni_manager/common/utils.ts
```diff
@@ -225,5 +225,19 @@ function cleanupUnitTest(): void {
     Container.restore(ExperimentStartupInfo);
 }
 
-export { getMsgDispatcherCommand, getLogDir, getExperimentRootDir, getDefaultDatabaseDir, mkDirP, delay, prepareUnitTest,
-    parseArg, cleanupUnitTest, uniqueString, randomSelect };
+/**
+ * Get IPv4 address of current machine
+ */
+function getIPV4Address(): string {
```
QuanluZhang: this function is not used in this pr, right?

src/nni_manager/rest_server/restValidationSchemas.ts
```diff
@@ -33,9 +33,19 @@ export namespace ValidationSchemas {
                 passphrase: joi.string()
             })),
             trial_config: joi.object({
-                gpuNum: joi.number().min(0).required(),
+                image: joi.string().min(1),
```
QuanluZhang: remove the white space :)

src/nni_manager/rest_server/restValidationSchemas.ts
```diff
@@ -33,9 +33,19 @@ export namespace ValidationSchemas {
                 passphrase: joi.string()
             })),
             trial_config: joi.object({
```
QuanluZhang: it might be better to improve this validation, this mixes the validation of trial's local/remote mode and pai mode.

tools/nnicmd/config_schema.py
```diff
@@ -47,7 +47,12 @@
 'trial':{
```
QuanluZhang: might be better to improve the validation

QuanluZhang:
would be great to pass the ci test

Makefile
yds05: We shouldn't remove these installs, it may be introduced by merge error and I will recover these lines. Thanks for pointing it out!

PR #130, opened 2018-09-27
Update setup.py Makefile and documents

Crysple:
- install requirements.txt under each tuner
- small changes in Makefile on Colorful output
- change pip install to python3 -m pip install to be more robust

### DIFF
--- a/docs/GetStarted.md
+++ b/docs/GetStarted.md
@@ -24,14 +24,6 @@
       chmod +x install.sh
       source install.sh
 
-
-* __Install NNI for all users__
-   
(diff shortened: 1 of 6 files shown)


docs/GetStarted.md
```diff
@@ -27,7 +27,7 @@
 
 * __Install NNI for all users__
    
-      sudo pip3 install -v --user git+https://github.com/Microsoft/nni.git@v0.1
+      sudo python3 -m pip install -v --user git+https://github.com/Microsoft/nni.git@v0.1
```
QuanluZhang: since you use `sudo`, why still use `--user`?

PR #136, opened 2018-09-28
Pai training service bug fix and enhancement

yds05:
1. Create nnioutput dir in container to store output files
2. Create a shell script to detect if nni package is already installed in PAI container. If not, pip install NN from github
3. Move trial.log from NNI_SYS_DIR to NNI_OUTPUT_DIR


src/nni_manager/training_service/pai/paiData.ts
```diff
@@ -47,12 +47,21 @@ export class PAITrialJobDetail implements TrialJobDetail {
     }
 }
 
+export const RAI_INSTALL_NNI_SHELL_FORMAT: string =
```
QuanluZhang: RAI -> PAI?

PR #138, opened 2018-09-28
Improve annotation

liuzhe-lz:
Modify generated search space key from "path/#123/type" to "path/__line123/type".
Use original user code directory (instead of generated code directory) as `codeDir` when search space is generated but NNI annotation is not used.


tools/nnicmd/launcher.py
```diff
@@ -189,13 +188,14 @@ def launch_experiment(args, experiment_config, mode, webuiport, experiment_id=No
     nni_config.set_config('restServerPid', rest_process.pid)
     # Deal with annotation
     if experiment_config.get('useAnnotation'):
-        path = os.path.join(tempfile.gettempdir(), 'nni', 'annotation', ''.join(random.sample(string.ascii_letters + string.digits, 8)))
+        path = os.path.join(tempfile.gettempdir(), 'nni', 'annotation')
+        path = tempfile.mkdtemp(dir=path)
         if os.path.isdir(path):
             shutil.rmtree(path)
         os.makedirs(path)
```
QuanluZhang: are these lines still necessary?

liuzhe-lz: I'm not sure if `mkdtemp` guarantees to create unique name on each call. The API doc is not very clear. I'll check the source code later.
By the way, the following code creates a default search space `json.dumps('')` (aka `'""'`) if the user does not provide search space info in any way. Who will use this? I think `None` or JSON `null` should be better than an JSON empty string.

liuzhe-lz: Confirmed. `mkdtemp` will deal with this.

QuanluZhang: It is possible that users do not provide search space file for their customized tuner (for example search space is hard-coded in their tuner). Agree, `None` or JSON `'null'` might be better.

liuzhe-lz: I see. I won't do the commit anyway. Leave it to @SparkSnail or someone else.

PR #139, opened 2018-09-29
Selectively install through pip

Crysple:
- install nni's sdk and trial_tool by add flag install-options='--sdk-only' in pip install
- something like:
pip install ---install-option='--sdk-only' -v git+https://github.com/Microsoft/nni.git@v0.2

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,24 +30,24 @@ def read(fname):
 class CustomInstallCommand(install):
     '''a customized install class in pip module'''
     user_options = install.user_options + [
-        ('target=', None, '<add it if you only want to install nni sdk only')
+        ('platform=', None, '<add it if you only want to install nni sdk only')
     ]
 
     def initialize_options(self):
+        self.platform = None
         install.initialize_options(self)
-        self.target = None
 
     def finalize_options(self):
-        print("in final", self.target)
+        print("in final", self.platform)
         install.finalize_options(self)
 
     def run(self):
-        #super().run()
-        if self.target == 'remote':
-            print("in run's if: %s"%self.target)
+        if self.platform == 'remote':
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -29,9 +29,35 @@ def read(fname):
...
+    user_options = install.user_options + [
+        ('sdk-only', None, '<add --sdk-only if you only want to install nni sdk only')
+    ]
+
+    def initialize_options(self):
+        install.initialize_options(self)
+        self.sdk_only = None
+
+    def install_requires(self):
+        self.install_requires_list = [
```
Crysple: - Move install_requires to here because pip install with install-option flag will apply to all install_requires if it is in the setup.
- It will failed because other packages' setup.py have no such flag

referring to https://github.com/pypa/pip/issues/1883

setup.py
```diff
@@ -29,9 +29,35 @@ def read(fname):
 
 class CustomInstallCommand(install):
     '''a customized install class in pip module'''
+    user_options = install.user_options + [
+        ('sdk-only', None, '<add --sdk-only if you only want to install nni sdk only')
```
yds05: if you want to only install nni sdk

setup.py
```diff
@@ -29,9 +29,35 @@ def read(fname):
...
+            'pyhdfs'
+        ]
+        for pkg in self.install_requires_list:
+            subprocess.run(['python3', '-m', 'pip', 'install', pkg], check=True)
+
     def run(self):
+        if self.sdk_only is None:
+            subprocess.run(['make', 'pip-install'], check=True)
+        self.install_requires()
         super().run()
```
yds05: call super().run() at first

PR #148, opened 2018-09-30
Multi-phase training service

chicm-ms:


.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
   - sudo sh -c 'PATH=/usr/local/node/bin:$PATH yarn global add serve'
 install:
   - make
-  - make install
+  - make dev-install
```
yds05: Why should we modify this line?

src/nni_manager/core/nniDataStore.ts
```diff
@@ -112,13 +114,18 @@ class NNIDataStore implements DataStore {
...
     public async getTrialJob(trialJobId: string): Promise<TrialJobInfo> {
-        const trialJobs = await this.queryTrialJobs(undefined, trialJobId);
+        const trialJobs: TrialJobInfo[] = await this.queryTrialJobs(undefined, trialJobId);
 
         return trialJobs[0];
     }
 
     public async storeMetricData(trialJobId: string, data: string): Promise<void> {
-        const metrics = JSON.parse(data) as MetricData;
+        this.log.debug(`storeMetricData: trialJobId: ${trialJobId}, data: ${data}`);
```
yds05: Unnecessary debug info, remove?

src/nni_manager/core/nniDataStore.ts
```diff
@@ -112,13 +114,18 @@ class NNIDataStore implements DataStore {
...
+        const metrics: MetricData = JSON.parse(data);
+        if (metrics.type === 'REQUEST_PARAMETER') {
```
yds05: Suggest to add comment to explain why we choose to skip to store 'REQUEST_PARAMETER' metric here

src/nni_manager/core/nniDataStore.ts
```diff
@@ -160,25 +167,56 @@ class NNIDataStore implements DataStore {
...
-        if (metrics.length > 1) {
-            this.log.error(`Found multiple final results for trial job: ${trialJobId}`);
+
+        const multiPhase: boolean | undefined = await this.isMultiPhase();
+
+        if (metrics.length > 1 && !multiPhase) {
+            this.log.error(`Found multiple FINAL results for trial job ${trialJobId}`);
+        }
+
+        return metrics[metrics.length - 1];
```
yds05: what happen if metrics.length == 0?

src/nni_manager/core/nniDataStore.ts
```diff
@@ -160,25 +167,56 @@ class NNIDataStore implements DataStore {
...
+    }
+
+    private async isMultiPhase(): Promise<boolean|undefined> {
```
yds05: Should change return type to Promise<boolean>, since we definitely return a boolean value according to line 185-189?

src/nni_manager/core/nniDataStore.ts
```diff
@@ -160,25 +167,56 @@ class NNIDataStore implements DataStore {
...
     }
 
+    private mergeHyperParameters(hyperParamList: string[], newParamStr: string): string[] {
+        const mergedHyperParams: any[] = [];
+        const newParam: any = JSON.parse(newParamStr);
+        for (const hyperParamStr of hyperParamList) {
+            const hyperParam: any = JSON.parse(hyperParamStr);
+            mergedHyperParams.push(hyperParam);
+        }
+        if (mergedHyperParams.filter((value: any) => { return value.parameter_index === newParam.parameter_index; }).length <= 0) {
```
yds05: how about simplifying 
mergedHyperParams.filter((value: any) => { return value.parameter_index === newParam.parameter_index; })

to 
mergedHyperParams.filter((value: any) => value.parameter_index === newParam.parameter_index) ?

src/nni_manager/core/nnimanager.ts
```diff
@@ -474,6 +477,22 @@ class NNIManager implements Manager {
...
+                const tunerCommand: any = JSON.parse(content);
+                assert(tunerCommand.parameter_index >= 0);
+                assert(tunerCommand.trial_job_id !== undefined);
+
+                const trialJobForm: TrialJobApplicationForm = {
+                    jobType: 'TRIAL',
+                    hyperParameters: {
+                        value: content,
+                        index: tunerCommand.parameter_index
+                    }
```
yds05: Suggest to decouple Command meta data(like trialJobId, parameter_index..etc) from command content, but since it's not a small refactor, you could do it after merging this PR.

src/sdk/pynni/nni/platform/local.py
```diff
@@ -35,10 +36,28 @@
 _log_file_path = os.path.join(_outputdir, 'trial.log')
 init_logger(_log_file_path)
 
+_param_index = 0
+
+def _send_request_parameter_metric():
```
yds05: Suggest to rename to request_next_parameter()

.travis.yml
chicm-ms: Thanks, the change is removed.

src/nni_manager/core/nniDataStore.ts
chicm-ms: Thanks, it is removed per comments.

src/nni_manager/core/nniDataStore.ts
chicm-ms: Thanks, comments added.

src/nni_manager/core/nniDataStore.ts
chicm-ms: It returns undefined as expected if metrics.length is 0

src/nni_manager/core/nniDataStore.ts
chicm-ms: Thanks, updated per comments.

src/nni_manager/core/nniDataStore.ts
chicm-ms: Thanks, updated per comments.

src/nni_manager/core/nnimanager.ts
chicm-ms: Thanks, will do it later.

src/sdk/pynni/nni/platform/local.py
chicm-ms: Thanks, renamed it per comments.

.travis.yml
chicm-ms: make install and make easy-install do not work here,  change back to make dev-install

PR #151, opened 2018-09-30
Update Makefile

Crysple:
- fix some bugs in Makefile about easy-install
- default install changes to install but not dev-install (the latter one just do some soft link and not really install nni into the install_path. For example, nni sdk will not be copied to python's site-package...)
- delete a redundant file and fix some markdown typo in readme


tools/README.md
```diff
@@ -1,36 +1,37 @@
-##NNI CTL
```
QuanluZhang: this file needs a big refactor

Makefile
```diff
@@ -203,21 +203,21 @@ install-dependencies: $(NODE_TARBALL) $(YARN_TARBALL) $(SERVE_TARBALL)
 .PHONY: install-python-modules
 install-python-modules:
 	#$(_INFO) Installing Python SDK $(_END)
-	cd src/sdk/pynni && python3 setup.py install $(PIP_MODE)
+	cd src/sdk/pynni && $(PIP_INSTALL) $(PIP_MODE) .
```
liuzhe-lz: `README.Makefile.md` should be updated because PIP is not optional anymore.

Crysple: At first, I was confused that why 'setup.py install' is used in install target and 'pip install' is used in dev-install target. And it seems nni cannot be uninstalled through pip if it's installed by setup.py, so I change it

liuzhe-lz: Because setuptools (setup.py) does not support editable (symlink) mode as far as I know.
Using pip looks fine to me anyway. Just sync the doc.

QuanluZhang: Thanks @liuzhe-lz . @Crysple  could you update the README file accordingly?

Crysple: > Thanks @liuzhe-lz . @Crysple could you update the README file accordingly?

Sure I will

Makefile
```diff
@@ -203,21 +203,22 @@ install-dependencies: $(NODE_TARBALL) $(YARN_TARBALL) $(SERVE_TARBALL)
...
 	
 	#$(_INFO) Installing nnictl $(_END)
-	cd tools && python3 setup.py install $(PIP_MODE)
+	cd tools && $(PIP_INSTALL) $(PIP_MODE) .
 
 .PHONY: install-node-modules
 install-node-modules:
 	mkdir -p $(INSTALL_PREFIX)/nni
+	rm -rf ${PWD}/src/nni_manager/dist/node_modules
```
QuanluZhang: this line is not correct, only delete the symbolic link.

PR #157, opened 2018-10-02
Support more data structure in tuner

xuehui1991:


src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -111,11 +111,20 @@ def handle_add_customized_trial(self, data):
...
         if data['type'] == 'FINAL':
+            value = None
             id_ = data['parameter_id']
+            
+            if isinstance(data['value'], float) or isinstance(data['value'], int):
+                value = data['value']
+            elif isinstance(data['value'], dict) and 'default' in data['value']:
+                value = data['value']['default']
+            else:
+                raise RuntimeError('The final result from Trial is dict and has no default key in Tuner.')
```
QuanluZhang: `RuntimeError('Incorrect final result: the final result should be float/int, or a dict which has key named "default"')`

src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -111,11 +111,20 @@ def handle_add_customized_trial(self, data):
 
     def handle_report_metric_data(self, data):
         if data['type'] == 'FINAL':
+            value = None
             id_ = data['parameter_id']
+            
+            if isinstance(data['value'], float) or isinstance(data['value'], int):
+                value = data['value']
+            elif isinstance(data['value'], dict) and 'default' in data['value']:
+                value = data['value']['default']
```
QuanluZhang: do we also need to check whether this value is float/int?

PR #172, opened 2018-10-08
Add enas from contributor

xuehui1991:


examples/trials/enas/README.md
```diff
@@ -0,0 +1,6 @@
+**Run ENAS in NNI**
+===
+
+Now we have a enas version run in nni from our contributors. Thanks our lovely contributors.
```
yds05: an enas

yds05: (Optional) 
How about:
Now we have an enas trial for NNI implemented by community contributors

PR #190, opened 2018-10-10
Support more search space and maxDuration format

lvybriage:
support more search space;
change maxDuration and still running style.

### DIFF
--- a/src/webui/src/components/Sessionpro.tsx
+++ b/src/webui/src/components/Sessionpro.tsx
@@ -510,7 +510,7 @@ class Sessionpro extends React.Component<{}, SessionState> {
                         className="tableButton"
                         onClick={this.downExperimentContent}
                     >
-                        Down Experiment
+                        Download Experiment
                     </Button>
                 </div>
             </div>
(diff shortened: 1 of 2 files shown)


src/webui/src/components/Sessionpro.tsx
```diff
@@ -88,6 +88,20 @@ class Sessionpro extends React.Component<{}, SessionState> {
         };
     }
 
+    convertTime = (num: number) => {
+        let hour = Math.floor(num / 3600 % 24);
```
chicm-ms: why mod 24 here?  it seems there is no day unit here.

src/webui/src/components/Sessionpro.tsx
```diff
@@ -88,6 +88,20 @@ class Sessionpro extends React.Component<{}, SessionState> {
...
+        if (num % 3600 === 0) {
+            return result = hour + 'h';
+        } 
+        let min = Math.floor(num / 60 % 60);
+        if (hour < 1) {
+            return result = min + '';
+        } else {
+            return result = hour + 'h  ' + min + 'min';
+        }
+    }
```
chicm-ms: This function can be simplified as following:
const hour =  Math.floor(num / 3600);
const min = Math.floor(num / 60 % 60);
return hour > 0 ? `${hour} h ${min} min` : `${min} min`

lvybriage: Thanks very much.

src/webui/src/components/Sessionpro.tsx
lvybriage: ok

PR #203, opened 2018-10-11
 Add api nni.get_sequence_id()

chicm-ms:


QuanluZhang:
it looks good overall. Since the functions of training services are becoming more and more, maybe we should make the interface between training service  and trial's sdk more clear.

src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -415,6 +418,16 @@ class LocalTrainingService implements TrainingService {
         const filepath: string = path.join(directory, `parameter_${hyperParameters.index}.cfg`);
         await fs.promises.writeFile(filepath, hyperParameters.value, { encoding: 'utf8' });
     }
+
+    private generateSequenceId(): number {
+        return this.trialSequenceId++;
+    }
+
+    private async writeSequenceIdFile(trialJobId: string): Promise<void> {
+        const trialJobDetail: LocalTrialJobDetail = <LocalTrialJobDetail>this.jobMap.get(trialJobId);
```
yds05: assert trialJobDetail !== undefined ?

src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -415,6 +418,16 @@ class LocalTrainingService implements TrainingService {
         const filepath: string = path.join(directory, `parameter_${hyperParameters.index}.cfg`);
         await fs.promises.writeFile(filepath, hyperParameters.value, { encoding: 'utf8' });
     }
+
+    private generateSequenceId(): number {
+        return this.trialSequenceId++;
+    }
+
+    private async writeSequenceIdFile(trialJobId: string): Promise<void> {
```
yds05: I suggest to put sequenceId into LocalTrialJobDetail, and make it as readonly, rather than generating it when calling writeSequenceIdFile function

src/nni_manager/training_service/local/localTrainingService.ts
chicm-ms: Thanks, assertion added.

src/nni_manager/training_service/local/localTrainingService.ts
chicm-ms: Thanks,  updated per your comments, sequenceId added into common TrialJobDetail.

PR #206, opened 2018-10-11
Fix OpenPAI training service failed issue after multiphase training code merged

yds05:
1. Fix OpenPAI training service failed issue after multiphase training code merged.

   Still use parameter.cfg as file name for the first parameter file, use parameter_1.cfg, parameter_2.cfg ... for successive parameter files whose index > 1, to keep backward compatibility with old version SDK (<=v0.2)
2. Assessor doc error correction.


examples/assessors/README.md
```diff
@@ -2,10 +2,10 @@
 
 *Assessor receive intermediate result from Trial and decide whether the Trial should be killed. Once the Trial experiment meets the early stop conditions, the assessor will kill the Trial.*
 
-So, if user want to implement a customized Assessor, she/he only need to:
+So, if want to implement a customized Assessor, you only need to:
```
QuanluZhang: how about "if users want to ...., they only need to:"?

yds05: sounds good. Will update

PR #216, opened 2018-10-15
Add idompotent support for get_parameters() in nni sdk

yds05:


src/nni_manager/core/nnimanager.ts
```diff
@@ -116,6 +116,10 @@ class NNIManager implements Manager {
         await this.storeExperimentProfile();
         this.log.debug('Setup tuner...');
 
+        if(expParams.multiPhase && this.trainingService.isMultiPhaseJobSupported) {
+            this.trainingService.setClusterMetadata('multiPhase', expParams.multiPhase.toString());
```
chicm-ms: Should we add this for resume experiment?

yds05: Yes, updated

PR #221, opened 2018-10-16
Dev exp stop more

QuanluZhang:
1. experiment stopping refactor 
2. ci refactor
3. support updating max trial number


azure-pipelines.yml
```diff
@@ -9,10 +9,10 @@ steps:
...
     displayName: 'Install python tools'
   - script: |
-      make easy-install
-      export PATH=$HOME/.nni/bin:$PATH
+      source install.sh
     displayName: 'Install dependencies'
   - script: |
       cd test/naive
-      PATH=$HOME/.local/nni/node/bin:$PATH python3 run.py
+      export PATH=$HOME/.local/bin:$PATH
```
yds05: Why we replace $HOME/.local/nni/node/bin with $HOME/.local/bin here?

install.sh
```diff
@@ -1,3 +1,7 @@
 #!/bin/bash
-make easy-install
+make install-dependencies
+make build
+make dev-install
+make install-examples
+make update-bash-config
```
yds05: why replace simple easy-install with expanded steps here?

src/nni_manager/core/nnimanager.ts
```diff
@@ -344,24 +338,138 @@ class NNIManager implements Manager {
...
     }
 
+    private async requestTrialJobsStatus(): Promise<number> {
+        const deferred: Deferred<number> = new Deferred<number>();
+        let finishedTrialJobNum: number = 0;
+        for (const trialJobId of Array.from(this.trialJobs.keys())) {
+            const trialJobDetail: TrialJobDetail = await this.trainingService.getTrialJob(trialJobId);
+            const oldTrialJobDetail: TrialJobDetail | undefined = this.trialJobs.get(trialJobId);
+            assert(oldTrialJobDetail);
+            if (oldTrialJobDetail !== undefined && oldTrialJobDetail.status !== trialJobDetail.status) {
```
yds05: assert(oldTrialJobDetail);  and oldTrialJobDetail !== undefined is duplicated. Need to leave just one.

src/nni_manager/core/nnimanager.ts
```diff
@@ -344,24 +338,138 @@ class NNIManager implements Manager {
...
+            }
+            const finishedTrialJobNum: number = await this.requestTrialJobsStatus();
+
+            allFinishedTrialJobNum += finishedTrialJobNum;
+            if (allFinishedTrialJobNum >= this.experimentProfile.params.maxTrialNum) {
+                // write this log for travis CI
+                this.log.info('Experiment done.');
+            }
+
+            const requestTrialNum: number = this.trialConcurrencyChange + finishedTrialJobNum;
```
yds05: the logic from line 403 to line 408 is not very straightforward, I suggest to add comment to explain the logic behind the code.

src/nni_manager/core/nnimanager.ts
```diff
@@ -344,24 +338,138 @@ class NNIManager implements Manager {
...
+                default:
+                // TO DO: add warning in log
+            }
+        }
+        deferred.resolve(finishedTrialJobNum);
+
+        return deferred.promise;
+    }
+
+    private async manageTrials(): Promise<void> {
```
yds05: manageTrials() is a big function. How about separate the logic of "dispatch new trials" and "check maxtrialnum && maxduration " into 2 private functions, and call them in manageTrials()

azure-pipelines.yml
Crysple: - Makefile will explicitly specify which node(and where) to use in $(BIN_PATH)/nnimanager:
- `echo '$(NODE) main.js $$@' >> $(BIN_PATH)/nnimanager`
- so it's unnecessary to add node/bin in PATH

src/nni_manager/core/nnimanager.ts
QuanluZhang: agree, but if I remove `oldTrialJobDetail !== undefined`, oldTrialJobDetail.status might be undefined, does not pass compiling.

src/nni_manager/core/nnimanager.ts
QuanluZhang: thanks, added comments.

src/nni_manager/core/nnimanager.ts
QuanluZhang: make sense, could I leave this code refactor in the next pr?:)

src/nni_manager/core/nnimanager.ts
QuanluZhang: removed assert.

Issue #166, opened 2018-10-07
“Hyper Parameter” is not show correctly in WebUI.


xuehui1991:

xuehui1991:
I will show the picture.

PR #225, opened 2018-10-16
 Fix bug: some trial jobs hyper params not stored  and error handling updates

chicm-ms:
https://github.com/Microsoft/nni/issues/166


src/nni_manager/core/nnimanager.ts
```diff
@@ -479,8 +479,6 @@ class NNIManager implements Manager {
                     };
                     const trialJobDetail: TrialJobDetail = await this.trainingService.submitTrialJob(trialJobAppForm);
                     this.trialJobsMaintainer.setTrialJob(trialJobDetail.id, Object.assign({}, trialJobDetail));
-                    // TO DO: to uncomment
-                    assert(trialJobDetail.status === 'WAITING');
```
QuanluZhang: why removing this assert? In which case does trainingservice return non 'WAITING' status?

chicm-ms: When the submitted job is happened to be scheduled immediately after submit,  the status of trialJobDetail is not waiting,  now change to use a snapshot jobdetail object created by job maintainer.

PR #226, opened 2018-10-17
Remove the gap in search space value array

lvybriage:


src/webui/src/components/Sessionpro.tsx
```diff
@@ -138,16 +138,28 @@ class Sessionpro extends React.Component<{}, SessionState> {
...
+                        let value = searchSpace[item]._value;
+                        switch (key) {
+                            case 'loguniform':
+                            case 'qloguniform':
+                                const a = Math.pow(10, value[0]);
+                                const b = Math.pow(10, value[1]);
+                                if (a < b) {
+                                    value = [a, b];
+                                } else {
+                                    value = [b, a];
```
chicm-ms: Maybe we should not change the order of the values specified by user in search space?

lvybriage: ok

src/webui/src/components/Sessionpro.tsx
```diff
@@ -138,16 +138,24 @@ class Sessionpro extends React.Component<{}, SessionState> {
...
-                                value = [b, a];
-                            }
-                            searchSpace[item]._value = value;
+                                searchSpace[item]._value = value;
+                                break;
+
+                            case 'quniform':
+                            case 'qnormal':
+                            case 'qlognormal':
+                                searchSpace[item]._value = [value[0], value[1]];
```
chicm-ms: This line (154) seems does nothing?

chicm-ms: Ok, got it.

Issue #231, opened 2018-10-17
nni installation failure when NODE_PATH exists or node<10.0.0 has installed


QuanluZhang:
**bug 1:**
In my environment, NODE_PATH already exits, e.g., /usr/lib/nodejs:/usr/lib/node_modules:/usr/share/javascript. This induces the error below:
`
  # Cleaning 
  rm -rf /usr/lib/nodejs:/usr/lib/node_modules:/usr/share/javascript
  rm -rf /tmp/nni-yarn
  rm -rf /home/zhac/.local/nni/serve
  mkdir -p /usr/lib/nodejs:/usr/lib/node_modules:/usr/share/javascript
  mkdir: cannot create directory ‘/usr/lib/nodejs:’: Permission denied
  Makefile:178: recipe for target 'install-dependencies' failed
  make: *** [install-dependencies] Error 1
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-req-build-r7w8srdk/setup.py", line 69, in <module>
      'install': CustomInstallCommand
    File "/home/zhac/anaconda3/lib/python3.7/site-packages/setuptools/__init__.py", line 140, in setup
      return distutils.core.setup(**attrs)
    File "/home/zhac/anaconda3/lib/python3.7/distutils/core.py", line 148, in setup
      dist.run_commands()
    File "/home/zhac/anaconda3/lib/python3.7/distutils/dist.py", line 966, in run_commands
      self.run_command(cmd)
    File "/home/zhac/anaconda3/lib/python3.7/distutils/dist.py", line 985, in run_command
      cmd_obj.run()
    File "/home/zhac/anaconda3/lib/python3.7/site-packages/wheel/bdist_wheel.py", line 238, in run
      self.run_command('install')
    File "/home/zhac/anaconda3/lib/python3.7/distutils/cmd.py", line 313, in run_command
      self.distribution.run_command(command)
    File "/home/zhac/anaconda3/lib/python3.7/distutils/dist.py", line 985, in run_command
      cmd_obj.run()
    File "/tmp/pip-req-build-r7w8srdk/setup.py", line 34, in run
      subprocess.run(['make', 'pip-install'], check=True)
    File "/home/zhac/anaconda3/lib/python3.7/subprocess.py", line 468, in run
      output=stdout, stderr=stderr)
  subprocess.CalledProcessError: Command '['make', 'pip-install']' returned non-zero exit status 2.
error
  Failed building wheel for NNI
`
[...]

PR #246, opened 2018-10-18
update Makefile

Crysple:
- To prevent from confusing the NODE_PATH in Makefile with the previously existing environment variable NODE_PATH in the system.
- add prefix 'NNI_' to all potentially confusing variables in Makefile

related to issue #231 bug1


QuanluZhang:
this pr addresses issue #231

PR #254, opened 2018-10-19
Support tensorboard for pai

SparkSnail:


src/nni_manager/common/trainingService.ts
```diff
@@ -78,6 +78,13 @@ interface HostJobDetail {
     readonly status: string;
 }
 
+interface ITensorBoardManager {
+    taskQueue: string[];
+    addCopyDataTask(trialJobId: string): Promise<void>;
+    getLocalDirectory(trialJobId: string): string;
+    copyDataFromHdfs(trialJobId: string): Promise<void>;
```
QuanluZhang: this is only for pai training service, is it proper to add it here in this way?

QuanluZhang:
I think this implementation is not very elegant, because it tightly couples multiple components.

PR #268, opened 2018-10-24
Support nnictl tensorboard

SparkSnail:
Support tensorboard in nnictl of local and remote mode.


tools/nnicmd/tensorboard_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    '''stop tensorboard'''
+    experiment_id = check_experiment_id(args)
+    experiment_config = Experiments()
+    experiment_dict = experiment_config.get_all_experiments()
+    config_file_name = experiment_dict[experiment_id]['fileName']
+    nni_config = Config(config_file_name)
+    tensorboard_pid_list = nni_config.get_config('tensorboardPidList')
+    if tensorboard_pid_list:
+        for tensorboard_pid in tensorboard_pid_list:
+            cmds = ['kill', '-9', str(tensorboard_pid)]
```
yds05: I suggest to catch exception and show error to user.

PR #272, opened 2018-10-25
Sdk update

chicm-ms:
Rename nni.get_parameters() to nni.get_next_parameter()


docs/howto_2_CustomizedTuner.md
```diff
@@ -61,7 +61,7 @@ If the you implement the ```generate_parameters``` like this:
         # your code implements here.
         return {"dropout": 0.3, "learning_rate": 0.4}
 ```
-It's means your Tuner will always generate parameters ```{"dropout": 0.3, "learning_rate": 0.4}```. Then Trial will receive ```{"dropout": 0.3, "learning_rate": 0.4}``` this object will using ```nni.get_parameters()``` API from NNI SDK. After training of Trial, it will send result to Tuner by calling ```nni.report_final_result(0.93)```. Then ```receive_trial_result``` will function will receied these parameters like：
+ It's means your Tuner will always generate parameters ```{"dropout": 0.3, "learning_rate": 0.4}```. Then Trial will receive ```{"dropout": 0.3, "learning_rate": 0.4}``` by calling API ```nni.get_next_parameter()```. Once the trial ends with a result (normally some kind of metrics), it can send the result to Tuner by calling API ```nni.report_final_result()```, such as ```nni.report_final_result(0.93)```. Then your Tuner's ```receive_trial_result``` function will receied the result like：
```
QuanluZhang: It's means -> It means
such as -> for example

tools/nni_annotation/examples/mnist_with_annotation.py
```diff
@@ -247,6 +247,7 @@ def generate_defualt_params():
 
 
 if __name__ == '__main__':
+    """@nni.next_parameter"""
```
QuanluZhang: is this line correct?

QuanluZhang:
could you also update nni annotation's document?

src/sdk/pynni/nni/platform/local.py
```diff
@@ -49,13 +49,16 @@ def request_next_parameter():
...
 
-def get_parameters():
+def get_next_parameter():
     global _param_index
     params_file_name = ''
     if _multiphase and (_multiphase == 'true' or _multiphase == 'True'):
         params_file_name = ('parameter_{}.cfg'.format(_param_index), 'parameter.cfg')[_param_index == 0]
     else:
-        params_file_name = 'parameter.cfg'
+        if _param_index > 0:
```
yds05: What will be the case if _param_index < 0?

docs/howto_2_CustomizedTuner.md
chicm-ms: Thanks, updated.

src/sdk/pynni/nni/platform/local.py
chicm-ms: Thanks, assertion added.

chicm-ms:
> could you also update nni annotation's document?

OK, updated.

tools/nni_annotation/examples/mnist_with_annotation.py
chicm-ms: Thanks, updated.

tools/nni_annotation/examples/mnist_with_annotation.py
```diff
@@ -247,6 +247,7 @@ def generate_defualt_params():
 
 
 if __name__ == '__main__':
+    """@nni.next_parameter()"""
```
QuanluZhang: what is the difference between @nni.next_parameter() and @nni.get_next_parameter()?

chicm-ms: Thanks, it is a mistake.

PR #274, opened 2018-10-26
move reward extraction logic to tuner

leckie-chn:
* move reward extraction logic to tuner
* update doc accordingly
* update pylintrc, to fix numpy.random.RandomState Error
* update .gitignore file


src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -111,20 +111,8 @@ def handle_add_customized_trial(self, data):
 
     def handle_report_metric_data(self, data):
         if data['type'] == 'FINAL':
-            value = None
             id_ = data['parameter_id']
```
yds05: I suggest we simply describe the desired schema for final metric data here.

leckie-chn: agreed

docs/howto_2_CustomizedTuner.md
```diff
@@ -27,12 +27,12 @@ class CustomizedTuner(Tuner):
     def __init__(self, ...):
         ...
     
-    def receive_trial_result(self, parameter_id, parameters, reward):
+    def receive_trial_result(self, parameter_id, parameters, value):
```
QuanluZhang: how about 'value' -> 'result'? @xuehui1991

src/sdk/pynni/nni/tuner.py
```diff
@@ -93,3 +93,16 @@ def _on_exit(self):
 
     def _on_error(self):
         pass
+
+    def extract_scalar_reward(self, value, scalar_key='default'):
+        if isinstance(value, float) or isinstance(value, int):
```
QuanluZhang: ok, we still support float and int, right?

src/sdk/pynni/nni/tuner.py
```diff
@@ -93,3 +93,16 @@ def _on_exit(self):
...
+            value = value
+        elif isinstance(value, dict) and scalar_key in value:
+            value = value[scalar_key]
+            if isinstance(value, float) or isinstance(value, int):
+                pass
+            else:
+                raise RuntimeError('Incorrect final result: the final result for %s should be float/int, or a dict which has a key named "default" whose value is float/int.' % str(self.__class__))
+        else:
+            raise RuntimeError('Incorrect final result: the final result for %s should be float/int, or a dict which has a key named "default" whose value is float/int.' % str(self.__class__)) 
+        return value
```
QuanluZhang: if you throw error anyway, how about change the code as below:
if isinstance(value, dict):
    value = value[scalar_key]
if isinstance(value, float) or instance(value, int):
    return value
else:
    raise ....

src/sdk/pynni/nni/tuner.py
yds05: I suggest to differ 'return' value from 'param' value to eliminate confusion, such as:

if isinstance(value, float) or isinstance(value, int):
        resultValue = value
        elif isinstance(value, dict) and scalar_key in value:
            resultValue = value[scalar_key]
            if isinstance(resultValue, float) or isinstance(resultValue, int):
                pass
            else:
                raise RuntimeError('Incorrect final result: the final result for %s should be float/int, or a dict which has a key named "default" whose value is float/int.' % str(self.__class__))
        else:
            raise RuntimeError('Incorrect final result: the final result for %s should be float/int, or a dict which has a key named "default" whose value is float/int.' % str(self.__class__)) 
        return resultValue

leckie-chn: Yes, I just moved those logic from dispatcher to tuner, to ensure that all the previous examples & trial apis won't be changed.

src/sdk/pynni/nni/tuner.py
```diff
@@ -93,3 +93,16 @@ def _on_exit(self):
...
     def _on_error(self):
         pass
+
+    def extract_scalar_reward(self, value, scalar_key='default'):
+        if isinstance(value, float) or isinstance(value, int):
+            reward = value
+        elif isinstance(value, dict) and scalar_key in value:
+            reward = value[scalar_key]
+            if isinstance(reward, float) or isinstance(reward, int):
+                pass
```
chicm-ms: Suggest to check type more straight forward, such as:
```
if not isinstance(reward, (float, int)):
    raise...
```

leckie-chn: done

PR #282, opened 2018-10-30
Show error information and fix paramiko installation

SparkSnail:


tools/nnicmd/launcher.py
```diff
@@ -36,6 +36,20 @@
 import random
 import string
 
+def print_log_content(config_file_name):
+    '''print log information'''
+    stdout_full_path = os.path.join(NNICTL_HOME_DIR, config_file_name, 'stdout')
```
yds05: can you use a constant module to define stdout_full_path/stderr_full_path? 
I remember these two paths are also used elsewhere, so suggest to keep only on declaration.

tools/nnicmd/launcher.py
```diff
@@ -36,6 +36,20 @@
...
+    stderr_full_path = os.path.join(NNICTL_HOME_DIR, config_file_name, 'stderr')
+    print_normal(' Stderr:')
+    stderr_cmds = ['cat', stderr_full_path]
+    stderr_content = check_output(stderr_cmds)
+    print(stderr_content.decode('utf-8'))
+    print('\n\n')
+    print_normal(' Stdout:')
+    stdout_cmds = ['cat', stdout_full_path]
+    stdout_content = check_output(stdout_cmds)
+    print(stdout_content.decode('utf-8'))
```
QuanluZhang: would be better to first print stdout, then stderr.

PR #285, opened 2018-10-31
Quick fix resume logic

SparkSnail:


tools/nnicmd/launcher.py
```diff
@@ -334,18 +326,8 @@ def resume_experiment(args):
...
-                    experiment_id = key
-                    experiment_endTime = experiment_dict[key]['endTime']
-                else:
-                    if cmp_time(experiment_dict[key]['endTime'], experiment_endTime) > 0:
-                        experiment_id = key
-                        experiment_endTime = experiment_dict[key]['endTime']
-        if experiment_id is None:
-            print_error('There is no experiment stopped!')
-            exit(1)
+        print_error('Please set experiment id! You could use \'nnictl resume {id}\' to resume a stopped experiment!')
```
QuanluZhang: also add 'you could use nnictl experiment list all to show experiments

tools/nnicmd/launcher.py
```diff
@@ -205,8 +205,9 @@ def set_experiment(experiment_config, mode, port, config_file_name):
         return response
     else:
         _, stderr_full_path = get_log_path(config_file_name)
-        with open(stderr_full_path, 'a+') as fout:
-            fout.write(json.dumps(json.loads(response.text), indent=4, sort_keys=True, separators=(',', ':')))
+        if response:
```
QuanluZhang: if this is true, does this mean response must have .text?

SparkSnail: yes.

PR #286, opened 2018-10-31
Tgs salt example

chicm-ms:


QuanluZhang:
suggest to change the folder's name from tgs-salt to kaggle-tgs-salt

chicm-ms:
> suggest to change the folder's name from tgs-salt to kaggle-tgs-salt

Thanks, done.

PR #317, opened 2018-11-02
 Uniform the names of python modules

goooxu:


src/nni_manager/common/log.ts
```diff
@@ -70,17 +70,23 @@ class Logger {
     private DEFAULT_LOGFILE: string = path.join(getLogDir(), 'nnimanager.log');
     private level: number = DEBUG;
     private bufferSerialEmitter: BufferSerialEmitter;
+    private writble: Writable;
```
chicm-ms: Is this typo on purpose?

yds05: typo, writable?

src/nni_manager/core/nnimanager.ts
```diff
@@ -342,7 +341,7 @@ class NNIManager implements Manager {
 
     private async periodicallyUpdateExecDuration(): Promise<void> {
         let count: number = 1;
-        for (; ;) {
+        while (this.status.status !== 'STOPPING') {
```
QuanluZhang: Is it possible that state transfer from 'STOPPING' to 'STOPPED' is very fast, so that this loop misses the 'STOPPING' state.
Suggest using `while (this.status.status !== 'STOPPING' && this.status.status !== 'STOPPED')` or `while (this.status.status !== 'STOPPED')`

src/nni_manager/core/nnimanager.ts
```diff
@@ -396,10 +395,7 @@ class NNIManager implements Manager {
             throw new Error('Error: tuner has not been setup');
         }
         let allFinishedTrialJobNum: number = 0;
-        for (; ;) {
-            if (this.status.status === 'STOPPING') {
-                break;
-            }
+        while (this.status.status !== 'STOPPING') {
```
QuanluZhang: similar problem, and this is my code....

src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -295,7 +297,9 @@ class LocalTrainingService implements TrainingService {
 
     public cleanUp(): Promise<void> {
         this.stopping = true;
-
+        for(const stream of this.streams) {
+            stream.destroy();
+        }
```
QuanluZhang: Is it proper to close all the opened streams at the very end?

goooxu: I think that's OK by now, will find a better way in future.

PR #331, opened 2018-11-06
Update examples

QuanluZhang:
update mnist-annotation

### DIFF
--- a/examples/trials/mnist-annotation/mnist.py
+++ b/examples/trials/mnist-annotation/mnist.py
@@ -180,10 +180,10 @@ def main(params):
     test_acc = 0.0
     with tf.Session() as sess:
         sess.run(tf.global_variables_initializer())
-        """@nni.variable(nni.choice(50, 250, 500), name=batch_num)"""
-        batch_num = params['batch_num']
-        for i in range(batch_num):
-            batch = mnist.train.next_batch(batch_num)
+        """@nni.variable(nni.choice(1, 4, 8, 16, 32), name=batch_size)"""
+        batch_size = params['batch_size']
+        for i in range(params['batch_num']):
+            batch = mnist.train.next_batch(batch_size)
             """@nni.variable(nni.choice(1, 5), name=dropout_rate)"""
             dropout_rate = params['dropout_rate']
             mnist_network.train_step.run(feed_dict={mnist_network.images: batch[0],
@@ -224,7 +224,8 @@ def generate_defualt_params():
         'pool_size': 2,
         'hidden_size': 1024,
         'learning_rate': 1e-4,
-        'batch_num': 200}
+        'batch_num': 2000,
+        'batch_size': 32}
     return params
 
 
--- a/examples/trials/mnist-annotation/mnist.py
+++ b/examples/trials/mnist-annotation/mnist.py
@@ -157,8 +157,8 @@ def main(params):
     '''
     # Import data
     mnist = input_data.read_data_sets(params['data_dir'], one_hot=True)
-    print('Mnist download data down.')
-    logger.debug('Mnist download data down.')
+    print('Mnist download data done.')
+    logger.debug('Mnist download data done.')
 
     # Create the model
     # Build the graph for the deep net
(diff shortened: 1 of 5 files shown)


examples/trials/mnist/config_assessor.yml
```diff
@@ -2,7 +2,7 @@ authorName: default
 experimentName: example_mnist
 trialConcurrency: 1
 maxExecDuration: 1h
-maxTrialNum: 1
+maxTrialNum: 100
```
yds05: a little concern about why would we increase to 100 for the sample config

examples/trials/mnist-annotation/mnist.py
```diff
@@ -180,11 +180,11 @@ def main(params):
...
-        """@nni.variable(nni.choice(50, 250, 500), name=batch_num)"""
-        batch_num = params['batch_num']
-        for i in range(batch_num):
-            batch = mnist.train.next_batch(batch_num)
-            """@nni.variable(nni.choice(1, 5), name=dropout_rate)"""
+        """@nni.variable(nni.choice(1, 4, 8, 16, 32), name=batch_size)"""
+        batch_size = params['batch_size']
+        for i in range(params['batch_num']):
+            batch = mnist.train.next_batch(batch_size)
+            """@nni.variable(nni.choice(0.1, 0.5, 1), name=dropout_rate)"""
```
chicm-ms: Normally dropout rate 1 does not need to be tried, the output should be 0 if dropout function is implemented correctly.

examples/trials/mnist-annotation/mnist.py
```diff
@@ -180,11 +180,11 @@ def main(params):
...
             dropout_rate = params['dropout_rate']
```
chicm-ms: This dropout_rate variable is not used.

examples/trials/mnist-annotation/mnist.py
```diff
@@ -180,11 +180,11 @@ def main(params):
...
             mnist_network.train_step.run(feed_dict={mnist_network.images: batch[0],
                                                     mnist_network.labels: batch[1],
```
chicm-ms: mnist_network.keep_prob: dropout_rate
Literally, keep_prob should be 1-dropout_rate,  is this a mistake?

examples/trials/mnist/config_assessor.yml
QuanluZhang: because 1 is too small. assessor would kill unpromising trials early, so it can run more trials. If you think it is too large as an example, is 20 or 30 better?

examples/trials/mnist-smartparam/mnist.py
```diff
@@ -180,9 +180,9 @@ def main(params):
     test_acc = 0.0
     with tf.Session() as sess:
         sess.run(tf.global_variables_initializer())
-        batch_num = nni.choice(50, 250, 500, name='batch_num')
-        for i in range(batch_num):
-            batch = mnist.train.next_batch(batch_num)
+        batch_size = nni.choice(1, 4, 8, 16, 32, name='batch_size')
+        for i in range(2000):
+            batch = mnist.train.next_batch(batch_size)
             dropout_rate = nni.choice(1, 5, name='dropout_rate')
```
chicm-ms: dropout_rate should in [0, 1)

examples/trials/mnist-annotation/mnist.py
QuanluZhang: sure, how about 0, 0.1, 0.5?

chicm-ms: Yes, it is fine.

examples/trials/mnist-annotation/mnist.py
QuanluZhang: thanks a lot, fixed, and also fixed the same problem in the other examples.

examples/trials/mnist/config_assessor.yml
yds05: yes, I think that will be better.

PR #347, opened 2018-11-09
Update nnictl.py

noklam:
Add a more obvious option to stop all experiments, the help text does not mention how to stop all experiments at once.

### DIFF
--- a/tools/nni_cmd/nnictl.py
+++ b/tools/nni_cmd/nnictl.py
@@ -74,6 +74,7 @@ def parse_args():
     #parse stop command
     parser_stop = subparsers.add_parser('stop', help='stop the experiment')
     parser_stop.add_argument('id', nargs='?', help='the id of experiment')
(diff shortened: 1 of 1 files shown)


tools/nni_cmd/nnictl.py
```diff
@@ -74,6 +74,7 @@ def parse_args():
     #parse stop command
     parser_stop = subparsers.add_parser('stop', help='stop the experiment')
     parser_stop.add_argument('id', nargs='?', help='the id of experiment')
+    parser_stop.add_argument('all', nargs='?', default='all', help='stop all experiments')
```
QuanluZhang: If adding this line, there would be two positional arguments, which is not accurate because `stop` only needs one positional argument, even for `all` (e.g., nnictl stop all). I would suggest to modify `id`'s help info to tell users how to use `all`.

noklam: You are right, I am not familiar with the parser. I change the help text instead, thx.

PR #356, opened 2018-11-12
Fix wrong example gdbt

noklam:
I find that the auto-gdbt example is wrong and some of the UI is misleading. It will show accuracy even though it is actually rmse. It is a problem because we want to minimize rmse and thus the optmization direction is completely the opposite.

Is there a way to set optimizing with a parameter similiar to larger_is_better = True?

image


noklam:
Sorry I will re-open a separate PR, I think I mess up my repository.

PR #357, opened 2018-11-12
Fix wrong example gdbt, minimize rmse instead

noklam:
image
I find that the auto-gdbt example is wrong and some of the UI is misleading. It will show accuracy even though it is actually rmse. It is a problem because we want to minimize rmse and thus the optmization direction is completely the opposite.

Is there an easy way to choose to minimize or maximize a metric?


QuanluZhang:
Thanks for your contribution @noklam . This is the problem of webui, rather than the gbdt example. As shown in config.yml, tuner minimizes rmse. However, webui still sorted the final result (i.e., rmse) from high to low. We will fix this problem soon.

PR #359, opened 2018-11-13
Update ci with new built-in tuner and assessor

Crysple:
- fix tuner and assessor's unit-test
- add medianstop, batchtuner to ci


test/sdk_test.py
```diff
@@ -11,56 +11,53 @@
...
             'optimize_mode': 'maximize'
         }
     }
     dump_yml_content(config_path, experiment_config)
 
-def test_builtin_tuner(tuner_name):
-    remove_files(['sdk_tuner_test/nni_tuner_result.txt'])
-    switch_tuner(tuner_name)
+def test_builtin_dispatcher(dispatch_type, dispatch_name):
+    remove_files(['sdk_test/nni_tuner_result.txt'])
```
QuanluZhang: nni_tuner_result.txt? what about assessor?

test/sdk_test.py
```diff
@@ -11,56 +11,49 @@
...
     
     assert experiment_status, 'Failed to finish in 30 sec'
 
-def run():
-    to_remove = ['tuner_search_space.json', 'tuner_result.txt', 'assessor_result.txt']
-    remove_files(to_remove)
-
-    for tuner_name in TUNER_LIST:
+def run(dispatch_type):
+    LIST = TUNER_LIST if dispatch_type == 'Tuner' else ASSESSOR_LIST
```
yds05: You assume here dispatch_type SHOULD be 'Assessor' if it's not 'Tuner', but I suggest we should also check if dispatch_type is 'Assessor' if it's not 'Tuner', and throw error if the validation failed.

PR #372, opened 2018-11-16
Asynchronous dispatcher

chicm-ms:
Add multiThread config flag on top level of config yml. If multiThread is enabled, dispatcher creates threads to process IPC commands.

multiThreads document will be added later.


src/nni_manager/core/nnimanager.ts
```diff
@@ -542,9 +544,29 @@ class NNIManager implements Manager {
         this.dispatcher.sendCommand(REPORT_METRIC_DATA, metric.data);
     }
 
+    private requestTrialJobs(jobNum: number): void {
+        if (jobNum < 1) {
```
yds05: Do you think we need to assert here? or just skip is enough

src/nni_manager/core/nnimanager.ts
```diff
@@ -542,9 +544,29 @@ class NNIManager implements Manager {
...
+            return;
+        }
+        if (this.dispatcher === undefined) {
+            throw new Error('Dispatcher error: tuner has not been setup');
+        }
+        if (this.experimentProfile.params.multiThread) {
+            for (let i: number = 0; i < jobNum; i++) {
+                this.dispatcher.sendCommand(REQUEST_TRIAL_JOBS, '1');
```
yds05: Suggest to add comment to explain why we ask for 1 parameter in multiThread mode

src/nni_manager/core/nnimanager.ts
chicm-ms: Skip is better here, checking can be done outside if necessary.

src/nni_manager/core/nnimanager.ts
chicm-ms: Thanks,  comments added.

src/nni_manager/core/nnimanager.ts
```diff
@@ -425,10 +427,12 @@ class NNIManager implements Manager {
                 if (this.customizedTrials.length > 0) {
                     const hyperParams: string | undefined = this.customizedTrials.shift();
                     this.dispatcher.sendCommand(ADD_CUSTOMIZED_TRIAL_JOB, hyperParams);
-                } else {
-                    this.dispatcher.sendCommand(REQUEST_TRIAL_JOBS, '1');
+                    requestTrialNum --;
```
QuanluZhang: why `requestTrialNum --`?

src/sdk/pynni/nni/__main__.py
```diff
@@ -28,6 +28,7 @@
 import importlib
```
QuanluZhang: could you pass pylint for all the modified python files?

QuanluZhang: please ignore this comment, not ready now.

chicm-ms: OK, thanks

src/nni_manager/core/nnimanager.ts
QuanluZhang: I got the reason, but this might be not correct? because requestTrialNum in line 430 is the same as requestTrialNum in line 425.

chicm-ms: This requestTrialNum-- is used to ensure that custom trials share same concurrent trial number with non custom trials.

chicm-ms: Updated, logic is same, but looks clearer.

PR #374, opened 2018-11-19
NNICTL set classArgs as optional

SparkSnail:
In nnictl, classArgs is not required, now set it as optional for some kind of tuner and assessor may not require classArgs.


QuanluZhang:
please also update doc accordingly

tools/nni_cmd/config_schema.py
```diff
@@ -34,10 +34,7 @@
 'useAnnotation': bool,
 'tuner': Or({
     'builtinTunerName': Or('TPE', 'Random', 'Anneal', 'Evolution', 'SMAC', 'BatchTuner', 'GridSearch'),
-    'classArgs': {
-        'optimize_mode': Or('maximize', 'minimize'),
-        Optional('speed'): int
-        },
+    Optional('classArgs'): dict,
```
QuanluZhang: since the spec of tpe/smac is different from random/gridsearch/batchtuner, is it possible to check their spec separately?

PR #378, opened 2018-11-19
add gpuNum check for local TS

leckie-chn:


src/nni_manager/training_service/local/localTrainingServiceForGPU.ts
```diff
@@ -78,7 +78,7 @@ class LocalTrainingServiceForGPU extends LocalTrainingService {
     }
 
     protected onTrialJobStatusChanged(trialJob: LocalTrialJobDetailForGPU, oldStatus: TrialJobStatus): void {
-        if (trialJob.gpuIndices !== undefined && trialJob.gpuIndices.length !== 0) {
+        if (trialJob.gpuIndices !== undefined && trialJob.gpuIndices.length !== 0 && this.gpuScheduler !== undefined) {
             if (oldStatus === 'RUNNING' && trialJob.status !== 'RUNNING') {
                 for (const index of trialJob.gpuIndices) {
                     this.availableGPUIndices[index] = false;
```
QuanluZhang: why assigning `false`?

QuanluZhang: if the gpu becomes available, it will be assigned `false`?

src/nni_manager/training_service/local/localTrainingServiceForGPU.ts
```diff
@@ -91,10 +91,12 @@ class LocalTrainingServiceForGPU extends LocalTrainingService {
         trialJobDetail: TrialJobDetail,
         resource: { gpuIndices: number[] }): { key: string; value: string }[] {
         const variables: { key: string; value: string }[] = super.getEnvironmentVariables(trialJobDetail, resource);
-        variables.push({
-            key: 'CUDA_VISIBLE_DEVICES',
-            value: resource.gpuIndices.join(',')
-        });
+        if (this.gpuScheduler !== undefined) {
```
QuanluZhang: it might be better to set CUDA_VISIBLE_DEVICES to be "" when required gpu number is 0.

QuanluZhang: @yds05 @chicm-ms

src/nni_manager/training_service/local/localTrainingServiceForGPU.ts
QuanluZhang: @yds05

yds05: I'm not very confident about the motivation of assigning false here because I'm not the original author of LocalTrainingService, but I think Yan's change is safe here because if this.gpuScheduler is undefined(means gpuNum is set to 0 in nniconfig file), we should not allocate GPU for trialJob.

src/nni_manager/training_service/local/localTrainingServiceForGPU.ts
yds05: Yes, I agree. 
@leckie-chn , how about change to 

variables.push({
           key: 'CUDA_VISIBLE_DEVICES',
          value: this.gpuScheduler === undefined ? '' : resource.gpuIndices.join(',')
            });

Otherwise, All the gpu devices will be visible to the training process even though gpuNum is set to 0.

leckie-chn: @yds05 code added, see blow.

yds05: line 94 can be removed now

Issue #243, opened 2018-10-18
Support for running on Mac OS


leckie-chn:
Currently nni only supports running on Linux, but running on Mac is also important.

leckie-chn:
encountered following error when running `nnictl create --config examples/trials/mnist/config.yml`
```
[2018-11-19 00:28:22] DEBUG (mnist_AutoML) {'dropout_rate': 0.562259326513507, 'conv_size': 2, 'hidden_size': 124, 'batch_size': 8, 'learning_rate': 0.01}
[2018-11-19 00:28:22] WARNING (tensorflow) From mnist.py:151: read_data_sets (from tensorflow.contrib.learn.python.learn.datasets.mnist) is deprecated and will be removed in a future version.
Instructions for updating:
Please use alternatives such as official/mnist/dataset.py from tensorflow/models.
[2018-11-19 00:28:22] WARNING (tensorflow) From /Users/test/Library/Python/3.6/lib/python/site-packages/tensorflow/contrib/learn/python/learn/datasets/mnist.py:260: maybe_download (from tensorflow.contrib.learn.python.learn.datasets.base) is deprecated and will be removed in a future version.
Instructions for updating:
Please write your own downloading logic.
[2018-11-19 00:28:22] WARNING (tensorflow) From /Users/test/Library/Python/3.6/lib/python/site-packages/tensorflow/contrib/learn/python/learn/datasets/base.py:252: _internal_retry.<locals>.wrap.<locals>.wrapped_fn (from tensorflow.contrib.learn.python.learn.datasets.base) is deprecated and will be removed in a future version.
Instructions for updating:
Please use urllib or similar directly.
[2018-11-19 00:28:22] ERROR (mnist_AutoML) <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:841)>
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/urllib/request.py", line 1318, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py", line 1239, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/http/client.py", line 1285, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
[...]

PR #386, opened 2018-11-22
mac support with local & remote mode

leckie-chn:
* issue https://github.com/Microsoft/nni/issues/243


src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -169,7 +169,7 @@ class LocalTrainingService implements TrainingService {
                 this.setTrialJobStatus(trialJob, 'FAILED');
                 try {
                     const state: string = await fs.promises.readFile(path.join(trialJob.workingDirectory, '.nni', 'state'), 'utf8');
-                    const match: RegExpMatchArray | null = state.trim().match(/^(\d+)\s+(\d+)$/);
+                    const match: RegExpMatchArray | null = state.trim().match(/^(\d+)\s+(\d+)/);
```
QuanluZhang: could you provide an example of `state` string?

QuanluZhang:
looks good.

src/nni_manager/common/utils.ts
```diff
@@ -268,5 +268,13 @@ function getIPV4Address(): string {
     throw Error('getIPV4Address() failed because no valid IPv4 address found.')
 }
 
-export { generateParamFileName, getMsgDispatcherCommand, getLogDir, getExperimentRootDir, 
+function getRemoteTmpDir(os_spec: string): string {
```
yds05: use camel naming convention, like osSpec, but I suggest to rename to osType which is a little bit more clear

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -66,8 +66,10 @@ class RemoteMachineTrainingService implements TrainingService {
     private log: Logger;
     private isMultiPhase: boolean = false;
     private trialSequenceId: number;
+    private remoteOS: string;
```
yds05: private readonly

PR #393, opened 2018-11-23
Add nniManagerIp in nnictl and trainingService

SparkSnail:
Add nniManager Ip in nnictl, pai TrainingService and kubeflow TrainingService.
If users set nniManagerIp, pai and kubeflow will use this ip instead of using getIPV4() function.
Web UI will also use this nniManagerIp.


tools/nni_cmd/config_schema.py
```diff
@@ -32,6 +32,7 @@
 Optional('searchSpacePath'): os.path.exists,
 Optional('multiPhase'): bool,
 Optional('multiThread'): bool,
+Optional('nniManagerIp'): str,
```
QuanluZhang: I suggest to put nniManagerIp in the spec of the required trainingservice.

src/nni_manager/training_service/pai/paiConfig.ts
```diff
@@ -120,4 +120,14 @@ export class NNIPAITrialConfig extends TrialConfig{
...
         this.outputDir = outputDir;
     }
-}
\ No newline at end of file
+}
+
+/**
+ * the ip of nni manager
+ */
+export class NNIManagerIpConfig {
```
yds05: duplicate code

PR #394, opened 2018-11-26
fix sdk for NoMoreTrial status

Crysple:


src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -99,12 +99,11 @@ def handle_request_trial_jobs(self, data):
         ids = [_create_parameter_id() for _ in range(data)]
         params_list = self.tuner.generate_multiple_parameters(ids)
 
+        for i, _ in enumerate(params_list):
+            send(CommandType.NewTrialJob, _pack_parameter(ids[i], params_list[i]))
         # when parameters is None.
-        if len(params_list) == 0:
+        if len(params_list) < len(ids):
             send(CommandType.NoMoreTrialJobs, _pack_parameter(ids[0], ''))
```
QuanluZhang: Is it proper to send no_more_trial when `params_list` is smaller than `len(ids)`? @chicm-ms @xuehui1991

QuanluZhang:
did you check all the tuners?

src/sdk/pynni/nni/msg_dispatcher.py
chicm-ms: This handle_request_trial_jobs  is a blocking call even for multi thread, It is reasonable to send NoMoreTrialJobs if len(params_list) < len(ids)

PR #395, opened 2018-11-26
modify loguniform and lognormal

Crysple:
change the definition and implementation of logXXX (loguniform and lognormal)


src/sdk/pynni/nni/evolution_tuner/parameter_expressions.py
```diff
@@ -65,7 +65,7 @@ def loguniform(low, high, random_state):
     high: an float that represent an upper bound
     random_state: an object of numpy.random.RandomState
     '''
-    return np.exp(uniform(low, high, random_state))
+    return 10 ** (uniform(np.log10(mu), np.log10(sigma), random_state))
```
chicm-ms: what is mu and sigma?

Crysple: Sorry I forgot to modify the variable name... Will fix, thanks~

Crysple:
Blocking now...
There are some problems about this pr. I'll check it first

leckie-chn:
what about `qloguniform` & `qlognormal` ? I suppose they should be consistent with `loguniform` & `lognormal`

QuanluZhang:
what about hyperopt? I think we should also do the corresponding modification to hyperopt.

docs/SearchSpaceSpec.md
```diff
@@ -39,13 +39,15 @@ The candidate type and value for variable is here:
    * Suitable for a discrete value with respect to which the objective is still somewhat "smooth", but which should be bounded both above and below. If you want to uniformly choose integer from a range [low, high], you can write `_value` like this: `[low, high, 1]`.
 <br/>
 
-* {"_type":"loguniform","_value":[low, high]}
-   * Which means the variable value is a value drawn according to exp(uniform(low, high)) so that the logarithm of the return value is uniformly distributed.
-   * When optimizing, this variable is constrained to the interval [exp(low), exp(high)].
+* {"_type":"loguniform","_value":[low, high, base]}
```
xuehui1991: I just wonder, do we need to change the search_space.json new design? because I can't tell did I need to add "base" value in the future?

src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
```diff
@@ -61,6 +62,12 @@ def json2space(in_x, name=ROOT):
             if _type == 'choice':
                 out_y = eval('hp.hp.'+_type)(name, _value)
             else:
+                if _type in ['loguniform', 'qloguniform']:
+                    if len(_value) == 3:
+                        _value = list(map(lambda x: math.log(x, _value[-1]), _value[:2]))
+                    else:
+                        _value = np.log10(_value[:2])
+                    _type = 'uniform'
```
xuehui1991: I don't think you could use uniform to instead qloguniform, because q is a step number, it will constraint the value be sampled.

docs/SearchSpaceSpec.md
QuanluZhang: there is default value here, so no bad to support it, right?

PR #409, opened 2018-11-28
Add codeDir file count validation for setClusterConfig

yds05:
Add codeDir file count validation for setClusterConfig
1. Add util method to check file count under a directory
2. In setclusterconfig method, check if file count is larger than 1000 for the codeDir in nniconfig. If yes, reject the setClusterconfig request.


QuanluZhang:
I suggest to move this validation to nnictl, and in ts backend we throw warning rather than error.

PR #430, opened 2018-12-03
update loguniform for smac

Crysple:
since loguniform in smac behaves differently, it is necessary to unify the behaviors of all tuner's loguniform.


src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
```diff
@@ -63,7 +64,8 @@ def generate_pcs(nni_search_space_content):
                                 json.dumps(search_space[key]['_value']),
                                 json.dumps(search_space[key]['_value'][0])))
                         elif search_space[key]['_type'] == 'loguniform':
-                            pcs_fd.write('%s real %s [%s] log\n' % (
+                            search_space[key]['_value'] = list(np.round(np.log(search_space[key]['_value']), 10))
```
QuanluZhang: would be better to briefly explain why you make this change.

QuanluZhang: I mean np.round

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -170,13 +172,19 @@ def generate_multiple_parameters(self, parameter_id_list):
         '''
```
QuanluZhang: should also update the function `generate_parameters`.

src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
Crysple: okay, I will add some comments

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
Crysple: I will add it. But it seems that nni will never invoke generate_parameters actually if you have overwrite generate_parameters already

QuanluZhang: yes, but still we should make `generate_parameters` correct.

PR #438, opened 2018-12-04
Update doc

xuehui1991:
update doc about anneal
move to docs


QuanluZhang:
looks good. Since you moved the file into docs/, could you fix all the broken links that reference this file and all the broken links in this file?

xuehui1991:
> looks good. Since you moved the file into docs/, could you fix all the broken links that reference this file and all the broken links in this file?

fixed

xuehui1991:
> looks good. Since you moved the file into docs/, could you fix all the broken links that reference this file and all the broken links in this file?

fixed

PR #440, opened 2018-12-04
update makefile & doc for pypi & installation

leckie-chn:


deployment/pypi/Makefile
```diff
@@ -1,25 +1,45 @@
...
 .PHONY: upload
 upload:
 	python3 -m pip install --user --upgrade twine
-	python3 -m twine upload dist/*
\ No newline at end of file
+	python3 -m twine upload dist/*
+
+.PHONY: clean
+clean:
+	-rm -rf $(CWD)../../src/sdk/pynni/dist/*.whl
```
chicm-ms: what does '-rm ...' mean?

leckie-chn: same as 'rm', just used in Makefile. see https://github.com/Microsoft/nni/blob/da76b6b0b63058ca8314ff7cd4dbd491828cf6b3/Makefile#L122-L130

chicm-ms: Thanks, got it.

PR #445, opened 2018-12-04
Fix multiphase error message in nnimanager.log

chicm-ms:


chicm-ms:
Fix following error message in nnimanger.log:
[2018-12-04T08:20:54.844Z] ERROR TypeError: Cannot read property 'params' of undefined
    at NNIDataStore.isMultiPhase (/mnt/d/Repos/nni/src/nni_manager/dist/core/nniDataStore.js:169:108)

PR #455, opened 2018-12-05
added search trial by id function

yhung119:
I changed the name searchTrialNo to searchTrial. Now it checks whether the search value is a substring of one of the "Trail Ids". I chose to do substring so the user doesn't need to remember the whole id string and can get the trial. If that's something you don't want, i can change it to match exactly easily.

One issue I noticed during training is that the table stop updating once I searched something. Originally the table and the chart are updated when the status of a trail is finished or a new trial is started. 
Should I create a new issue?


QuanluZhang:
sure, go ahead.

PR #471, opened 2018-12-12
Update annotation to support displaying real choice in searchspace

Crysple:
- Before if you use annotation, searchspace will only display choice indices in type choice or function_choice.
- Now it will display either the real value of the choice or the name of the function_choice
- Fix exception printing cuz exception.arg might not be a string


Crysple:
Note that the example 'smartparamter' will be temporarily disabled
Further modification should be done

tools/nni_annotation/search_space_generator.py
```diff
@@ -66,23 +66,22 @@ def visit_Call(self, node):  # pylint: disable=invalid-name
...
+            name = '__line' + str(node.lineno)
             specified_name = False
 
         if func in ('choice', 'function_choice'):
-            # arguments of `choice` may contain complex expression,
-            # so use indices instead of arguments
-            args = list(range(len(node.args)))
+            # we will use keys in the dict as the choices
+            assert len(node.args) == 1, 'Smart parameter has arguments other than dict'
+            args = [key.n if type(key) is ast.Num else key.s for key in node.args[0].keys]
```
QuanluZhang: add comments

tools/nni_annotation/search_space_generator.py
```diff
@@ -66,23 +66,22 @@ def visit_Call(self, node):  # pylint: disable=invalid-name
...
 
-        if node.keywords:
+        keys = [keyword.arg for keyword in node.keywords]
+        if 'name' in keys:
             # there is a `name` argument
-            assert len(node.keywords) == 1, 'Smart parameter has keyword argument other than "name"'
-            assert node.keywords[0].arg == 'name', 'Smart paramater\'s keyword argument is not "name"'
-            assert type(node.keywords[0].value) is ast.Str, 'Smart parameter\'s name must be string literal'
-            name = node.keywords[0].value.s
+            name_index = keys.index('name')
```
QuanluZhang: add the deleted lines back

tools/nni_annotation/testcase/annotated/dir/simple.py
```diff
@@ -1,13 +1,14 @@
...
+
 def max_pool(k):
     pass
 
 
 h_conv1 = 1
-conv_size = nni.choice(2, 3, 5, 7, name='conv_size')
-h_pool1 = nni.function_choice(lambda : max_pool(h_conv1), lambda : avg_pool
-    (h_conv2, h_conv3), name='max_pool')
+conv_size = nni.choice({2: 2, 3: 3, 5: 5, 7: 7}, name='conv_size')
```
xuehui1991: I am a little confuse of this change. Could you explain?

Crysple: emmmm. Currently it's hard to display the real value in searchspace with nothing changed, so quanlu decided to discard the smartparam temporarily, which has not been public in docs.
I mean, this modification is a temporary move and it is implicit in nni's internal code and will not affect annotation _(which means that the interface of annotation will not be changed)_ but will only affect smartparams _(if you still want to use smartparam, you should write it in this {key: value} format, where the key is the same as the value)_

src/sdk/pynni/tests/test_smartparam.py
```diff
@@ -26,37 +29,37 @@
...
 
     def test_default_name_func(self):
-        val = nni.function_choice(
-            lambda: max(1, 2, 3),
-            lambda: 2 * 2  # NOTE: assign this line number to lineno2
-        )
+        val = nni.function_choice({
+            'max': lambda: max(1, 2, 3),
+            'min': lambda: min(1, 2)  # NOTE: assign this line number to lineno2
+        })
```
QuanluZhang: should also test 2 * 2.
would be better to test something like max() + min()

src/sdk/pynni/tests/test_smartparam.py
```diff
@@ -26,37 +29,37 @@
...
-            'test_smartparam/__line{:d}/function_choice'.format(lineno2): 0
+            'test_smartparam/func/function_choice': 'bar',
+            'test_smartparam/__line{:d}/function_choice'.format(lineno2): 'max'
         }
         nni.trial._params = { 'parameter_id': 'test_trial', 'parameters': params }
 
 
     def test_specified_name(self):
-        val = nni.choice('a', 'b', 'c', name = 'choice1')
+        val = nni.choice({'a': 'a', 'b': 'b', 'c': 'c'}, name = 'choice1')
```
QuanluZhang: add pure number test.

PR #472, opened 2018-12-13
[Kubeflow training service] Use Kubernete API server to replace kubectl dependency

yds05:
<h3>Motivation</h3>
Right now we use kubectl and generate k8s resource yaml file to submit job, delete job and query job. However, that introduces dependency on kubectl and command line execution. Actually, kubectl calls k8s API server to complete CLI actions. So migrating from kubectl to K8S apiserver is more straight-forward.

<h3>Solution</h3>
1. We use  [Node.js client by GoDaddy](https://github.com/godaddy/kubernetes-client) as K8S rest client. 
2. Write kubernetesApiClient class to define operator rest client for tf-operator and pytorch-operator. 
3. Add apiVersion config entry (required) in nni kubeflow config, so that we can support different tf/pytorch operator versions, like v1alpha2 and v1beta2. 
4. Replace kubectl call with kubernetesApiClient call.


src/nni_manager/config/pytorchjob-crd-v1alpha2.json
```diff
@@ -0,0 +1,17 @@
+{
```
QuanluZhang: how about putting these json files under training_service/kubeflow?

yds05: Sure. Will move

yds05: Done. Thanks for the comments.

PR #474, opened 2018-12-14
quick fix create secret in kubeflowTrainingService

SparkSnail:
rest api of kubernetes does not use base64 to encode chars, now use base64 to encode username and then create secret.


tools/nni_cmd/config_schema.py
```diff
@@ -162,6 +162,7 @@
         }
     },{
         'operator': Or('tf-operator', 'pytorch-operator'),
+        'apiVersion': str,
```
yds05: Why add this line?

SparkSnail: to keep same with nfs.

PR #477, opened 2018-12-17
Add curve fitting assessor

PurityFan:
Add new curve fitting assessor


src/sdk/pynni/nni/curvefitting_assessor/curvefitting_assessor.py
```diff
@@ -0,0 +1,78 @@
+# Copyright (c) Microsoft Corporation
+# All rights reserved.
+#
```
xuehui1991: could you add your paper referenced and update the README?

PurityFan: Added README doc and submitted it to a new pr, pls go to: https://github.com/Microsoft/nni/pull/481

PR #478, opened 2018-12-17
Download log

chicm-ms:


src/nni_manager/rest_server/nniRestServer.ts
```diff
@@ -50,6 +51,7 @@ export class NNIRestServer extends RestServer {
         this.app.use(express.static('static'));
         this.app.use(bodyParser.json());
         this.app.use(this.API_ROOT_URL, createRestHandler(this));
+        this.app.use('/logs', express.static(getLogDir()));
```
yds05: should you consider to define a static readonly variable for '/logs', like what you did for this.API_ROOT_URL?

chicm-ms: Thanks, updated.

PR #491, opened 2018-12-19
Add ut to ci & Fix some ut & Add some testcase

Crysple:


src/nni_manager/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "nni",
-  "version": "NNI_VERSION",
+  "version": "1.0.0",
```
chicm-ms: If change this to 1.0.0,  we should remove following line from make file:
sed -ie 's/NNI_VERSION/$(NNI_VERSION)/' $(NNI_PKG_FOLDER)/package.json

chicm-ms: @goooxu

test/unittest.sh
```diff
@@ -1,15 +1,29 @@
...
-chmod +x nni_annotation.sh
-./nni_annotation.sh
 
 # -------------For typescrip unittest-------------
-chmod +x nni_manager.sh
-./nni_manager.sh
+cd ${CWD}/../src/nni_manager
+echo ""
+echo "===========================Testing: nni_manager==========================="
+sed -ie 's/NNI_VERSION/1.0.0/' package.json
```
chicm-ms: This version modification should be done in install scripts, not ut script.

Issue #500, opened 2018-12-20
Experiment status is done when there is a running trial job


SparkSnail:
When max trial number is 1, and the experiment status is done when the trial is not finished.

PR #504, opened 2018-12-20
fix state transition

QuanluZhang:
image

@lvybriage


QuanluZhang:
this pr fixed the issue https://github.com/Microsoft/nni/issues/500

PR #523, opened 2018-12-25
Quick fix remote bug

SparkSnail:
Remote TrainingService, passwd is not required when users set sshKeyPath.


husimple:
@SparkSnail  Hi, as you said that remote TrainingService, passwd is not required when users set sshKeyPath ,but when I unset passwd in config_remote.yml .
ERROR: Failed! Error is: {"statusCode":400,"error":"Bad Request","message":"child \"body\" fails because [child \"machine_list\" fails because [\"machine_list\" at position 0 fails because [child \"passwd\" fails because [\"passwd\" is not allowed to be empty]]]]"}

QuanluZhang:
@husimple thanks for reporting this issue, could you also create a github issue for this problem?

husimple:
@QuanluZhang Do you know how to  set sshKeyPath?

PR #526, opened 2018-12-27
Make it feasible for annotation whether to add an extra line "nni.get_next_parameter()" or not

Crysple:
- if user adds it, it's ok
- if user does not add it, we will get automatically

Note: it is not applicable for multi-phase in annotation


QuanluZhang:
please add one function: if users use `nni.get_next_parameter()`, our annotation pre-compile should output warning to tell users that `nni.get_next_parameter()` is deprecated.

Crysple:
> please add one function: if users use `nni.get_next_parameter()`, our annotation pre-compile should output warning to tell users that `nni.get_next_parameter()` is deprecated.

Has added

PR #530, opened 2018-12-27
Add frameworkcontroller document

SparkSnail:
1. Add frameworkcontroller document.
2. Fix other document small issues.


docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
+**Run an Experiment on FrameworkController**
+===
+NNI supports running experiment using [FrameworkController](https://github.com/Microsoft/frameworkcontroller), called frameworkcontroller mode. FrameworkController is built to orchestrate all kinds of applications on Kubernetes, you could use frameworkcontroller as a training service to run your experiment.
```
yds05: Now you can use frameworkcontroller as the training service to run NNI experiment.

docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Set up Kubernetes Service and kubeconfig
+FrameworkController has same prerequisites as kubeflow mode except that you don't need to install kubeflow. Please refer the [document](./KubeflowMode.md) to set up your kubernetes cluster and other prerequisites for nni.
```
yds05: FrameworkController has similar prerequisites as [kubeflow mode](https://github.com/Microsoft/nni/blob/master/docs/KubeflowMode.md)

docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Set up FrameworkController
+Follow the [guideline](https://github.com/Microsoft/frameworkcontroller/tree/master/example/run) to set up the frameworkcontroller in the kubernetes cluster, nni support frameworkcontroller by the statefulset mode.
```
yds05: remove 'the' before frameworkcontroller.

docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Design
+Please refer the design of [kubeflow training service](./KubeflowMode.md), frameworkcontroller training service pipeline is similar to kubeflow training service.
```
yds05: 'to kubeflow training service' could be removed

docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
...
+  keyVault:
+    vaultName: {your_vault_name}
+    name: {your_secert_name}
+  azureStorage:
+    accountName: {your_storage_account_name}
+    azureShare: {your_azure_share_name}
+```
+Note: You should explicitly set `trainingServicePlatform: frameworkcontroller` in nni config yaml file if you want to start experiment in frameworkcontrollerConfig mode. 
+
+The trial's config format for nni frameworkcontroller mode is a simple version of frameworkcontroller's offical config, you could refer the [tensorflow example of frameworkcontroller](https://github.com/Microsoft/frameworkcontroller/blob/master/example/framework/scenario/tensorflow/cpu/tensorflowdistributedtrainingwithcpu.yaml) for deeply understanding.
```
yds05: change to 'for deep understanding'

PurityFan:
LGTM!

yqwang-ms:
Can we refer kubeflow doc as fewer as possible? It is better to refer a common k8s doc instead of kubeflow doc 😆

Besides, what is the doc to support other frameworks except for tf? And it would be better if we can mention the strength that frameworkcontroller can support all these frameworks.

And what is the plan to add the related project frameworkcontroller and the how-to in the nni project home readme? :)

yqwang-ms:
BTW, I see you added the frameworkcontroller's framework crd. By design, it will be created by frameworkcontroller itself, so frameworkcontroller's user, such as nni, shouldn't care about it. So, I wonder why you need maintain the crd file in nni project?

docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
...
+Trial configuration in frameworkcontroller mode have the following configuration keys:
+* taskRoles: you could set multiple task roles in config file, and each task role is a basic unit to process in kubernetes cluster.
+   * name: the name of task role specified, like "worker", "ps", "master".
+   * taskNum: the replica number of the task role.
+   * command: the users' command to be used in the container.
+   * gpuNum: the number of gpu device used in container.
+   * cpuNum: the number of cpu device used in container.
+   * memoryMB: the memory limitaion to be specified in container.
+   * image: the docker image used to create pod and run the program.
+   * frameworkAttemptCompletionPolicy: the policy to run framework, please refer the [user-manual](https://github.com/Microsoft/frameworkcontroller/blob/master/doc/user-manual.md) to get the specific information.
```
yqwang-ms: why not also expose retrypolicy in your nni spec?

SparkSnail: This is the first version of FrameworkcontrollerTrainingService, add we will support more spec in next version.

yqwang-ms: OK, so what is current retrypolicy for nni? And where is the hardcoded retrypolicy ?

SparkSnail: NNI does not set retrypolicy spec in code. Beside, we will add a document to tell users how to add their needed spec in nni and how to contribute to nni's code.

SparkSnail:
@yqwang-ms Thanks for your comment. The docs about kubeflow is https://github.com/Microsoft/nni/blob/dev-doc/docs/KubeflowMode.md, there is only tf operator doc now, pytorch operator will be added to this doc. I will update the doc as you suggested, and update the pr.
As for "add the related project frameworkcontroller ", we will discuss the plan in meeting, and it will be added soon.
I add crd file in project because nni use ts kubernetes rest client, and the code need to use crd template to create job.

yqwang-ms:
Got it :)
And seems currently you only support single machine training, so why not directly support distributed training? (to support it you need to use frameworkbarrier and enable fancyRetryPolicy, see https://github.com/Microsoft/frameworkcontroller/blob/master/example/framework/scenario/tensorflow/cpu/tensorflowdistributedtrainingwithcpu.yaml)

docs/FrameworkControllerMode.md
QuanluZhang: i suggest the link to be https://github.com/Microsoft/frameworkcontroller/blob/master/doc/user-manual.md#frameworkattemptcompletionpolicy
A concern is this link is more like a specification, we should give example to show why completionpolicy is necessary. For example, ps does not stop, only worker stops. this completionpolicy helps stop ps.

docs/FrameworkControllerMode.md
```diff
@@ -0,0 +1,84 @@
+**Run an Experiment on FrameworkController**
+===
+NNI supports running experiment using [FrameworkController](https://github.com/Microsoft/frameworkcontroller), called frameworkcontroller mode. FrameworkController is built to orchestrate all kinds of applications on Kubernetes, now you can use frameworkcontroller as the training service to run NNI experiment.
+
+## Set up Kubernetes Service and kubeconfig
+FrameworkController has similar prerequisites as kubeflow mode, like K8S installation and storage preparation, but you don't need to install kubeflow. Please refer Kubeflow mode [document](./KubeflowMode.md) to set up your kubernetes cluster and other prerequisites for nni.
```
QuanluZhang: I suggest to copy the corresponding steps in ./KubeflowMode.md here.

PR #538, opened 2018-12-28
To install the whole nni in an virtual environment

Crysple:


Crysple:
- Colorful output on Mac
- Modify installation priority
  - Installation
    - if python is system-wide and user has no privilege ----> install nni locally (that is ~/.local/)
    - otherwise install in current python's site-package (for situation where user is a root user or wants to install nni in his or her own installed python, such as python in a virtual environment)
  - Launch
    - if python is system-wide python, give priority to using local nni if it exists
    - otherwise use nni installed in current python's site-package (for situation where nni is installed globally or in a self-installed python)
- Remove `--user` while installing nni through pip cuz it will always install nni to ~/.local(if one wants to install in a virtual environment, the flag will cause some unexpected things). Users should have the ability to decide themselves whether to add this flag.
- Remove `source .bashrc` in install.sh. This will refresh the bash and might cause something unwanted. For example, deactivate the virtual environment.....

Crysple:
Additions:
- Solve the permission problem happened when different users installed nni in the same machine

Makefile
```diff
@@ -1,32 +1,38 @@
...
-
+SHELL := /bin/bash
 PIP_INSTALL := python3 -m pip install --no-cache-dir
 PIP_UNINSTALL := python3 -m pip uninstall
 
 ## Colorful output
-_INFO := $(shell echo -e '\e[1;36m')
-_WARNING := $(shell echo -e '\e[1;33m')
-_END := $(shell echo -e '\e[0m')
+_INFO := $(shell echo -e '\033[1;36m')
```
leckie-chn: have you done regression test on Linux about this colorful output?

Crysple: Yes, Linux also support it.
It is because in OS 'Esc' is only allowed to be encoded to '\033'. However in Linux both '\e' and '\033' are allowed.

tools/nni_cmd/launcher.py
```diff
@@ -72,16 +72,26 @@ def start_rest_server(port, platform, mode, config_file_name, experiment_id=None
...
+    # If system-wide python is used, we will give priority to using user-sitepackage given that nni exists there
+    if python_sitepackage.startswith('/usr') or python_sitepackage.startswith('/Library'):
+        local_python_dir = str(Path(site.getusersitepackages()).parents[2])
+        entry_file = os.path.join(local_python_dir, 'nni', 'main.js')
+        entry_dir = os.path.join(local_python_dir, 'nni')
+    else:
+        # If this python is not system-wide python, we will use its site-package directly
+        sys_wide_python = False
+
+    if not sys_wide_python or not os.path.isfile(entry_file):
```
QuanluZhang: it is possible `entry_file` is undefined.

PR #551, opened 2018-12-29
[Logging architecture refactor] Remove unused metrics related code in nni trial_tools, support kubeflow mode for logging architecture refactor

yds05:


src/sdk/pynni/nni/platform/local.py
```diff
@@ -36,7 +36,7 @@
     os.makedirs(_outputdir)
 
 _nni_platform = os.environ['NNI_PLATFORM']
-if _nni_platform != 'pai':
+if _nni_platform not in ['pai', 'kubeflow']:
```
SparkSnail: add frameworkcontroller.

yds05: Sure. Thanks for reminding. 
Already created an issue to track: https://github.com/Microsoft/nni/issues/552

PR #559, opened 2019-01-02
Code coverage report

chicm-ms:
To see python code coverage for integration test:
1. cd nni/test
2. run it.sh
3. check reports in coverhtml directory


src/nni_manager/core/test/import_all.test.ts
```diff
@@ -0,0 +1,15 @@
+import * as glob from 'glob';
+
+// Istanbul only generates report for used/imported files, the files are not used/imported by test cases
+// are not included in code coverage reports.
+// This is a workaround to import all files in order to show all source files in code coverage reports.
+
+glob.sync('**/*.ts').forEach((file) => {
```
yds05: some files in node_modules folder also have *.test.ts file. Should we consider to skip these files?

Issue #558, opened 2019-01-02
The trial process is still alive after the nnictl stop command.


leelaylay:
The trial process is still running after the nnictl stop command, so it takes so much resources of the computer and I have to kill it by force.

PS: My `trainingServicePlatform` is remote.

QuanluZhang:
I suspect that it is because of this line: 
https://github.com/Microsoft/nni/blob/master/src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts#L266
Could @SparkSnail or someone take a look?

SparkSnail:
fixed in https://github.com/Microsoft/nni/pull/564

PR #564, opened 2019-01-03
Fix remote TrainingService bug, change forEach to "for of"

SparkSnail:
trial job could not be stopped in remote machine when experiment is stopped, because awit/async does not work normally in forEach, refer https://codeburst.io/javascript-async-await-with-foreach-b6ba62bbf404.
fix issue: https://github.com/Microsoft/nni/issues/558

### DIFF
--- a/src/nni_manager/training_service/kubernetes/kubernetesTrainingService.ts
+++ b/src/nni_manager/training_service/kubernetes/kubernetesTrainingService.ts
@@ -93,7 +93,7 @@ abstract class KubernetesTrainingService {
         return Promise.resolve(jobs);
     }
 
-    public getTrialJob(trialJobId: string): Promise<TrialJobDetail> {
+    public async getTrialJob(trialJobId: string): Promise<TrialJobDetail> {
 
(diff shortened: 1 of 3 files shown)


src/nni_manager/training_service/kubernetes/kubernetesTrainingService.ts
```diff
@@ -81,14 +81,14 @@ abstract class KubernetesTrainingService {
         }
     }
 
-    public listTrialJobs(): Promise<TrialJobDetail[]> {
+    public async listTrialJobs(): Promise<TrialJobDetail[]> {
         const jobs: TrialJobDetail[] = [];
         
-        this.trialJobsMap.forEach(async (value: KubernetesTrialJobDetail, key: string) => {
+        for (const [key, value] of this.trialJobsMap) {
```
goooxu: key, value without type definition, could it pass tslint ?

SparkSnail: destructuring assignment does not support declare type.

Issue #535, opened 2018-12-28
nni installation issue


QuanluZhang:
nni installation from source does not support installation from different users on the same machine, because they download node and yarn to the same path (i.e., /tmp/)

egg-west:
I guess it can be solved by editing Makefile?

Crysple:
> I guess it can be solved by editing Makefile? More specifically, in line 41-42.
> And is it the same as issue #468 ?

It's the same problem in different locations. I am planning to use username to create a folder in /tmp for every user to install or use NNI.

egg-west:
> > I guess it can be solved by editing Makefile? More specifically, in line 41-42.
> > And is it the same as issue #468 ?
> 
> It's the same problem in different locations. I am planning to use username to create a folder in /tmp for every user to install or use NNI.

Sounds good.

Issue #468, opened 2018-12-11
Permission denied when different users running annotation on the same machine


Crysple:
Known issue. It is because annotation will create tmp directory under /tmp/annoatation

scarlett2018:
Closing the issue as it had released with 0.5.

PR #567, opened 2019-01-04
fix annotation permission deny

Crysple:
- Similar to resolving permissions issues during installation, here we create a folder for every user under.
- So it guarantees that the user has write access to the folder


egg-west:
Solved these I think?
#535 
#468

Crysple:
> Solved these I think?
> #535
> #468

No, another problem is related to installation. So I fixed it in my pr #535 and tested them together

PR #568, opened 2019-01-04
Dev weight sharing

leckie-chn:
* example for weight sharing on ga_squad
* tutorial for weight sharing & async dispatcher
* fix ga_squad tuner bug
* test for weight sharing & async dispatcher


docs/AdvancedNAS.md
```diff
@@ -0,0 +1,111 @@
+# Tutorial for Advanced Neural Architecture Search
```
xuehui1991: I guess weight sharing is a advanced feature. Do we need to put in "AdvancedNAS.md"  or "AdvancedFeature.md"? I am not sure about that..

docs/AdvancedNAS.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Weight Sharing among trials
+Currently we recommend sharing weights through NFS (Network File System), which supports sharing files across machines, and is light-weighted, (relatively) efficient. We also welcome contributions from the community on more efficient techniques.
+
+### NFS Setup
+In NFS, files are physically stored on a server machine, and trials on the client machine can read/write those files in the same way that they access local files.
+
+#### Install NFS on server machine
+First, install NFS server:
+```bash
```
xuehui1991: can we write a script which including all the commands? (suggestion

docs/AdvancedNAS.md
```diff
@@ -0,0 +1,111 @@
...
+sudo mkdir -p /tmp/nni/shared
+sudo echo "/tmp/nni/shared *(rw,sync,no_subtree_check,no_root_squash)" >> /etc/exports
+sudo service nfs-kernel-server restart
+```
+You can check if the above directory is successfully exported by NFS using `sudo showmount -e localhost`
+
+#### Install NFS on client machine
+First, install NFS client:
+```bash
```
xuehui1991: same here.

docs/AdvancedNAS.md
```diff
@@ -0,0 +1,111 @@
...
+```
+Then create & mount the mounted directory of shared files:
+```bash
+sudo mkdir -p /mnt/nfs/nni/
+sudo mount -t nfs 10.10.10.10:/tmp/nni/shared /mnt/nfs/nni
+```
+where `10.10.10.10` should be replaced by the real IP of NFS server machine in practice.
+
+### Example code for trial
+In our example, we assign each layer a `hash_id` to identify whether a previously trained model weight is sharable,and construct the tensorflow graph using `hash_id` as the variable scope name:
```
xuehui1991: shall we split the detail of example from this general document? or add an example link? (Because I think it too specific..)

docs/AdvancedNAS.md
```diff
@@ -0,0 +1,111 @@
...
+            ...
+        ]
+    },
+    "restore_dir": "/mnt/nfs/nni/ga_squad/87",
+    "save_dir": "/mnt/nfs/nni/ga_squad/95"
+}
+```
+
+### Tuner customization for sharing policy
+We recommend implementing sharing policy for customized tuner through the calculation of `Layer.hash_id`. In our example, a layer is sharable iff. the configurations of the layer itself and all its previous layers are not changed. For details, see `Layer.update_hash` and `Graph.update_hash` function in [graph.py](../examples/tuners/weight_sharing/ga_customer_tuner/graph.py)
```
xuehui1991: what's the 'Layer.hash_id'? does all the customized tuner need them? what if I don't have concept of "layer" in my case?

docs/AdvancedNAS.md
leckie-chn: Maybe we can, both for server machine & client machine. I suppose kubeflow also depends on NFS. Maybe we can just tell the users that they need to install & config NFS first. And let themselves Bing the instructions.

docs/AdvancedNAS.md
leckie-chn: This tutorial is not just about weight sharing, but also async dispatcher mode. But I agree with you that we can put AdvancedNAS under AdvancedFeatures. Open to futher discussion.

docs/AdvancedNAS.md
xuehui1991: I think as a user, he may not want to read many material. He just want to quickly find the solution. So shall we tell him how to do weight sharing in a simple scenario? Instead of jump into advanced practice?

leckie-chn: I assume that the features of weight sharing & async dispatcher themselves are for advanced users. Can you find any senarios other than NAS?

PR #570, opened 2019-01-05
Add dependent package installation for tuners before start restful server

egg-west:
I tried SMAC tuner on mnist trial. And it seems everything works fine until I failed to open WebUI. After figuring out that SMAC has 2 packages to install, I think the launcher should deal with this and show the error rather than pretend to working fine.

So I add this check in `launcher.py` . And to keep the maintain convenient, I copied the map from tuners' name to their module in `nni` to `constatnts.py`.

### DIFF
--- a/tools/nni_cmd/constants.py
+++ b/tools/nni_cmd/constants.py
@@ -82,3 +82,16 @@ COLOR_RED_FORMAT = '\033[1;31;31m%s\033[0m'
 COLOR_GREEN_FORMAT = '\033[1;32;32m%s\033[0m'
 
 COLOR_YELLOW_FORMAT = '\033[1;33;33m%s\033[0m'
+
+ModuleName = {
+    'TPE': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Random': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Anneal': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Evolution': 'nni.evolution_tuner.evolution_tuner',
+    'SMAC': 'nni.smac_tuner.smac_tuner',
+    'BatchTuner': 'nni.batch_tuner.batch_tuner',
+    'Medianstop': 'nni.medianstop_assessor.medianstop_assessor',
+    'GridSearch': 'nni.gridsearch_tuner.gridsearch_tuner',
+    'NetworkMorphism': 'nni.networkmorphism_tuner.networkmorphism_tuner',
+    'Curvefitting': 'nni.curvefitting_assessor.curvefitting_assessor'
+}
--- a/tools/nni_cmd/launcher.py
+++ b/tools/nni_cmd/launcher.py
@@ -21,9 +21,10 @@
 
 import json
 import os
+import sys
 import shutil
 import string
-from subprocess import Popen, PIPE, call, check_output
+from subprocess import Popen, PIPE, call, check_output, check_call
 import tempfile
 from nni_annotation import *
 from .launcher_utils import validate_all_content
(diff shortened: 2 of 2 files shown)


tools/nni_cmd/constants.py
```diff
@@ -81,4 +81,17 @@
...
+    'TPE': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Random': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Anneal': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Evolution': 'nni.evolution_tuner.evolution_tuner',
+    'SMAC': 'nni.smac_tuner.smac_tuner',
+    'BatchTuner': 'nni.batch_tuner.batch_tuner',
+    'Medianstop': 'nni.medianstop_assessor.medianstop_assessor',
+    'GridSearch': 'nni.gridsearch_tuner.gridsearch_tuner',
+    'NetworkMorphism': 'nni.networkmorphism_tuner.networkmorphism_tuner',
+    'Curvefitting': 'nni.curvefitting_assessor.curvefitting_assessor'
```
QuanluZhang: Medianstop and Curvefitting are not tuner.

egg-west: Thanks, fixed.

tools/nni_cmd/constants.py
```diff
@@ -81,4 +81,15 @@
...
\ No newline at end of file
+COLOR_YELLOW_FORMAT = '\033[1;33;33m%s\033[0m'
+
+ModuleName = {
+    'TPE': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Random': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Anneal': 'nni.hyperopt_tuner.hyperopt_tuner',
+    'Evolution': 'nni.evolution_tuner.evolution_tuner',
+    'SMAC': 'nni.smac_tuner.smac_tuner',
+    'BatchTuner': 'nni.batch_tuner.batch_tuner',
```
xuehui1991: Seems this part has conflict with master now

tools/nni_cmd/launcher.py
```diff
@@ -272,6 +273,17 @@ def set_experiment(experiment_config, mode, port, config_file_name):
...
     '''follow steps to start rest server and start experiment'''
     nni_config = Config(config_file_name)
+
+    # check packages for tuner
+    if experiment_config.get('tuner') and experiment_config['tuner'].get('builtinTunerName'):
+        tuner_name = experiment_config['tuner']['builtinTunerName']
+        module_name = ModuleName[tuner_name]
+        try:
+            check_call([sys.executable, '-c', 'import %s'%(module_name)])
+        except:
```
xuehui1991: Shall we only catch Exception like "ImportError" or "ModuleNotFoundError" here? :)

tools/nni_cmd/constants.py
egg-west: You mean the customer tuner?

tools/nni_cmd/launcher.py
egg-west: Sure, fixed.

tools/nni_cmd/constants.py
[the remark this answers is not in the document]
egg-west: Good idea, fixed.

tools/nni_cmd/launcher.py
```diff
@@ -272,6 +274,17 @@ def set_experiment(experiment_config, mode, port, config_file_name):
...
+
+    # check packages for tuner
+    if experiment_config.get('tuner') and experiment_config['tuner'].get('builtinTunerName'):
+        tuner_name = experiment_config['tuner']['builtinTunerName']
+        module_name = ModuleName[tuner_name]
+        try:
+            check_call([sys.executable, '-c', 'import %s'%(module_name)])
+        except ModuleNotFoundError as e:
+            print_error('The tuner %s should be installed through nnictl'%(tuner_name))
+            exit(1)
```
xuehui1991: @chicm-ms  do you have any concern about exit(1)?

xuehui1991: @chicm-ms do you have any concern here?

chicm-ms: @xuehui1991 Do you mean the exit(1)?  Since here is in launcher, the rest server won't start if the checking fails, it won't be a big problem for code coverage tool here.  And there are already a couple of exit calls in launcher.

PR #571, opened 2019-01-06
Fix hyperband

QuanluZhang:
1. fix the bug that hyperband does not support final/intermediate metric data in `dict` format.
2. make hyperband also use final metric data, not only intermediate metric data.

### DIFF
--- a/src/sdk/pynni/nni/hyperband_advisor/hyperband_advisor.py
+++ b/src/sdk/pynni/nni/hyperband_advisor/hyperband_advisor.py
@@ -214,7 +214,7 @@ class Bracket():
         self.num_configs_to_run.append(len(hyper_configs))
         self.increase_i()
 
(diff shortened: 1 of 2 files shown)


src/sdk/pynni/nni/hyperband_advisor/hyperband_advisor.py
```diff
@@ -214,6 +208,14 @@ def _record_hyper_configs(self, hyper_configs):
         self.num_configs_to_run.append(len(hyper_configs))
         self.increase_i()
 
+def extract_scalar_reward(value, scalar_key='default'):
+    if isinstance(value, float) or isinstance(value, int):
```
xuehui1991: what's the difference between this function and the extract_scalar_reward function in the other file?

QuanluZhang: the same.

PR #578, opened 2019-01-08
 Fix a race condidtion issue in trial_keeper for reading log from pipe

yds05:


tools/nni_trial_tool/trial_keeper.py
```diff
@@ -65,7 +65,7 @@ def main_loop(args):
     while True:
         retCode = process.poll()
         # child worker process exits and all stdout data is read
-        if retCode is not None and log_pipe_stdout.is_read_completed == True:
+        if retCode is not None and log_pipe_stdout.set_process_exit() and log_pipe_stdout.is_read_completed == True:
```
chicm-ms: It seems log_pipe_stdout.set_process_exit()  is always true,  better to directly call it.

yds05: put set_process_exit() into the expression is just a chain call, to make sure if retCode is not None, set_process_exit() will be called then log_pipe_stdout.is_read_completed will be judged.

PR #582, opened 2019-01-08
Doc fix: formats and typo.

squirrelsc:


QuanluZhang:
suggest to change back to CRLF

PurityFan:
Great job! 
But I am a little worried, because this pr and the branch of dev-doc modify the same document, such as "Overview". And this two modification seems to be difficult to merge due to lots of **conflict**, I am wondering if we will eventually have a big trouble when merging?

.gitignore
```diff
@@ -67,3 +67,5 @@ typings/
 __pycache__
 build
 *.egg-info
+
```
leelaylay: I think that we can remove `.vscode` in gitignore.

docs/GetStarted.md
```diff
@@ -2,27 +2,35 @@
...
+  git
+  wget
+  ```
+
+  python pip should also be correctly installed. You could use "python3 -m pip -v" to check in Linux.
 
-    python pip should also be correctly installed. You could use "python3 -m pip -v" to check in Linux.
-    
-    * Note: we don't support virtual environment in current releases.
+  * Note: we don't support virtual environment in current releases.
```
leelaylay: It seems that virtual environment like conda is supported if we install it in the right way.

.gitignore
squirrelsc: .vscode is for vs code local settings. It's better to keep it local, instead of sharing.

docs/GetStarted.md
squirrelsc: There is no change on original content, I just changed indent of those lines. We can review content again, and get it correctly.

PR #593, opened 2019-01-10
specify nni-sdk version in setup.py

Crysple:
otherwise it won't be upgraded


xuehui1991:
Shall we close this PR? @Crysple

PR #595, opened 2019-01-11
update README in metis and update RuntimeError info

xuehui1991:


src/sdk/pynni/nni/metis_tuner/metis_tuner.py
```diff
@@ -134,8 +134,8 @@ def update_search_space(self, search_space):
                     self.x_bounds[idx] = key_range
```
QuanluZhang: suggest to throw error if it is not number here.

src/sdk/pynni/nni/metis_tuner/metis_tuner.py
```diff
@@ -132,10 +132,15 @@ def update_search_space(self, search_space):
                     self.x_types[idx] = 'range_continuous'
                 elif key_type == 'choice':
                     self.x_bounds[idx] = key_range
+
+                    for key_value in key_range:
+                        if (not isinstance(key_value, int)) and (not isinstance(key_value, float)):
```
Crysple: actually this can be reduced to isinstance(key_value, [int, float]).
But it does not matter.

xuehui1991: Good suggestion!

PR #598, opened 2019-01-11
Integration test refactoring

chicm-ms:
1.  Refactor integration test, new IT test cases can be added by creating config.yml
2.  Test examples in integration test
3.  Update some examples to make them can be finished in shorter time for integration test.


chicm-ms:
Cancel, replaced by another commit.

PR #599, opened 2019-01-11
Update Dockfile

leelaylay:
1.One of my suggestion is to add this line `ARG CACHEBUST=1`  to make sure to run the command below without docker cache every time.
```docker
RUN python3 -m pip --no-cache-dir install nni
```

The docker build command changes the meantime.
```bash
docker build --build-arg CACHEBUST=$(date+%s) -t nni/nni .
```

2.Considering the present docker does not work in nni local, shall we offer the nni local runtime(node and yarn) in docker?


Mitchell0069:
+

PR #609, opened 2019-01-15
Update release note

xuehui1991:
Update the README.md in the home page.


QuanluZhang:
could you also check are there other places that are still 0.4?

PR #619, opened 2019-01-17
update pipeline

chicm-ms:


test/config_test/examples/mnist.test.yml
```diff
@@ -2,7 +2,7 @@ authorName: nni
 experimentName: default_test
 maxExecDuration: 5m
 maxTrialNum: 2
-trialConcurrency: 2
+trialConcurrency: 1
```
xuehui1991: why change this?

chicm-ms: This is a temp workaround for this issue:  https://github.com/Microsoft/nni/issues/621

PR #620, opened 2019-01-17
add SQuAD_evolution_examples.md

xuehui1991:
Will refine later.


docs/SQuAD_evolution_examples.md
```diff
@@ -0,0 +1,312 @@
...
+    optimize_mode: maximize
+trial:
+  command: python3 trial.py
+  codeDir: ~/nni/examples/trials/ga_squad
+  gpuNum: 0
+```
+
+In the "trial" part, if you want to use GPU to perform the architecture search, change `gpuNum` from `0` to `1`. You need to increase the `maxTrialNum` and `maxExecDuration`, according to how long you want to wait for the search result.
+
+`trialConcurrency` is the number of trials running concurrently, which is the number of GPUs you want to use, if you are setting `gpuNum` to 1.
```
chicm-ms: ``` which is the number of GPUs you want to use, if you are setting gpuNum to 1.```
This is true, but since trialConcurrency is not directly related to gpuNum,  suggest to remove it.

PR #625, opened 2019-01-17
Integration tests refactoring

chicm-ms:
1.  Add following examples into integration test:
     mnist
     mnist-annotation
     mnist-keras
     cifar10-pytorch
     sklearn-classification
     sklearn-regression
2.  Update most of above examples to allow configuring steps/epochs through command line parameters.
3.  Add  multiphase,  multi-thread,  metrics test into integration test.
4.  New test cases (such as examples test) can be added into  integration test by creating a yml configuration file under directory  ```nni/test/config_test``` .
5. Run integration test in customized agent pool "NNI CI GPU" .


azure-pipelines.yml
```diff
@@ -1,35 +1,44 @@
...
+  - script: python3 -m pip install --upgrade pip setuptools --user
     displayName: 'Install python tools'
   - script: |
       source install.sh
     displayName: 'Install nni toolkit via source code'
+  - script: |
+      python3 -m pip install sklearn --user
+      python3 -m pip install torchvision --user
+      python3 -m pip install keras --user
+      python3 -m pip install tensorflow==1.9.0 --user
```
yds05: Suggest to install tensorflow-gpu 1.10, to be consistent with NNI official Docker image.

chicm-ms: Thanks, updated.

azure-pipelines.yml
```diff
@@ -1,35 +1,45 @@
...
 
   steps:
-  - script: python3 -m pip install --upgrade pip setuptools
+  - script: python3 -m pip install --upgrade pip setuptools --user
     displayName: 'Install python tools'
   - script: |
       source install.sh
     displayName: 'Install nni toolkit via source code'
+  - script: |
+      python3 -m pip install scikit-learn==0.20.0 --user
```
xuehui1991: if for test example, we need to install lightgbm and pandas.

chicm-ms: Currently only test following examples:
mnist
mnist-annotation
mnist-keras
cifar10-pytorch
sklearn-classification
sklearn-regression

When a new example is added,  we should add dependencies here for the new example.

PR #630, opened 2019-01-18
update dev-doc to sphinx

leckie-chn:
Build homepage using sphinx instead of mkdocs

To view doc changes locally, you can 
* install dependencies using `python3 -m pip install -r docs/requirements.txt`
* build html using `cd docs/ && make html`
then you can see built webpage under `docs/_build/html`


docs/RELEASE.md
```diff
@@ -1,6 +1,35 @@
-# Release 0.4 - 12/6/2018
-
-## Major Features
+# ChangeLog
+
```
xuehui1991: please add v0.5 release...

leckie-chn: Actually I reverted the merge with v0.5 branch before initiating this PR. changes for v0.5 should not come in dev-doc branch currently. At least there should be a seperate merge PR.

Issue #626, opened 2019-01-17
Fail to find nni under both "%s" and "%s"' % (local_python_dir, python_dir)


gonghuiyun:
![nni_fail](https://user-images.githubusercontent.com/12653453/51316141-c12ada00-1a8e-11e9-94b3-3db8b17da14c.png)

gonghuiyun:
> It seems nni is not correctly installed.
> Could you describe how did you install nni?
> Your python version?
> Did you installed nni out of python virtualenv before?
> Could you list the directory: ~/.local/ and <your virtualenv>/ ? Is there nni/ directory there?

1)my OS:Ubuntu 16.04
Firstly I activate tensorflow env in anaconda, then I use pip command:
(tensorflow) abc@abc-System-Product-Name:~$ python3 -m pip install --user --upgrade nni
install information:
![install](https://user-images.githubusercontent.com/12653453/51390376-11c33580-1b6a-11e9-9b33-f758c755d5f0.png)
2)my python version:
![python vi](https://user-images.githubusercontent.com/12653453/51390427-3e774d00-1b6a-11e9-9a3e-9ebd627064ec.png)
3)I'm not sure is anaconda belong to python virtualenv?I installed nni in anaconda env.
4)~/.local/ 
![local](https://user-images.githubusercontent.com/12653453/51390619-be9db280-1b6a-11e9-9fc9-bcba17aa7cd3.png)
I don't find any nni directory in my anaconda path:
~/anaconda3/envs/tensorflow/lib/python3.6/site-packages

QuanluZhang:
@gonghuiyun If you use `--user` for the installation in virtualenv, nni will be installed under `~/.local/lib/python` rather than in your virtualenv. This is the property of python package installation. Due to a bug in nni v0.5, nnictl cannot find the nni installed under `~/.local/lib/python`. We will fix this bug very soon. **You can work around this bug by installing nni in virtualenv through the command** `python3 -m pip install --upgrade nni`.

gonghuiyun:
> @gonghuiyun If you use `--user` for the installation in virtualenv, nni will be installed under `~/.local/lib/python` rather than in your virtualenv. This is the property of python package installation. Due to a bug in nni v0.5, nnictl cannot find the nni installed under `~/.local/lib/python` in virtualenv. We will fix this bug very soon (#631). **You can workaround this bug by installing nni in virtualenv through the command** `python3 -m pip install --upgrade nni`.

It seems still don't work after I fix the command.

[...]

PR #631, opened 2019-01-18
Fix installation

Crysple:
there's two virtual environment ('virtualenv' and conda's virtual environment)
while in package 'virtualenv', site has no attribute 'getsitepackages()'
Fortunately, in 'virtualenv', there's an environment variable called VIRTUAL_ENV which specifies the location of the current python


Crysple:
related to #626

tools/nni_cmd/launcher.py
```diff
@@ -75,24 +75,36 @@ def start_rest_server(port, platform, mode, config_file_name, experiment_id=None
...
+        if python_sitepackage.startswith('/usr') or python_sitepackage.startswith('/Library'):
+            python_dir, entry_file = get_installation_path(0)
+        else:
+            python_dir, entry_file = get_installation_path(1)
         # Nothing is found
-        if not os.path.isfile(entry_file):
-            raise Exception('Fail to find nni under both "%s" and "%s"' % (local_python_dir, python_dir))
+        if python_dir is None:
+            raise Exception('Fail to find nni under python packages')
+    entry_dir = os.path.join(python_dir, 'nni')
```
SparkSnail: What is the purpose of entry_dir?

tools/nni_cmd/launcher.py
```diff
@@ -75,24 +75,36 @@ def start_rest_server(port, platform, mode, config_file_name, experiment_id=None
...
+    if os.getenv('VIRTUAL_ENV'):
+        # if 'virtualenv' package is used, site.getsitepackages is missed 
+        python_dir = os.getenv('VIRTUAL_ENV')
         entry_file = os.path.join(python_dir, 'nni', 'main.js')
-        entry_dir = os.path.join(python_dir, 'nni')
+    else:
+        # If system-wide python is used, we will give priority to using user-sitepackage given that nni exists there
+        python_sitepackage = site.getsitepackages()[0]
+        if python_sitepackage.startswith('/usr') or python_sitepackage.startswith('/Library'):
+            python_dir, entry_file = get_installation_path(0)
```
SparkSnail: I suggest only return "python_dir" in this function, and generate "entry_dir" and "entry_file" outside.

tools/nni_cmd/launcher.py
```diff
@@ -75,24 +75,40 @@ def start_rest_server(port, platform, mode, config_file_name, experiment_id=None
...
+            entry_file = os.path.join(python_dir, 'nni', 'main.js')
+            if os.path.isfile(entry_file):
+                return python_dir
+            return None
+
+        sitepackages = [site.getusersitepackages(), site.getsitepackages()[0]]
+        python_dir = _generate_installation_path(sitepackages[which_first])
+        if python_dir is not None:
+            return python_dir
+        return _generate_installation_path(sitepackages[1-which_first])
```
chicm-ms: Not easy to understand,  suggest to refactor this piece to make it more straightforward.     For example maybe remove the which_first param  and directly pass sitepackages...

Crysple: Good suggestion. Has modified. How about now?

chicm-ms: 1. It is much better now.  Since the start_rest_server is already too long.  It would be better if you extract the logic to acquire entry_file into a seperate function.
2. It seems the entry_dir is not used.

Crysple: - Good suggestion. Has move that piece of code to a new function
- `entry_dir` is used when invoking Popen in the next several lines so I kept it

PR #644, opened 2019-01-22
 Configurable nniManager log path and log level

chicm-ms:
1. Configure nniManager log path
2. Configure nniManager log level
3. Add --debug command line parameter to nnictl


tools/nni_cmd/config_schema.py
```diff
@@ -33,6 +33,8 @@
 Optional('multiPhase'): bool,
 Optional('multiThread'): bool,
 Optional('nniManagerIp'): str,
+Optional('logDir'): os.path.isdir,
+Optional('logLevel'): Or('debug', 'info', 'warning', 'error', 'critical'),
```
QuanluZhang: please update doc accordingly

PR #645, opened 2019-01-23
Fix broken links of Tuners and Assessor

PurityFan:


PurityFan:
The environment is chaotic

PR #652, opened 2019-01-23
Refactoring nnimanager log

chicm-ms:
1. change default log level from debug to info
2. Refactoring some nnimanager log info
3. Remove tensorboard code
4. Fix trace/fatal log level bug
5. Fix nnimanager setStatus minor issue


src/nni_manager/core/nnimanager.ts
```diff
@@ -148,6 +148,7 @@ class NNIManager implements Manager {
     }
 
     public async resumeExperiment(): Promise<void> {
+        this.log.info(`Resuming experiment: ${this.experimentProfile.id}`);
```
yds05: I suggest we go through other places to make sure all the logs follow correct log level.

chicm-ms: Gone through training services and database, more log updated.

PR #660, opened 2019-01-25
fix format of doc, change nni to NNI, yaml to yml.

squirrelsc:
fix some indents for translation tools.


docs/EnableAssessor.md
```diff
@@ -33,7 +33,7 @@ trial:
 For our built-in assessors, you need to fill two fields: `builtinAssessorName` which chooses NNI provided assessors (refer to [here]() for built-in assessors), `optimize_mode` which includes maximize and minimize (you want to maximize or minimize your trial result).
 
 ## Using user customized Assessor
-You can also write your own assessor following the guidance [here](). For example, you wrote an assessor for `examples/trials/mnist-annotation`. You should prepare the yaml configure below:
```
leckie-chn: suggest using YAML instead of yaml: https://en.wikipedia.org/wiki/YAML

docs/ExperimentConfig.md
```diff
@@ -1,7 +1,7 @@
 # Experiment config reference
 
 A config file is needed when create an experiment, the path of the config file is provide to nnictl.
-The config file is written in yaml format, and need to be written correctly.
```
leckie-chn: see above

PR #671, opened 2019-01-28
Update customized assessor doc

chicm-ms:


docs/Customize_Assessor.md
```diff
@@ -8,7 +8,7 @@ If you want to implement a customized Assessor, there are three things for you t
 
 1) Inherit a assessor of a base Assessor class
```
QuanluZhang: could you remove "-" from the top title?

chicm-ms: Thanks, updated.

PR #684, opened 2019-01-30
Update for reference

xuehui1991:
update docstring in tuner


leckie-chn:
@xuehui1991 I suggest having a look at `sdk_reference.html` to verify the effect of those docstring changes.

PR #707, opened 2019-02-01
Fix a race condition bug that does not store Trial Job cancel status correctly

yds05:
1. For PAITrainingService, we should rely on the status returned from PAI rest server to change trial job's status, rather than setting it's cancelling status in cancelTrialJob() method, because jobInfoCollector will then reset job's status based on PAI rest server's result. 
2. For RemoteMachineTrainingService, it almost the same, we should not set trial's status in cancelTrialJob() method.

### DIFF
--- a/src/nni_manager/training_service/pai/paiTrainingService.ts
+++ b/src/nni_manager/training_service/pai/paiTrainingService.ts
@@ -325,6 +325,9 @@ class PAITrainingService implements TrainingService {
             }
         };
 
+        // Set trialjobDetail's early stopped field, to mark the job's cancellation source
+        trialJobDetail.isEarlyStopped = isEarlyStopped;
+
         request(stopJobRequest, (error: Error, response: request.Response, body: any) => {
             if (error || response.statusCode >= 400) {
                 this.log.error(`PAI Training service: stop trial ${trialJobId} to PAI Cluster failed!`);
@@ -333,9 +336,6 @@ class PAITrainingService implements TrainingService {
                 deferred.resolve();
             }
         });
-
-        // Set trialjobDetail's early stopped field, to mark the job's cancellation source
-        trialJobDetail.isEarlyStopped = isEarlyStopped;
 
         return deferred.promise; 
     }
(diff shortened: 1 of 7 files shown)


src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -324,18 +324,19 @@ class PAITrainingService implements TrainingService {
...
             } else {
-                if (isEarlyStopped) {
-                    trialJobDetail.status = 'EARLY_STOPPED';
-                }
                 deferred.resolve();
             }
         });
 
+        // Set trialjobDetail's early stopped field, to mark the job's cancellation source
+        trialJobDetail.isEarlyStopped = isEarlyStopped;
```
QuanluZhang: I think this line has to be put at the beginning of this function.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -280,7 +280,9 @@ class RemoteMachineTrainingService implements TrainingService {
             const jobpidPath: string = this.getJobPidPath(trialJob.id);
             try {
                 await SSHClientUtility.remoteExeCommand(`pkill -P \`cat ${jobpidPath}\``, sshClient);
-                trialJob.status = getJobCancelStatus(isEarlyStopped);
+                //TODO: delete and move set USER_CANCELLED/EARLY_STOP in getTrialJob
+                // Mark the toEarlyStop tag here
+                trialJob.isEarlyStopped = isEarlyStopped;
```
QuanluZhang: this line should be put before the `await SSHClientUtility` (i.e., line 282)

src/nni_manager/training_service/pai/paiTrainingService.ts
yds05: Not very sure why we should put this line to the beginning. 
The beginning of this function has many condition checking. If we put this line to the beginning of this function but condition checking failed, the trial job status may be wrong.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
yds05: Done

src/nni_manager/training_service/pai/paiTrainingService.ts
QuanluZhang: agree, then we should put this line before line 328 `request(...`

yds05: Yes Sure.

PR #716, opened 2019-02-08
fix bug in smac that configspace cannot deal with complex searchspace

QuanluZhang:
in nni 0.5 and higher, mnist-annotation example cannot be tuned by SMAC tuner, because SMAC cannot parse categorical values which include ",() ". In this pr, we fixed this problem.

### DIFF
--- a/src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
+++ b/src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
@@ -53,7 +53,6 @@ def generate_pcs(nni_search_space_content):
                             pcs_fd.write('%s categorical {%s} [%s]\n' % (
                                 key, 
                                 json.dumps(list(range(choice_len)))[1:-1], 
-                                #json.dumps(search_space[key]['_value'][0])))
                                 json.dumps(0)))
                             if key in categorical_dict:
                                 raise RuntimeError('%s has already existed, please make sure no search space with the same key.' % key)
--- a/src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
+++ b/src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
@@ -41,6 +41,7 @@ def generate_pcs(nni_search_space_content):
     # parameter_name real [min_value, max_value] [default value] log
     # https://automl.github.io/SMAC3/stable/options.html
     '''
+    categorical_dict = {}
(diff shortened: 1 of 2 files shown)


src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
```diff
@@ -41,17 +41,22 @@ def generate_pcs(nni_search_space_content):
...
                         if search_space[key]['_type'] == 'choice':
+                            choice_len = len(search_space[key]['_value'])
                             pcs_fd.write('%s categorical {%s} [%s]\n' % (
                                 key, 
-                                json.dumps(search_space[key]['_value'])[1:-1], 
-                                json.dumps(search_space[key]['_value'][0])))
+                                json.dumps(list(range(choice_len)))[1:-1], 
+                                json.dumps(0)))
+                            if key in categorical_dict:
+                                raise RuntimeError('%s has already existed, please make sure no search space with the same key.' % key)
```
yds05: change 'the same' to 'duplicate' may be more precise for this case? like:
'please make sure search space has no duplicate key'

src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
```diff
@@ -41,17 +41,22 @@ def generate_pcs(nni_search_space_content):
...
     '''
+    categorical_dict = {}
     search_space = nni_search_space_content
     with open('param_config_space.pcs', 'w') as pcs_fd:
         if isinstance(search_space, dict):
             for key in search_space.keys():
                 if isinstance(search_space[key], dict):
                     try:
                         if search_space[key]['_type'] == 'choice':
+                            choice_len = len(search_space[key]['_value'])
```
yds05: Is it possible that user's search space has no 'choice' value?

src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
QuanluZhang: good suggestion, thanks

src/sdk/pynni/nni/smac_tuner/convert_ss_to_scenario.py
QuanluZhang: yes, it is. If there is no choice, no need to do convert for categorical in smac

PR #717, opened 2019-02-09
Improve the performance of the curve fitting assessor

PurityFan:
Mainly made the following changes to the logic of the assesor：
1. Change the artificial maxfev to the default value (the default value of the number of iterations maxfev is 100*(N+1))
2. Add the 'gap' hyperparameter, able to reduce the number of assessments to prevent blockage.
3. If the fitting and predicting time exceeds 60s, will print a WARNING log.


src/sdk/pynni/nni/curvefitting_assessor/curvefitting_assessor.py
```diff
@@ -69,21 +74,32 @@ def assess_trial(self, trial_job_id, trial_history):
...
         return AssessResult.Good or AssessResult.Bad
         '''
+        self.trial_job_id = trial_job_id
         self.trial_history = trial_history
         curr_step = len(trial_history)
         if curr_step < self.start_step:
             return AssessResult.Good
+        if (curr_step - self.start_step) / self.gap < self.judgment_num:
+            return AssessResult.Good
+        self.judgment_num = (curr_step - self.start_step) / self.gap
```
QuanluZhang: is self.judgment_num a float? Then it almost never runs line 83

PurityFan: Turn it into divisible

src/sdk/pynni/nni/curvefitting_assessor/curvefitting_assessor.py
```diff
@@ -100,21 +105,32 @@ def assess_trial(self, trial_job_id, trial_history):
...
             curvemodel = CurveModel(self.target_pos)
             predict_y = curvemodel.predict(trial_history)
             logger.info('Prediction done. Trial job id = ', trial_job_id, '. Predict value = ', predict_y)
             if predict_y is None:
                 logger.info('wait for more information to predict precisely')
                 return AssessResult.Good
             standard_performance = self.completed_best_performance * self.threshold
+
+            end_time = datetime.datetime.now()
+            if (end_time - start_time).seconds > 3600:
```
chicm-ms: It seems confusing,  3600s  == 60s?

PurityFan: Yes, 60s

src/sdk/pynni/nni/curvefitting_assessor/model_factory.py
```diff
@@ -110,7 +109,8 @@ def filter_curve(self):
         std = np.std(predict_data)
         for model in tmp_model:
             y = self.predict_y(model, self.target_pos)
-            if y < median + 3 * std and y > median - 3 * std:
+            epsilon = self.point_num / 10 * std
```
chicm-ms: Is it possible that std is zero?     ( For example,  np.std([5,5] is zero)
If yes,  you can add a very small number to 10*std.

PR #727, opened 2019-02-13
Fix pypi package

SparkSnail:
1. add nni_gpu_tool in setup.py
2. add exclude=['test'] in setup.py


deployment/pypi/setup.py
```diff
@@ -48,11 +48,12 @@
     long_description_content_type = 'text/markdown',
     license = 'MIT',
     url = 'https://github.com/Microsoft/nni',
-    packages = setuptools.find_packages('../../tools') + setuptools.find_packages('../../src/sdk/pynni')
+    packages = setuptools.find_packages('../../tools', exclude=['tests']) + setuptools.find_packages('../../src/sdk/pynni', exclude=['tests']),
```
leckie-chn: only need to exclude https://github.com/Microsoft/nni/tree/master/src/sdk/pynni/tests. 
So first `exclude=['tests']` should be deleted.

SparkSnail: removed the first `exclude=['tests']`

PR #733, opened 2019-02-14
fix some absolute links to relative.

squirrelsc:


PurityFan:
I have checked all the changed links, they can work.

PR #744, opened 2019-02-18
move docs under en_US

leckie-chn:


QuanluZhang:
The contributing part looks good.

PR #760, opened 2019-02-19
Add different tuner config files for config_test

leelaylay:
All test passed locally.


leelaylay:
All tuners except NetworkMorphism passes test because the port is not absolutely released in the integration test.

PR #764, opened 2019-02-20
update Chinese file path as en_US

squirrelsc:


docs/.gitignore
```diff
@@ -1 +1 @@
-*/_build
\ No newline at end of file
+**/_build
```
leckie-chn: what's the point of replacing * with ** here?

squirrelsc: ** means any folder depth, * means one level.

squirrelsc:
They are mapping with en_us version, when en_us version is removed, they can be removed together.

leckie-chn:
> They are mapping with en_us version, when en_us version is removed, they can be removed together.

@squirrelsc could you please just delete the mapping? After all, some of the documents is for developers, such as `deployment/docker/README.md`, which do not need a Chinese version.

PR #765, opened 2019-02-20
fix doc bug for referencing old version docstring

leckie-chn:
will be able to generate reference from the right docstring.


PurityFan:
I have tested this modification, the reference will automatically be modified with the modification of the source docstring.

docs/requirements.txt
```diff
@@ -4,4 +4,8 @@ sphinx-markdown-tables==0.0.9
 sphinx-rtd-theme==0.4.2
 sphinxcontrib-websupport==1.1.0
 recommonmark==0.5.0
-nni==0.5
\ No newline at end of file
+hyperopt
+json_tricks
```
xuehui1991: I wonder this package is for docs or for sdk?

leckie-chn: those dependencies is added so that Tuners can be successfully installed without installing NNI first.

PR #773, opened 2019-02-22
move readmes for tuners under docs directory

leckie-chn:
* also remove dependencies for smac
* will create link from tuner directory to docs


squirrelsc:
Some of tuner file has no special information, they can be removed.

leckie-chn:
> Some of tuner file has no special information, they can be removed.

@QuanluZhang said he will amplify those documents later.

PR #792, opened 2019-02-27
Update build pipeline

chicm-ms:
Manuall set version number


deployment/deployment-pipelines.yml
```diff
@@ -7,16 +26,45 @@ jobs:
     matrix:
       Python36:
         PYTHON_VERSION: '3.6'
+  variables:
+    valid_version: 'false'
 
   steps:
   - script: |
+      echo $(build_version)
+      if [[ $(build_version) =~ v[0-9](.[0-9])+ ]]; then
```
yds05: suggest to limit the number of .[0-9], like:
v[0-9](.[0-9]){1,3} is enough

chicm-ms: Thanks, updated.

PR #796, opened 2019-02-28
Set max number of trialConcurrency in remoteTrainingService

SparkSnail:


tools/nni_cmd/launcher_utils.py
```diff
@@ -209,12 +209,17 @@ def validate_annotation_content(experiment_config, spec_key, builtin_name):
                 exit(1)
             validate_search_space_content(experiment_config)
 
-def validate_machine_list(experiment_config):
-    '''Validate machine list'''
+def validate_remote_content(experiment_config):
+    '''Validate the content of remoteTrainingService'''
+    #node ssh client does not support large number of open channels for a connection, refer https://github.com/mscdex/ssh2/issues/219 
+    if experiment_config['trainingServicePlatform'] == 'remote' and experiment_config['trialConcurrency'] > 6:
```
chicm-ms: 1. Suggest to change the maximum value to 5
2.  It's better to describe the error in positive /neutral way, such as :  The maximum value of trialConcurrency for remote training service is 5.

QuanluZhang: 6 is too small, it is common that a server has 8 gpus. Is it possible to make it large?

leelaylay: I am wonder if it is possible to open just one ssh channel to upload(sftp) many files
in one queue. Just my suggestion. :-)
[sshClientUtility.ts](https://github.com/Microsoft/nni/blob/6d591989fe93c664d50d143d78474ad95ad09806/src/nni_manager/training_service/remote_machine/sshClientUtility.ts#L84)

yds05: @QuanluZhang  I discussed with Shinai offline. It's just a mitigation, and we have a final solution to work around this issue, but that requires bigger development effort and  will be sent out in a separate PR.

QuanluZhang: @yds05 got it, thanks

SparkSnail:
this issue is fixed in https://github.com/Microsoft/nni

Issue #801, opened 2019-03-02
Getting keyerror using annotation in a class


JohnAllen:
**Short summary about the issue/question**:

I'm getting this:
```
File "/home/xhaojing/.local/lib/python3.6/site-packages/nni/trial.py", line 53, in get_current_parameter
    return _params['parameters'][tag]
KeyError: 'Reward/failure_reward/choice
```
when I use an annotation in a class imported to my main training file.

This is the code that throws it:
```
class Reward:
    def __init__(self, data):
        self.Data = data
        """@nni.variable(nni.choice(-1, -0.5, -0.1, -0.01, -0.001, -0.0001), name=failure_reward)"""
        failure_reward = -.01
        self.failure_reward = failure_reward
```
`
**Brief what process you are following**: 
This is a local machine, 
**How to reproduce it**: 
Instantiate a class (that runs the code above) in a main training script like so:
`Reward = Reward(data)`

**nni Environment**:
- nni version: 0.5.1
- nni mode(local|pai|remote): local
- OS: Ubuntu 18
- python version: 3.6.7
- is conda or virtualenv used?: NO
- is running in docker?: NO

JohnAllen:
Figured it out.  Some files were in a separate dir from main training file and NNI was looking for `lib/some_annotated_var` and not `some_annotated_var`.  Moved files into same dir and everything worked.  

I would call it a bug though that you have to have files in the same dir.  Or what am I missing?

Crysple:
Hi @ @JohnAllen, thanks for reporting this.
It seems that the `inspect` library is running differently than expected. Originally we'd like to include it's path in it's key so there won't be two keys with the same name. I'll fix this soon. As you say, currently you can move them into the same dir as a temporary measure

chicm-ms:
Fixed.

scarlett2018:
@Crysple @yds05 - confirm to have this in v0.6.

PR #806, opened 2019-03-04
fix annotation key-error

Crysple:
### Overview

Related to #801 

When use annotation in a file in a sub-directory and import it from the main program, there's a search space key inconsistency between search space generator and NNI sdk. The latter one use `inspect` library to get module name in a wrong way (without importing path)

### Background

Note that NNI's annotation variable consists of three parts like:
`key = '{}/{}/{}'.format(module, name, function_type)`
where the first one is the module path of the file that uses annotation, the second one is the annotated variable name given by user, and the third one is the search space type like `choice` or `uniform`. Below we only consider the module part of its key.

### Comparison

For example, if the main program is `main.py` and users use annotation in `subpkg_main.py`, their key is: 

- NNI API:
  - Before fix: `subpkg_main`
  - After fix: `subpkg.subpkg_main`
- Search Space Generator:
  - Before fix: `subpkg.subpkg_main`
  - After fix: `subpkg.subpkg_main`
```
├── main.py
└── subpkg
    ├── __init__.py
    └──  subpkg_main.py
```


QuanluZhang:
@Crysple let's rethink the construction of the key.

QuanluZhang:
looks good

Crysple:
Waiting for further unit test modification

PR #807, opened 2019-03-04
Support version check of nni

SparkSnail:
1. check nni version in trialkeeper, to make sure the version of trialkeeper is consistent with trainingService
2. add a debug mode in config file


tools/nni_trial_tool/trial_keeper.py
```diff
@@ -103,6 +104,32 @@ def main_loop(args):
...
+    except:
+        #package nni does not exist
+        pass
+    if trial_keeper_version is None:
+        try:
+            trial_keeper_version = pkg_resources.get_distribution('nni-tool').version
+        except:
+            #package nni-tool does not exist
+            pass
+    if trial_keeper_version is None:
```
xuehui1991: "if not trial_keeper_version"？（pylint suggests...)

tools/nni_trial_tool/trial_keeper.py
```diff
@@ -103,6 +104,32 @@ def main_loop(args):
 def trial_keeper_help_info(*args):
     print('please run --help to see guidance')
 
+def check_version(args):
+    trial_keeper_version = None
+    try:
+        trial_keeper_version = pkg_resources.get_distribution('nni').version
+    except:
+        #package nni does not exist
```
xuehui1991: just catch an expected exception?   and log something here...

tools/nni_cmd/nnictl.py
```diff
@@ -34,7 +34,10 @@
 
 def nni_info(*args):
     if args[0].version:
-        print(pkg_resources.get_distribution('nni').version)
+        try:
+            print(pkg_resources.get_distribution('nni').version)
+        except:
+            print_error('Get version failed, please use `pip3 list | grep nni` to check nni version!')
```
xuehui1991: can you just catch the expected exception?

tools/nni_cmd/launcher.py
```diff
@@ -271,6 +271,8 @@ def set_experiment(experiment_config, mode, port, config_file_name):
         request_data['tuner'] = experiment_config['tuner']
         if 'assessor' in experiment_config:
             request_data['assessor'] = experiment_config['assessor']
+    if experiment_config.get('versionCheck') is not None:
```
xuehui1991: if experiment_config.get('versionCheck'):

tools/nni_trial_tool/trial_keeper.py
SparkSnail: changed to `if not trial_keeper_version`

tools/nni_trial_tool/trial_keeper.py
SparkSnail: Yes, changed to catch exception and log warning information.

tools/nni_cmd/nnictl.py
SparkSnail: Yes, changed to catch exception.

tools/nni_cmd/launcher.py
SparkSnail: experiment_config['versionCheck'] could be true, false or None, I need to set request_data values if it is not None. If change to `if experiment_config.get('versionCheck'):`, it will not set values when experiment_config['versionCheck'] is false.

xuehui1991: got it.

src/nni_manager/common/manager.ts
```diff
@@ -36,6 +36,7 @@ interface ExperimentParams {
     trainingServicePlatform: string;
     multiPhase?: boolean;
     multiThread?: boolean;
+    versionCheck?: boolean;
```
chicm-ms: Is it necessary to add a swith in config file for version check?

leelaylay:
It seems that `Building wheel for nni` has exceeded the maximum execution time of 60 in the pipeline.

src/nni_manager/common/manager.ts
SparkSnail: we will discuss this feature offline.

SparkSnail: change this param to be "debug" as discussed offline with @yds05

Issue #621, opened 2019-01-17
Mnist and mnist-annotation example download data failed in azure pipeline integration test env


chicm-ms:
This issue can be reproduced on azure pipeline, but can not be reproduced on local system.
Can be work around by change trial concurrency to 1:

run mnist or mnist-annotation on azure pipeline, set trial concurrency to 2,  the first trial is succeeded, but the second trial failed to download data.

[...]

PR #815, opened 2019-03-05
Fix tensorflow MNIST dataset download bug in multiple trials

leelaylay:
Fix for #621


chicm-ms:
Could you please update trialConcurrency to 2 and maxTrialNum to 4 in test/config_test/examples/mnist.test.yml and mnist-annotation.yml, so that this fix can be verified by IT?

leelaylay:
@chicm-ms  OK,  I will do it later.

PR #820, opened 2019-03-08
Support setting dispatcher  log level

SparkSnail:


src/sdk/pynni/nni/common.py
```diff
@@ -52,6 +67,20 @@ def write(self, s):
...
+    if env_args.log_level:
+        level_number = logLevelNameMap.get(env_args.log_level)
+        if level_number:
+            if level_number >= DEBUG:
+                log_level = logging.DEBUG
+            elif level_number >= INFO:
+                log_level = logging.INFO
+            elif level_number >= WARNING:
+                log_level = logging.WARNING
+            elif level_number >= ERROR:
```
chicm-ms: Suggest to simplify this section, just put logging.xxx into the map.

SparkSnail: Good suggestion, updated.

PR #822, opened 2019-03-08
Fix integration test dependencies

chicm-ms:


test/pipelines-it-kubeflow.yml
```diff
@@ -38,6 +38,16 @@ jobs:
...
 
+  - script: |
+      python3 -m pip install scikit-learn==0.20.0 --user
+      python3 -m pip install torch==0.4.1 --user
+      python3 -m pip install torchvision==0.2.1 --user
+      python3 -m pip install keras==2.1.6 --user
+      python3 -m pip install tensorflow-gpu==1.10.0 --user
+      sudo apt-get install swig -y
+      nnictl package install --name=SMAC
+    displayName: 'Install dependencies for integration tests'
```
yds05: Should we consider to add mode into the displyaName message, like 'Install dependencies for integration tests in Kubeflow mode'

test/pipelines-it-pai.yml
```diff
@@ -38,6 +38,16 @@ jobs:
...
 
+  - script: |
+      python3 -m pip install scikit-learn==0.20.0 --user
+      python3 -m pip install torch==0.4.1 --user
+      python3 -m pip install torchvision==0.2.1 --user
+      python3 -m pip install keras==2.1.6 --user
+      python3 -m pip install tensorflow-gpu==1.10.0 --user
+      sudo apt-get install swig -y
+      nnictl package install --name=SMAC
+    displayName: 'Install dependencies for integration tests'
```
yds05: The same concern as above.

test/pipelines-it-kubeflow.yml
chicm-ms: Thanks, updated.

test/pipelines-it-pai.yml
chicm-ms: Thanks, updated.

test/pipelines-it-kubeflow.yml
```diff
@@ -38,6 +38,16 @@ jobs:
       source install.sh
     displayName: 'Install nni toolkit via source code'
 
+  - script: |
```
Crysple: I remember I saw a practice somewhere, which says that it's better to use a shell script and run it in the configuration file instead, if there's too many commands in CI. For example, a sh or a requirements.txt might be better

yds05: My feeling is this snippet doesn't contain too many commands, and wrapping it inline into the yml is clean and easy to maintain. We can consider to extract it into a seperate shell file once the command list become too large, like more than 50 lines, e.g.

chicm-ms: I also found that it is better for debug and maintain to use seperate commands instead of sh file in pipeline. When the tasks are in one pipeline file, it is a good practice to eliminate dup code by refactoring tasks. In the future, I will look into how to reuse tasks across pipelines when it is necessary. 
 Thanks for discussion.

PR #826, opened 2019-03-11
Test if CI will fail given sdk's unittest fail

Crysple:
Just for test
Do not merge


chicm-ms:
Looks like the problem is on source unittest.sh,  the source command does not return exit code as we expected.

PR #833, opened 2019-03-12
Refactor pipeline

chicm-ms:
Move local integrate test from PR pipeline into a sepearate pipeline which is scheduled daily.


test/pipelines-it-local.yml
```diff
@@ -0,0 +1,38 @@
...
+- job: 'integration_test_local_ubuntu'
+
+  steps:
+  - script: python3 -m pip install --upgrade pip setuptools --user
+    displayName: 'Install python tools'
+  - script: |
+      source install.sh
+    displayName: 'Install nni toolkit via source code'
+  - script: |
+      python3 -m pip install scikit-learn==0.20.0 --user
```
xuehui1991: shall we put them in a script?

chicm-ms: Thanks, I will refactor this part together with other integration test pipelines later, use shared tasks or sh to eliminate dup code.

Issue #842, opened 2019-03-13
Support shmMB parameter in pai platform


SparkSnail:
PAI support setting shmMB in job parameter, NNI will support this configuration.
[refer](https://github.com/Microsoft/pai/blob/d26f762301f0de2efd263b84548f958b453d979b/docs/job_tutorial.md)

PR #847, opened 2019-03-13
Support setting shmMB in pai config

SparkSnail:
https://github.com/Microsoft/nni/issues/842


chicm-ms:
should we add shmMB and virtualCluster to this doc? https://github.com/Microsoft/nni/blob/master/docs/en_US/PAIMode.md

SparkSnail:
> should we add shmMB and virtualCluster to this doc? https://github.com/Microsoft/nni/blob/master/docs/en_US/PAIMode.md

Updated the document.

PR #849, opened 2019-03-13
fix typo in readme

codemayq:
use "完整"  instead of "完成" in this sentence and context.


squirrelsc:
Thanks for contribution, but translated content should be updated from below website. I updated as your suggestion, and it will be merged in next release.

https://crowdin.com/project/vs-ai/zh-CN#

Issue #867, opened 2019-03-18
Training service error: spawn /bin/sh EMFILE


BabakAp:
**Short summary about the issue/question**:
Long running jobs result in EMFILE error at TrialNum ~4000. It seems to be due to node leaking.

`lsof | grep node | wc -l`
> 44968

`lsof | grep node`
There are thousands of lines similar to this:

> node      15731                  user 106r      REG                8,1        0   4457298 /home/user/nni/experiments/TPuECjPB/trials/FTP4I/.nni/metrics

**Brief what process you are following**: 
1. Create experiment, running locally on an Ubuntu machine 
2. Let it run >4000 trials.

**How to reproduce it**: 

**nni Environment**:
- nni version: 0.5.2.1
- nni mode(local|pai|remote): local
- OS: Ubuntu 16.04.6 LTS
- python version: 3.6.8
- is conda or virtualenv used?:  conda
- is running in docker?: No

**Anything else we need to know**:

SparkSnail:
This issue may caused by using stream to read files when a trial is running, and this stream is not closed when a trial job is finished. Fix this issue in https://github.com/Microsoft/nni/pull/885

BabakAp:
I ran experiments in the same environment, with both the git master and 0.6 release. The problem still happens at ~4000 trials.
The number of open files steadily increases as an experiment runs. The program is keeping the metrics file of _every_ trial open (lsof | grep node shows thousands of lines like this: /experiments/{experiment ID}/trials/{trial ID}/.nni/metrics). When it reaches the number of open files limit set by the OS, the error happens again.

SparkSnail:
Hi, this error is fixed in https://github.com/microsoft/nni/pull/1189, and I've verified that it works on my environment.

BabakAp:
I can confirm that with nni 0.9.1 the issue is not reproducible in the same environment. Thank you for the great work!

PR #885, opened 2019-03-19
Fix localTrainingService stream

SparkSnail:
fix issue https://github.com/Microsoft/nni/issues/867


src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -345,6 +345,17 @@ class LocalTrainingService implements TrainingService {
...
             this.onTrialJobStatusChanged(trialJob, oldStatus);
+            //if job is not running, destory job stream
+            if(['SUCCEEDED', 'FAILED', 'USER_CANCELED', 'SYS_CANCELED', 'EARLY_STOPPED'].includes(newStatus)) {
+                if(this.jobStreamMap.has(trialJob.id)) {
+                    const stream = this.jobStreamMap.get(trialJob.id);
+                    if(!stream) {
+                        throw new Error(`Could not find stream in trial ${trialJob.id}`);
+                    }
+                    stream.destroy();
+                    this.jobStreamMap.delete(trialJob.id);
```
chicm-ms: This is not a good place to destroy job stream ,  it is better to put it into onTrialJobStatusChanged.

SparkSnail: moved to onTrialJobStatusChanged.

src/nni_manager/training_service/local/localTrainingServiceForGPU.ts
```diff
@@ -78,6 +78,17 @@ class LocalTrainingServiceForGPU extends LocalTrainingService {
...
     protected onTrialJobStatusChanged(trialJob: LocalTrialJobDetailForGPU, oldStatus: TrialJobStatus): void {
+        //if job is not running, destory job stream
+        if(['SUCCEEDED', 'FAILED', 'USER_CANCELED', 'SYS_CANCELED', 'EARLY_STOPPED'].includes(trialJob.status)) {
+            if(this.jobStreamMap.has(trialJob.id)) {
+                const stream = this.jobStreamMap.get(trialJob.id);
+                if(!stream) {
+                    throw new Error(`Could not find stream in trial ${trialJob.id}`);
+                }
+                stream.destroy();
+                this.jobStreamMap.delete(trialJob.id);
```
chicm-ms: I assume this is not specific for GPU,   looks like should move to parent class.

SparkSnail: moved to parent class

PR #887, opened 2019-03-20
Update Trial's doc and move HowToDebug to Tutorial

Crysple:
Add output_dir in trial's doc

### DIFF
--- a/docs/en_US/Trials.md
+++ b/docs/en_US/Trials.md
@@ -130,7 +130,7 @@ useAnnotation: true
 
 In NNI, every trial has a dedicated directory for them to output their own data. In each trial, an environment variable `NNI_OUTPUT_DIR` is exported. Under this directory, you could find each trials code, data and other possible log. In addition, each trial's log will be re-directed to a file name `trial.log` under that directory.
 
-Besides, trial's converted code is in another temporary directory if NNI Annotation is used. You can check that in a file name `run.sh` under the directory indicated by `NNI_OUTPUT_DIR`. The second line of this file will change directory to the actual directory where code is located.
+Besides, trial's converted code is in another temporary directory if NNI Annotation is used. You can check that in a file named `run.sh` under the directory indicated by `NNI_OUTPUT_DIR`. The second line of this file will change directory to the actual directory where code is located.
 
 For more information, please refer to [HowToDebug](HowToDebug.md)
 
--- a/docs/en_US/Trials.md
+++ b/docs/en_US/Trials.md
@@ -125,6 +125,15 @@ In the YAML configure file, you need to set *useAnnotation* to true to enable NN
 useAnnotation: true
 ```
 
+
+## Where is my trials?
+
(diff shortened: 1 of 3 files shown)


docs/en_US/Trials.md
```diff
@@ -125,6 +125,28 @@ In the YAML configure file, you need to set *useAnnotation* to true to enable NN
 useAnnotation: true
 ```
 
+
+## Where are my trials?
+
+In NNI, every trial has a dedicated directory for them to output their own data. In each trial, an environment variable called `NNI_OUTPUT_DIR` is exported. Under this directory, you could find each trials code, data and other possible log. In addition, each trial's log will be re-directed to a file name `trial.log` under that directory.
+
+Besides, trial's converted code is in another temporary directory if NNI Annotation is used. You can check that in a file named `run.sh` under the directory indicated by `NNI_OUTPUT_DIR`. The second line of this file will change directory to the actual directory where code is located. Below is an example of `run.sh`:
```
SparkSnail: In local mode, trial.log and run.sh is in `NNI_OUTPUT_DIR`. But in remote, pai and kubeflow mode, `NNI_OUTPUT_DIR` specifies the trial's output directory in remote machines,  and NNI may not copy run.sh and code files to this folder, and trial.log is transmitted to local machine in ~/nni/experiments/{expid}/log/trial_{id}.log.

docs/en_US/Tutorials.rst
```diff
@@ -12,4 +12,5 @@ Tutorials
     WebUI
     Training Platform<training_services>
     How to use docker<HowToUseDocker>
-    advanced
\ No newline at end of file
+    advanced
```
xuehui1991: I think we could change "advanced" to "Advanced"

docs/en_US/Trials.md
```diff
@@ -125,6 +125,28 @@ In the YAML configure file, you need to set *useAnnotation* to true to enable NN
...
+```shell
+#!/bin/bash
+cd /tmp/user_name/nni/annotation/tmpzj0h72x6 #This is the actual directory
+export NNI_PLATFORM=local
+export NNI_SYS_DIR=/home/user_name/nni/experiments/K1unkStq/trials/nrbb2
```
leelaylay: K1unkStq? nrbb2?
I suggest that we should replace it with $Experiment_ID$ and $Trial_ID$.

Crysple: Well, I'll fix this

docs/en_US/Tutorials.rst
Crysple: Well, it is related to the name of that specific rst file so if I change it here, I'll have to rename that file. Moreover, there are still some files whose name's first letter is lowercase, do you think it's necessary to change them all?

docs/en_US/Trials.md
Crysple: I know, but it's really hard to say cuz there's no such easily-expressed rules clear enough to a new user....
So I might put a constrain to the above explanation that it is only applicable in local mode. And in other modes, this environmental variable only stands for trial's output directory ?

docs/en_US/Tutorials.rst
xuehui1991: Or we could discuss in the scrum and then make a decision.

Issue #843, opened 2019-03-13
Enable ealy stopped trials data for tuners.


scarlett2018:
If a trial is early stopped, no metric data is sent to tuner. Some users also want tuner to know the metric data of those early stopped trials.

1. NNImanager find the last metric data when a trial is killed, and send this metric data to tuner. Let webui show the last intermediate result as final result.
2. Two ways: 
    - Advisor with functionalities of tuner and assessor; 
    - Also send early stop final state with final intermediate data to tuner.
3. Tuner tolerates the duplicate messages of early stop.

Issue #841, opened 2019-03-13
Tuner and assessor compete for the same message queue, assessor may starve tuner.


scarlett2018:
1. Add a warning message (easy to find this message) if the pipe queue is too long.
2. Two pipe, one for tuner and one for assessor.
3. One working thread for tuner, and one working thread for assessor.

PR #891, opened 2019-03-20
Route tuner and assessor commands to 2 seperate queues

chicm-ms:
1. Route tuner and assessor commands to 2 seperate queues  https://github.com/Microsoft/nni/issues/841
2. Allow tuner to leverage intermediate result when trial is early stopped.  https://github.com/Microsoft/nni/issues/843


src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -157,7 +153,21 @@ def handle_trial_end(self, data):
                 self.assessor.trial_end(trial_job_id, data['event'] == 'SUCCEEDED')
         return True
 
+    def _handle_final_metric_data(self, data):
+        '''
+        Call tuner to process final results
```
xuehui1991: I think we could format this docstring~

src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -157,7 +153,21 @@ def handle_trial_end(self, data):
...
+        id_ = data['parameter_id']
+        value = data['value']
+        if id_ in _customized_parameter_ids:
+            self.tuner.receive_customized_trial_result(id_, _trial_params[id_], value)
+        else:
+            self.tuner.receive_trial_result(id_, _trial_params[id_], value)
+
     def _handle_intermediate_metric_data(self, data):
+        '''
+        Call assessor to process intermediate results
```
xuehui1991: same here. You could reference the docstring in xgboost.

src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -187,5 +197,21 @@ def _handle_intermediate_metric_data(self, data):
...
+            # notify tuner
+            _logger.debug('env var: NNI_INCLUDE_INTERMEDIATE_RESULTS: [%s]', os.environ.get('NNI_INCLUDE_INTERMEDIATE_RESULTS'))
+            if os.environ.get('NNI_INCLUDE_INTERMEDIATE_RESULTS') == 'true':
+                self._earlystop_notify_tuner(data)
         else:
             _logger.debug('GOOD')
+
+    def _earlystop_notify_tuner(self, data):
+        '''
+        Send last intermediate result as final result to tuner in case the
```
xuehui1991: same here.

src/sdk/pynni/nni/msg_dispatcher_base.py
```diff
@@ -19,72 +19,126 @@
...
+        '''
+        Process commands in command queues.
+        '''
+        while True:
+            try:
+                # set timeout to ensure self.stopping is checked periodically
+                command, data = command_queue.get(timeout=3)
+                try:
+                    self.process_command(command, data)
+                except Exception as e:
```
xuehui1991: shall we only catch the expected exception?

test/tuner_test.py
```diff
@@ -77,7 +77,7 @@ def run(dispatch_type):
     for dispatcher_name in dipsatcher_list:
         try:
             # sleep 5 seconds here, to make sure previous stopped exp has enough time to exit to avoid port conflict
-            time.sleep(5)
+            time.sleep(6)
```
xuehui1991: why change this? shall we change the comments at the same time? "# sleep 5 seconds here, to make sure previous stopped exp has enough time to exit to avoid port conflict"....

src/nni_manager/core/test/ipcInterfaceTerminate.test.ts
```diff
@@ -106,7 +106,7 @@ describe('core/ipcInterface.terminate', (): void => {
                 assert.ok(!procError);
                 deferred.resolve();
             },
-            2000);
+            5000);
```
xuehui1991: why change this?

docs/en_US/ExperimentConfig.md
```diff
@@ -227,12 +227,17 @@ machineList:
...
+
+  * __gpuNum__
 
       __gpuNum__ specifies the gpu number to run the tuner process. The value of this field should be a positive number.
 
       Note: users could only specify one way to set tuner, for example, set {tunerName, optimizationMode} or {tunerCommand, tunerCwd}, and could not set them both.
 
+  * __includeIntermeidateResults__
+
+      If __includeIntermeidateResults__ is true, the last intermediate results of the trials early stopped by assessor are sent to tuner as final results. The default value of __includeIntermeidateResults__ is false.
```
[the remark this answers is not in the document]
chicm-ms: This name is what we have for now as a discussion result in Teams,  it was proposed by Mao and Chi, if we can find a better one, it is still can be changed before release.

src/sdk/pynni/nni/msg_dispatcher.py
chicm-ms: Thanks , done.

src/sdk/pynni/nni/msg_dispatcher.py
chicm-ms: Thanks, updated.

src/sdk/pynni/nni/msg_dispatcher.py
chicm-ms: Thanks, updated.

src/sdk/pynni/nni/msg_dispatcher_base.py
chicm-ms: The process_command calls Tuner or Assessor code, we can not know the exception types it will raise,  if we do not capture them, they will be raise to nowhere because this code path is in a worker thread.

test/tuner_test.py
chicm-ms: Thanks, updated the comments.
The reason to change this is:  after this tuner/assessor queue change, if there is no commands in queues, it will still possible to wait up to 3 seconds after dispatcher receives TERMINATE command. So it may take a little bit longer time to end for a normal exepriment.

src/nni_manager/core/test/ipcInterfaceTerminate.test.ts
chicm-ms: Same reason as above to accommodate the queue change.

docs/en_US/ExperimentConfig.md
```diff
@@ -227,12 +227,17 @@ machineList:
...
+
+  * __gpuNum__
 
       __gpuNum__ specifies the gpu number to run the tuner process. The value of this field should be a positive number.
 
       Note: users could only specify one way to set tuner, for example, set {tunerName, optimizationMode} or {tunerCommand, tunerCwd}, and could not set them both.
 
+  * __includeIntermediateResults__
+
+      If __includeIntermediateResults__ is true, the last intermediate results of the trials early stopped by assessor are sent to tuner as final results. The default value of __includeIntermediateResults__ is false.
```
QuanluZhang: the last intermediate result of the trial that is early stopped by assessor is sent to tuner as final result.

chicm-ms: Thanks, updated.

src/sdk/pynni/nni/msg_dispatcher_base.py
```diff
@@ -19,72 +19,123 @@
...
+            if self.stopping and (_worker_fast_exit_on_terminate or command_queue.empty()):
+                break
+
+    def enqueue_command(self, command, data):
+        """Enqueue command into command queues
+        """
+        if command == CommandType.TrialEnd or (command == CommandType.ReportMetricData and data['type'] == 'PERIODICAL'):
+            self.assessor_command_queue.put((command, data))
+        else:
+            self.default_command_queue.put((command, data))
```
QuanluZhang: please make sure the queue put is thread-safe, because both main thread and assessor thread will put record to this queue

QuanluZhang:
looks good. One more thing is that according to our discussion, trial end event will also be sent to tuner, we can implement this function in the next pr.

src/sdk/pynni/nni/msg_dispatcher_base.py
chicm-ms: OK, merge for integration test, will check this later.

chicm-ms:
> looks good. One more thing is that according to our discussion, trial end event will also be sent to tuner, we can implement this function in the next pr.

OK

src/sdk/pynni/nni/msg_dispatcher_base.py
chicm-ms: Just checked:
Queue is thread safe.
https://docs.python.org/3/library/queue.html
The queue module implements multi-producer, multi-consumer queues. It is especially useful in threaded programming when information must be exchanged safely between multiple threads. The Queue class in this module implements all the required locking semantics. It depends on the availability of thread support in Python; see the threading module.

Reference:    https://github.com/python/cpython/blob/3.7/Lib/queue.py  self.not_full lock is used with put method.

PR #901, opened 2019-03-21
Support remoteLoggingType

SparkSnail:
If user set remoteloggingType in config file, log content will not be transmitted from trialkeeper


tools/nni_cmd/config_schema.py
```diff
@@ -36,6 +36,7 @@
 Optional('logDir'): os.path.isdir,
 Optional('debug'): bool,
 Optional('logLevel'): Or('trace', 'debug', 'info', 'warning', 'error', 'fatal'),
+Optional('remoteLoggingType'): Or('http', 'none'),
```
chicm-ms: Suggest not to put this on top level,  maybe under trial or traininig service , and rename it to loggingType or a better name.

SparkSnail: As discussed offline, we will change this param to `logCollection`

PR #905, opened 2019-03-22
export trial status to csv

leckie-chn:
using `nnictl trial export xxxxxxxx -f hello.csv`

test code pending discussion with @chicm-ms

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ setup(
         'hyperopt',
         'json_tricks',
         'numpy',
+        'pandas',
         'psutil',
         'pyyaml',
         'requests',
--- a/tools/nni_cmd/nnictl_utils.py
+++ b/tools/nni_cmd/nnictl_utils.py
@@ -18,13 +18,12 @@
 # DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+import csv
 import os
 import psutil
 import json
 import datetime
 import time
-
-import pandas as pd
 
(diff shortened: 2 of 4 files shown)


QuanluZhang:
please update doc or docstring accordingly

tools/nni_cmd/nnictl_utils.py
```diff
@@ -433,3 +436,38 @@ def monitor_experiment(args):
...
+    elif isinstance(reward, dict):
+        dict_ret = {**hparam, **reward}
+    else:
+        raise ValueError("Invalid finalMetricsData format: {}".format(trial_data['finalMetricData'][0]['data']))
+    return dict_ret
+
+def export_experiment_2csv(args):
+    """export experiment metadata to csv
+    :param: csv file path to export
+    """
```
QuanluZhang: it would be better to use the following format:
"""export experiment metadata to csv
    
    Parameters
    ----------
    args: string
        csv file path to export

    Returns
    -------
    null
    """

tools/nni_cmd/nnictl_utils.py
```diff
@@ -433,3 +436,38 @@ def monitor_experiment(args):
         except Exception as exception:
             print_error(exception)
             exit(1)
+
+
+def parse_trial_data(trial_data):
+    """output: Dict"""
+    hparam = json.loads(trial_data['hyperParameters'][0])['parameters']
```
QuanluZhang: how to deal with multi-phase enabled experiments?

leckie-chn: it's not taken into account yet, for intermediate result variate length sequence. I don't know how to convert to a fixed csv column yet.

setup.py
```diff
@@ -50,6 +50,7 @@ def read(fname):
         'hyperopt',
         'json_tricks',
         'numpy',
+        'pandas',
```
QuanluZhang: could you remove this line?

PR #906, opened 2019-03-22
Fix version check

SparkSnail:
There is one kind of version string like 'v0.5.2-gews11f', it is generated by installing from source code.
In current trialKeeper, use exact version match, and this version string will cause code break in msranni/nni image, because our offical image use clean number version.
Change the logic to fuzzy match, only match the main number of nni.


tools/nni_trial_tool/trial_keeper.py
```diff
@@ -109,22 +109,25 @@ def check_version(args):
...
     else:
-        nni_log(LogType.Info,  'NNI version is {}'.format(args.version))
+        regular = re.compile('v?(?P<version>[0-9](\.[0-9]){0,2}).*')
+        try:
+            trial_keeper_version = regular.search(trial_keeper_version).group('version')
+            nni_log(LogType.Info, 'trial_keeper_version is {0}'.format(trial_keeper_version))
+            training_service_version = regular.search(args.version).group('version')
+            nni_log(LogType.Info, 'training_service_version is {0}'.format(training_service_version))
+            if trial_keeper_version != training_service_version:
+                nni_log(LogType.Error, 'Version does match!')
```
chicm-ms: typo:  Version does match

tools/nni_trial_tool/trial_keeper.py
```diff
@@ -109,22 +109,25 @@ def check_version(args):
...
-        os._exit(1)
     else:
-        nni_log(LogType.Info,  'NNI version is {}'.format(args.version))
+        regular = re.compile('v?(?P<version>[0-9](\.[0-9]){0,2}).*')
+        try:
+            trial_keeper_version = regular.search(trial_keeper_version).group('version')
+            nni_log(LogType.Info, 'trial_keeper_version is {0}'.format(trial_keeper_version))
+            training_service_version = regular.search(args.version).group('version')
+            nni_log(LogType.Info, 'training_service_version is {0}'.format(training_service_version))
+            if trial_keeper_version != training_service_version:
```
chicm-ms: should we use match instead of search and equal?

SparkSnail: As offline discussed, we use match to get the number group, and judge whether the number of trial_keeper is equal to training_service.

PR #911, opened 2019-03-25
Update hyperband

PurityFan:
* Update Hyperband
1. Change s = s_max after the s = 0 iterations, let Hyperband run continuously to get better experiment result.
2. Change "STEPS" to "TRIAL_BUDGET" to make the arg more clear. Modified related docs.
3. Unified mnist.py in mnist-hyperband with mnist.


examples/trials/mnist-hyperband/mnist.py
```diff
@@ -207,38 +208,31 @@ def main(params):
...
-        'pool_size': 2,
-        'hidden_size': 1024,
-        'learning_rate': 1e-4,
-        'batch_size': 32}
-    return params
-
+def get_params():
+    ''' Get parameters from command line '''
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--data_dir", type=str, default='/tmp/tensorflow/mnist/input_data', help="data directory")
```
xuehui1991: I don't think we need to update like argparse. Because you only use the default value.

examples/trials/mnist-hyperband/mnist.py
```diff
@@ -207,38 +208,31 @@ def main(params):
...
-        params.update(RCV_PARAMS)
-        '''
-        If you use Hyperband, among the hyperparameters (i.e., key-value pairs) received by a trial, 
-        there is one more key called `STEPS` besides the hyperparameters defined by user. 
-        By using this `STEPS`, the trial can control how long it runs.
-        '''
-        params['batch_num'] = RCV_PARAMS['STEPS'] * 10
+        tuner_params = nni.get_next_parameter()
+        logger.debug(tuner_params)
+        tuner_params['batch_num'] = tuner_params['STEPS'] * 100
```
xuehui1991: I think change the 'STEPS' to other name is better.

examples/trials/mnist-hyperband/mnist.py
PurityFan: This is just to keep the other parts consistent with the MNIST example, then users can find the difference between the trial code examples(to know where and how to control the budget).

examples/trials/mnist-hyperband/mnist.py
PurityFan: Done, change to TRIAL_BUDGET by quanlu's suggestion.

examples/trials/mnist-hyperband/mnist.py
xuehui1991: I think it's easy code to understand and we don't have to change this part...

PurityFan: In addition to this non-uniformity, the unification here is mainly to **change the batch_num 10 times to 100 times**, otherwise the result of Hyperband example's is always lower than TPE/SMAC.

PurityFan: Different training codes can cause a lot of misunderstandings about Tuner's performance. I think this part of the version may be better than the original? (although it doesn't matter)

xuehui1991: If you want to change, could you consider to change all the mnist in examples?

PurityFan: Will start a new pr to do this.

PR #913, opened 2019-03-25
[WebUI] Default metric is latest or final result, !local, show trial log hyperlink or address

lvybriage:
In trial detail's table, default metric's value is trial's final result when trial's status is succeed, default metric's value is intermediate result's latest val when trial's status isn't succeed.
when user set logCollection, according value to decide to show log address or anchor.


chicm-ms:
Looks good for me, please check with Scarlett for the new column name "Intermediate"

lvybriage:
> Looks good for me, please check with Scarlett for the new column name "Intermediate"



> Looks good for me, please check with Scarlett for the new column name "Intermediate"

Ok.

PR #914, opened 2019-03-25
Fix remote integration test pipeline

SparkSnail:


test/remote_docker.py
```diff
@@ -21,17 +21,26 @@ def find_port():
         port = random.randint(10000, 20000)
     return port
 
+def find_wheel_package(dir):
+    '''Find the wheel package uploaded to this machine'''
+    for file_name in os.listdir(dir):
+        if file_name.endswith('.whl'):
```
yds05: Suggest to use more strict rule to match NNI wheel package, rather than simple .whl suffix.

SparkSnail: changed to use `regular = re.compile('^nni-.*\.whl$')`

Issue #903, opened 2019-03-21
Remove Tuner's unnecessary ClassArg requirements


PurityFan:
Currently Tuner has some ClassArg that is not necessary.

For example:
`tuner:
  builtinTunerName: Random
  classArgs:
    optimize_mode: maximize`

In fact, random doesn‘t need optimize_mode.
This option is required, that is to say, if you don't add this item, you will get an error.
This is neither reasonable nor affects the user experience. It is **recommended to change to optional**.

xuehui1991:
Good suggestion!

xuehui1991:
#916 will fix this issue. Since we don't need to know the optimize_mode in random, remove "optimize_mode" in Random tuner.

xuehui1991:
Already merge into the master branch.  Close this issue.

PR #916, opened 2019-03-26
Fix Issue #903

xuehui1991:
remove the parameter "optimize_mode" of Random tuner


xuehui1991:
@leelaylay  thank you very much!

PR #921, opened 2019-03-26
Add section of how to load best model for Network Morphism 

leelaylay:
1. Add doc of How to load the best model for Network Morphism (as well as a general method to save and load model)
2. Fix parameter size calculation bug related to this [commit](https://github.com/jhfjhfj1/autokeras/commit/64f87ea651094c46e2f7414d25e681626f796aeb).
3. Add `utils.py` for each trial example in `examples/trials/network_morphism`.


docs/en_US/networkmorphismTuner.md
```diff
@@ -54,6 +54,52 @@ net = build_graph_from_json(RCV_CONFIG)
 nni.report_final_result(best_acc)
 ```
 
+If you want to save and load the best model, the following methods are recommended.
```
xuehui1991: I recommend  we use bold font here...  "**load the best model**"

leelaylay: 👌

src/sdk/pynni/nni/networkmorphism_tuner/layers.py
```diff
@@ -254,7 +254,7 @@ def export_weights_keras(self, keras_layer):
         keras_layer.set_weights((self.weights[0].T, self.weights[1]))
 
     def size(self):
```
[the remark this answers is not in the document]
leelaylay: Original Conv2d layer parameter calculation is wrong. Please refer to [this](https://medium.com/@iamvarman/how-to-calculate-the-number-of-parameters-in-the-cnn-5bd55364d7ca).

The details of the pull request:
```diff
- return self.filters * self.kernel_size * self.kernel_size + self.filters
+ return (self.input_channel * self.kernel_size * self.kernel_size + 1) * self.filters
```

examples/trials/network_morphism/cifar10/cifar10_pytorch.py
```diff
@@ -27,8 +27,7 @@
 import torch.optim as optim
 import torchvision
 
-sys.path.append("../")
-from network_morphism import utils
+import utils
```
QuanluZhang: this is offical package or your local package? is it better to make it clear?

leelaylay: [`utils`](https://github.com/Microsoft/nni/blob/master/examples/trials/network_morphism/utils.py) is the local package in this example. I find that original code `cifar10_pytorch.py` works fine in local mode, but it doesn't work in the remote mode or PAI cluster because NNI will only copy the files in the same folder as `config.yml`.

QuanluZhang: how about `import .utils`?

QuanluZhang: ok, no problem here

Issue #931, opened 2019-03-27
"nnictl stop" show endTime is "N/A" when experiment's endtime is number


lvybriage:
**Short summary about the issue/question**:

![image](https://user-images.githubusercontent.com/35484733/55069426-fbcc7680-50be-11e9-8534-1c6aa4fd6d14.png)

**Brief what process you are following**: 
I run some experiments, when I want to stop them I find that their end time is "N/A".
But their status is "Done" and end time is a number. So it may be a bug.

**How to reproduce it**: 
 Run two experiment, when experiment's status is done, you can input nnictl stop to reproduce it.

**nni Environment**:
- nni version: v0.5.2
- nni mode(local|pai|remote): --
- OS: --
- python version: --
- is conda or virtualenv used?:  --
- is running in docker?: --

**Anything else we need to know**:

scarlett2018:
Seems already fixed in #934 .

PR #934, opened 2019-03-28
Fix nnictl experiment list command

SparkSnail:
fix issue https://github.com/Microsoft/nni/issues/931


tools/nni_cmd/nnictl_utils.py
```diff
@@ -31,24 +31,49 @@
      EXPERIMENT_MONITOR_INFO, TRIAL_MONITOR_HEAD, TRIAL_MONITOR_CONTENT, TRIAL_MONITOR_TAIL
 from .common_utils import print_normal, print_error, print_warning, detect_process
 
-def update_experiment_status():
+def get_experiment_time(port):
+    '''get the startTime and endTime of an experiment'''
+    response = rest_get(experiment_url(port), 20)
```
xuehui1991: shall we use the default value of the parameter, instead of using "20"?

SparkSnail: Fixed, changed to use `REST_TIME_OUT`.

PR #935, opened 2019-03-28
Fix for MetisTuner

leelaylay:
1. Remove the debug information which can cause error.
2. Increase 'alpha' parameter of GaussianProcessRegressor to make sure a positive definite matrix.
3. Increase the probability of exploration from 0.1 to 0.9 to select from next config.

### DIFF
--- a/src/sdk/pynni/nni/metis_tuner/metis_tuner.py
+++ b/src/sdk/pynni/nni/metis_tuner/metis_tuner.py
@@ -396,7 +396,7 @@ class MetisTuner(Tuner):
             if next_candidate is not None:
                 outputs = self._pack_output(next_candidate['hyperparameter'])
             else:
-                random_parameter = _rand_init(self.x_bounds, self.x_types, 1)[0]
+                random_parameter = _rand_init(x_bounds, x_types, 1)[0]
                 outputs = self._pack_output(random_parameter)
         self.history_parameters.append(outputs)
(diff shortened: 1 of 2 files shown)


src/sdk/pynni/nni/metis_tuner/metis_tuner.py
```diff
@@ -65,7 +65,7 @@ class MetisTuner(Tuner):
     """
 
     def __init__(self, optimize_mode="maximize", no_resampling=True, no_candidates=True,
-                 selection_num_starting_points=600, cold_start_num=10, exploration_probability=0.1):
+                 selection_num_starting_points=600, cold_start_num=10, exploration_probability=0.9):
```
QuanluZhang: why changing from 0.1 to 0.9, seems like a big change.

leelaylay: I talked with @xuehui1991 about it before. Metis Tuner will do more exploration with next_config.

QuanluZhang: is it better to expose this configuration to users? or you think 0.9 is a good choice for different applications.

leelaylay: Sure, I do think 0.9 is just a good choice.

Issue #953, opened 2019-04-03
Update document of nnictl


PurityFan:
1. Check if there are bugs in the requirements, especially required and optional
2. Add some example for a better user guide

PurityFan:
Fixed in #961

PR #961, opened 2019-04-07
Fix bug in NNICTLDOC and add several examples

PurityFan:
Fix issue #953

### DIFF
--- a/docs/en_US/NNICTLDOC.md
+++ b/docs/en_US/NNICTLDOC.md
@@ -50,22 +50,22 @@ nnictl support commands:
 
   * Examples
 
-    `create a new experiment with the port: 8080`
+    > create a new experiment with the default port: 8080
 
     ```bash
     nnictl create --config nni/examples/trials/mnist/config.yml
     ```
 
-    `create a new experiment with specified port of restful server`
+    > create a new experiment with specified port of restful server
 
     ```bash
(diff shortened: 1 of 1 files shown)


SparkSnail:
Looks good to me, thanks for the update!

docs/en_US/NNICTLDOC.md
```diff
@@ -47,13 +48,34 @@ nnictl support commands:
...
+    > create a new experiment with the default port: 8080
+
+    ```bash
+    nnictl create --config nni/examples/trials/mnist/config.yml
+    ```
+
+    > create a new experiment with specified port of restful server
+
+    ```bash
+    nnictl create --config nni/examples/trials/mnist/config.yml --port $8088
```
chicm-ms: why '$8088' instead of '8088' ?

PurityFan: In a regular expression "$" means replaceable, let the user know that this value is replaceable.

chicm-ms: '8088' is a literal value, it is weird to be used as a variable,  either use plain literal value '8088' ,  or use [PORT] here to imply it is a replacable variable to be consistent with other parts.

PurityFan: Remove "$", now use 8088 as an example, and use [experiment_id] to refer a replaceable value.

PR #967, opened 2019-04-09
Extract common functions in sdk ("extract_scalar_reward")

PurityFan:
Extract common functions "extract_scalar_reward" in sdk and update references in Tuner and Advisor.


examples/tuners/ga_customer_tuner/customer_tuner.py
```diff
@@ -24,6 +24,7 @@
 import numpy as np
 
 from nni.tuner import Tuner
+from nni.utils import *
```
chicm-ms: suggest to import only required function, avoid import *

PurityFan: Done

PR #969, opened 2019-04-10
update multi-phase doc

QuanluZhang:
Make multi-phase doc more clear.

### DIFF
--- a/docs/en_US/multiPhase.md
+++ b/docs/en_US/multiPhase.md
@@ -2,33 +2,20 @@
 
 Typically each trial job gets a single configuration (e.g., hyperparameters) from tuner, tries this configuration and reports result, then exits. But sometimes a trial job may wants to request multiple configurations from tuner. We find this is a very compelling feature. For example:
 
-1. Job launch takes tens of seconds in some training platform. If a configuration takes only around a minute to finished, running only one configuration in a trial job would be every inefficient. An appealing way is that a trial job requests a configuration and finished it, then requests another configuration and run. The extreme case is that a trial job can run infinite configurations. If you set concurrency to be for example 6, there would be 6 __long running__ jobs keeping trying different configurations.
+1. Job launch takes tens of seconds in some training platform. If a configuration takes only around a minute to finish, running only one configuration in a trial job would be every inefficient. An appealing way is that a trial job requests a configuration and finishes it, then requests another configuration and run. The extreme case is that a trial job can run infinite configurations. If you set concurrency to be for example 6, there would be 6 __long running__ jobs keeping trying different configurations.
 
 2. Some types of models have to be trained phase by phase, the configuration of next phase depends on the results of previous phase(s). For example, to find the best quantization for a model, the training procedure is often as follows: the auto-quantization algorithm (i.e., tuner in NNI) chooses a size of bits (e.g., 16 bits), a trial job gets this configuration and trains the model for some epochs and reports result (e.g., accuracy). The algorithm receives this result and makes decision of changing 16 bits to 8 bits, or changing back to 32 bits. This process is repeated for a configured times.
 
-The above cases can be supported by the same feature, i.e., multi-phase execution. To support those cases, basically a trial job should be able to request multiple configurations from tuner. Tuner is aware of whether two configuration requests are from the same trial job or different ones. Also in multi-phase a trial job can report multiple final results. is there restriction on sequence???
+The above cases can be supported by the same feature, i.e., multi-phase execution. To support those cases, basically a trial job should be able to request multiple configurations from tuner. Tuner is aware of whether two configuration requests are from the same trial job or different ones. Also in multi-phase a trial job can report multiple final results. 
 
-__To use multi-phase experiment, please follow below steps:__
+Note that, currently multi-phase only supports one pattern of calling `nni.get_next_parameter()` and `nni.report_final_result()`: __call the former one, then call the later one; and repeat this pattern__. If calling `nni.get_next_parameter()` multiple times consecutively, calling `nni.report_final_result()` after that is reporting the final result of the latest retrieved configuration by `nni.get_next_parameter()`.
 
-1. Implement nni.multi_phase.MultiPhaseTuner. For example, this [ENAS tuner](https://github.com/countif/enas_nni/blob/master/nni/examples/tuners/enas/nni_controller_ptb.py) is a multi-phase Tuner which implements nni.multi_phase.MultiPhaseTuner. While implementing your MultiPhaseTuner, you may want to use the trial_job_id parameter of generate_parameters method to generate hyper parameters for each trial job.
+__Write trial code which leverages multi-phase:__
 
-1. Set `multiPhase` field to `true`, and configure your tuner implemented in step 1 as customized tuner in configuration file, for example:
-
-    ```yaml
-    ...
-    multiPhase: true
-    tuner:
-      codeDir: tuners/enas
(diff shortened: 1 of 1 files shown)


docs/en_US/multiPhase.md
```diff
@@ -1,16 +1,14 @@
...
+2. Some types of models have to be trained phase by phase, the configuration of next phase depends on the results of previous phase(s). For example, to find the best quantization for a model, the training procedure is often as follows: the auto-quantization algorithm (i.e., tuner in NNI) chooses a size of bits (e.g., 16 bits), a trial job gets this configuration and trains the model for some epochs and reports result (e.g., accuracy). The algorithm receives this result and makes decision of changing 16 bits to 8 bits, or changing back to 32 bits. This process is repeated for a configured times.
 
-Multi-phase experiments refer to experiments whose trial jobs request multiple hyper parameters from tuner and report multiple final results to NNI.
+The above cases can be supported by the same feature, i.e., multi-phase execution. To support those cases, basically a trial job should be able to request multiple configurations from tuner. Tuner is aware of whether two configuration requests are from the same trial job or different ones. Also in multi-phase a trial job can report multiple final results. is there restriction on sequence???
```
chicm-ms: What is this?

squirrelsc: too much question marks here.

docs/en_US/multiPhase.md
```diff
@@ -1,36 +1,21 @@
 ## Create multi-phase experiment
```
squirrelsc: how about "Create multi-phase trial"?

docs/en_US/multiPhase.md
```diff
@@ -1,36 +1,21 @@
...
 
-Multi-phase experiments refer to experiments whose trial jobs request multiple hyper parameters from tuner and report multiple final results to NNI.
+The above cases can be supported by the same feature, i.e., multi-phase execution. To support those cases, basically a trial job should be able to request multiple configurations from tuner. Tuner is aware of whether two configuration requests are from the same trial job or different ones. Also in multi-phase a trial job can report multiple final results. 
 
-To use multi-phase experiment, please follow below steps:
+Note that, currently multi-phase only supports one pattern of calling `nni.get_next_parameter()` and `nni.report_final_result()`: __call the former one, then call the later one; and repeat this pattern__. If calling `nni.get_next_parameter()` multiple times consecutively, calling `nni.report_final_result()` after that is reporting the final result of the latest retrieved configuration by `nni.get_next_parameter()`.
```
squirrelsc: How about below?
Note that, **`nni.get_next_parameter()` and `nni.report_final_result()` must be called sequentially**. If `nni.get_next_parameter()` is called multiple times, and then `nni.report_final_result()` is called once, the result is associated to last configuration, which is retrieved from the last get_next_parameter call. So there there is no result associated to previous get_next_parameter calls. It may cause some multi-phase algorithm broken.

docs/en_US/multiPhase.md
```diff
@@ -40,4 +25,14 @@ To use multi-phase experiment, please follow below steps:
         # report final result somewhere for the parameter retrieved above
         nni.report_final_result()
         # ...
+    # ...
     ```
+To enable multi-phase, you should also add `multiPhase: true` in your experiment yaml configure file. If this line is not added, `nni.get_next_parameter()` would always return the same configuration. For all the built-in tuners/advisors, you can use multi-phase in your trial code without modification of tuner/advisor spec in the yaml configure file.
```
squirrelsc: it needs a title like below, or people may ignore the change.
__update experiment YAML configuration__

docs/en_US/multiPhase.md
```diff
@@ -40,4 +25,14 @@ To use multi-phase experiment, please follow below steps:
...
+
+__Write a tuner that leverages multi-phase:__
+
+Before writing a multi-phase tuner, we highly suggest you to go through  [Customize Tuner](https://nni.readthedocs.io/en/latest/Customize_Tuner.html). Different from writing a normal tuner, your tuner needs to inherit from `MultiPhaseTuner` (in nni.multi_phase_tuner). The key difference between `Tuner` and `MultiPhaseTuner` is that the methods in MultiPhaseTuner are aware of additional information, that is, `trial_job_id`. With this information, the tuner could know which trial is requesting a configuration, and which trial is reporting results. This information provides enough flexibility for your tuner to deal with different trials and different phases. For example, you may want to use the trial_job_id parameter of generate_parameters method to generate hyperparameters for a specific trial job.
+
+Of course, to use your multi-phase tuner, you should add `multiPhase: true` in your experiment yaml configure file.
```
squirrelsc: It's already said in previous content, and should be a bullet also.

docs/en_US/multiPhase.md
```diff
@@ -40,4 +25,14 @@ To use multi-phase experiment, please follow below steps:
...
+
+__Write a tuner that leverages multi-phase:__
+
+Before writing a multi-phase tuner, we highly suggest you to go through  [Customize Tuner](https://nni.readthedocs.io/en/latest/Customize_Tuner.html). Different from writing a normal tuner, your tuner needs to inherit from `MultiPhaseTuner` (in nni.multi_phase_tuner). The key difference between `Tuner` and `MultiPhaseTuner` is that the methods in MultiPhaseTuner are aware of additional information, that is, `trial_job_id`. With this information, the tuner could know which trial is requesting a configuration, and which trial is reporting results. This information provides enough flexibility for your tuner to deal with different trials and different phases. For example, you may want to use the trial_job_id parameter of generate_parameters method to generate hyperparameters for a specific trial job.
```
squirrelsc: How about add an example? That would be more clear and simpler.

docs/en_US/multiPhase.md
QuanluZhang: thanks for the suggestion. updated

docs/en_US/multiPhase.md
```diff
@@ -1,36 +1,21 @@
...
 
-To use multi-phase experiment, please follow below steps:
+Note that, currently multi-phase only supports one pattern of calling `nni.get_next_parameter()` and `nni.report_final_result()`: __call the former one, then call the later one; and repeat this pattern__. If calling `nni.get_next_parameter()` multiple times consecutively, calling `nni.report_final_result()` after that is reporting the final result of the latest retrieved configuration by `nni.get_next_parameter()`.
 
-1. Implement nni.multi_phase.MultiPhaseTuner. For example, this [ENAS tuner](https://github.com/countif/enas_nni/blob/master/nni/examples/tuners/enas/nni_controller_ptb.py) is a multi-phase Tuner which implements nni.multi_phase.MultiPhaseTuner. While implementing your MultiPhaseTuner, you may want to use the trial_job_id parameter of generate_parameters method to generate hyper parameters for each trial job.
+__Write trial code which leverages multi-phase:__
```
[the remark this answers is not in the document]
QuanluZhang: maybe i still do not make it clear. To use multi-phase, users do not have to implement a tuner. our built-in tuners all support multi-phase.

docs/en_US/multiPhase.md
QuanluZhang: updated

docs/en_US/multiPhase.md
QuanluZhang: yes, you are right, let's put this task in future release.

docs/en_US/multiPhase.md
QuanluZhang: Let's first simply highlight it :)

Issue #592, opened 2019-01-10
All trail jobs status stays on 'waiting' for long time on PAI platform 


chicm-ms:
All trail jobs status stays on 'waiting'  until all trial jobs are submitted on PAI training service, while trialConcurrency is large, such as 50, it may take a few minutes to submit all trial jobs.  During the period that trial jobs being submitted,  the status of trial jobs do not change.

The root cause is:
1. It is a little bit slow to submit trial job with pai training service.
2. Trial job status query logic and job submission logic in a same sequential loop.

chicm-ms:
Fixed.

PR #973, opened 2019-04-10
Pai training service uses job queue for submitting jobs

chicm-ms:
issue #592


src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -330,10 +206,10 @@ class PAITrainingService implements TrainingService {
             uri: `http://${this.paiClusterConfig.host}/rest-server/api/v1/user/${this.paiClusterConfig.userName}/jobs/${trialJobDetail.paiJobName}/executionType`,
             method: 'PUT',
             json: true,
-            body: {'value' : 'STOP'},
+            body: {value: 'STOP'},
             headers: {
-                "Content-Type": "application/json",
-                "Authorization": 'Bearer ' + this.paiToken
+                'Content-Type': 'application/json',
+                Authorization: `Bearer ${this.paiToken}`
```
QuanluZhang: one uses "'" and the other does not?

src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -478,41 +357,219 @@ class PAITrainingService implements TrainingService {
...
+            if (restServer.getErrorMessage) {
+                throw new Error(restServer.getErrorMessage);
+            }
+            await delay(3000);
+        }
+    }
+
+    private async submitJobLoop(): Promise<void> {
+        while (!this.stopping) {
+            while (this.jobQueue.length > 0) {
```
QuanluZhang: If does not check this.stopping within this while loop, it might induce a situation that the experiment cannot be stopped, because there are still a lot of trials waiting for submission.

chicm-ms: thanks, updated.

src/nni_manager/training_service/pai/paiTrainingService.ts
chicm-ms: This is updated per tslint hint, because there is '-' in Content-Type, must be quoted.

QuanluZhang: got it.

PR #976, opened 2019-04-11
Fix paiToken update logic

SparkSnail:
When update paiToken after experiment is started, do not throw error.


src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -529,7 +529,14 @@ class PAITrainingService implements TrainingService {
 
     private async statusCheckingLoop(): Promise<void> {
         while (!this.stopping) {
-            await this.updatePaiToken();
+            try{
+                await this.updatePaiToken();
```
QuanluZhang: if timeout, there is no log written.

SparkSnail: > if timeout, there is no log written.

when catch error, log it out here.

Issue #709, opened 2019-02-02
`nnictl --version` does not work with `make dev-install`


QuanluZhang:
In `make dev-install`, nni python package is not installed (using sym link instead), thus, `nnictl --version` cannot find nni's version.

leckie-chn:
new observation: after `make dev-install` using `nnictl --verison`:
```
nnictl --version

Traceback (most recent call last):

  File "/home/yann/.local/bin/nnictl", line 11, in <module>

    load_entry_point('nni', 'console_scripts', 'nnictl')()

  File "/home/yann/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 480, in load_entry_point

    return get_distribution(dist).load_entry_point(group, name)

  File "/home/yann/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2691, in load_entry_point

    return ep.load()

  File "/home/yann/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2322, in load

    return self.resolve()

  File "/home/yann/.local/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2328, in resolve

    module = __import__(self.module_name, fromlist=['__name__'], level=0)

ImportError: No module named 'nni_cmd'
```

leckie-chn:
seems related to this pypi issue: https://github.com/pypa/pip/issues/3160

leelaylay:
The master branch has not fixed it yet.

QuanluZhang:
will resolve this issue in v0.7

PR #977, opened 2019-04-11
Fix bugs of dev-install caused by composition of the two packages

Crysple:
PR for #709, which is related to an open [https://github.com/pypa/pip/issues/3160](issue) of pypi.
That is, editable install does not work with package_directory (or namespace or sub-package)

Currently we figure out a temporary remedy.
When users want to `dev-install`, we'll firstly create **soft links** to **all python package** and put them all in a newly built directory under the root directory. Then we modify `setup.py` and throw it into that directory and run `pip install` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,8 +111,7 @@ dev-install:
 
 .PHONY: uninstall
 uninstall:
-	-$(PIP_UNINSTALL) -y nni
-	-$(PIP_UNINSTALL) -y nnictl
+	-cd build && $(PIP_UNINSTALL) -y nni
 	-rm -rf $(NNI_PKG_FOLDER)
 	-rm -f $(BIN_FOLDER)/node
 	-rm -f $(BIN_FOLDER)/nnictl
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,18 @@ install-python-modules:
 .PHONY: dev-install-python-modules
 dev-install-python-modules:
 	#$(_INFO) Installing Python SDK $(_END)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -164,7 +164,18 @@ install-python-modules:
...
+	ln -sf ../tools/nni_annotation build/nni_annotation
+	ln -sf ../tools/nni_cmd build/nni_cmd
+	ln -sf ../tools/nni_trial_tool build/nni_trial_tool
+	ln -sf ../tools/nni_gpu_tool build/nni_gpu_tool
+	cp setup.py build/
+	cp README.md build/
+	sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' build/setup.py
+	sed -ie 's/src\/sdk\/pynni\/nni/nni/g' build/setup.py
+	sed -ie 's/tools\///g' build/setup.py
+	cd build && $(PIP_INSTALL) $(PIP_MODE) -e .
```
QuanluZhang: please update `make uninstall` and `make clean` accordingly

Crysple: Have added

Makefile
```diff
@@ -111,8 +111,7 @@ dev-install:
 
 .PHONY: uninstall
 uninstall:
-	-$(PIP_UNINSTALL) -y nni
-	-$(PIP_UNINSTALL) -y nnictl
+	-cd build && $(PIP_UNINSTALL) -y nni
```
QuanluZhang: please also remove the `build` directory.

Crysple: Have removed

PR #984, opened 2019-04-15
Chinese translation

squirrelsc:

PR #990, opened 2019-04-16
Format other final keys and fix bugs in hyper-parameter graph

lvybriage:
1. Decimal format of metrics other than default on the Web UI #984 
2. fix hyper-parameter graph upper limit and randint upper limit [0-value)
3. help user choose maximal or minimal if he set it in tuner classArgs optimize_mode field
4. remaining field > 0, too long set ...

### DIFF
--- a/src/webui/src/components/trial-detail/TableList.tsx
+++ b/src/webui/src/components/trial-detail/TableList.tsx
@@ -386,16 +386,23 @@ class TableList extends React.Component<TableListProps, TableListState> {
                         key: item,
                         width: 150,
                         render: (text: string, record: TableObj) => {
-                            return (
-                                <div>
-                                    {
-                                        record.acc
-                                            ?
-                                            record.acc[item]
-                                            :
-                                            '--'
+                            const temp = record.acc;
+                            let decimals = 0;
+                            let other = '';
+                            if (temp !== undefined) {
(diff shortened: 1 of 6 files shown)


src/webui/src/components/trial-detail/TableList.tsx
```diff
@@ -386,16 +386,23 @@ class TableList extends React.Component<TableListProps, TableListState> {
...
-                                            ?
-                                            record.acc[item]
-                                            :
-                                            '--'
+                            const temp = record.acc;
+                            let decimals = 0;
+                            let other = '';
+                            if (temp !== undefined) {
+                                if (temp[item].toString().indexOf('.') !== -1) {
+                                    decimals = temp[item].toString().length - temp[item].toString().indexOf('.') - 1;
```
chicm-ms: You can use one line to implement this block:
```
return (<div> record.acc[item] === undefined? '--' :  record.acc[item].toFixed(6)  </div>)
```

lvybriage: 如果改成这样，用户return的data是0.9这种，我们也会在末尾追加五个0呀。为了避免这种情况，所以在上面判断了位数。so...

chicm-ms: OK, got it, thanks!

Issue #920, opened 2019-03-26
Refactor dispatcher / trial job env var settings


chicm-ms:
**Short summary about the issue/question**:
Refactor dispatcher / trial job env var settings

**Brief what process you are following**: 

Currently dispatcher and trial job share a common environment variable settings _load_env_args in sdk common.py.  It is not straitfoward to understand which env var is used by dispatcher or trial job,  This piece should be refactored to make it clearer.

**How to reproduce it**: 

**nni Environment**:
- nni version:
- nni mode(local|pai|remote):
- OS:
- python version:
- is conda or virtualenv used?: 
- is running in docker?:

**Anything else we need to know**:

chicm-ms:
Fixed.

PR #993, opened 2019-04-17
Refactor env var

chicm-ms:
issue: https://github.com/Microsoft/nni/issues/920


QuanluZhang:
looks great, very elegant.

PR #1009, opened 2019-04-22
Fix bug bash of import data feature

PurityFan:
Solution: add a conversion function to convert the parameter into an acceptable format in TPE

### DIFF
--- a/src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
+++ b/src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
@@ -394,6 +394,9 @@ class HyperoptTuner(Tuner):
             _params = trial_info["parameter"]
             assert "value" in trial_info
             _value = trial_info['value']
+            if not _value:
+                logger.info("Useless trial data, value is %s, skip this trial data." %_value)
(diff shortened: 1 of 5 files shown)


src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
```diff
@@ -139,6 +139,27 @@ def json2vals(in_x, vals, out_y, name=ROOT):
         for i, temp in enumerate(in_x):
             json2vals(temp, vals[i], out_y, name + '[%d]' % i)
 
+def nni_params2tuner_params(in_x, parameter):
```
xuehui1991: the name of this function is a little weird? why we need the prefix "nni"?

PurityFan: I also confuse the name of this function. This function converts the form of hyperparameters in NNI into the form in TPE tuner. Do you have any suggestions for the name of this function?

PurityFan: Rename to "params2tuner_params"

PR #1010, opened 2019-04-23
Fix windows pypi pipeline and add win32 version

demianzhang:
1. Add Windows 32-bit version for source install and pip install.
2. Change the pypi pipeline script to powershell

Tested in virtual machine Windows 10 32-bit using virtualbox.


deployment/pypi/README.md
```diff
@@ -53,9 +53,10 @@ This is the PyPI build and upload tool for NNI project.
   ```
 
 * __How to build__
-
+ 
+  parameter `version_os` is used to choose Windows version 64-bit or 32-bit, `version_ts` is used to build for test or not. 
   ```bash
-  powershell ./install.ps1
+  powershell ./install.ps1 -version_os [64/32] -version_ts [$True/$False]
```
chicm-ms: Can you remove the -version_ts option from doc?   it could be change or removed from code.

demianzhang: I can not understand the change. Could you give more details?

chicm-ms: -version_ts option is an implementation detail,  should not be mentioned in document.

demianzhang: It decides whether to build pypi package for test or release. Should we remove the parameters -version_os and -version_ts?

PR #1025, opened 2019-04-27
Update check_ready_to_run in windows local mode

demianzhang:


tools/nni_gpu_tool/gpu_metrics_collector.py
```diff
@@ -25,9 +25,13 @@
 from xml.dom import minidom
 
 def check_ready_to_run():
-    #TODO check process in windows
     if sys.platform == 'win32':
-        return True
+        pgrep_output = subprocess.check_output('wmic process where "CommandLine like \'%nni_gpu_tool.gpu_metrics_collector%\' and name like \'%python%\'" get processId')
```
yds05: Could you describe the process and motivation of 'check ready to run'?

demianzhang: It will check whether the nni_gpu_tool.gpu_metrics_collector is running or not excluding the self process. if it return false, it will not continue to get information and write to gpu_metrics.

tools/nni_gpu_tool/gpu_metrics_collector.py
```diff
@@ -25,9 +25,13 @@
...
+        pidList = pgrep_output.decode("utf-8").strip().split()
+        pidList.pop(0) # remove the key word 'ProcessId'
+        pidList = list(map(int, pidList))
+        pidList.remove(os.getpid())
+        return len(pidList) == 0
     pgrep_output =subprocess.check_output('pgrep -fx \'python3 -m nni_gpu_tool.gpu_metrics_collector\'', shell=True)
```
chicm-ms: Can you use 'else' for  linux and mac logic?

tools/nni_gpu_tool/gpu_metrics_collector.py
yds05: Got it.

Issue #1037, opened 2019-04-30
Trial won't run under Windows when gpuNum is set to 1


0x8A63F77D:
**Short summary about the issue/question**:
The trial won't run under Windows when gpuNum is set to 1. The trail keeps waiting. When I set gpuNum to 0, it works normally.
**Brief what process you are following**: 

**How to reproduce it**: 
My own code didn't work. I tried on the mnist example and it didn't work if gpuNum = 1.

**nni Environment**:
- nni version:0.7
- nni mode(local|pai|remote):local
- OS:Windows 10 
- python version:3.7
- is conda or virtualenv used?: miniconda
- is running in docker?:no

**need to update document(yes/no)**:

**Anything else we need to know**:

demianzhang:
Thanks for raising this issue. Could you provide more details of your installation, nnimanager.log, trial stderr?  I can not reproduce this error now.

0x8A63F77D:
nnimanager.log
[2019-4-30 14:26:29] INFO [ 'Datastore initialization done' ]
[2019-4-30 14:26:29] INFO [ 'Rest server listening on: http://0.0.0.0:8080' ]
[2019-4-30 14:26:29] INFO [ 'Construct local machine training service.' ]
[2019-4-30 14:26:29] INFO [ 'RestServer start' ]
[2019-4-30 14:26:29] INFO [ 'RestServer base port is 8080' ]
[2019-4-30 14:26:33] INFO [ 'NNIManager setClusterMetadata, key: trial_config, value: {"command":"c:\\\\Miniconda3\\\\envs\\\\tensorflow\\\\python.exe mnist.py","codeDir":"D:\\\\Users\\\\0x8A63F77D\\\\Documents\\\\GitHub\\\\nni\\\\examples\\\\trials\\\\mnist\\\\.","gpuNum":1}' ]
[2019-4-30 14:26:33] INFO [ 'required GPU number is 1' ]
[2019-4-30 14:26:35] INFO [ 'Starting experiment: kDj198ea' ]
[2019-4-30 14:26:35] INFO [ 'Change NNIManager status from: INITIALIZED to: RUNNING' ]
[2019-4-30 14:26:35] INFO [ 'Add event listeners' ]
[2019-4-30 14:26:35] INFO [ 'Run local machine training service.' ]
[2019-4-30 14:26:36] INFO [ 'NNIManager received command from dispatcher: ID, ' ]
[...]

PR #1043, opened 2019-05-05
Fix Windows available gpu

demianzhang:
issue #1037 
For Windows, there may be inevitable processes occupying the gpu while gpu utilization is low. The change is to use all gpus in Windows rather than filtering out the gpus that no process running on. The risk is that OOM error may happen due to the gpu memory is not enough for the new job.


crackcomm:
It fixed the problem for me as described in #1037. Trials are now running.

Issue #1036, opened 2019-04-30
[bug] The error message is empty when failed to create kubeflow config.


gaocegege:
**Short summary about the issue/question**:

The error message is empty when failed to create kubeflow config.

**Brief what process you are following**: 

Using the kubeconfig to set up an experiment with kubeflow training platform. The output is:

```
ERROR: Failed! Error is: {}
```

The root cause of the issue is that the response [here](https://github.com/Microsoft/nni/blob/master/tools/nni_cmd/launcher.py#L242) is empty.

**How to reproduce it**: 

Create a local k8s cluster, and use it to create kubeflow experiments.

```bash
nnictl create --config ./examples/trials/mnist/config_kubeflow.yml -p 8081
```

**nni Environment**:
- nni version: master
- nni mode(local|pai|remote): kubeflow
- OS: Ubuntu
- python version: 3.6
- is conda or virtualenv used?: No
- is running in docker?: No

**need to update document(yes/no)**:

**Anything else we need to know**:

The response code of http://localhost:8081/api/v1/nni/experiment/cluster-metadata is 500, and the response body is empty, I think.

The request body is 

```
{'kubeflow_config': {'operator': 'tf-operator', 'apiVersion': 'v1alpha2', 'storage': 'nfs', 'nfs': {'server': '10.10.10.10', 'path': '/var/nfs/general'}}}
```

gaocegege:
Found the reason in manager log:

```
[4/30/2019, 9:58:32 AM] FATAL [ 'Mount NFS 10.10.10.10:/var/nfs/general to /home/gaocegege/nni/experiments/a2D9yNon/trials-nfs-tmp failed, error is ChildProcessError: Command failed: sudo mount 10.10.10.10:/var/nfs/general /home/gaocegege/nni/experiments/a2D9yNon/trials-nfs-tmp\nmount: /home/gaocegege/nni/experiments/a2D9yNon/trials-nfs-tmp: bad option; for several filesystems (e.g. nfs, cifs) you might need a /sbin/mount.<type> helper program.\n `sudo mount 10.10.10.10:/var/nfs/general /home/gaocegege/nni/experiments/a2D9yNon/trials-nfs-tmp` (exited with error code 32)' ]
```

While I cannot get anything in the nnictl output.

SparkSnail:
Hi, this error dues to your nfs path in config file is not valid, you should change your nfs path in config file to your own nfs server path, not 10.10.10.10 in example.

[...]

PR #1045, opened 2019-05-05
Show error message when setClusterMetadata in kubeflow and frameworkcontroller

SparkSnail:
Fix issue https://github.com/Microsoft/nni/issues/1036


src/nni_manager/training_service/kubernetes/frameworkcontroller/frameworkcontrollerTrainingService.ts
```diff
@@ -237,23 +237,29 @@ class FrameworkControllerTrainingService extends KubernetesTrainingService imple
...
+                        let nfsFrameworkControllerClusterConfig = <FrameworkControllerClusterConfigNFS>this.fcClusterConfig;
+                        await this.createNFSStorage(
+                            nfsFrameworkControllerClusterConfig.nfs.server,
+                            nfsFrameworkControllerClusterConfig.nfs.path
+                        );
+                    } 
+                } catch(error) {
+                    this.log.error(error);
+                    return Promise.reject(new Error(error));
+                }
```
chicm-ms: I assume the problem is caused by that the type of error is string?  if yes, the problem has been address by this PR, https://github.com/microsoft/nni/pull/1064 , it is not necessary to handle this issue case by case.

PR #1049, opened 2019-05-06
Dev nas interface -- document

QuanluZhang:


docs/en_US/GeneralNasInterfaces.md
```diff
@@ -0,0 +1,142 @@
...
+
+[__TODO__] Simple example of NAS on NNI.
+
+### Weight sharing
+
+Sharing weights among chosen architectures (i.e., trials) could speedup model search. For example, properly inheriting weights of completed trials could speedup the converge of new trials. One-Shot NAS (e.g., ENAS, Darts) is more aggressive, the training of different architectures (i.e., subgraphs) shares the same copy of the weights in full graph.
+
+![](../img/nas_weight_share.png)
+
+We believe weight sharing (transferring) plays a key role on speeding up NAS, while finding efficient ways of sharing weights is still a hot research topic. We provide a key-value store for users to store and load weights. Tuners and Trials use a provided KV client lib to access the storage.
```
gaocegege: I have a question here.

Is the feature invasive for user code? Do the users need to write the CRUD KV store logic in their code?

QuanluZhang: Yes, we provide CRUD KV interface to users, this is kind of lower level APIs. To make programming more convenient, we may provide higher level APIs which are wrappers of CRUD KV logic. The detailed design will be refactored during supporting one-shot NAS algorithms.

Issue #1061, opened 2019-05-08
[bug]Local TrainingService read gpu metrics file before it is created


SparkSnail:
**Short summary about the issue/question**:

**Brief what process you are following**: 

**How to reproduce it**: 

**nni Environment**:
- nni version: v0.7
- nni mode(local|pai|remote):
- OS: Linux
- python version:
- is conda or virtualenv used?: 
- is running in docker?:

**need to update document(yes/no)**:

**Anything else we need to know**:

szymonzmyslony:
Does anybody know what this might be caused by or how should I go around solving it?

SparkSnail:
This bug is caused because nni create gpu_metric file slower than reading process, so nni will read gpu_metrics file before it is created. NNI will catch this error, and the experiment will continue to run after the gpu_metrics file is created, but there will be an error message in nniManager.log. This bug will be fixed in pr https://github.com/microsoft/nni/pull/1068

PR #1068, opened 2019-05-10
Fix gpu detector in localTrainingService

SparkSnail:
Fix issue https://github.com/microsoft/nni/issues/1061


src/nni_manager/training_service/local/gpuScheduler.ts
```diff
@@ -100,12 +100,20 @@ class GPUScheduler {
...
+                if (cmdresult && cmdresult.stdout) {
+                    this.gpuSummary = <GPUSummary>JSON.parse(cmdresult.stdout);
+                } else {
+                    this.log.error('Could not get gpu metrics information!');
+                }
+            } else{
+                this.log.warning('gpu_metrics file does not exist!')
+            }
+        } catch(err) {
+            this.log.error(err);
```
chicm-ms: Please remove this new outside try catch if it is unnecessary.

SparkSnail: removed.

PR #1075, opened 2019-05-14
Support paiTrainingService on windows

SparkSnail:


scarlett2018:
For #1075

test/generate_ts_config.py
```diff
@@ -73,7 +73,7 @@ def update_training_service_config(args):
 
 def convert_command():
     '''convert command by platform'''
-    if sys.platform != 'win32':
+    if sys.platform != 'win32' or args.ts:
```
chicm-ms: why add this?

src/nni_manager/main.ts
```diff
@@ -169,4 +169,12 @@ process.on('SIGTERM', async () => {
...
         process.exit(hasError ? 1 : 0);
     }
+}
+
+process.on('SIGBREAK', async () => {
+    await cleanUp();
+})
+
+process.on('SIGTERM', async () => {
+    await cleanUp();
```
QuanluZhang: this has different logic from pr 1073

test/generate_ts_config.py
SparkSnail: this function is executed only if the platform if local, refactor logic,

src/nni_manager/main.ts
SparkSnail: fixed, changed to use logic in pr 1073

PR #1093, opened 2019-05-21
Fix smac import data

QuanluZhang:


src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -258,4 +262,45 @@ def generate_multiple_parameters(self, parameter_id_list):
...
+        for trial_info in data:
+            self.logger.info("Importing data, current processing progress %s / %s", _completed_num, len(data))
+            # simply validate data format
+            assert "parameter" in trial_info
+            _params = trial_info["parameter"]
+            assert "value" in trial_info
+            _value = trial_info['value']
+            if not _value:
+                self.logger.info("Useless trial data, value is %s, skip this trial data.", _value)
+                continue
```
xuehui1991: you could try to avoid "continue" here...

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -258,4 +262,45 @@ def generate_multiple_parameters(self, parameter_id_list):
...
+        """
+        _completed_num = 0
+        for trial_info in data:
+            self.logger.info("Importing data, current processing progress %s / %s", _completed_num, len(data))
+            # simply validate data format
+            assert "parameter" in trial_info
+            _params = trial_info["parameter"]
+            assert "value" in trial_info
+            _value = trial_info['value']
+            if not _value:
```
xuehui1991: when the _value is None?

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -258,4 +262,45 @@ def generate_multiple_parameters(self, parameter_id_list):
...
+                if key in self.loguniform_key:
+                    _params[key] = np.log(value)
+                elif key in self.categorical_dict:
+                    if value in self.categorical_dict[key]:
+                        _params[key] = self.categorical_dict[key].index(value)
+                    else:
+                        self.logger.info("The value %s of key %s is not in search space.", str(value), key)
+                        valid_entry = False
+                        break
+            if not valid_entry:
```
xuehui1991: Just skip the invalid entry?

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
QuanluZhang: when the trial is failed. or users fill None on this field.

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
QuanluZhang: why?

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
QuanluZhang: logged in line 291?

src/sdk/pynni/nni/batch_tuner/batch_tuner.py
```diff
@@ -88,13 +85,40 @@ def generate_parameters(self, parameter_id):
...
+        self.count = -1
+
+        _completed_num = 0
+        for trial_info in data:
+            logger.info("Importing data, current processing progress %s / %s", _completed_num, len(data))
+            # simply validate data format
+            assert "parameter" in trial_info
+            _params = trial_info["parameter"]
+            assert "value" in trial_info
+            _value = trial_info['value']
```
leckie-chn: local variable `_param` / `_value` will not be actually stored?

QuanluZhang: they are stored in `runhistory`, no need to store them again in this file.

src/sdk/pynni/nni/batch_tuner/batch_tuner.py
```diff
@@ -88,13 +85,40 @@ def generate_parameters(self, parameter_id):
...
+            a list of dictionarys, each of which has at least two keys, 'parameter' and 'value'
+        """
+        if len(self.values) == 0:
+            logger.info("Search space has not been initialized, skip this data import")
+            return
+
+        self.values = self.values[(self.count+1):]
+        self.count = -1
+
+        _completed_num = 0
```
chicm-ms: In NNI code, normally there is no underscore "_" at the begining of local variable names,  suggest to remove it for consistency.

src/sdk/pynni/nni/batch_tuner/batch_tuner.py
leckie-chn: but `runhistory` only exists in 'smac_tuner'.

QuanluZhang: yes, so the problem is?

leckie-chn: I still can't see a data structure in `batch tuner` that store the history data.

QuanluZhang: because batch tuner has already used the imported data, it is reflected on removed configurations from the configuration list/array.

Issue #1024, opened 2019-04-26
Website became very slowly when I run an experiment with many trials


lvybriage:
**Short summary about the issue/question**:
  WebUI became very slowly when I run an experiment with 2000 trials. A trial has 30 intermediate result.

**Brief what process you are following**: 
Run a minist experiment with 2000 trials.

**How to reproduce it**: 
Run a minist experiment with 2000 trials, and then watch the website.

**nni Environment**: /
- nni version: 0.7
- nni mode(local|pai|remote): local
- OS: linux
- python version: /
- is conda or virtualenv used?: /
- is running in docker?: /

**need to update document(yes/no)**: no

**Anything else we need to know**: /

lvybriage:
In pr #1095, I ensure that the detail page only render one graph rather than four graph render together.
And then I will take the feather #1106 to make the page more smooth.

PR #1095, opened 2019-05-21
Use scu to avoid unnecessary rendering

lvybriage:
also, use this._isMounted to check the component status.
For issue #1024

### DIFF
--- a/src/webui/src/components/TrialsDetail.tsx
+++ b/src/webui/src/components/TrialsDetail.tsx
@@ -27,6 +27,11 @@ interface TrialDetailState {
     entriesInSelect: string;
     searchSpace: string;
     isMultiPhase: boolean;
+    isTableLoading: boolean;
+    whichGraph: string;
+    hyperCounts: number; // user click the hyper-parameter counts
+    durationCounts: number;
+    intermediateCounts: number;
 }
 
 class TrialsDetail extends React.Component<{}, TrialDetailState> {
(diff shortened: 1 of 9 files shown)


src/webui/src/components/trial-detail/Intermeidate.tsx
```diff
@@ -204,28 +213,73 @@ class Intermediate extends React.Component<IntermediateProps, IntermediateState>
...
-        if (isFilter === true) {
-            const pointVal = this.pointInput !== null ? this.pointInput.value : '';
-            const minVal = this.minValInput !== null ? this.minValInput.value : '';
-            if (pointVal === '' && minVal === '') {
-                this.drawIntermediate(nextProps.source);
+    componentWillReceiveProps(nextProps: IntermediateProps, nextState: IntermediateState) {
+        const { isFilter, filterSource } = nextState;
+        const { whichGraph, source } = nextProps;
+        
+        if (whichGraph === '4') {
```
yds05: what does '4' mean in this context?

lvybriage: '4' means that user is looking at intermediate result graph. I'll just render intermediate graph rather then all graphs.

src/webui/src/components/TrialsDetail.tsx
```diff
@@ -241,13 +255,19 @@ class TrialsDetail extends React.Component<{}, TrialDetailState> {
...
                 break;
             case '50':
-                this.setState(() => ({ entriesTable: 50 }));
+                if (this._isMounted) {
+                    this.setState(() => ({ entriesTable: 50 }));
+                }
                 break;
             case '100':
-                this.setState(() => ({ entriesTable: 100 }));
+                if (this._isMounted) {
```
chicm-ms: It seems unnecessary to use 3 cases for 20, 50 and 100, these 3 cases can be combined into this.setState(() => ({ entriesTable: value }));

lvybriage: Ok, I'll change. Thanks!

Issue #902, opened 2019-03-21
[Improvement] Searchspace's type "randint" is better to support lower bound


PurityFan:
**Short summary about the issue/question**:
Currently, the search space in NNI only supports setting upper bound. 
If I want to set {batch_size: [32, 256]}, I can't choose randint, I have to use:
"batch_size":{"_type":"quniform", "_value":[32, 256, 1]}
But this will return a float, and you need to convert it to int in your code.

Suggest give the chose to the user, "randint" support lower bound, like 
{“_type”:”randint”,”_value”:[low, high]}
If you want to use the original one, you can set {“_type”:”randint”,”_value”:[0, high]}
This can enhance its generalization

**nni Environment**:
- nni version: 0.5.2

xuehui1991:
Good suggestion!

squirrelsc:
I need this also, for example in random forest, number of leaves in a node can be a range from 10 to 100, but I don't want it randomized to 0, 1, 2.

xuehui1991:
Actually, we might make a mistake about the search space type of "randint".  

"randint" Which means the variable value is a random integer in the range [0, upper). The semantics of this distribution is that there is **no more correlation in the loss function** between nearby integer values, as compared with more distant integer values.   

So I think "randint" is more like "choice". @squirrelsc @PurityFan  in your scenario, we might use quniform to instead.

PurityFan:
But the hyperparameter gets from quniform get is a **float**. 
So if user have to get a rand int from `[32, 256]`, use `arg['batch_size'] + 32`(by randint) or `int(arg['batch_size'])`(by quniform) is bothersome

squirrelsc:
other concern is the choice is not good for big range like [32, 256), it means more than 200 choices need to be written.

silverlining21:
> But the hyperparameter gets from quniform get is a **float**.
> So if user have to get a rand int from `[32, 256]`, use `arg['batch_size'] + 32`(by randint) or `int(arg['batch_size'])`(by quniform) may be bothersome

cant be more agree~

PR #1096, opened 2019-05-21
Chang to randint to support lower bound

xuehui1991:
fix issue:
https://github.com/microsoft/nni/issues/902


QuanluZhang:
Could you check how this modification affects other tuners and advisors?

docs/en_US/AnnotationSpec.md
```diff
@@ -36,8 +36,8 @@ There are 10 types to express your search space as follows:
 
 * `@nni.variable(nni.choice(option1,option2,...,optionN),name=variable)`
   Which means the variable value is one of the options, which should be a list The elements of options can themselves be stochastic expressions
-* `@nni.variable(nni.randint(upper),name=variable)`
-  Which means the variable value is a random integer in the range [0, upper).
+* `@nni.variable(nni.randint(lower, upper),name=variable)`
```
QuanluZhang: lower and upper should be integers? or could be float?

src/sdk/pynni/nni/utils.py
```diff
@@ -94,9 +96,32 @@ def convert_dict2tuple(value):
...
+                value = in_x[NodeType.VALUE]
+                value.append(1)
+
+                in_x[NodeType.TYPE] = 'quniform'
+                in_x[NodeType.VALUE] = value
+ 
+            elif in_x[NodeType.TYPE] == 'choice':
+                randint_to_quniform(in_x[NodeType.VALUE])
+            else:
+                pass
```
QuanluZhang: might better to remove else and pass

docs/en_US/AnnotationSpec.md
QuanluZhang: Could you also check whether the chosen value is integer or float?

src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
```diff
@@ -232,6 +232,8 @@ def update_search_space(self, search_space):
         search_space : dict
         """
         self.json = search_space
+        randint_to_quniform(self.json)
```
QuanluZhang: I checked smac, hyperband, and bohb, the same modification would work. Could add this line in them?

docs/en_US/AnnotationSpec.md
QuanluZhang: I asked this because users may want to know the value's type, for example, an integer cannot be used as array index. It would be better to explicitly tell users.

xuehui1991: The check may in another PR.

src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
xuehui1991: I think that owner add them would be better.

src/sdk/pynni/nni/utils.py
```diff
@@ -94,9 +96,30 @@ def convert_dict2tuple(value):
...
+                in_x[NodeType.TYPE] = 'quniform'
+                in_x[NodeType.VALUE] = value
+ 
+            elif in_x[NodeType.TYPE] == 'choice':
+                randint_to_quniform(in_x[NodeType.VALUE])
+        else:
+            for key in in_x.keys():
+                randint_to_quniform(in_x[key])
+    elif isinstance(in_x, list):
+        for _, temp in enumerate(in_x):
```
chicm-ms: Since the underscore ('_') is not used,  both the underscore and enumerate can be removed.

docs/en_US/SearchSpaceSpec.md
```diff
@@ -36,9 +36,9 @@ All types of sampling strategies and their parameter are listed here:
     - Anneal
     - Evolution
 
-* {"_type":"randint","_value":[upper]}
+* {"_type":"randint","_value":[lower, upper]}
```
chicm-ms: What  will happen if a user already have a search space using old randint? such as  {"_type":"randint","_value":[100]} ?
will the trial still can run?

xuehui1991: we cannot support old randint now.

chicm-ms: Got it,  if the existing trial with old randint can not run anymore,  we may want to describe this change on our release note.

QuanluZhang: how about backward compatibility? if users only write `upper`, the default value of `lower` is 0.

silverlining21:
I need this feature~

docs/en_US/SearchSpaceSpec.md
silverlining21: further more, randint support step could be much better,  
`
"_type": "randint"
"_value": [10, 20, 2]
`

xuehui1991:
@silverlining21 @QuanluZhang  Thanks for your suggestion! Will consider in next PR.

PR #1107, opened 2019-05-24
import finished trial data to tuner when experiment is resumed

QuanluZhang:


tools/nni_cmd/nnictl_utils.py
```diff
@@ -482,25 +461,25 @@ def export_trials_data(args):
...
             if args.type == 'json':
-                json_records = []
-                for trial in records:
-                    value = trial.pop('reward', None)
-                    trial_id =  trial.pop('id', None)
-                    json_records.append({'parameter': trial, 'value': value, 'id': trial_id})
-            with open(args.path, 'w') as file:
-                if args.type == 'csv':
-                    writer = csv.DictWriter(file, set.union(*[set(r.keys()) for r in records]))
+                f_obj = open(args.path, 'w')
```
xuehui1991: suggest using "with open(...) as file" to make sure you have the same style with line 279.

QuanluZhang: good suggestion.

tools/nni_cmd/nnictl_utils.py
```diff
@@ -482,25 +461,25 @@ def export_trials_data(args):
...
+                f_obj.write(response.text)
+                f_obj.close()
+            else:
```
leckie-chn: suggest using
```python
elif args.type=='csv':
...
else:
    raise ValueError("export file type not supported, please use json or csv instead")
```
 here

QuanluZhang: good

SparkSnail: Do not suggest to use raise ValueError(''), it is not friendly for users in command line terminal. Use print_error('') instead.

QuanluZhang: @SparkSnail thanks!

SparkSnail: > suggest using
> 
> ```python
> elif args.type=='csv':
> ...
> else:
>     raise ValueError("export file type not supported, please use json or csv instead")
> ```
> 
> here

Do not suggest to use `raise ValueError('')`, it is not friendly for users in command line terminal. Use `print_error('')` and `exit(1)` instead.

PR #1117, opened 2019-05-28
Update NNI on Windows doc

demianzhang:


docs/en_US/Installation.md
```diff
@@ -48,9 +48,9 @@ Currently we support installation on Linux, Mac and Windows(local, remote and pa
   you can install NNI as administrator or current user as follows:
 
   ```bash
-  git clone -b v0.7 https://github.com/Microsoft/nni.git
+  git clone -b v0.8 https://github.com/Microsoft/nni.git
```
xuehui1991: I wonder why we don't use "[Release]" instead?

docs/en_US/NniOnWindows.md
```diff
@@ -23,9 +23,11 @@ Set-ExecutionPolicy -ExecutionPolicy Unrestricted
 * __Install NNI through source code__
```
squirrelsc: It looks duplicated with homepage, you can add a link here only.

demianzhang: have fixed

PR #1125, opened 2019-05-28
re-organize links & Fix link err 

suiguoxin:
Re-organize links for detailed descriptions of the tuners and assessors, put link of the algorithm at the end of every "Suggested Scenario";
Fix link err in HpoComparision.md.


docs/en_US/CommunitySharings/HpoComparision.md
```diff
@@ -5,17 +5,17 @@ Comparison of Hyperparameter Optimization algorithms on several problems.
...
-All algorithms run in NNI local environment。
+- [Random Search](../BuiltinTuner.md)
+- [Grid Search](../BuiltinTuner.md)
+- [Evolution](../BuiltinTuner.md)
+- [Anneal](../BuiltinTuner.md)
+- [Metis](../BuiltinTuner.md)
+- [TPE](../BuiltinTuner.md)
+- [SMAC](../BuiltinTuner.md)
+- [HyperBand](../BuiltinTuner.md)
+- [BOHB](../BuiltinTuner.md)
```
QuanluZhang: please keep the previous version here.

suiguoxin: The previous version leads to 404.  The format **'xxx.md#position'** is not supported

xuehui1991: @suiguoxin if you change to “(../BuiltinTuner.md#BOHB)", it will work here.

suiguoxin: “(../BuiltinTuner.html#BOHB)" works.

suiguoxin: i have added a commit to fix it

xuehui1991: @suiguoxin Seems you should use ".md" instead of ".html"

suiguoxin: @xuehui1991  “(../xxx.md#position)" doesn't work since it will not be translated into 'xxx.html#position' automatically. So here we need to use ".html" directly.

xuehui1991: @suiguoxin  Can you open it?    [https://github.com/microsoft/nni/blob/master/docs/en_US/BuiltinTuner.md#Random](url)

xuehui1991: @suiguoxin Can you open this link? [https://github.com/microsoft/nni/blob/master/docs/en_US/BuiltinTuner.md#Random](url)

suiguoxin: delete #section from cross-file links to make links work in both readthedocs and github docs @xuehui1991 @QuanluZhang

PR #1147, opened 2019-06-03
a simple debug tool for general nas programming interface

QuanluZhang:
added a new nnictl command: `nnictl trial codegen <exp_id> --trial_id <trial_id>` to generate an independently executable (without nni) trial code for the trial <trial_id>.

Currently, this command is only for general nas programming interface, will expand this command to the whole nni annotation in future.


examples/trials/mnist-nas/mnist.py
```diff
@@ -9,7 +9,6 @@
 import tensorflow as tf
 from tensorflow.examples.tutorials.mnist import input_data
 
-import nni
 import operators as op
```
xuehui1991: please add config_test for v0.8 release

tools/nni_annotation/__init__.py
```diff
@@ -121,3 +135,22 @@ def _expand_file_annotations(src_path, dst_path):
                 raise RuntimeError(src_path + ' ' + '\n'.join(str(arg) for arg in exc.args))
             else:
                 raise RuntimeError('Failed to expand annotations for %s: %r' % (src_path, exc))
+
+def _generate_specific_file(src_path, dst_path, exp_id, trial_id, module):
+    with open(src_path) as src, open(dst_path, 'w') as dst:
+        try:
+            with open(os.path.expanduser('~/nni/experiments/%s/trials/%s/parameter.cfg'%(exp_id, trial_id))) as fd:
```
xuehui1991: can you make sure this path exist in windows?

tools/nni_annotation/specific_code_generator.py
```diff
@@ -0,0 +1,365 @@
...
+import ast
+import astor
+from nni_cmd.common_utils import print_warning
+
+# pylint: disable=unidiomatic-typecheck
+
+para_cfg = None
+prefix_name = None
+
+def parse_annotation_mutable_layers(code, lineno):
```
xuehui1991: I think you could add some unit-test to these functions after v0.8 release.
Because their logic is complex and hard to maintain by different people

tools/nni_annotation/specific_code_generator.py
```diff
@@ -0,0 +1,365 @@
...
+            elif k.id == 'optional_inputs':
+                assert not fields['optional_inputs'], 'Duplicated field: optional_inputs'
+                assert type(value) is ast.List, 'Value of optional_inputs should be a list'
+                var_names = [astor.to_source(var).strip() for var in value.elts]
+                chosen_inputs = para_cfg[mutable_id][mutable_layer_id]['chosen_inputs']
+                elts = []
+                for i in chosen_inputs:
+                    index = var_names.index(i)
+                    elts.append(value.elts[index])
+                #optional_inputs = ast.Dict(keys=var_names, values=value.elts)
```
xuehui1991: remove unused code.

tools/nni_annotation/specific_code_generator.py
QuanluZhang: good suggestion, will do that.

examples/trials/mnist-nas/mnist.py
QuanluZhang: sure

tools/nni_annotation/__init__.py
QuanluZhang: good point. Let me check how to fix

docs/en_US/GeneralNasInterfaces.md
```diff
@@ -1,4 +1,6 @@
-# General Programming Interface for Neural Architecture Search
+# General Programming Interface for Neural Architecture Search (experimental feature)
+
+_*This is an experimental feature, currently, we only implemented the general NAS programming interface. Weight sharing and one-shot NAS will be supported in the following releases._
```
xuehui1991: weight sharing already supported....

docs/en_US/GeneralNasInterfaces.md
```diff
@@ -24,6 +26,8 @@ When designing the following model there might be several choices in the fourth
 
 There are two ways to write annotation for this example. For the upper one, `input` of the function calls is `[[],[out3]]`. For the bottom one, `input` is `[[out3],[]]`.
 
+__Debugging__: We provided an `nnictl trial codegen` command to help debugging your code of NAS programming on NNI. If your trial with trial_id `XXX` in your experiment `YYY` is failed, you could run `nnictl trial codegen YYY --trial_id XXX` to generate an executable code for this trial under your current directory. With this code, you can directly run the trial command without NNI to check why this trial is failed. Basically, this command is to compile your trial code and replace the NNI NAS code with the real chosen layers and inputs.
```
xuehui1991: I have a question here.  Does it only generate one trial code? What if other trials (in this search space) may fail?

tools/nni_annotation/__init__.py
QuanluZhang: checked, it also works on windows.

docs/en_US/GeneralNasInterfaces.md
QuanluZhang: let me make it more clear...

tools/nni_annotation/specific_code_generator.py
```diff
@@ -0,0 +1,364 @@
...
+        if not fields['optional_inputs']:
+            optional_inputs = ast.List(elts=[])
+        inputs = ast.List(elts=[fixed_inputs, optional_inputs])
+
+        func_call.args.append(inputs)
+        node = ast.Assign(targets=[layer_output], value=func_call)
+        nodes.append(node)
+    return nodes
+
+def parse_annotation(code):
```
Crysple: Looks like this file shares lots of similar code with the code_generator？

QuanluZhang: yes, will do refactor in the next release, which might be your work :)

docs/en_US/GeneralNasInterfaces.md
QuanluZhang: updated doc

tools/nni_annotation/specific_code_generator.py
```diff
@@ -0,0 +1,364 @@
...
+
+para_cfg = None
+prefix_name = None
+
+def parse_annotation_mutable_layers(code, lineno):
+    """Parse the string of mutable layers in annotation.
+    Return a list of AST Expr nodes
+    code: annotation string (excluding '@')
+    """
+    global para_cfg
```
Crysple: It seems that there's need to declare these two variables as global variable if their value are not modified in this function.

tools/nni_annotation/specific_code_generator.py
Crysple: lol, haode

tools/nni_annotation/specific_code_generator.py
QuanluZhang: removed, thanks.

leckie-chn:
seems that there's no good example for using NAS interface with `nni.get_next_parameter`. At least I don't see any examples/docs for teaching users how to parse & generate network structure according to the output of tuner. Strongly suggest you add this to next release :)

QuanluZhang:
> seems that there's no good example for using NAS interface with `nni.get_next_parameter`. At least I don't see any examples/docs for teaching users how to parse & generate network structure according to the output of tuner. Strongly suggest you add this to next release :)

sorry for making this unclear, this NAS interface follows nni annotation rule, no need to use `nni.get_next_parameter`. I can think about how to make it more clear.

leckie-chn:
> > seems that there's no good example for using NAS interface with `nni.get_next_parameter`. At least I don't see any examples/docs for teaching users how to parse & generate network structure according to the output of tuner. Strongly suggest you add this to next release :)
> 
> sorry for making this unclear, this NAS interface follows nni annotation rule, no need to use `nni.get_next_parameter`. I can think about how to make it more clear. `nni.mutable_layers` not only expresses the neural architecture search space, but also indicate how each trial's neural architecture should be.

just out of curious. Did you considered the scenario of tuning architecture & hyper-parameter simultaneously? Does it mean that in this scenario the users can only use NNI annotation?

QuanluZhang:
> > > seems that there's no good example for using NAS interface with `nni.get_next_parameter`. At least I don't see any examples/docs for teaching users how to parse & generate network structure according to the output of tuner. Strongly suggest you add this to next release :)
> > 
> > 
> > sorry for making this unclear, this NAS interface follows nni annotation rule, no need to use `nni.get_next_parameter`. I can think about how to make it more clear. `nni.mutable_layers` not only expresses the neural architecture search space, but also indicate how each trial's neural architecture should be.
> 
> just out of curious. Did you considered the scenario of tuning architecture & hyper-parameter simultaneously? Does it mean that in this scenario the users can only use NNI annotation?

This is a good question. Tuning both architecture and hyper-parameter at the same time would be inevitable in real application scenarios. Yes, users can use NNI annotation in these scenarios. If they use `nni.get_next_parameter`, code might be unnecessarily complex because it means neural architecture is expressed through hyper-parameters, they have to parse the hyper-parameters to build the network first. We can have further discussion on this.

Issue #1106, opened 2019-05-24
Enhancement: ability to adjust rendering interval.


lvybriage:
**Background**
When running experiments that have 2000+ trials, the web page is overloaded for frequent rendering. We need the ability to adjust rendering interval (manually or automatically).

**Without this feature, how does current nni work**：slow render

**Components that may involve changes**: webui

**Brief description of your proposal if any**:--

yangou1988:
Add auto refresh setting drop-down & Refresh button  at the top navigation area.

- Users can click the drop down and choose auto refresh 10s, 30s, 60s or disable it.
- Users can also click the Refresh button to refresh manually.

overview
![Web 1920 – 2](https://user-images.githubusercontent.com/44491713/58446384-22e10c00-8133-11e9-996e-3d0853b67045.png)

Drop down
![image](https://user-images.githubusercontent.com/44491713/58446409-32f8eb80-8133-11e9-9f18-5b224f5156ac.png)

After secleted 'Refresh every 30s'
![image](https://user-images.githubusercontent.com/44491713/58446432-4efc8d00-8133-11e9-9a9e-53bb1d7cdfe2.png)

PR #1181, opened 2019-06-17
Enhancement: ability to adjust rendering interval 

lvybriage:
Fix issue #1106


src/webui/package.json
```diff
@@ -9,11 +9,12 @@
     "copy-to-clipboard": "^3.0.8",
     "echarts": "^4.1.0",
     "echarts-for-react": "^2.0.14",
-    "react": "^16.4.2",
-    "react-dom": "^16.4.2",
+    "react": "^16.7.0-alpha.2",
+    "react-dom": "^16.7.0-alpha.2",
```
chicm-ms: Is this upgrade necessary?  since they are alpha versions.

lvybriage: I've forgot why change it. I'll test with before package.json again.

lvybriage: Had changed back.

lvybriage: If don't use react alpha version, will met the problem "_react.default.useContext is not a function". I haven't found a better solution for it.

chicm: ok

chicm-ms: ok

PR #1185, opened 2019-06-20
fix negative time number in local mode when trial time is short

LeonardoWang:
fix negative time number in local mode when trial time is short， when a trial finish in 1 second, last 3 number of  finish time will be set "000", which is smaller than the start time


QuanluZhang:
@LeonardoWang should also add the corresponding modification on webui.

PR #1189, opened 2019-06-21
Fix EMFILE  error in local training service

SparkSnail:
Fix issue: https://github.com/microsoft/nni/issues/867
Refer: https://github.com/Juul/tail-stream/issues/20


src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -355,7 +355,8 @@ class LocalTrainingService implements TrainingService {
         this.log.info('Stopping local machine training service...');
         this.stopping = true;
         for (const stream of this.jobStreamMap.values()) {
-            stream.destroy();
+            stream.end(0)
```
QuanluZhang: could be better to add the github issue. what do you think @chicm-ms @SparkSnail ?

SparkSnail: As discussed with @chicm-ms , I've add related issue in code.

Issue #1192, opened 2019-06-23
hyper-parameter not shown correctly in WebUI when there is only one hyper parameter


suiguoxin:
<**Short summary about the issue/question**:
hyper-parameter not shown correctly in WebUI when there is only one parameter

**Brief what process you are following**: 
Run experiment where there is only one hyper parameter defined in search space. 

**How to reproduce it**: 

**nni Environment**:
- nni version: 0.8
- nni mode(local|pai|remote): local
- OS:
- python version: 3
- is conda or virtualenv used?:  conda
- is running in docker?: no

**need to update document(yes/no)**: no

**Anything else we need to know**:

PR #1200, opened 2019-06-24
Show default metric in hyper-parameter

lvybriage:
Fix issue #1192 
render：

无标题


src/webui/src/components/trial-detail/Para.tsx
```diff
@@ -213,6 +214,10 @@ class Para extends React.Component<ParaProps, ParaState> {
 
             }
         }
+        parallelAxis.push({
+            dim: i,
+            name: 'default metric'
+        });
```
xuehui1991: can we show other metrics effect (corresponding to different params)?

lvybriage: I'll create issue to follow up this.

PR #1201, opened 2019-06-24
Implement enas-mode and oneshot-mode for NAS interface

Crysple:
Related issue #1159, used to run ENAS

# API Changes

In the new version of NAS interface, users need to specify the NAS mode in their Config.yml for NNI under the `trial` section. Like this:
```diff
trial:
  command: sh ./macro_cifar10.sh
  codeDir: ./
  gpuNum: 1
+ #choice: classic_mode, enas_mode, oneshot_mode
+ nasMode: oneshot_mode
```

Currently there are three modes available:

- `classic_mode`: The same as the original version of NAS interface
- `enas_mode`: An efficient mode developed for tensorflow to run ENAS (introduced in the following)
- `oneshot_mode`: A mode for users to train a full graph of a search space, where all the inputs and operations will be chosen and summed together

# Different modes

### Overview
In tensorflow, users need to build graph first and then create a session to run that graph. In the previous version, the graph of a trial will be determined (to be a sub-graph) once it receives a parameter configuration from the tuner. That is to say, the graph of this trial will not change even if it receives other parameter configurations in the future.

So in this mode we will create and use tensorflow variable as signals, and tensorflow conditional functions to control the search space (full-graph) to be more flexible, which means it can be changed into different sub-graphs (multiple times) depending on these signals.

### API Changes

Users need to add `nni.get_next_parameter(session)` before they invoke the session.run function:

Note that they need to **pass their tensorflow session as an arg** into this function. An example might be:
```python
for _ in range(num):
    """@nni.get_next_parameter(self.session)"""
    loss, _ = self.session.run([loss_op, train_op])
    """@nni.report_final_result(loss)"""
```

## oneshot-mode
We provide a mode for users to train [oneshot model](http://proceedings.mlr.press/v80/bender18a/bender18a.pdf).
image
In this mode, every inputs and operations will be chosen. As suggested in the paper, a dropout method is implemented to the inputs for every layer. The dropout rate is set to r^(1/k), where 0 < r < 1 is a hyper-parameter of the model (default to 0.01) and k is number of optional inputs for a specific layer. The higher the fan-in, the more likely each possible input is to be dropped out. However, the probability of dropping out all optional_inputs of a layer is kept constant regardless of its fan-in. Suppose r = 0.05. If a layer has k = 2 optional_inputs then each one will independently be dropped out with probability 0.051/2 ≈ 0.22 and will be retained with probability 0.78. If a layer has k = 7 optional_inputs then each one will independently be dropped out with probability 0.051/7 ≈ 0.65 and will be retained with probability 0.35. In both cases, the probability of dropping out all of the layer's optional_inputs is 5%.


src/nni_manager/rest_server/restValidationSchemas.ts
```diff
@@ -51,6 +51,7 @@ export namespace ValidationSchemas {
                 command: joi.string().min(1),
                 virtualCluster: joi.string(),
                 shmMB: joi.number(),
+                nasMode: joi.string().valid('classic_mode', 'enas_mode', 'oneshot_mode'),
```
xuehui1991: what's the meaning of "enas-mode" or "one-shot mode"? do we have any doc about that?

QuanluZhang: is this nasMod field optional?

src/sdk/pynni/nni/nas_utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+def classic_mode(
+        mutable_id,
+        mutable_layer_id,
+        funcs,
+        funcs_args,
+        fixed_inputs,
+        optional_inputs,
+        optional_input_size):
+    '''Execute the chosen function and inputs directly'''
```
QuanluZhang: => '''Execute the chosen function and inputs directly. In this mode, the trial code is only running the chosen subgraph (i.e., the chosen ops and inputs), without touching the full model graph'''

src/nni_manager/rest_server/restValidationSchemas.ts
Crysple: Originally I set it to be optional so that the previous code can run. I'll check again.

Crysple: > what's the meaning of "enas-mode" or "one-shot mode"? do we have any doc about that?

A simple doc has been post in the conversation overview of this PR. I'm now modifying the detailed document.

src/sdk/pynni/nni/nas_utils.py
Crysple: Will change

QuanluZhang:
please pass pylint

src/sdk/pynni/nni/nas_utils.py
```diff
@@ -0,0 +1,154 @@
...
+def enas_mode(
+        mutable_id,
+        mutable_layer_id,
+        funcs,
+        funcs_args,
+        fixed_inputs,
+        optional_inputs,
+        optional_input_size,
+        tf):
+    '''Build a tensorflow graph including all functions and optional_inputs using signal varaibles.
```
QuanluZhang: '''For enas mode, we build the full model graph in trial but only run a subgraph, this is implemented by masking inputs and branching ops. Specifically, based on the received subgraph (through nni.get_next_parameter), it can be known which inputs should be masked and which op should be executed.'''

src/sdk/pynni/nni/nas_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        mutable_id,
+        mutable_layer_id,
+        funcs,
+        funcs_args,
+        fixed_inputs,
+        optional_inputs,
+        optional_input_size,
+        tf):
+    '''Execute all the functions using all the optional_inputs, where a dropout will be implemented
+    to optional_inputs.'''
```
QuanluZhang: '''Similar to enas mode, oneshot mode also builds the full model graph. The difference is that oneshot mode does not receive subgraph, instead, it uses dropout to randomly dropout inputs and ops.'''

src/sdk/pynni/nni/nas_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    if trial._params is None:
+        trial.get_next_parameter()
```
QuanluZhang: what are these two lines used for?

src/sdk/pynni/nni/nas_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    optional_inputs = list(optional_inputs.values())
+    inputs_num = len(optional_inputs)
+    # Calculate dropout rate according to the formular r^(1/k), where r is a hyper-parameter and k is the number of inputs
+    if inputs_num > 0:
+        rate = 0.01 ** (1 / inputs_num)
+        noise_shape = [inputs_num] + [1] * len(optional_inputs[0].get_shape())
```
QuanluZhang: what is this line used for?

tools/nni_annotation/code_generator.py
```diff
@@ -316,16 +320,19 @@ def _visit_string(self, node):
             return node  # not an annotation, ignore it
 
         if string.startswith('@nni.get_next_parameter'):
-            deprecated_message = "'@nni.get_next_parameter' is deprecated in annotation due to inconvenience. Please remove this line in the trial code."
-            print_warning(deprecated_message)
+            call_node = parse_annotation(string[1:]).value
```
QuanluZhang: please add comments here: it is used in enas mode as it needs to retrieve the next subgraph for training.

tools/nni_annotation/code_generator.py
```diff
@@ -369,5 +377,8 @@ def parse(code):
         if type(nodes[i]) is ast.ImportFrom and nodes[i].module == '__future__':
             last_future_import = i
     nodes.insert(last_future_import + 1, import_nni)
+    if nas_mode in ['enas_mode', 'oneshot_mode']:
```
QuanluZhang: please add comments here: enas and oneshot modes for tensorflow need tensorflow module, so we import it here.

src/sdk/pynni/nni/nas_utils.py
Crysple: NNI requires to get_next_parameter before report a result. But the parameter will not be used in this mode

src/sdk/pynni/nni/nas_utils.py
Crysple: To dropout an entire input but dropout different inputs independently

tools/nni_annotation/code_generator.py
Crysple: resolve

PR #1216, opened 2019-06-26
Fix nnictl commands

SparkSnail:


tools/nni_cmd/nnictl_utils.py
```diff
@@ -486,13 +488,13 @@ def get_platform_dir(config_content):
...
             port = machine.get('port')
-            dir_list.append(host + ':' + str(port) + '/tmp/nni/experiments')
+            dir_list.append(host + ':' + str(port) + '/tmp/nni')
     elif platform == 'pai':
         pai_config = config_content.get('paiConfig')
         host = config_content.get('paiConfig').get('host')	
         user_name = config_content.get('paiConfig').get('userName')
         output_dir = config_content.get('trial').get('outputDir')
-        dir_list.append('hdfs://{0}:9000/{1}/nni/experiments'.format(host, user_name))
+        dir_list.append('hdfs://{0}:9000/{1}/nni'.format(host, user_name))
```
squirrelsc: The code actually is using http://xxx/webhdfs to access files. There can be changed like below format:

'server: {0}, path: {1}/nni'.format(host, user_name)

SparkSnail: fixed, changed to `'server: {0}, path: {1}/nni'.format(host, user_name)`

PR #1254, opened 2019-07-04
add doc for one-shot training modes of NAS

QuanluZhang:


docs/en_US/GeneralNasInterfaces.md
```diff
@@ -83,9 +85,99 @@ Accordingly, a specified neural architecture (generated by tuning algorithm) is
...
+for _ in range(num):
+    """@nni.get_next_parameter(self.session)"""
+    loss, _ = self.session.run([loss_op, train_op])
+    """@nni.report_final_result(loss)"""
+```
+Here, `get_next_parameter` needs an arg which is the session variable that you use to train your model.
+
+**oneshot_mode**: following the training approach in [this paper][6]. Different from enas_mode which trains the full graph by training large numbers of subgraphs, in oneshot_mode the full graph is built and dropout is added to candidate inputs and also added to candidate ops' outputs. Then this full graph is trained like other DL models. [Detailed Description](#OneshotMode). (currently only supported on tensorflow). 
+
+To use oneshot_mode, you should add one more field in the `trial` config as shown below. In this mode, no need to specify tuner in the config file as it does not need tuner. (Note that you still need to specify a tuner (any tuner) in the config file in v0.9, not anymore in future releases.)
```
xuehui1991: "not anymore in future releases"? (Remove? because we cannot make sure now.)

docs/en_US/GeneralNasInterfaces.md
```diff
@@ -83,9 +85,99 @@ Accordingly, a specified neural architecture (generated by tuning algorithm) is
...
+```diff
+trial:
+    command: your command to run the trial
+    codeDir: the directory where the trial's code is located
+    gpuNum: the number of GPUs that one trial job needs
++   #choice: classic_mode, enas_mode, oneshot_mode
++   nasMode: oneshot_mode
+```
+
+**darts_mode**: following the training approach in [this paper][3]. It is similar to oneshot_mode. There are two differences, one is that darts_mode only add architecture weights to the outputs of candidate ops, the other is that it trains model weights and architecture weights in an interleaved manner. [Detailed Description](#DartsMode). (not supported yet).
```
xuehui1991: "one is that darts_mode only add architecture weights to the outputs of candidate ops"? (I cannot understand here...)

docs/en_US/GeneralNasInterfaces.md
QuanluZhang: For oneshot_mode, in v0.9, user has to specify a tuner. However, specifying tuner would make this mode strange, so we will remove this constraint in the next release.

docs/en_US/GeneralNasInterfaces.md
QuanluZhang: sure, I will make it more clear. Can the detailed description make you understand?

docs/en_US/AdvancedFeature/GeneralNasInterfaces.md
```diff
@@ -83,9 +85,109 @@ Accordingly, a specified neural architecture (generated by tuning algorithm) is
...
+To use enas_mode, you should add one more field in the `trial` config as shown below.
+```diff
+trial:
+    command: your command to run the trial
+    codeDir: the directory where the trial's code is located
+    gpuNum: the number of GPUs that one trial job needs
++   #choice: classic_mode, enas_mode, oneshot_mode
++   nasMode: enas_mode
+```
+Similar to classic_mode, in enas_mode you need to specify a tuner for nas, as it also needs to receive subgraphs from tuner (or controller using the terminology in the paper). Since this trial job needs to receive multiple subgraphs from tuner, each one for a mini-batch, two lines need to be added to the trial code to receive the next subgraph (i.e., `nni.training_update`) and report the report the result of the current subgraph. Below is an example:
```
leckie-chn: typo: `report the report the result` => `report the result`

docs/en_US/GeneralNasInterfaces.md
QuanluZhang: good suggestion, let's remove it

docs/en_US/AdvancedFeature/GeneralNasInterfaces.md
QuanluZhang: fixed, thanks!

Issue #1098, opened 2019-05-22
BOHB experiment - scipy version


VincentMatthys:
Issue : 

BOHB experiment started (so says `nnictl` ) but no experiment runs

```
INFO: Successfully started experiment!
-----------------------------------------------------------------------
The experiment id is JhglNJSn
The Web UI urls are: http://172.17.0.2:8080   http://127.0.0.1:8080
```

Looking for logs in `nni/experiments/JhglNJSn/log/nnimanager.log`:
```
[5/22/2019, 10:24:56 AM] ERROR [ 'Read GPU summary failed with error: ',
  { ChildProcessError: Command failed: tail -n 1 /tmp/nni/script/gpu_metrics
  tail: cannot open '/tmp/nni/script/gpu_metrics' for reading: No such file or directory
   `tail -n 1 /tmp/nni/script/gpu_metrics` (exited with error code 1)
      at callback (/home/appuser/.local/nni/node_modules/child-process-promise/lib/index.js:33:27)
      at ChildProcess.exithandler (child_process.js:296:5)
      at ChildProcess.emit (events.js:182:13)
      at maybeClose (internal/child_process.js:962:16)
      at Socket.stream.socket.on (internal/child_process.js:381:11)
      at Socket.emit (events.js:182:13)
      at Pipe._handle.close (net.js:606:12)
    name: 'ChildProcessError',
    code: 1,
    childProcess:
     { ChildProcess: [Function],
       fork: [Function: fork],
       _forkChild: [Function: _forkChild],
       exec: [Function: exec],
       execFile: [Function: execFile],
       spawn: [Function: spawn],
       spawnSync: [Function: spawnSync],
       execFileSync: [Function: execFileSync],
       execSync: [Function: execSync] },
    stdout: '',
    stderr:
     'tail: cannot open \'/tmp/nni/script/gpu_metrics\' for reading: No such file or directory\n' } ]
```

But the real problem lies in `nni/experiments/JhglNJSn/log/dispatcher.log`
```
[05/22/2019, 10:24:57 AM] DEBUG (nni.main/MainThread) START
[05/22/2019, 10:24:57 AM] ERROR (nni.main/MainThread) cannot import name 'factorial'
Traceback (most recent call last):
  File "/home/appuser/.local/lib/python3.5/site-packages/nni/__main__.py", line 202, in <module>
    main()
  File "/home/appuser/.local/lib/python3.5/site-packages/nni/__main__.py", line 137, in main
[...]

PR #1290, opened 2019-07-10
Fix bug of nnictl about stopping experiment and update statsmodels version

Crysple:
## 1
issue #1098
incompatability between scipy and statsmodel --> refer to [this issue](https://github.com/statsmodels/statsmodels/issues/5759)
since statsmodel has been updated from 0.9.0 to 0.10.0　

## 2 
issue #1292
there's a bug about nnictl stop experiment in this PR #1216 .  `all` remains an option in args in `nnictl.py` but is used as a boolean value in `nnictl_utils.py`

### DIFF
--- a/tools/nni_cmd/nnictl_utils.py
+++ b/tools/nni_cmd/nnictl_utils.py
@@ -151,7 +151,7 @@ def parse_ids(args):
             exit(1)
         else:
             result_list = running_experiment_list
-    elif args.all:
+    elif args.all != 'all':
         result_list = running_experiment_list
     elif args.id.endswith('*'):
         for id in running_experiment_list:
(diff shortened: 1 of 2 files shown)


tools/nni_cmd/nnictl_utils.py
```diff
@@ -151,7 +151,7 @@ def parse_ids(args):
             exit(1)
         else:
             result_list = running_experiment_list
```
xuehui1991: do you need to consider change line 554 and line 434? because "args_all" look likes a boolean variable in these lines.

Crysple: In the `nnictl.py`, `all` is treated as one value of args.id under the `stop_experiment` subparser but is a boolean value `--all` under other subparser so they are different. According to the current document, user need to type something like `nnictl list experiment --all` to list all experiments. But they only need to type `nnictl stop all` to stop all running experiment. 
Originally I'd like to change the command to `nnictl stop --all` so they might be a little more consistent but I think that it has become customary to type `nnictl stop all` so I discard that thought.

Issue #1207, opened 2019-06-25
Succeeded trial doesn't have final result but webui show ×××(FINAL)


lvybriage:
**Short summary about the issue/question**:
I run an experiment,
 from webui overview table, I see that image. The trial didn't have final result.
![无标题](https://user-images.githubusercontent.com/35484733/60077202-077e1680-975c-11e9-92b8-2e90e1be0f3e.png)
from detail page, trial had final result!
![1](https://user-images.githubusercontent.com/35484733/60077219-12d14200-975c-11e9-8160-8ecc4f8aaa1e.png)
from rest api, I found that the trial did not have final result, like this pictrue
![3](https://user-images.githubusercontent.com/35484733/60077274-309ea700-975c-11e9-8f73-dc60078eb2b6.png)
So, it's confused. I found that detail table's final result was the latest intermediate result other than final result. But it rendered 0.160111(FINAL)
![5](https://user-images.githubusercontent.com/35484733/60077343-5166fc80-975c-11e9-90c3-a0a32895428a.png)

**Brief what process you are following**: 

**How to reproduce it**: 

**nni Environment**: 
- nni version: 0.8
- nni mode(local|pai|remote): pai
- OS: --
- python version: --
- is conda or virtualenv used?:  --
- is running in docker?: --

**need to update document(yes/no)**: no

**Anything else we need to know**: webui show wrong message

PR #1293, opened 2019-07-11
Fix final metrics collection when metrics data is not at beginning of line

chicm-ms:
Fix issue #1207


test/metrics_test/trial.py
```diff
@@ -2,9 +2,14 @@
 import nni
 
 if __name__ == '__main__':
-    hyper_params = nni.get_next_parameter()
-
+    nni.get_next_parameter()
     for i in range(10):
+        if i % 2 == 0:
```
xuehui1991: why we add these code?

tools/nni_trial_tool/log_utils.py
```diff
@@ -134,7 +134,7 @@ def __init__(self, logger, log_collection, log_level=logging.INFO):
         self._is_read_completed = False
         self.process_exit = False
         self.log_collection = log_collection
-        self.log_pattern = re.compile(r'^NNISDK_MEb\'.*\'$')
+        self.log_pattern = re.compile(r'NNISDK_MEb\'.*\'$')
```
QuanluZhang: I suggest to keep '^', it makes the matching more restrict.

test/metrics_test/trial.py
chicm-ms: Except local mode, for other training services, metrics are printed into stdout log file with special MARK 'NNISDK_ME'.
the purpose of this code is to test whether following 2 cases can be handled correctly:
1.  Intermediate result is not at the begining of a line, such as:
`report intermediate result without end of line.NNISDK_MExxx`
2.  Intermediate result is  at the begining of a line, such as:
```
report intermediate result.
NNISDK_MExxx
```

tools/nni_trial_tool/log_utils.py
```diff
@@ -172,11 +172,19 @@ def run(self):
...
+            metrics = None
+            if search_result:
+                collect_line = line[0:search_result.span()[0]] + line[search_result.span()[1]:]
+                metrics = search_result.group(0)
+            else:
+                collect_line = line
+
+            if self.log_collection != 'none':
+                self.queue.put(collect_line)
+            elif metrics is not None:
```
QuanluZhang: elif -> if?

chicm-ms: Thanks for good catch, there was a bug.  Fixed now.

tools/nni_trial_tool/log_utils.py
yds05: The issue is caused by ^, since that case doesn't have \r\n

PR #1318, opened 2019-07-17
Python wrapper for rest api

chicm-ms:


QuanluZhang:
@ liuzhe

QuanluZhang:
looks great! please add unit test and doc.

src/sdk/pycli/nnicli/nni_client.py
```diff
@@ -0,0 +1,133 @@
...
+def _create_process(cmd):
+    if sys.platform == 'win32':
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE, creationflags=subprocess.CREATE_NEW_PROCESS_GROUP)
+    else:
+        process = subprocess.Popen(cmd, stdout=subprocess.PIPE)
+
+    while process.poll() is None:
+        output = process.stdout.readline()
+        if output:
+            print(output.decode('utf-8').strip())
```
liuzhe-lz: If we simply print the output, why not use `stdout=None`?

src/sdk/pycli/nnicli/nni_client.py
```diff
@@ -0,0 +1,133 @@
...
+    global _api_endpoint
+    _api_endpoint = endpoint
+
+def _check_endpoint():
+    if _api_endpoint is None:
+        raise AssertionError("Please call set_endpoint to specify nni endpoint")
+
+def _nni_rest_get(api_path, response_type='json'):
+    _check_endpoint()
+    uri = os.path.join(_api_endpoint, API_ROOT_PATH, api_path)
```
liuzhe-lz: Does this work on windows?

src/sdk/pycli/nnicli/nni_client.py
liuzhe-lz: Should we verify the return code?

liuzhe-lz: Should we check the return code?

chicm-ms:
> looks great! please add unit test and doc.

Integration test added.
A simple document is added into the notebook tutorial in this PR.

src/sdk/pycli/nnicli/nni_client.py
chicm-ms: Thanks, fixed, integration test added for Windows local mode.

src/sdk/pycli/nnicli/nni_client.py
chicm-ms: Thanks,  return code checking added.
While using `stdout=None`,  within jupyter notebook, it prints the nnictl ouput to jupyter backend console, which is not desired behavior. With current code, it can print nnictl output to notebook.

PR #1326, opened 2019-07-18
Fix includeIntermediateResults field in config_schema

SparkSnail:
apply includeIntermediateResults field for all tuners


tools/nni_cmd/config_schema.py
```diff
@@ -80,6 +80,7 @@ def setPathCheck(key):
     },
     ('BatchTuner', 'GridSearch', 'Random'): {
         'builtinTunerName': setChoice('builtinTunerName', 'BatchTuner', 'GridSearch', 'Random'),
+        Optional('includeIntermediateResults'): setType('includeIntermediateResults', bool),
```
xuehui1991: I think we could hold a discussion about refactoring these logic.

Issue #1339, opened 2019-07-21
Hyperband/BOHB not running from the most promising configs


suiguoxin:
After the n configurations is finished, Hyperband/BOHB choose top n/eta configurations and runs them. These n/eta configurations trials should be lanced **from the top to the bottom**, while the current version lance them **in reverse**.

suiguoxin:
![bohb](https://user-images.githubusercontent.com/12380769/61587269-639c5500-abb9-11e9-8166-1c48c777d6e8.png)

suiguoxin:
![bohb](https://user-images.githubusercontent.com/12380769/61587985-82edaf00-abc6-11e9-9d60-a2b2293e1e21.png)

here R = 9, eta= 3, concurrency=1. we can see that the 10th trial has the same config with **the third best** of the previous 9, which is not reasonable.

PR #1340, opened 2019-07-21
Fix hyperband pop issue

suiguoxin:
fix issue #1339


suiguoxin:
bohb_fixed

here R = 9, eta= 3, concurrency=1. we can see that the 10th trial has the same config with the top one of the previous 9

PR #1349, opened 2019-07-23
Add test code for `nnictl stop` command

Crysple:
Including `nnictl stop`, `nnictl stop id` and `nnictl stop all`

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -22,6 +22,7 @@ import contextlib
 import collections
 import json
 import os
+import socket
 import sys
 import subprocess
 import requests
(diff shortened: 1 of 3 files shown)


test/utils.py
```diff
@@ -76,10 +78,13 @@ def setup_experiment(installed=True):
             pypath = ':'.join([sdk_path, cmd_path])
         os.environ['PYTHONPATH'] = pypath
 
-def fetch_nni_log_path(experiment_url):
+def get_experiment_id(experiment_url):
+    experiment_id = requests.get(experiment_url).json()['id']
+    return experiment_id
+
+def get_nni_log_path(experiment_url):
```
SparkSnail: why refactor this place?

liuzhe-lz:
I think we should better test `nnictl stop` for each experiment status.
`type ExperimentStatus = 'INITIALIZED' | 'RUNNING' | 'ERROR' | 'STOPPING' | 'STOPPED' | 'DONE' | 'NO_MORE_TRIAL' | 'TUNER_NO_MORE_TRIAL';`
I believe most of them are easy to test, though `INITIALIZED` and `STOPPING` may be hard.

test/utils.py
Crysple: For convenience. Since I need a new function to get experiment id to be used in other place
and the first two lines of this function is to get experiment id, so...
BTW, I change `fetch` to `get` to be more consistent with other functions' name

Crysple:
> I think we should better test `nnictl stop` for each experiment status.
> `type ExperimentStatus = 'INITIALIZED' | 'RUNNING' | 'ERROR' | 'STOPPING' | 'STOPPED' | 'DONE' | 'NO_MORE_TRIAL' | 'TUNER_NO_MORE_TRIAL';`
> I believe most of them are easy to test, maybe except `INITIALIZED` and `STOPPING`.
> We can skip them for now if it's to hard.

Good point. But I think it's more urgent to write test code for many other `nnictl` commands like `nnictl experiment show/list/status/export`(with a lot of args, (also acts a test for output of RestServer)), `nnictl trial ls/kill`, `nnictl update`, `nnictl config`, `nnictl --version` and so on, provided that there's sufficient labor.
We might firstly write some simple test to cover most commands, to be Triggered By Every PR. Then gradually enrich these code and expand these tests and Trigger them in Daily Build.

Issue #1332, opened 2019-07-18
Hyper-parameter page broken


ultmaster:
**Short summary about the issue/question**:

I'm trying to use NNI to train my network, and what the first few trials show "SUCCEEDED", I can no longer visit the hyper-parameter page in training details.

Then I press F12 to see what's going on. It seems like some "undefined" bug.

```
DOMException: Failed to execute 'addColorStop' on 'CanvasGradient': The value provided ('undefined') could not be parsed as a color.
    at o.getGradient (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:852514)
    at t.brush (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:86745)
    at g._doPaintEl (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:2283310)
    at g._doPaintList (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:2282292)
    at g._paintList (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:2281297)
    at g.refresh (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:2280095)
    at p.refreshImmediately (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:451438)
    at p.flush (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:451567)
    at W.B.setOption (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:8912)
    at t.n.renderEchartDom (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:2707289)
ha @ react-dom.production.min.js:198
n.callback @ react-dom.production.min.js:210
ia @ react-dom.production.min.js:193
na @ react-dom.production.min.js:193
ja @ react-dom.production.min.js:217
Ba @ react-dom.production.min.js:220
(anonymous) @ react-dom.production.min.js:250
t.unstable_runWithPriority @ scheduler.production.min.js:18
Is @ react-dom.production.min.js:250
Ts @ react-dom.production.min.js:249
Os @ react-dom.production.min.js:248
Ds @ react-dom.production.min.js:251
On @ react-dom.production.min.js:85
react-dom.production.min.js:248 Uncaught DOMException: Failed to execute 'addColorStop' on 'CanvasGradient': The value provided ('undefined') could not be parsed as a color.
    at o.getGradient (http://10.150.144.124:20718/static/js/main.565fbd70.js:1:852514)
[...]

PR #1350, opened 2019-07-23
[fix issue#1332] Double check for trial does not have final result in webui

lvybriage:
#1332

### DIFF
--- a/src/webui/src/components/public-child/IntermediateVal.tsx
+++ b/src/webui/src/components/public-child/IntermediateVal.tsx
@@ -30,7 +30,12 @@ class IntermediateVal extends React.Component<IntermediateValProps, {}> {
                 }
             }
             if (status === 'SUCCEEDED') {
-                result = `${result} (FINAL)`;
+                // some trial haven't final result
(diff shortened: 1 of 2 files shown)


suiguoxin:
@ultmaster

ultmaster:
I'm kind of confused. The bug seems to be fixed, but the interface looks like this in case there is no final result.

image

As you see, the default metrics in the bottom shows FINAL (but I didn't submit FINAL result); so if it is FINAL for real, why is it not shown on the hyper-parameter chart?

To reproduce this test, you only need to comment one line here and rerun MNIST as instructed: https://github.com/microsoft/nni/blob/a5fa235190c54f840fdb82b31ca26e1800e8d465/examples/trials/mnist/mnist.py#L207

Maybe I made some mistake; contact me if needed.

lvybriage:
@ultmaster I'll continue to fix this confuse in detail table. If trial's status is succeed, and it haven't final result, will show that xxx (LATEST).

lvybriage:
image

ultmaster:
Issue fixed. @suiguoxin

Issue #1331, opened 2019-07-18
add get_experiment_id() function python API


suiguoxin:
rename nni.get_sequence_id() to nni.get_trial_id ?

QuanluZhang:
when re-submit
- trial_id change
- sequence_id do not change 

trial_id : str
sequence_id: num

QuanluZhang:
@ liuzhe

PR #1353, opened 2019-07-23
issue #1331

suiguoxin:
- add python API get_experiment_id
- add python API get_trial_id
- related doc and test


liuzhe-lz:
Looks great

PR #1358, opened 2019-07-24
#1286 Re submit

suiguoxin:
- only failed jobs can be re-submitted
- no trial_id returned when re-submitted successfully
- else raise err


src/nni_manager/core/nniDataStore.ts
```diff
@@ -141,6 +141,7 @@ class NNIDataStore implements DataStore {
 
     public async getTrialJob(trialJobId: string): Promise<TrialJobInfo> {
         const trialJobs: TrialJobInfo[] = await this.queryTrialJobs(undefined, trialJobId);
+        assert(trialJobs.length <= 1);
```
xuehui1991: if trialJobs.length == 0, then asseration is right but will crash when take value "trialJobs[0]" ?

src/sdk/pynni/nni/msg_dispatcher_base.py
```diff
@@ -168,6 +169,9 @@ def handle_initialize(self, data):
     def handle_request_trial_jobs(self, data):
         raise NotImplementedError('handle_request_trial_jobs not implemented')
 
+    def handle_resubmit_trial_job(self, data):
+        raise NotImplementedError('handle_resubmit_trial_job not implemented')
+
     def handle_update_search_space(self, data):
        raise NotImplementedError('handle_update_search_space not implemented')
```
liuzhe-lz: Irrelevant to this PR. The indentation is mistyped.

src/nni_manager/common/manager.ts
```diff
@@ -103,6 +103,7 @@ abstract class Manager {
     public abstract exportData(): Promise<string>;
 
     public abstract addCustomizedTrialJob(hyperParams: string): Promise<void>;
+    public abstract resubmitTrialJob(hyperParams: string): Promise<void>;
     public abstract cancelTrialJobByUser(trialJobId: string): Promise<void>;
```
chicm-ms: the parameter should be trial job id as we discussed.

src/nni_manager/core/nnimanager.ts
```diff
@@ -135,6 +135,42 @@ class NNIManager implements Manager {
         return this.dataStore.storeTrialJobEvent('ADD_CUSTOMIZED', '', hyperParams);
     }
 
+    public async resubmitTrialJob(data: string): Promise<void> {
+        // get trail job info by jobId
+        const trialJobId = JSON.parse(data).job_id
```
chicm-ms: The parameter should be trial job id, it should not need to be parsed inside this function.

src/nni_manager/core/nnimanager.ts
```diff
@@ -135,6 +135,42 @@ class NNIManager implements Manager {
...
+            hyperParameter = trialJobInfo.hyperParameters[0]
+        } else {
+            hyperParameter = trialJobInfo.hyperParameters.find(hp=>JSON.parse(hp).parameter_index == 0) as string
+        }
+
+        if(trialJobInfo.status != 'FAILED'){
+            return Promise.reject(
+                new Error('only failed trials can be resubmitted')
+            );
+        }
```
chicm-ms: This status checking can be moved up before retrieving hyperParameter.

src/sdk/pynni/nni/msg_dispatcher_base.py
suiguoxin: why irrelevant ? Which indentation ?

src/nni_manager/core/nnimanager.ts
suiguoxin: fixed

src/nni_manager/core/nnimanager.ts
suiguoxin: fixed

src/nni_manager/common/manager.ts
suiguoxin: fixed

src/nni_manager/core/nniDataStore.ts
suiguoxin: Here is to avoid confict with this PR https://github.com/microsoft/nni/pull/1288/files. I don't exactly why this assert is added @chicm-ms

src/sdk/pynni/nni/msg_dispatcher_base.py
suiguoxin: fixed

PR #1363, opened 2019-07-24
fix webui start bug

lvybriage:


src/webui/package.json
```diff
@@ -39,7 +39,7 @@
     "lodash": "^4.17.13",
     "lodash.template": "^4.5.0",
     "js-yaml": "^3.13.1",
-    "webpack-dev-server": "^3.1.11",
+    "webpack-dev-server": "3.0.0",
```
xuehui1991: “^”?

lvybriage: ^3.0.0 still have this question:
image

PR #1372, opened 2019-07-26
[Fix issue#1180] Support resubmit FAILED trial

lvybriage:
- FAILED job can be resubmitted

- resubmit button will be disabled when experiment status is DONE | RUNNING | STOPPED or the job  resubmit successfully


lvybriage:
This PR is closed because of having new design.

Issue #1359, opened 2019-07-24
Can nnictl stop by port?


xuehui1991:
**What would you like to be added**:
Can "nnictl" stop experiment by port? because the experiment id is hard to remember.

When we start the new experiment in the same port will be got this message:
![image](https://user-images.githubusercontent.com/6280746/61763532-50cb8f80-ae08-11e9-92b6-0a7218f2ec6d.png)

Could nni tell me the expeirment_id in the same port directly? So I don't have to search the experiment_id.

Thank you!

**Why is this needed**:

**Without this feature, how does current nni work**：

**Components that may involve changes**:

**Brief description of your proposal if any**:

PR #1384, opened 2019-07-29
Let nnictl support stopping by port

liuzhe-lz:
Issue #1359 

~~Maybe we should update docs as well.~~
~~But I don't know where.~~


SparkSnail:
> Issue #1359
> 
> Maybe we should update docs as well.
> But I don't know where.

NNICTL doc is in https://github.com/microsoft/nni/blob/master/docs/en_US/Tutorial/Nnictl.md#stop, please update the corresponding doc.

Issue #1376, opened 2019-07-26
SMAC cannot be installed if nni is installed in dev mode


suiguoxin:
as title

suiguoxin:
- check err info
- check why some dependencies are not installed when installing SMAC

PR #1385, opened 2019-07-29
Add package install error information

SparkSnail:
https://github.com/microsoft/nni/issues/1376


tools/nni_cmd/launcher.py
```diff
@@ -360,9 +360,13 @@ def launch_experiment(args, experiment_config, mode, config_file_name, experimen
         module_name = AdvisorModuleName.get(package_name)
     if package_name and module_name:
         try:
-            check_call([sys.executable, '-c', 'import %s'%(module_name)], stdout=PIPE, stderr=PIPE)
+            stdout_full_path, stderr_full_path = get_log_path(config_file_name)
+            with open(stdout_full_path, 'a+') as stdout_file, open(stderr_full_path, 'a+') as stderr_file:
```
QuanluZhang: why use `a+` rather than 'w'?

tools/nni_cmd/launcher.py
```diff
@@ -360,9 +360,13 @@ def launch_experiment(args, experiment_config, mode, config_file_name, experimen
...
+                check_call([sys.executable, '-c', 'import %s'%(module_name)], stdout=stdout_file, stderr=stderr_file)
         except CalledProcessError as e:
-            print_error('%s should be installed through \'nnictl package install --name %s\''%(package_name, package_name))
+            print_error('some errors happen when import package %s.' %(package_name))
+            print_log_content(config_file_name)
+            print_error('If %s is not installed, it should be installed through \'nnictl package install --name %s\''%(package_name, package_name))
```
QuanluZhang: would be better to check whether this package needs installation then print this error message, because a package which does not need installation may also have import error.

tools/nni_cmd/launcher.py
SparkSnail: the stdout_file and stderr_file is a common file, will also be used in other place of nnictl. `w` will override the content.

tools/nni_cmd/launcher.py
SparkSnail: fixed.

PR #1389, opened 2019-07-30
Set gpuNum as optional

SparkSnail:


liuzhe-lz:
It seems GPUs are still available when `CUDA_VISIBLE_DEVICES` is set to 0, while Tensorflow will only use 1 GPU in this case.
Can we use this as the default setting?

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -601,12 +601,15 @@ class RemoteMachineTrainingService implements TrainingService {
         let command: string;
         // Set CUDA_VISIBLE_DEVICES environment variable based on cuda_visible_device
         // If no valid cuda_visible_device is defined, set CUDA_VISIBLE_DEVICES to empty string to hide GPU device
-        if (typeof cuda_visible_device === 'string' && cuda_visible_device.length > 0) {
-            command = `CUDA_VISIBLE_DEVICES=${cuda_visible_device} ${this.trialConfig.command}`;
+        if (this.trialConfig.gpuNum === undefined) {
```
yds05: Add comments to explain if gpuNum is undefined we will not set CUDA_VISIBLE_DEVICE?

docs/en_US/Tutorial/ExperimentConfig.md
```diff
@@ -266,7 +266,7 @@ machineList:
 
   * __gpuNum__
 
-      __gpuNum__ specifies the gpu number to run the tuner process. The value of this field should be a positive number.
+      __gpuNum__ specifies the gpu number to run the tuner process. The value of this field should be a positive number. If the field is not set, NNI will not set `CUDA_VISIBLE_DEVICE` in script, and will not manage gpu resource.
```
QuanluZhang: => in script (that is, will not control the visibility of GPUs on trial command through `CUDA_VISIBLE_DEVICES`), and ...

CUDA_VISIBLE_DEVICE => CUDA_VISIBLE_DEVICES

src/nni_manager/training_service/remote_machine/gpuScheduler.ts
```diff
@@ -49,6 +49,9 @@ export class GPUScheduler {
      * @param requiredGPUNum required GPU number
      */
     public scheduleMachine(requiredGPUNum: number, trialJobDetail : RemoteMachineTrialJobDetail) : RemoteMachineScheduleResult {
```
QuanluZhang: is the type `number` for requiredGPUNum still correct?

QuanluZhang: number -> number | undefined ?

QuanluZhang:
what is the behavior in kubeflow/pai mode when gpuNum is not specified?

docs/en_US/Tutorial/ExperimentConfig.md
SparkSnail: fixed.

src/nni_manager/training_service/remote_machine/gpuScheduler.ts
SparkSnail: fixed.

SparkSnail:
> what is the behavior in kubeflow/pai mode when gpuNum is not specified?

In kubeflow/pai mode, the gpuNum field is required, since the gpu schedule policy does not relay on `CUDA_VISIBLE_DEVICES`.

SparkSnail:
> It seems GPUs are still available when `CUDA_VISIBLE_DEVICES` is set to 0, while Tensorflow will only use one GPU in this case.
> Can we use this as default setting?
> 
> Why did we use `CUDA_VISIBLE_DEVICES=-1`? @QuanluZhang

 `CUDA_VISIBLE_DEVICES=-1` is set to block all of gpus.

PR #1392, opened 2019-07-31
[Fixed issue#1218] Show the best metric curve during search progress in webui

lvybriage:
1. when user set `optimize_mode: maximize`, best metric graph is: 
image
 and default metric graph is:
image

2. when user set `optimize_mode: minimize`, best metric graph is: 
image

3. the filed `optimize_mode: minimize |  maximize` is from tuner.classArgs.optimize_mode


lvybriage:
This pr is closed for another realization.

Issue #1220, opened 2019-06-27
nnictl can't find config files of kubeflow and frameworkcontroller


yayamao-ponyai:
**Short summary about the issue/question**:

nnictl can't find config files of kubeflow and frameworkcontroller

**Brief what process you are following**: 

nnictl create -c examples/trials/mnist/config_kubeflow.yml
nnictl create -c examples/trials/mnist/config_frameworkcontroller.yml

**How to reproduce it**: 

```console
nnictl create -c examples/trials/mnist/config_kubeflow.yml
INFO: expand searchSpacePath: search_space.json to /nni/examples/trials/mnist/search_space.json 
INFO: expand codeDir: . to /nni/examples/trials/mnist/. 
INFO: Starting restful server...
INFO: Successfully started Restful server!
INFO: Setting kubeflow config...
ERROR: Failed! Error is: {"error":"ENOENT: no such file or directory, open './config/kubeflow/tfjob-crd-v1alpha2.json'"}
```

```console
nnictl create -c examples/trials/mnist/config_frameworkcontroller.yml
INFO: expand searchSpacePath: search_space.json to /nni/examples/trials/mnist/search_space.json 
INFO: expand codeDir: . to /nni/examples/trials/mnist/. 
INFO: Starting restful server...
INFO: Successfully started Restful server!
INFO: Setting frameworkcontroller config...
ERROR: Failed! Error is: {"error":"ENOENT: no such file or directory, open './config/frameworkcontroller/frameworkcontrollerjob-crd-v1.json'"}
```

**nni Environment**:
- nni version: 0.8
- nni mode(local|pai|remote): kubeflow or frameworkcontroller with minikube
- OS: ubuntu 16.04
- python version: python3.5
- is conda or virtualenv used?: test on both local and virtualenv
- is running in docker?: no

**need to update document(yes/no)**: no

**Anything else we need to know**: no

yayamao-ponyai:
Just a friendly ping, @scarlett2018 @yds05 have you guys identify this can be reproduced, if so, do you have a plan to fix it?

scarlett2018:
Adding @SparkSnail as well.

SparkSnail:
This problem can be reproduced if install NNI by pip command, will fix it as soon as possible. You could install NNI from source code for temp workaround.

PR #1404, opened 2019-08-02
Fix pypi package

SparkSnail:
config folder is missed when build pypi package. 
fix issue https://github.com/microsoft/nni/issues/1220


liuzhe-lz:
I don't think we should maintain a separate Makefile for pip.
But fine, this is a bugfix.

PR #1412, opened 2019-08-05
Eject from react-scripts-ts-antd and bump webui dependencies version

liuzhe-lz:
To solve the security alert, we must upgrade to webpack v4, which is incompatible with `react-scripts-ts-antd`.
I have checked the repo of `react-scripts-ts-antd` as well as it's upstream `react-scripts-ts`, and confirmed that both of them are no longer maintained.
So I believe it's time to `eject`.

**Note: This PR requires a clean build!** (i.e. `make clean` before `make`)
Please check if it works with your devices.

Known issues:
* The version of react and its type annotation are not perfectly matched.
* Some plugins are deprecated or unmaintained in webpack v4.
* The scripts are auto-generated and are somewhat redundant.
* **The memory-hungry uglify plugin is temporarily disabled to pass the CI pipeline. We should enable webpack's built-in uglifier in next release.**


liuzhe-lz:
For reviewers:
This PR, especially files in `src/webui/scripts` and `src/webui/config` folders, are mostly generated with `yarn eject` command.
Instead of manually reading diff files, please pull the branch to your device and try it out.

Issue #1375, opened 2019-07-26
It cannot work with nni-0.9 version but work with nni-0.7 version


peipei-pig:
I use nni-0.7 version it can work very well.But today I update new version:
(1) pip uninstall nni-0.7-py3-none-manylinux1_x86_64.whl
(2) pip install nni-0.9-py3-none-manylinux1_x86_64.whl
then I run script cmd: nnictl create --config config.xml 

I find some errors  in nnimanager.log.
[2019-7-26 15:01:00] INFO [ 'Starting experiment: UCMnCLlE' ]
[2019-7-26 15:01:00] INFO [ 'Change NNIManager status from: INITIALIZED to: RUNNING' ]
[2019-7-26 15:01:00] INFO [ 'Add event listeners' ]
[2019-7-26 15:01:00] INFO [ 'Run local machine training service.' ]
[2019-7-26 15:01:00] WARNING [ 'gpu_metrics file does not exist!' ]
[2019-7-26 15:01:01] INFO [ 'NNIManager received command from dispatcher: ID, ' ]
[2019-7-26 15:01:05] ERROR [ 'Read GPU summary failed with error: ',
  SyntaxError: Unexpected end of JSON input
      at JSON.parse (<anonymous>)
      at GPUScheduler.updateGPUSummary (/usr/local/miniconda3/nni/training_service/local/gpuScheduler.js:69:40) ]
[2019-7-26 15:01:10] ERROR [ 'Error: This socket has been ended by the other party\n    at Socket.writeAfterFIN [as write] (net.js:402:12)\n    at IpcInterface.sendCommand (/usr/local/minic
onda3/nni/core/ipcInterface.js:47:38)\n    at NNIManager.pingDispatcher (/usr/local/miniconda3/nni/core/nnimanager.js:282:29)' ]
[2019-7-26 15:01:10] INFO [ 'Change NNIManager status from: RUNNING to: ERROR' ]
 
If I reinstall nni-0.7 version,it can  work well again. How can I solve this problem？

**nni Environment**:
- nni version:nni-0.9-py3-none-manylinux1_x86_64.whl 
- nni mode(local|pai|remote):local
- OS:ubuntu16.04
- python version:Python 3.6.5 :: Anaconda, Inc.
- is conda or virtualenv used?:conda 
- is running in docker?:yes

liuzhe-lz:
Please provide your dispatcher's log (`dispatcher.log` in the same folder of `nnimanager.log`).
[...]

PR #1418, opened 2019-08-06
Report the error more gracefully when nvidia-smi not exist

liuzhe-lz:
Currently `nni_gpu_tool` crashes when `nvidia-smi` is not available.
This PR makes it output empty GPU status (`"gpuCount": 0`) instead.
In local mode, when `gpuNum` set to non-zero and `gpuCount` is zero, the experiment fail.
In remote mode, when a server's `gpuCount` is zero, a **warning** message will be written to log file. (If you think it should be an "error", please leave comment.)

Update:
Previously the script creates `gpu_metrics` file and then changes it's permission to `777`. This will fail when the file is owned by another user, even if it's `777`.
Now it creates `gpu_metrics` with `umask` 0.

Relate to: issue #1375


tools/nni_gpu_tool/gpu_metrics_collector.py
```diff
@@ -48,11 +49,13 @@ def main(argv):
     with open(os.path.join(metrics_output_dir, "gpu_metrics"), "w") as outputFile:
         pass
     os.chmod(os.path.join(metrics_output_dir, "gpu_metrics"), 0o777)
-    cmd = 'nvidia-smi -q -x'
+    cmd = 'nvidia-smi -q -x'.split()
     while(True):
         try:
-            smi_output = subprocess.check_output(cmd, shell=True)
+            smi_output = subprocess.check_output(cmd)
             parse_nvidia_smi_result(smi_output, metrics_output_dir)
```
QuanluZhang: there is try catch within `parse_nvidia_smi_result`

QuanluZhang: and it uses 'a' in `open`

liuzhe-lz: This is to catch `subprocess.check_output` when `nvidia-smi` not exist

src/nni_manager/training_service/local/gpuScheduler.ts
```diff
@@ -54,6 +54,9 @@ class GPUScheduler {
             } catch (error) {
                 this.log.error('Read GPU summary failed with error: ', error);
             }
+            if (this.gpuSummary.gpuCount === 0) {
+                throw new Error('GPU not available. Please check your CUDA configuration');
```
SparkSnail: If the gpu_metrics_collector update gpu_metric file slowly, will gpuSummary.gpuCount updated to a positive number after several detection?

PR #1422, opened 2019-08-07
Chinese Translation

squirrelsc:


liuzhe-lz:
Don't forget to squash commits before merging...

squirrelsc:
> Don't forget to squash commits before merging...

yes, it's default option.

Issue #1416, opened 2019-08-05
Failed trials show a huge negative duration on macOS


ultmaster:
**Short summary about the issue/question**:

Duration shows a huge negative number when the trial gets failed.

![Screen Shot 2019-08-06 at 00 11 45](https://user-images.githubusercontent.com/8463288/62479012-1afeb180-b7df-11e9-9b51-4f25ef07c154.png)

**Brief what process you are following**: 

**How to reproduce it**: Install nii using pip and run mnist example.

**nni Environment**:
- nni version: 0.9.1.1
- nni mode(local|pai|remote): local
- OS: macOS Mojave
- python version: Python 3.6
- is conda or virtualenv used?: virtualenv
- is running in docker?: no

**need to update document(yes/no)**: no

**Anything else we need to know**:

I'm pretty sure it's a problem related to Mac or macOS.

ultmaster:
https://superuser.com/questions/599072/how-to-get-bash-execution-time-in-milliseconds-under-mac-os-x

The problem is `date +%s%3N` is not supported in macOS. So needs an extra `if` for platform darwin.

PR #1434, opened 2019-08-07
Fix the issue that date nanoseconds does not work under macOS

ultmaster:
#1416


src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -521,9 +521,14 @@ class LocalTrainingService implements TrainingService {
                 `$NOW_DATE = "$NOW_DATE" + (Get-Date -Format fff).ToString()`,
                 `Write $LASTEXITCODE " " $NOW_DATE  | Out-File ${path.join(workingDirectory, '.nni', 'state')} -NoNewline -encoding utf8`);
         } else {
-            script.push(
-                `eval ${localTrailConfig.command} 2>${path.join(workingDirectory, 'stderr')}`,
-                `echo $? \`date +%s%3N\` >${path.join(workingDirectory, '.nni', 'state')}`);
+            script.push(`eval ${localTrailConfig.command} 2>${path.join(workingDirectory, 'stderr')}`);
+            if (process.platform == 'darwin') {
```
chicm-ms: Better to use "===" instead of "=="

ultmaster: fixed

PR #1436, opened 2019-08-08
Set TPE classArgs as optional

SparkSnail:


suiguoxin:
Other tuners have the same problem, right ?

QuanluZhang:
@suiguoxin depending on the code of each tuner. would be better to make tuner/advisor the same behavior.

PR #1438, opened 2019-08-08
Metis ss

suiguoxin:
use same pattern of quniform/randint as in other tuners

### DIFF
--- a/src/sdk/pynni/nni/metis_tuner/metis_tuner.py
+++ b/src/sdk/pynni/nni/metis_tuner/metis_tuner.py
@@ -121,7 +121,7 @@ class MetisTuner(Tuner):
                 key_range = search_space[key]['_value']
                 idx = self.key_order.index(key)
                 if key_type == 'quniform':
-                    if key_range[2] == 1 and key_range[0].is_interger() and key_range[1].is_interger():
+                    if key_range[2] == 1 and key_range[0].is_integer() and key_range[1].is_integer():
                         self.x_bounds[idx] = [key_range[0], key_range[1]+1]
                         self.x_types[idx] = 'range_int'
                     else:
--- a/src/sdk/pynni/nni/metis_tuner/metis_tuner.py
+++ b/src/sdk/pynni/nni/metis_tuner/metis_tuner.py
@@ -121,13 +121,12 @@ class MetisTuner(Tuner):
                 key_range = search_space[key]['_value']
                 idx = self.key_order.index(key)
                 if key_type == 'quniform':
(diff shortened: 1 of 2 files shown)


src/sdk/pynni/nni/metis_tuner/metis_tuner.py
```diff
@@ -121,13 +121,12 @@ def update_search_space(self, search_space):
...
-                    if key_range[2] == 1:
-                        self.x_bounds[idx] = [key_range[0], key_range[1]]
+                    if key_range[2] == 1 and key_range[0].is_interger() and key_range[1].is_interger():
+                        self.x_bounds[idx] = [key_range[0], key_range[1]+1]
                         self.x_types[idx] = 'range_int'
                     else:
-                        bounds = []
-                        for value in np.arange(key_range[0], key_range[1], key_range[2]):
-                            bounds.append(value)
+                        low, high, q = key_range
```
QuanluZhang: so `q` must be an integer?

suiguoxin: no, why this question ?

src/sdk/pynni/nni/metis_tuner/metis_tuner.py
```diff
@@ -121,13 +121,12 @@ def update_search_space(self, search_space):
                 key_range = search_space[key]['_value']
                 idx = self.key_order.index(key)
                 if key_type == 'quniform':
-                    if key_range[2] == 1:
-                        self.x_bounds[idx] = [key_range[0], key_range[1]]
+                    if key_range[2] == 1 and key_range[0].is_interger() and key_range[1].is_interger():
```
ultmaster: Why `interger` instead of `integer`?

src/sdk/pynni/nni/metis_tuner/metis_tuner.py
QuanluZhang: because the type in line 131 is `discrete_int`

src/sdk/pynni/nni/metis_tuner/metis_tuner.py
suiguoxin: fixed

src/sdk/pynni/nni/metis_tuner/metis_tuner.py
suiguoxin: "discreate_int" doesn't mean that the value will be int. here is to deal with the index of the generated list

Issue #1440, opened 2019-08-08
[V1.0 bug-bash] The concurrent number in input box update delay on WebUI


rabbit008:
**Short summary about the issue/question**:

**Brief what process you are following**: 
![image](https://user-images.githubusercontent.com/28092647/62696469-0dc60a80-ba0b-11e9-9902-53d2771739ff.png)
When I edit the concurrency value in WebUI and commit , the input box always delay to update and show the original value, but the profile always update in time

windows -->Run the example in the official tutorial
**How to reproduce it**: 
windows -->Run the example in the official tutorial

**nni Environment**:
- nni version:1.0
- nni mode(local|pai|remote):local
- OS:windows 
- python version:3.7
- is conda or virtualenv used?: no
- is running in docker?:no

**need to update document(yes/no)**:no

**Anything else we need to know**:no

Issue #1437, opened 2019-08-08
[V1.0 bug-bash] Show current phase is different in overview page and detail page


lvybriage:
in overview page:
![image](https://user-images.githubusercontent.com/35484733/62677202-17864880-b9e0-11e9-9101-0cd7bdf5d0b0.png)

in detail page:
![image](https://user-images.githubusercontent.com/35484733/62677219-2c62dc00-b9e0-11e9-89d4-766af68a5b05.png)

PR #1451, opened 2019-08-09
fix bugs, sort, concurrency, multi

lvybriage:
- default metric column sort [issue](https://github.com/microsoft/nni/issues/1427)
- rename: optimization -> Optimization, Download -> View
- detail table: column保持选过的状态
- update concurrency: linux-数字闪现，windows-更新的是上一次的数值 ( Windows&Linux测过 ) [issue](https://github.com/microsoft/nni/issues/1440)

multiphase： 
- current phase在succeed table同步 [issue](https://github.com/microsoft/nni/issues/1437)
- download exp results bug

### DIFF
--- a/src/webui/src/components/trial-detail/TableList.tsx
+++ b/src/webui/src/components/trial-detail/TableList.tsx
@@ -4,7 +4,7 @@ import ReactEcharts from 'echarts-for-react';
 import { Row, Table, Button, Popconfirm, Modal, Checkbox, Select, Icon } from 'antd';
 const Option = Select.Option;
 const CheckboxGroup = Checkbox.Group;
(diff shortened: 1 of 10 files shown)


src/webui/src/components/trial-detail/TableList.tsx
```diff
@@ -316,8 +318,8 @@ class TableList extends React.Component<TableListProps, TableListState> {
                 value: item
             });
         });
-        Object.keys(columnSelected).map(key => {
-            const item = columnSelected[key];
+        Object.keys(columnList).map(key => {
```
liuzhe-lz: Well... I think most of `Object.keys()` are unnecessary and can be replaced with `forEach()` or something else...

lvybriage: Ok, will change in next version.

PR #1456, opened 2019-08-12
fix bug when duplicated final results

QuanluZhang:


src/sdk/pynni/nni/hyperopt_tuner/hyperopt_tuner.py
```diff
@@ -315,6 +315,10 @@ def receive_trial_result(self, parameter_id, parameters, value, **kwargs):
                 rval = self.CL_rval
             else:
                 rval = self.rval
+                # ignore duplicated reported final result (due to aware of intermedate result)
+                if parameter_id not in self.running_data:
```
squirrelsc: It's better to run left logic, even there is a duplicated result. Because it's previous logic, and latest result should be better than previous ones.

QuanluZhang: let's leave it in the next code refactor.

PR #1460, opened 2019-08-13
fix bug of hyper-parameter broken when have not succeeded trial

lvybriage:


src/webui/src/components/trial-detail/Para.tsx
```diff
@@ -238,15 +238,19 @@ class Para extends React.Component<ParaProps, ParaState> {
                         },
                         axisLabel: {
                             formatter: function (value: string) {
-                                const length = value.length;
-                                if (length > 16) {
-                                    const temp = value.split('');
-                                    for (let m = 16; m < temp.length; m += 17) {
-                                        temp[m] += '\n';
+                                if (value !== undefined) {
```
liuzhe-lz: The type annotation says it can't be `undefined`?

src/webui/src/components/trial-detail/Para.tsx
```diff
@@ -238,15 +238,19 @@ class Para extends React.Component<ParaProps, ParaState> {
...
+                                    const length = value.length;
+                                    if (length > 16) {
+                                        const temp = value.split('');
+                                        for (let m = 16; m < temp.length; m += 17) {
+                                            temp[m] += '\n';
+                                        }
+                                        return temp.join('');
```
liuzhe-lz: I think it's better to split the string into 17-char chunks with `substring`/`slice` and then join them with `\n`.

src/webui/src/components/trial-detail/Para.tsx
lvybriage: Ok, will add it.

src/webui/src/components/trial-detail/Para.tsx
lvybriage: Thanks, I will try it in next version.

Issue #1461, opened 2019-08-13
Add portList to PAI mode configurations


ultmaster:
**What would you like to be added**: Add an optional field `portList` to configurations when using PAI mode.

**Why is this needed**: I want to use see my tensorboard on PAI. Currently I follow the practice below: launch a tensorboard in the script submitted to PAI in the background, then launch my training. However, I cannot access my tensorboard without a open port. PAI has provided an option, so that I can access the port using something like `$PAI_PORT_LIST_main_0_tensorboard`, but this option is not accessible from the interface that NNI has provided.

**Without this feature, how does current nni work**: Indeed we can find a workaround like downloading the event files and view it locally; but I can imagine adding this feature will provide great convenience for not only tensorboard, but also visdom and other web-dashboard-based tools.

**Components that may involve changes**: trainingServices logic, configuration settings.

**Brief description of your proposal if any**:

ultmaster:
Merged

PR #1467, opened 2019-08-14
Add portList config in PAI trainingService

SparkSnail:
1. https://github.com/microsoft/nni/issues/1461
2. fix error link in supportTrainingService.md


ultmaster:
It's working now, but the process is not straightforward (at least to me). Posting an example here:

Let's say you want to launch a tensorboard in the mnist example. So the first step is to write a wrapper script of `mnist.py`.

```bash
pip3 install tensorboard
export TENSORBOARD_PORT=PAI_PORT_LIST_${PAI_CURRENT_TASK_ROLE_NAME}_0_tensorboard
tensorboard --logdir . --port ${!TENSORBOARD_PORT} &
python3 mnist.py
```

According to [this discussion](https://stackoverflow.com/questions/917260/can-var-parameter-expansion-expressions-be-nested-in-bash), you need to use indirect expansion `${!XXX}` to get the port here, and therefore it's also required to use bash instead of sh in the next step.

Besides, `config_pai.yml` needs to specify that "it wants a port".

```yaml
trial:
  command: bash launch_pai.sh
  portList:
    - label: tensorboard
      beginAt: 0
      portNumber: 1
```

docs/en_US/TrainingService/PaiMode.md
```diff
@@ -55,6 +55,32 @@ Compared with [LocalMode](LocalMode.md) and [RemoteMachineMode](RemoteMachineMod
     * Optional key. Set the shmMB configuration of OpenPAI, it set the shared memory for one task in the task role.
 * authFile
     * Optional key, Set the auth file path for private registry while using PAI mode, [Refer](https://github.com/microsoft/pai/blob/2ea69b45faa018662bc164ed7733f6fdbb4c42b3/docs/faq.md#q-how-to-use-private-docker-registry-job-image-when-submitting-an-openpai-job), you can prepare the authFile and simply provide the local path of this file, NNI will upload this file to HDFS for you.
+* portList  
+    * Optional key. Set the portList configuration of OpenPAI, it specifies a list of port used in container, [Refer](https://github.com/microsoft/pai/blob/b2324866d0280a2d22958717ea6025740f71b9f0/docs/job_tutorial.md#specification).  
+    The config schema in NNI is shown below:
+    ```
+    portList:
+      - label: test,
```
yds05: remove extra ','

SparkSnail: removed.

PR #1471, opened 2019-08-15
fix idx type err in smac categorical param conversion

suiguoxin:


src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -202,7 +202,7 @@ def convert_loguniform_categorical(self, challenger_dict):
                 converted_dict[key] = np.exp(challenger_dict[key])
             # convert categorical back to original value
             elif key in self.categorical_dict:
-                idx = challenger_dict[key]
+                idx = int(challenger_dict[key])
```
QuanluZhang: why challenger_dict[key] is not int?

suiguoxin: > why challenger_dict[key] is not int?

That's a variable returned by the original smac package

QuanluZhang:
close as no reproduced

Issue #1475, opened 2019-08-17
Hyper-parameter page broken when there is null in values


ultmaster:
**Short summary about the issue/question**:

When there is null in values, hyper-parameter page fails.

An example search space:

```json
{
  "sequence_step_size": {
    "_type": "choice",
    "_value": [null, 10, 20, 50]
  }
}
```

Relevent error trace:

```
backend.js:1 TypeError: Cannot read property 'toString' of null
    at Para_Para.hyperParaPic (Para.tsx:171)
    at Para_Para.reInit (Para.tsx:386)
    at Para_Para.componentDidMount (Para.tsx:507)
    at Vh (react-dom.production.min.js:216)
    at Zh (react-dom.production.min.js:220)
    at react-dom.production.min.js:250
    at Object.exports.unstable_runWithPriority (scheduler.production.min.js:18)
    at Fi (react-dom.production.min.js:250)
    at Di (react-dom.production.min.js:249)
    at Yh (react-dom.production.min.js:248)

Para.tsx:171 Uncaught TypeError: Cannot read property 'toString' of null
    at Para_Para.hyperParaPic (Para.tsx:171)
    at Para_Para.reInit (Para.tsx:386)
    at Para_Para.componentDidMount (Para.tsx:507)
    at Vh (react-dom.production.min.js:216)
    at Zh (react-dom.production.min.js:220)
    at react-dom.production.min.js:250
    at Object.exports.unstable_runWithPriority (scheduler.production.min.js:18)
    at Fi (react-dom.production.min.js:250)
    at Di (react-dom.production.min.js:249)
    at Yh (react-dom.production.min.js:248)
```

**nni Environment**:
- nni version: 1.0 bug-bash round 2
- nni mode(local|pai|remote): local
- OS: Win
- python version: 3.7
- is conda or virtualenv used?: conda
- is running in docker?: no

**need to update document(yes/no)**: no

**Anything else we need to know**:

ultmaster:
It's also broken when there are no values.

Error info:

```
backend.js:1 TypeError: Cannot read property 'length' of undefined
    at formatter (Para.tsx:241)
    at axisHelper.js:328
    at addItem (axisTickLabelBuilder.js:333)
    at makeLabelsByNumericCategoryInterval (axisTickLabelBuilder.js:328)
    at makeCategoryLabelsActually (axisTickLabelBuilder.js:112)
    at makeCategoryLabels (axisTickLabelBuilder.js:89)
[...]

PR #1479, opened 2019-08-19
Rewrite search space documentation for choice

ultmaster:
#1475 

According to @lvybriage , there are still some problems and ambiguities when displaying on hyper-parameter page if we allow users to define a search space of choice like `[null, "128", 256, 512]`. The quick yet elegant solution is to add a "disclaimer" in the documentation that a search space like this might cause undefined behaviors.

This PR also refines the docs of that part by fixing a few issues and typos.


QuanluZhang:
good!

PR #1491, opened 2019-08-26
Change `receive_customized_trial_result` API

liuzhe-lz:
Change `receive_customized_trial_result` API to make customized trials not that special.
Tuners will not receive results of customized trials by default.
They can call `self.accept_customized_trials()` if they support customizing hyper-parameters. After which `receive_trial_result` will receive results from customized trials as well as algorithm-generated trials.

NODE: This change is somewhat not backward compatible.
Old tuners are still functional, but they will not receive results from customized trials anymore.


src/sdk/pynni/nni/tuner.py
```diff
@@ -57,19 +57,21 @@ def generate_multiple_parameters(self, parameter_id_list, **kwargs):
 
     def receive_trial_result(self, parameter_id, parameters, value, **kwargs):
         """Invoked when a trial reports its final result. Must override.
+        By default this only reports results of algorithm-generated hyper-parameters.
+        Use `accept_customized_trials()` to receive results from user-added parameters.
         parameter_id: int
         parameters: object created by 'generate_parameters()'
         reward: object reported by trial
         """
+        # TODO: kwargs doc?
```
xuehui1991: remove this line?

liuzhe-lz: I'll add the doc later this week.

xuehui1991: But it should not appear in our repo....

PR #1522, opened 2019-09-04
Mnist torch example

suiguoxin:
- add torch mnist example 
- change  links for mnist examples in README


suiguoxin:
Difference from mnist-tensorflow example:
- in mnist-tensorflow example, it reports the intermediate result every 100 batches, no matter the size of batch
- in this example, it reports  the intermediate result every time it traverses the training data

QuanluZhang:
could you also add it into IT?

PR #1523, opened 2019-09-04
add config test for mnist torch exp

suiguoxin:


examples/trials/mnist-pytorch/mnist.py
```diff
@@ -108,17 +108,16 @@ def main(args):
...
-        logger.debug('test accuracy %g', test_acc)
-        logger.debug('Pipe send intermediate result done.')
-
-    test_acc = test(args, model, device, test_loader)
-    # report final result
-    nni.report_final_result(test_acc)
-    logger.debug('Final result is %g', test_acc)
-    logger.debug('Send final result done.')
-
+        if epoch < args['epochs']:
```
xuehui1991: what's the difference between the original version and this change(111-120 line)? Seems are same?  Why change here

suiguoxin: In the original version, the final result will be reported twice: once as intermediate-result, once as final result. This leads to a duplicated display in the learning curve of WebUI.

PR #1526, opened 2019-09-06
update config schema for PPOTuner

QuanluZhang:


tools/nni_cmd/config_schema.py
```diff
@@ -149,6 +149,13 @@ def setPathCheck(key):
             Optional('trials_per_update'): setNumberRange('trials_per_update', int, 0, 99999),
             Optional('epochs_per_update'): setNumberRange('epochs_per_update', int, 0, 99999),
             Optional('minibatch_size'): setNumberRange('minibatch_size', int, 0, 99999),
+            Optional('ent_coef'): setType('ent_coef', float),
```
xuehui1991: Do these parameters have range?

QuanluZhang: no need to set the range, because these parameters are only for advanced users, I assume they know what they are doing.

PR #1554, opened 2019-09-17
Chinese translation

squirrelsc:


squirrelsc:
> MultiPhase.md's en_US file is also changed?

Yes, en_us changes are already in master branch.

PR #1555, opened 2019-09-17
Added patch to close file in common.py

farziengineer:


src/sdk/pynni/nni/common.py
```diff
@@ -52,21 +52,21 @@ def init_logger(logger_file_path, log_level_name='info'):
     logger_file_path: path of logger file (path-like object).
     """
     log_level = log_level_map.get(log_level_name, logging.INFO)
-    logger_file = open(logger_file_path, 'w')
-    fmt = '[%(asctime)s] %(levelname)s (%(name)s/%(threadName)s) %(message)s'
-    logging.Formatter.converter = time.localtime
-    formatter = logging.Formatter(fmt, _time_format)
-    handler = logging.StreamHandler(logger_file)
-    handler.setFormatter(formatter)
+    with open(logger_file_path, 'w') as logger_file:
```
xuehui1991: Seems there is some error in IT.  Try to push this PR again?

farziengineer: Do you mean there is some problem in that particular line?

QuanluZhang:
@farziengineer thanks for your contribution, but this log file cannot be closed because nni writes log during running.

PR #1563, opened 2019-09-20
KSE Quantizer implementation

Cjkkkk:
a kse quantizer implementation using current framework( have to override default compress method)


QuanluZhang:
let's discuss the details of this implementation

PR #1575, opened 2019-09-27
Sync model compression doc and implementation

liuzhe-lz:
TODO: It seems PyTorch does not use the glossary "op". Maybe we should use "layer".


docs/en_US/Compressor/Pruner.md
```diff
@@ -1,41 +1,38 @@
...
-
-This is one basic pruner: you can set a target sparsity level(expressed as a fraction, 0.6 means we will prune 60%). 
+This is one basic pruner: you can set a target sparsity level (expressed as a fraction, 0.6 means we will prune 60%). 
 
 We first sort the weights in the specified layer by their absolute values. And then mask to zero the smallest magnitude weights until the desired sparsity level is reached.
 
 ### Usage
 
 Tensorflow code
 ```
```
QuanluZhang: how about also adding the `import` line here?

PR #1577, opened 2019-09-27
ignore dot git folder

SeanKim:
when doing setup for remote nni, counting .git folder may make total number of folder larger than 1,000(nni file transfer threshold). generally we don't need to include .git folder for remote execution for code. so I excluded .git folder when counting and zipping files.

### DIFF
--- a/src/nni_manager/common/utils.ts
+++ b/src/nni_manager/common/utils.ts
@@ -370,9 +370,9 @@ function countFilesRecursively(directory: string, timeoutMilliSeconds?: number):
     let fileCount: number = -1;
     let cmd: string;
     if(process.platform === "win32") {
-        cmd = `powershell "Get-ChildItem -Path ${directory} -Recurse -File | Measure-Object | %{$_.Count}"`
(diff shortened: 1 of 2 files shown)


SparkSnail:
> when doing setup for remote nni, counting .git folder may make total number of folder larger than 1,000(nni file transfer threshold). generally we don't need to include .git folder for remote execution for code. so I excluded .git folder when counting and zipping files.

Hi, thanks for your contribution. `.git` folder is unnecessary for most scene, and perhaps should be avoided to be counted in NNI. For my concern, NNI should provide a general solution for the folders which users do not want to count and upload to remote machine, not just a specific `.git` folder. Hard code for `.git` folder is not an elegance way, could you please refactor the code and provide a more general way to support ignoring more kinds of folders? Thanks.

liuzhe-lz:
Suggest two ways:
1. Follow `grep`'s practice (`--exclude-dir`). It may be hard to deal with pattern.
2. Ignore all hidden files and directories.

scarlett2018:
> Suggest two ways:
> 
> 1. Follow `grep`'s practice (`--exclude-dir`). It may be hard to deal with pattern.
> 2. Ignore all hidden files and directories.

@SeanKim - do you think the suggestions make senses? would you like to update the PR accordingly?thanks.

SparkSnail:
close this pr, if you have any question, please feel free to give feedback.

PR #1579, opened 2019-09-29
Fixed the sequencing of the font attributes

TimYagan:
Fixed the sequencing of the font attributes to ensure that the parent attributes are not overwritten

### DIFF
--- a/src/webui/src/index.css
+++ b/src/webui/src/index.css
@@ -19,8 +19,8 @@ time, mark, audio, video {
 	margin: 0;
 	padding: 0;
 	border: 0;
+	font: inherit;	
 	font-size: 100%;
-	font: inherit;
 }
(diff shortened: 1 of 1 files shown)


lvybriage:
@TimYagan 

Hi, TimYagan. I think it isn't important to change it. Element such as 'div' could have own CSS style rather than to inherit it's parent element style.

lvybriage:
> Fixed the sequencing of the font attributes to ensure that the parent attributes are not overwritten

Do you mean that we should override parent element style?

TimYagan:
Thanks for the replies.

I meant that there was a possibility that the sequencing of the css
elements could have overwritten parent attributes, but perhaps this is
rare. If anything, this is a super low probability event.

PR #1590, opened 2019-10-08
Resolve comments in PR 1571

liuzhe-lz:


src/sdk/pynni/tests/test_compressor.py
```diff
@@ -1,9 +1,9 @@
 from unittest import TestCase, main
-import nni.compression.tensorflow as tf_compressor
-import nni.compression.torch as torch_compressor
+import tensorflow as tf
 import torch
 import torch.nn.functional as F
-import tensorflow as tf 
+import nni.compression.tensorflow as tf_compressor
+import nni.compression.torch as torch_compressor
```
QuanluZhang: this import order cannot pass pylint

liuzhe-lz: It can.

src/sdk/pynni/nni/compression/tensorflow/builtin_pruners.py
```diff
@@ -32,8 +32,8 @@ class AGP_Pruner(Pruner):
     """
     def __init__(self, config_list):
         """
-            Configure Args
-                initial_sparsity:
+            Configure Args:
```
xuehui1991: please follow the docstring format like other tuners?

liuzhe-lz: Updated.
Followed format used in `msg_dispatcher.handle_report_metric_data`

xuehui1991: I think this format has some differences with the format that we used(We used the format like xgboost).   I not sure about that. @leckie-chn

liuzhe-lz: This does not document the parameter list, instead it documents the fields of one parameter.

ultmaster:
Would you add a link to compressor docs in table of contents? Currently, there's no way to see compressor other than searching for it.

PR #1569, opened 2019-09-26
Add example for customized advisor and some refactoring

ultmaster:
1. Add example for customized advisor
2. Add docs for MsgDispatcherBase
3. Prettify some codes in dispatcher and tests (will add more tests for builtin advisors and tuners later)

Needs your help/advice: 

https://github.com/ultmaster/nni/blob/1beecd9454f87d50cfb996169d48f6a7a6d56203/src/sdk/pynni/nni/msg_dispatcher_base.py#L202

https://github.com/ultmaster/nni/blob/1beecd9454f87d50cfb996169d48f6a7a6d56203/src/sdk/pynni/nni/msg_dispatcher_base.py#L226

PR #1594, opened 2019-10-09
Fix hyperband

chicm-ms:
Fix a problem caused by this pr:  https://github.com/microsoft/nni/pull/1569


ultmaster:
Seems my fault. :(

Issue #1595, opened 2019-10-09
example mnist-pytorch fails


chicm-ms:
example mnist-pytorch failed with following errors:
config file: https://github.com/microsoft/nni/blob/master/test/config_test/examples/mnist-pytorch.test.yml

```
Traceback (most recent call last):
  File "mnist.py", line 157, in <module>
    main(params)
  File "mnist.py", line 91, in main
    transforms.Normalize((0.1307,), (0.3081,))
  File "/home/chec/anaconda3/envs/nni/lib/python3.7/site-packages/torchvision/datasets/mnist.py", line 46, in __init__
    self.download()
  File "/home/chec/anaconda3/envs/nni/lib/python3.7/site-packages/torchvision/datasets/mnist.py", line 120, in download
    gzip.GzipFile(file_path) as zip_f:
  File "/home/chec/anaconda3/envs/nni/lib/python3.7/gzip.py", line 163, in __init__
    fileobj = self.myfileobj = builtins.open(filename, mode or 'rb')
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tensorflow/mnist/input_data/raw/train-images-idx3-ubyte.gz'
```

Then change maxTrialNum from 4 to 2, trialConcurrency from 2 to 1, got following errors:

```
Traceback (most recent call last):
  File "mnist.py", line 157, in <module>
    main(params)
  File "mnist.py", line 91, in main
    transforms.Normalize((0.1307,), (0.3081,))
  File "/home/chec/anaconda3/envs/nni/lib/python3.7/site-packages/torchvision/datasets/mnist.py", line 46, in __init__
    self.download()
  File "/home/chec/anaconda3/envs/nni/lib/python3.7/site-packages/torchvision/datasets/mnist.py", line 128, in download
    read_image_file(os.path.join(self.root, self.raw_folder, 'train-images-idx3-ubyte')),
  File "/home/chec/anaconda3/envs/nni/lib/python3.7/site-packages/torchvision/datasets/mnist.py", line 299, in read_image_file
    return torch.from_numpy(parsed).view(length, num_rows, num_cols)
RuntimeError: PyTorch was compiled without NumPy support
```

chicm-ms:
Manully upgrade pytorch from 0.4.1 to 1.2.0,  second problem solved.

chicm-ms:
problem fixed.

PR #1596, opened 2019-10-09
Fix mnist pytorch

chicm-ms:
Fix first problem  for issue: https://github.com/microsoft/nni/issues/1595

For the second problem, pytorch version will be upgraded for pipeline with another PR.


liuzhe-lz:
I don't fully understand. Where does each trial have to use separate directory?

chicm-ms:
> I don't fully understand. Where does each trial have to use separate directory?

The problem is described in issue #1595 ,  on local training services,  sometimes the example fails because multiple trial jobs download data into one single data_dir.

liuzhe-lz:
OK. Looks a bit awkward though...

PR #1597, opened 2019-10-10
Remove unneccessary dependencies in pipeline

chicm-ms:


test/pipelines-it-kubeflow.yml
```diff
@@ -39,11 +39,6 @@ jobs:
     displayName: 'Install nni toolkit via source code'
 
   - script: |
-      python3 -m pip install scikit-learn==0.20.0 --user
-      python3 -m pip install torch==0.4.1 --user
-      python3 -m pip install torchvision==0.2.1 --user
-      python3 -m pip install keras==2.1.6 --user
-      python3 -m pip install tensorflow==1.12.0 --user
```
QuanluZhang: these packages are not necessary any more or installed at somewhere else?

chicm-ms: They are not necessary at all,  the test trial jobs are  running on remote kubeflow training service.   These dependencies are installed on local kubeflow client (nnimanager).

liuzhe-lz: Seems this installs dependencies to host environment but the trials are running in container?
If that's the case, then `nnictl package install` are also unnecessary?

Issue #1568, opened 2019-09-26
IT pipeline to show more log when experiment fails


chicm-ms:
![image](https://user-images.githubusercontent.com/38930155/65654338-03a7ac00-e04b-11e9-8693-d924eacd8d4a.png)

Better to show  nnimanager log, dispatcher log, trial job log for failed experiment, so that we can know why the pipe line failed.

chicm-ms:
PR #2190 submitted.

chicm-ms:
Fixed.

PR #1602, opened 2019-10-11
Show error log of failed trial jobs for integration tests

chicm-ms:
Fix issue https://github.com/microsoft/nni/issues/1568


test/utils.py
```diff
@@ -107,23 +108,34 @@ def get_experiment_status(status_url):
...
+    return failed_jobs
+
+def print_failed_job_log(training_service, trial_jobs_url):
+    '''Print job log of FAILED trial jobs'''
+    trial_jobs = get_failed_trial_jobs(trial_jobs_url)
+    for trial_job in trial_jobs:
+        if training_service == 'local':
+            log_filename = trial_job['stderrPath'].split(':')[-1]
+        else:
+            log_filename = os.path.join(get_nni_log_dir(EXPERIMENT_URL), 'trial_' + trial_job['id'] + '.log')
```
QuanluZhang: for the training services other than local, how can you get trial log's content?

QuanluZhang: you enabled log collection for these tests?

QuanluZhang: I am curious about that why in logcollection mode trial's log is stored under `experiment_id/log/` rather than `experiment_id/trials/trial_id/`?

chicm-ms: The logCollection for IT are enabled by this PR in order to retrieve logs.
@SparkSnail @QuanluZhang   It could be a little bit confusing if collected logs are put under `experiment_id/trials/trial_id/` since the directory is used by training service, and it seems not good to put it under  `experiment_id/log`,  maybe it's better to put it under `experiment_id/log/log_collection` ?

QuanluZhang: I think `experiment_id/log/log_collection/` is better than `experiment_id/log/`. But I don't get why `experiment_id/trials/trail_id/` is not proper? Is logcollection also a component of training service?

chicm-ms: `experiment_id/log/log_collection/` provides some kind of hints that the logs are related to logCollection.
Just check the code, the remote trainning service uses `/tmp/nni/exepriments/experiment_id/trials/trial_id`  as trial dir, it is not same as this one,  so either are OK for me.

QuanluZhang: personally, I prefer putting log under `experiment_id/trials/trial_id/`

chicm-ms: @QuanluZhang   Updated, changed to log directory to `experiment_id/trials/trial_id/`

PR #1607, opened 2019-10-14
fix built-in pruner bugs

tanglang96:
initial mask may have different device with layer weights, to make them consistent


liuzhe-lz:
Resubmitted as #1612

PR #1611, opened 2019-10-16
example mnist was not compatible with tersorflowv2

xinshouke:
replace line 9  with tensorflow.compat.v1,the cause  was not able to  compatible with tersorflowv2

### DIFF
--- a/examples/trials/mnist/mnist.py
+++ b/examples/trials/mnist/mnist.py
@@ -6,7 +6,8 @@ import math
 import tempfile
 import time
 
-import tensorflow as tf
+import tensorflow.compat.v1 as tf
+tf.disable_v2_behavior()
 from tensorflow.examples.tutorials.mnist import input_data
 
 import nni


liuzhe-lz:
This change will make it incompatible with tensorflow lower than 1.13.
We will discuss how to resolve the issue soon.

xinshouke:
> This change will make it incompatible with tensorflow lower than 1.13.
> We will discuss how to resolve the issue soon.

Oh,maybe I thought it must support tensorflow v2 while using python >= 3.5...I rushed advance a little.

liuzhe-lz:
I totally agree it should support tensorflow v2.
But we want to support lower versions at the same time...

liuzhe-lz:
We have added a [TFv2 example](https://github.com/microsoft/nni/tree/master/examples/trials/mnist-tfv2) and we are working on using v2 as default.

PR #1613, opened 2019-10-16
Fix web UI bugs found in bug bash

liuzhe-lz:
Fixed bugs:
  - Dispatcher log does not refresh automatically
  - "Add column" button is not functional on initializing
  - Empty search result breaks detail page
  - Trials fail before sending first metric will not be listed

### DIFF
--- a/src/webui/src/components/Modal/LogDrawer.tsx
+++ b/src/webui/src/components/Modal/LogDrawer.tsx
@@ -27,7 +27,7 @@ class LogDrawer extends React.Component<LogDrawerProps, LogDrawerState> {
         this.state = {
             nniManagerLogStr: null,
             dispatcherLogStr: null,
-            isLoading: true
+            isLoading: true,
         };
     }
 
--- a/src/webui/src/components/Modal/LogDrawer.tsx
+++ b/src/webui/src/components/Modal/LogDrawer.tsx
@@ -8,89 +8,46 @@ import MonacoHTML from '../public-child/MonacoEditor';
 import '../../static/style/logDrawer.scss';
 
 interface LogDrawerProps {
-    isVisble: boolean;
     closeDrawer: () => void;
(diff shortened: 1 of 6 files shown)


src/webui/src/components/Modal/LogDrawer.tsx
```diff
@@ -8,89 +8,46 @@ import MonacoHTML from '../public-child/MonacoEditor';
...
+            downFile(this.state.dispatcherLogStr, 'dispatcher.log');
+        }
     }
 
     dispatcherHTML = () => {
         return (
             <div>
                 <span>Dispatcher Log</span>
-                <span className="refresh" onClick={this.getdispatcherLogmessage}>
+                <span className="refresh" onClick={this.manualRefresh}>
```
lvybriage: manualRefresh方法会同时更新dispatcher和nnimanager log, 所以当用户留在某一个tab页面时，其实两个页面会同步更新，这样当log内容很大的时候，会不会卡？

src/webui/src/components/Modal/LogDrawer.tsx
```diff
@@ -139,7 +82,7 @@ class LogDrawer extends React.Component<LogDrawerProps, LogDrawerState> {
                     closable={false}
                     destroyOnClose={true}
                     onClose={closeDrawer}
-                    visible={isVisble}
+                    visible={true}
```
lvybriage: why `visible` always is true？

liuzhe-lz: Because when it's not visible, `LogDrawer` will not render.

src/webui/src/components/Modal/LogDrawer.tsx
liuzhe-lz: It will not increase delay more than double. So I believe it's acceptable.

PR #1570, opened 2019-09-26
Add comprehensive tests for tuners

ultmaster:
This PR addresses the first point of #1543.

This PR:

1. Check the table of search space support among tuners.
2. Clarify some limitations discovered by the tests.
3. Unify all `optimize_mode` to be `maximize` by default.
4. ~Refactor the "dependency graph" of some tuners.~
5. ~Modify `.gitignore`.~
6. Fix rewrite NAS space behavior on user defined search space.
7. Uniform can accept `low == high` now.

This PR will not:

1. Add tests for advisors.
2. Add precondition checks for tuners.
3. Refactor redundant code in tuners.

PR #1637, opened 2019-10-21
Rewrite NAS space for tuners other than hyperopt

ultmaster:
#1570 is merged. This PR will do the rest to support mutable_layer in most tuners.


src/sdk/pynni/tests/test_tuner.py
```diff
@@ -118,7 +118,7 @@ def search_space_test_all(self, tuner_factory, supported_types=None, ignore_type
             search_space_all = json.load(fp)
         if supported_types is None:
             supported_types = ["choice", "randint", "uniform", "quniform", "loguniform", "qloguniform",
-                               "normal", "qnormal", "lognormal", "qlognormal"]
+                               "normal", "qnormal", "lognormal", "qlognormal", "mutable"]
```
xuehui1991: since the `_type` is `mutable_layer`, why we are not consistent with the naming rule here?

ultmaster: Because I used some tricks on deciding whether the testcase should be successful or not, which is based on "splitting with underscore".

ultmaster: Nevertheless refactored to get rid of this trick.

ultmaster:
Close due to incompatible API change. This item is yet to be discussed.

PR #1650, opened 2019-10-24
Support keyboard-interactive mode in ssh client

SparkSnail:


src/nni_manager/training_service/remote_machine/remoteMachineData.ts
```diff
@@ -231,6 +232,8 @@ export class SSHClientManager {
           .on('error', (err: Error) => {
             // SSH connection error, reject with error message
             deferred.reject(new Error(err.message));
+        }).on("keyboard-interactive", (name, instructions, lang, prompts, finish) => {
```
liuzhe-lz: Are these arguments required in ts? I'm not sure.

SparkSnail: we have to use the `finish` argument, so the others arguments are put here as placeholders.

PR #1653, opened 2019-10-25
[webui] customized trial

lvybriage:


src/webui/src/components/trial-detail/TableList.tsx
```diff
@@ -458,7 +502,14 @@ class TableList extends React.Component<TableListProps, TableListState> {
                         className="titleColumn"
                     />
                 </Modal>
+                {/* compare trials based message */}
                 <Compare compareRows={selectRows} visible={isShowCompareModal} cancelFunc={this.hideCompareModal} />
+                {/* clone trial parameters and could submit a customized trial */}
+                <Customize
+                    visible={isShowCustomizedModal}
```
liuzhe-lz: Suggest move this out. Only render this element when it's visible.

src/webui/src/components/Modal/CustomizedTrial.tsx
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    render() {
+        const { visible, closeCustomizeModal, hyperParameter } = this.props;
+        const { isShowSubmitSucceed, isShowSubmitFailed, isShowWarning, customID } = this.state;
+        const {
+            form: { getFieldDecorator },
+            // form: { getFieldDecorator, getFieldValue },
+        } = this.props;
+        const parameters = (hyperParameter !== '') ? JSON.parse(hyperParameter) : '';
```
liuzhe-lz: If your purpose is to mark empty hyper-parameters, `null` is better than `''`.

src/webui/src/components/trial-detail/TableList.tsx
```diff
@@ -458,7 +502,14 @@ class TableList extends React.Component<TableListProps, TableListState> {
...
+                    hyperParameter={trialParameter}
```
liuzhe-lz: I think it's good to pass as few props as possible. `Customize` only needs to know which trial (ID) it's copying.

src/webui/src/components/Modal/CustomizedTrial.tsx
```diff
@@ -0,0 +1,329 @@
...
+    closeFailedHint = () => {
+        // also close customized trial modal
+        if (this._isCustomizeMount === true) {
+            this.setState(() => ({ isShowSubmitFailed: false }));
+            this.props.closeCustomizeModal();
+        }
+    }
+
+    componentDidMount() {
+        this._isCustomizeMount = true;
```
liuzhe-lz: Official react doc says we should not check "is mounted" to update states.
Instead we should cancel handlers on unmount.
If there is no handler, than this check does nothing.

lvybriage: OK.

src/webui/src/components/Modal/CustomizedTrial.tsx
lvybriage: Ok.

src/webui/src/components/trial-detail/TableList.tsx
lvybriage: Ok.

src/webui/src/components/trial-detail/TableList.tsx
lvybriage: Ok.

src/webui/src/components/Modal/CustomizedTrial.tsx
```diff
@@ -0,0 +1,329 @@
...
+                    destroyOnClose={true}
+                    maskClosable={false}
+                    centered={true}
+                >
+                    {/* search space form */}
+                    <Row className="hyper-box">
+                        <Form>
+                            {
+                                Object.keys(parameters).map(item => {
+                                    return (
```
liuzhe-lz: `{ return ( blahblah ) }` is equal to `blahblah`

src/webui/src/components/trial-detail/TableList.tsx
lvybriage: > Suggest move this out. Only render this element when it's visible.

We can't see the `succeed modal` if `<Customize>` component's status is hidden. So...

src/webui/src/components/Modal/CustomizedTrial.tsx
```diff
@@ -0,0 +1,317 @@
...
+                        || customized[item] > searchSpace[item]._value[1]) {
+                        flag = true;
+                        return;
+                    }
+                }
+            }
+        });
+        if (flag !== false) {
+            // open the warning modal
+            this.setState(() => ({ isShowWarning: true, customParameters: customized }));
```
liuzhe-lz: `() =>` is unnecessary in all `setState`s.
`this.setState(() => ({ k: v }));` is effectively identical to `this.setState({ k: v });`.

chicm-ms:
Merged into PR 1695

PR #1670, opened 2019-10-30
Fix compressor op_types

chicm-ms:


examples/model_compress/README.md
```diff
@@ -20,7 +20,7 @@ configure_list = [{
     'start_epoch': 0,
     'end_epoch': 10,
     'frequency': 1,
-    'op_type': 'default'
+    'op_types': 'default'
```
QuanluZhang: missed square brackets?

chicm-ms: Thanks, updated.

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -35,7 +35,7 @@ def compress(self, model):
         for name, module in model.named_modules():
             layer = LayerInfo(name, module)
             config = self._select_config(layer)
-            if config is not None:
+            if config is not None and layer.type in config['op_types']:
```
QuanluZhang: _select_config already checked whether `layer.type in config['op_types']`?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -112,10 +121,6 @@ class Quantizer(Compressor):
     Base quantizer for pytorch quantizer
     """
 
-    def __call__(self, model):
```
QuanluZhang: why removing this func?

src/sdk/pynni/nni/compression/torch/compressor.py
chicm-ms: Good catch, removed.

src/sdk/pynni/nni/compression/torch/compressor.py
chicm-ms: There is an exactly same version already in parent class Compressor.

QuanluZhang: i see

Issue #1583, opened 2019-10-02
logDir: Allow to have spaces in the path to the directory


thomas-beznik:
Hello everyone!

**Short summary about the issue/question**:
I would like to save the experiments in a specific folder; I thus set the "logDir" in the config file to the path to this folder. The problem is that in the path to that folder there is a folder which has a space in its name ("My Drive"). The name of this last folder cannot be changed. Because of this space, I get the following error:

ERROR [ 'ChildProcessError: Command failed: mkdir -p /content/gdrive/My Drive/3DSegmentation/relu-framework/Sessions/Test/py8x7FgL/trials/Ugn5Y\nmkdir: cannot create directory ‘/content/gdrive/My’: Operation not supported\n `mkdir -p /content/gdrive/My Drive/3DSegmentation/relu-framework/Sessions/Test/py8x7FgL/trials/Ugn5Y` (exited with error code 1)\n    at callback (/usr/local/nni/node_modules/child-process-promise/lib/index.js:33:27)\n    at ChildProcess.exithandler (child_process.js:296:5)\n    at ChildProcess.emit (events.js:182:13)\n    at maybeClose (internal/child_process.js:962:16)\n    at Process.ChildProcess._handle.onexit (internal/child_process.js:251:5)' ]

Would it be possible to support paths with spaces in them ? How could I fix this myself in the meantime ?

Thank you for the help!

Thomas

SparkSnail:
Hi, thanks for raising the issue. NNI does not support spaces in logDir yet, we will support the feature in next release.

xinshouke:
I had met this issue now,will fix refer to the above later...

SparkSnail:
fixed in https://github.com/microsoft/nni/pull/1694, close this issue.

PR #1694, opened 2019-11-03
Support space in logDir

SparkSnail:
https://github.com/microsoft/nni/issues/1583


src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -380,9 +380,9 @@ class LocalTrainingService implements TrainingService {
         const envVariables: { key: string; value: string }[] = [
             { key: 'NNI_PLATFORM', value: 'local' },
             { key: 'NNI_EXP_ID', value: this.experimentId },
-            { key: 'NNI_SYS_DIR', value: trialJobDetail.workingDirectory },
+            { key: 'NNI_SYS_DIR', value: `'${trialJobDetail.workingDirectory}'` },
```
liuzhe-lz: I don't think the value of environment variables should to be quoted.
We should quote them when composing the command, not here.

SparkSnail: fixed.

PR #1695, opened 2019-11-03
Return sequence ID to web UI when add customized trials

liuzhe-lz:


src/nni_manager/core/nnimanager.ts
```diff
@@ -131,19 +129,34 @@ class NNIManager implements Manager {
...
         }
         if (this.currSubmittedTrialNum >= this.experimentProfile.params.maxTrialNum) {
-            return Promise.reject(
-                new Error('reach maxTrialNum')
-            );
+            return Promise.reject(new Error('reach maxTrialNum'));
         }
-        this.customizedTrials.push(hyperParams);
+
+        // TODO: NNI manager should not peek tuner's internal protocol, let's refactor this later
```
QuanluZhang: any idea about how to refactor?

QuanluZhang: multi-phase does not work in this implementation. but if the customized configuration is sent to tuner first, then multi-phase can be supported

src/nni_manager/core/nnimanager.ts
```diff
@@ -560,18 +573,7 @@ class NNIManager implements Manager {
...
-                if (this.customizedTrials.length > 0) {
-                    const hyperParams: string | undefined = this.customizedTrials.shift();
-                    this.dispatcher.sendCommand(ADD_CUSTOMIZED_TRIAL_JOB, hyperParams);
-                }
-            }
-
-            if (requestTrialNum - requestCustomTrialNum > 0) {
-                this.requestTrialJobs(requestTrialNum - requestCustomTrialNum);
-            }
+            this.requestTrialJobs(requestTrialNum);
```
QuanluZhang: a potential issue here. if does not count customized trial here, nnimanager may request more configurations from tuner than the training service can immediately run.

src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -136,7 +136,6 @@ def handle_add_customized_trial(self, data):
         # data: parameters
         id_ = _create_parameter_id()
         _customized_parameter_ids.add(id_)
```
QuanluZhang: better to comment that this member func is not supported

src/nni_manager/core/nnimanager.ts
liuzhe-lz: Customized trials now share the same queue with algorithm-generated trials.

src/sdk/pynni/nni/msg_dispatcher.py
liuzhe-lz: The hyphen in its name indicates that this is a private method and is not exposed to end users.

liuzhe-lz: This method does work. However the protocol has been changed so that this function should not send `NewTrialJob` command back to NNI manager.

liuzhe-lz: I should have documented this change in IPC protocol, but the protocol itself is not well documented I have not fully understood it yet.
I'm planning on clean up the protocol and write doc for it.

src/nni_manager/core/nnimanager.ts
liuzhe-lz: The refactor should be in dispatcher protocol. I'll do it after I understand the protocol clearly.
To me multi-phase is also a protocol problem so I don't know how to do it for now.

src/sdk/pynni/nni/msg_dispatcher.py
QuanluZhang: @liuzhe-lz please add an issue for this

QuanluZhang:
only reviewed backend code.

liuzhe-lz:
Included PR #1653

PR #1702, opened 2019-11-04
Change default CPU scheduler from random to round-robin

chicm-ms:


src/nni_manager/training_service/remote_machine/gpuScheduler.ts
```diff
@@ -28,20 +28,27 @@ import {
     parseGpuIndices, RemoteMachineMeta, RemoteMachineScheduleResult, RemoteMachineTrialJobDetail, ScheduleResultType, SSHClientManager
 } from './remoteMachineData';
 
+type SCHEDULE_POLICY_NAME = 'random' | 'round-robin';
```
SparkSnail: Can users change schedule policy?

chicm-ms: No for now,  there is no need to change it from round-robin to random for now.

PR #1703, opened 2019-11-05
add overview doc for NAS

QuanluZhang:


docs/en_US/NAS/Overview.md
```diff
@@ -0,0 +1,55 @@
...
+    1. When designing a neural network, the designer may have multiple choices for a layer, sub-model, or connection, and not sure which one or a combination performs the best. It would be appealing to have an easy way to express the candidate layers/sub-models they want to try. 
+    2. For the researchers who are working on automatic NAS, they want to have an unified way to express the search space of neural architectures. And making unchanged trial code adapted to different searching algorithms.
+
+For expressing neural architecture search space, we provide two APIs:
+
+```python
+# choose one ``op`` from ``ops``, for pytorch this is a module.
+# ops: for pytorch ``ops`` is a list of modules, for tensorflow it is a list of keras layers.
+# key: the name of this ``LayerChoice`` instance
+nni.nas.LayerChoice(ops, key)
```
squirrelsc: Could it expand ops here, so that user can understand the signature of ops more.

docs/en_US/NAS/Overview.md
```diff
@@ -0,0 +1,61 @@
+# NNI Programming Interface for Neural Architecture Search (NAS)
+
+*This is an experimental feature, programming APIs are almost done, NAS trainers are under intensive deveropment.*
```
ultmaster: development

docs/en_US/NAS/Overview.md
```diff
@@ -0,0 +1,61 @@
...
+                       dataset_valid=dataset_valid,
+                       batch_size=args.batch_size,
+                       log_frequency=args.log_frequency)
+# finding the best model from search space
+trainer.train()
+# export the best found model
+trainer.export_model()
+```
+
+Different trainers could have different input arguments depending on their algorithms. After training, users could export the best one of the found models through `trainer.export_model()`.
```
ultmaster: To minimize the effort of changing one trainers to another one, I think we should at least make some of the arguments common, e.g., `model`, `optimizer`, `loss`, `metrics`, `num_epochs`, `log_frequency`, `dataset`, `batch_size`...

docs/en_US/NAS/Overview.md
QuanluZhang: fixed, thanks.

docs/en_US/NAS/Overview.md
QuanluZhang: fixed, thanks

docs/en_US/NAS/Overview.md
QuanluZhang: let's keep it for now, will update after the design is done

Issue #1665, opened 2019-10-29
Temporary dir for gpu collection is not always unique among users


Markus92:
**Short summary about the issue/question**:
When running NNI, it puts the gpu metrics collector in /tmp/nni/script, which is not unique among users. So two people cannot run NNI simultaneously (for example on different GPU).

**Brief what process you are following**: 
User 1: runs NNI fine
User 2: runs NNI on different port, but it'll give status error: "Training service error: EACCES: permission denied, open '/tmp/nni/script/gpu_metrics_collector.sh'"

**How to reproduce it**: 
See above

**nni Environment**:
- nni version: 1.0 (but also present in HEAD)
- nni mode(local|pai|remote): local
- OS: Ubuntu 18.04LTS
- python version: 3.6
- is conda or virtualenv used?:  Yes (conda)
- is running in docker?: We run it in Singularity (github.com/sylabs/singularity/), with mounted /tmp as our containers itself are read-only. Redefining /tmp is not an option unfortunately (breaks too many other things it seems). Our containers are self-made from a ubuntu base, with conda, then `pip install nni`

**need to update document(yes/no)**:
Maybe

**Anything else we need to know**:
Bug can be traced to line 435 - 437 of /src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
Clearly `path.basename(os.homedir());` doesn't expand to username in our configuration, even though os.homedir() should work in our containers (homedir is mounted).

A solution would be to use either (/var)/run/user/$PID/, create a /tmp/$PID/ or add a random nonce to the path (for example trial id). Another solution would be allowing to specify your own temporary dir. Upstream at nodejs there might also be a fix possible to use the temporary dir supplied by systemd.

QuanluZhang:
fixed, close.

PR #1707, opened 2019-11-05
fix gpu script permission issue

liuzhe-lz:
Fix issue #1665 

Solution: Don't create script on Linux, use one-line command instead. Windows routine is not modified.

### DIFF
--- a/src/nni_manager/training_service/local/gpuScheduler.ts
+++ b/src/nni_manager/training_service/local/gpuScheduler.ts
@@ -28,7 +28,7 @@ import { String } from 'typescript-string-operations';
 import { getLogger, Logger } from '../../common/log';
 import { delay } from '../../common/utils';
 import { GPUInfo, GPUSummary } from '../common/gpuData';
-import { execKill, execMkdir, execRemove, execTail, getgpuMetricsCollectorScriptContent, getScriptName, runScript } from '../common/util';
+import { execKill, execMkdir, execRemove, execTail, runGpuMetricsCollector } from '../common/util';
 
 /**
  * GPUScheduler for local training service
@@ -101,12 +101,7 @@ class GPUScheduler {
      */
     private async runGpuMetricsCollectorScript(): Promise<void> {
         await execMkdir(this.gpuMetricCollectorScriptFolder, true);
-        //generate gpu_metrics_collector script
-        const gpuMetricsCollectorScriptPath: string =
-            path.join(this.gpuMetricCollectorScriptFolder, getScriptName('gpu_metrics_collector'));
-        const gpuMetricsCollectorScriptContent: string = getgpuMetricsCollectorScriptContent(this.gpuMetricCollectorScriptFolder);
-        await fs.promises.writeFile(gpuMetricsCollectorScriptPath, gpuMetricsCollectorScriptContent, { encoding: 'utf8' });
-        runScript(gpuMetricsCollectorScriptPath);
+        runGpuMetricsCollector(this.gpuMetricCollectorScriptFolder);
     }
 
     // tslint:disable:non-literal-fs-path
--- a/src/nni_manager/training_service/local/gpuScheduler.ts
+++ b/src/nni_manager/training_service/local/gpuScheduler.ts
@@ -43,7 +43,7 @@ class GPUScheduler {
     constructor() {
         this.stopping = false;
         this.log = getLogger();
-        this.gpuMetricCollectorScriptFolder = `${os.tmpdir()}/nni/script`;
+        this.gpuMetricCollectorScriptFolder = `${os.tmpdir()}/${os.userInfo().username}/nni/script`;
     }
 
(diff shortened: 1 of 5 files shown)


src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -480,23 +450,18 @@ class RemoteMachineTrainingService implements TrainingService {
...
-        await this.generateGpuMetricsCollectorScript(rmMeta.username);
         const nniRootDir: string = unixPathJoin(getRemoteTmpDir(this.remoteOS), 'nni');
         await SSHClientUtility.remoteExeCommand(`mkdir -p ${this.remoteExpRootDir}`, conn);
 
-        // Copy NNI scripts to remote expeirment working directory
-        const localGpuScriptCollectorDir: string = this.getLocalGpuMetricCollectorDir();
         // the directory to store temp scripts in remote machine
         const remoteGpuScriptCollectorDir: string = this.getRemoteScriptsPath(rmMeta.username);
-        await SSHClientUtility.remoteExeCommand(`mkdir -p ${remoteGpuScriptCollectorDir}`, conn);
+        await SSHClientUtility.remoteExeCommand(`(umask 0 ; mkdir -p ${remoteGpuScriptCollectorDir})`, conn);
```
QuanluZhang: no need to add umask, because this `unixPathJoin(getRemoteTmpDir(this.remoteOS), userName, 'nni', 'scripts');` is how `remoteGpuScriptCollectorDir` come from, `userName` in it.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -480,23 +450,18 @@ class RemoteMachineTrainingService implements TrainingService {
...
         await SSHClientUtility.remoteExeCommand(`chmod 777 ${nniRootDir} ${nniRootDir}/* ${nniRootDir}/scripts/*`, conn);
```
QuanluZhang: what is `nniRootDir` used for? @SparkSnail

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -480,23 +450,18 @@ class RemoteMachineTrainingService implements TrainingService {
     private async initRemoteMachineOnConnected(rmMeta: RemoteMachineMeta, conn: Client): Promise<void> {
         // Create root working directory after ssh connection is ready
         // generate gpu script in local machine first, will copy to remote machine later
```
QuanluZhang: this line can be removed

src/nni_manager/training_service/local/gpuScheduler.ts
```diff
@@ -101,12 +101,7 @@ class GPUScheduler {
      */
     private async runGpuMetricsCollectorScript(): Promise<void> {
         await execMkdir(this.gpuMetricCollectorScriptFolder, true);
```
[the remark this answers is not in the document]
QuanluZhang: `this.gpuMetricCollectorScriptFolder` is `${os.tmpdir()}/nni/script`, I think it is better to add username in it.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
SparkSnail: It is the dir used in remote machine of NNI. Normally it is `/tmp/nni` folder, the folder is used for nni trial logs and gpu script files.

src/nni_manager/training_service/local/gpuScheduler.ts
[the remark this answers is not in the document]
liuzhe-lz: `true` means `share=True`, so `umask` is 0 here. As a magic value `true` is hard to understand and I'll add some comments.
This folder is to store collected metrics. So it cannot have user name in it so long as the collector script runs exclusively. Consider user A starts a collector and it writes metrics to `/tmp/nni/A`; then user B will fail to start another collector and she does not know where the metrics locates in.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
liuzhe-lz: Oh, then we are in trouble.
As stated in comment above, one machine can only have one collector running, which makes it impossible for the second user to collect GPU metrics.

liuzhe-lz: According to blame this behaviour was introduced 10 months ago. I will be surprised if a bug can be hidden for so long.
I don't have a remote GPU machine for now. I'll check it on Monday.
I have confirmed `pgrep -fx` will detect process launched by other users though.

liuzhe-lz:
Updated. Now each user will have a seperate GPU metrics collector.

PR #1709, opened 2019-11-06
Installation test pipeline

chicm-ms:
Only test pip install for now, will add more installation test later.
This test is scheduled daily.

### DIFF
--- a/test/pipelines-it-installation.yml
+++ b/test/pipelines-it-installation.yml
@@ -28,6 +28,8 @@ jobs:
 
   steps:
   - script: |
+      python3 -V
+      python3 -m pip install --upgrade pip setuptools --user
       make dev-install
     displayName: 'make dev install'
   - script: |
--- a/test/pipelines-it-installation.yml
+++ b/test/pipelines-it-installation.yml
@@ -34,7 +34,7 @@ jobs:
     displayName: 'Install nni'
   - script: |
       cd examples/trials/mnist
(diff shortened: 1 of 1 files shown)


test/pipelines-it-installation.yml
```diff
@@ -0,0 +1,49 @@
...
+  strategy:
+    matrix:
+      Python36:
+        PYTHON_VERSION: '3.6'
+
+  steps:
+  - script: |
+      python3 -V
+      python3 -m pip install --upgrade pip setuptools --user
+      python3 -m pip install --upgrade nni --user
```
liuzhe-lz: This only tests released version right?

chicm-ms: yes.

liuzhe-lz: So when this test case fails, we have already published a broken release...

chicm-ms: This is just a start point, we can add more later, for example:
1. pre-release testpypi test
2. make dev install
3. install for different platforms (python version, virtual envs)
4. anything related installation not covered by IT.

test/pipelines-it-installation.yml
```diff
@@ -0,0 +1,49 @@
...
+    vmImage: 'windows-latest'
+  strategy:
+    matrix:
+      Python36:
+        PYTHON_VERSION: '3.6'
+
+  steps:
+  - script: |
+      python -V
+      python3 -m pip install --upgrade pip setuptools --user
```
SparkSnail: why `python3`?

chicm-ms: Thanks, fixed.

Issue #1589, opened 2019-10-08
ValueError: Out of range float values are not JSON compliant


martsalz:
**Short summary about the issue/question**:

When executing an experiment, the following error message appears for some trials:

`ValueError: Out of range float values are not JSON compliant`

What's the reason for this?

**nni Environment**:
- nni version: v0.8-320-g421065b
- nni mode(local|pai|remote): local
- OS: CentOS 7
- python version: Python 3.6
- is conda or virtualenv used?: virtualenv 
- is running in docker?: no

**Anything else we need to know**:

stderr:
```
Using TensorFlow backend.
Traceback (most recent call last):
  File "test.py", line 164, in <module>
    train(ARGS, RECEIVED_PARAMS)
  File "test.py", line 136, in train
    validation_data=(x_test, y_test), callbacks=[SendMetrics(), TensorBoard(log_dir=TENSORBOARD_DIR)])
  File "/home/msalz/venv_nni_dev/lib64/python3.6/site-packages/keras/engine/training.py", line 1178, in fit
    validation_freq=validation_freq)
  File "/home/msalz/venv_nni_dev/lib64/python3.6/site-packages/keras/engine/training_arrays.py", line 224, in fit_loop
    callbacks.on_epoch_end(epoch, epoch_logs)
  File "/home/msalz/venv_nni_dev/lib64/python3.6/site-packages/keras/callbacks.py", line 152, in on_epoch_end
    callback.on_epoch_end(epoch, logs)
  File "test.py", line 119, in on_epoch_end
    nni.report_intermediate_result(logs["val_loss"])
  File "/home/msalz/nni/build/nni/trial.py", line 81, in report_intermediate_result
    'value': metric
  File "/home/msalz/venv_nni_dev/lib64/python3.6/site-packages/json_tricks/nonp.py", line 99, in dumps
    primitives=primitives, fallback_encoders=fallback_encoders, **jsonkwargs).encode(obj)
  File "/usr/lib64/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
ValueError: Out of range float values are not JSON compliant
```
```
authorName: default
experimentName: example_mnist
maxExecDuration: 2h
maxTrialNum: 1000
trialConcurrency: 2
localConfig:
    useActiveGpu: true
    maxTrialNumPerGpu: 2
[...]

PR #1711, opened 2019-11-06
Allow serializing NaN in SDK

liuzhe-lz:
I found `json_tricks` has a parameter `allow_nan`, which can hopefully solve issue #1589 .
But this may introduce other problems, because NaN is not part of JSON standard and cannot be deserialized in major browsers or Node.js.

Update: Maybe we can use [JSON5](https://www.npmjs.com/package/json5) in TypeScript side. However our JSONs can be really large so we should assess its performance before using it.

The pipeline failed for "cannot find module sqlite3". Seems we have some version problems again. @chicm-ms


QuanluZhang:
this change may break nniDataStore and webui.

liuzhe-lz:
Implemented in PR #1958

PR #1713, opened 2019-11-07
Update trial doc string

chicm-ms:


src/sdk/pynni/nni/trial.py
```diff
@@ -43,15 +43,34 @@
...
+    Get the hyper paremeters generated by tuner. For a multiphase experiment, it returns a new group of hyper
+    parameters at each call of get_next_parameter. For a non-multiphase (multiPhase is not configured or set to False),
+    it returns hyper parameters only on the first call for each trial job, it returns None since second call.
+
+    Returns
+    -------
+    dict
+        A dict object contains the hyper parameters generated by tuner, the keys of the dict are defined in
+        search space.
+        Returns None if no more hyper parameters can be generated by tuner.
```
liuzhe-lz: When will a trial start if tuner has no more parameters?

src/sdk/pynni/nni/trial.py
```diff
@@ -60,19 +79,49 @@ def get_current_parameter(tag=None):
     return _params['parameters'][tag]
 
 def get_experiment_id():
+    """
+    Get experiment ID.
+
+    Returns
+    -------
+    str
+      Identifier of current experiment
```
liuzhe-lz: Four spaces.

src/sdk/pynni/nni/trial.py
chicm-ms: Updated some wording.   A trial starts before get_next_parameter is called inside trial. For non-multiphase trial, get_next_parameter returns None since the second call,  what is the suggestion?

src/sdk/pynni/nni/trial.py
chicm-ms: Thanks, updated.

src/sdk/pynni/nni/trial.py
```diff
@@ -60,19 +79,49 @@ def get_current_parameter(tag=None):
...
+    Returns
+    -------
+    str
+        Identifier of current trial job which is calling this API.
+    """
     return _trial_id
 
 def get_sequence_id():
+    """
+    Get trial job sequence nubmer. A sequence number is an integer value assigned to each trial job base on the
```
xuehui1991: What's the difference between trial id and sequence id?
Can you give an example here?

chicm-ms: Thanks, example and more details added.

PR #1719, opened 2019-11-08
NAS refactoring

squirrelsc:
1. use register/hook pattern to replace name depended hook, to make it's more clear and reduce troubleshooting from weak type.
2. rename PyTorchMutable to mutable for future extensibility, if tensorflow supports.


src/sdk/pynni/nni/nas/pytorch/mutables.py
```diff
@@ -15,12 +15,14 @@ class PyTorchMutable(nn.Module):
...
         super().__init__()
-        if key is not None:
-            self.key = key
-        else:
-            self.key = self.__class__.__name__ + str(global_mutable_counting())
+        for item_key, item_value in kwargs.items():
+            if item_key == "key":
+                self.key = kwargs.get(item_key, self.__class__.__name__ + str(global_mutable_counting()))
+            else:
+                self.__setattr__(item_key, item_value)
```
ultmaster: Could you elaborate on this else branch? For what purpose exactly?

src/sdk/pynni/nni/nas/pytorch/darts/mutator.py
```diff
@@ -3,16 +3,19 @@
...
+from nni.nas.pytorch.mutator import Mutator
 
 
-class DartsMutator(PyTorchMutator):
+class DartsMutator(Mutator):
 
     def before_build(self, model):
         self.choices = nn.ParameterDict()
+        self.register_on_init_hook(LayerChoice, self.on_init_layer_choice)
+        self.register_on_forward_hook(LayerChoice, self.on_forward_layer_choice)
```
ultmaster: I agree this is a better design, but probably we should include these two lines in `__init__`.

ultmaster:
Merge this before merge mine. I'll resolve the conflicts.

PR #1726, opened 2019-11-11
Show experiment name in nnictl experiment list

### LINKED ISSUE #1615: More information in experiment list
1. Maybe we should display the name of the experiment here:

```
$ nnictl experiment list
----------------------------------------------------------------------------------------
                Experiment information
Id: ScIo49BD    Status: RUNNING    Port: 9098    Platform: pai    StartTime: 2019/10/16 14:32:53    EndTime: N/A
Id: qp1Ngawp    Status: RUNNING    Port: 9090    Platform: pai    StartTime: 2019/10/16 14:46:11    EndTime: N/A
Id: ZJ8YhJM5    Status: RUNNING    Port: 10000    Platform: pai    StartTime: 2019/10/16 14:48:09    EndTime: N/A
```

Currently there is no way I can tell the difference other than examining them one by one.

2. Maybe we should ask users whether they are sure to stop all the experiments when they execute `nnictl stop --all`.

### LINKED ISSUE #1628: nnictl experiment list --all should show experiment name as well

At present 
```bash
nnictl experiment list --all
```
shows all the previous experiments however they are listed by experiment ID.

It would be good to display the experiment name as well

**Why is this needed**:

Since experiment Id is auto generated it is difficult to remember to which experiment it belongs. 

I have to go in all the experiments to figure out which is the one that I need to look at.

**Brief description of your proposal if any**:

One straightforward approach is to add a column with experiment names. These experiment names do not need to unique. User should be able to combine experiment name & date to figure out which one he/she is interested in.

Other approach is to have another command that lists a table that shows the mapping between experiment names and auto generated Ids.

ultmaster:
Fix #1615. Fix #1628.

Code in `nnictl` could need some refactor. (Many duplicates)


tools/nni_cmd/nnictl_utils.py
```diff
@@ -573,8 +581,13 @@ def experiment_list(args):
             print_warning('There is no experiment running...\nYou can use \'nnictl experiment list --all\' to list all experiments.')
     experiment_information = ""
     for key in experiment_id_list:
-        experiment_information += (EXPERIMENT_DETAIL_FORMAT % (key, experiment_dict[key]['status'], experiment_dict[key]['port'],\
-        experiment_dict[key].get('platform'), experiment_dict[key]['startTime'], experiment_dict[key]['endTime']))
+        experiment_information += EXPERIMENT_DETAIL_FORMAT % (key,
+                                                              experiment_dict[key].get('experimentName', 'N/A'),
```
xuehui1991: why add a new line in 585

ultmaster: To make it neat, symmetrical and beautiful.

tools/nni_cmd/config_utils.py
```diff
@@ -81,6 +81,8 @@ def add_experiment(self, expId, port, time, file_name, platform):
         self.experiments[expId]['status'] = 'INITIALIZED'
         self.experiments[expId]['fileName'] = file_name
         self.experiments[expId]['platform'] = platform
+        self.experiments[expId]['authorName'] = author_name
```
SparkSnail: what is `author_name` used for?

ultmaster: It's an important metadata. Save it for future use. Also okay to eject. Suggestions?

SparkSnail: Don't know if it will be used in the future, do not suggest to maintain unuseful information in metadata.

PR #1730, opened 2019-11-12
pdarts implementation (export is not included)

squirrelsc:


src/sdk/pynni/nni/nas/pytorch/mutables.py
```diff
@@ -109,11 +112,12 @@ def forward(self, optional_inputs, semantic_labels=None):
             "Length of the input list must be equal to number of candidates."
         if semantic_labels is None:
             semantic_labels = ["default_label"] * self.n_candidates
-        out, mask = self.mutator.on_forward(self, optional_inputs, semantic_labels)
+        out, mask = self.mutator.on_forward(
+            self, optional_inputs, semantic_labels)
```
ultmaster: why?

src/sdk/pynni/nni/nas/pytorch/mutables.py
```diff
@@ -56,8 +56,8 @@ def _check_built(self):
                 "Mutator not set for {}. Did you initialize a mutable on the fly in forward pass? Move to __init__"
                 "so that trainer can locate all your mutables. See NNI docs for more details.".format(self))
 
-    def __repr__(self):
-        return "{} ({})".format(self.name, self.key)
+    # def __repr__(self):
+    #     return "{} ({})".format(self.name, self.key)
```
ultmaster: Can you do a better repr instead of delete it?

src/sdk/pynni/nni/nas/pytorch/modules.py
```diff
@@ -0,0 +1,9 @@
+
+from torch import nn as nn
+
+
+class RankedModule(nn.Module):
```
ultmaster: What is RankedModule?

src/sdk/pynni/nni/nas/pytorch/pdarts/mutator.py
```diff
@@ -0,0 +1,99 @@
...
+            if switches[index] == False:
+                del(mutable.choices[index])
+                mutable.length -= 1
+
+        self.switches[mutable.key] = switches
+
+        self.choices[mutable.key] = nn.Parameter(
+            1.0E-3 * torch.randn(mutable.length))
+
+    def on_calc_layer_choice_mask(self, mutable: LayerChoice):
```
ultmaster: I believe this is never called.

src/sdk/pynni/nni/nas/pytorch/mutables.py
squirrelsc: the default __repr__ is good enough, it's easy to compare with original network structure.
If we want some additional information, it can be another side method.

src/sdk/pynni/nni/nas/pytorch/modules.py
squirrelsc: like cell, it marks difference rank/level of group. for example, in darts, cell is rank=1, and ops can be treat like rank=2. It can support macro/micro pattern better.

src/sdk/pynni/nni/nas/pytorch/mutables.py
squirrelsc: Auto format. We can use same tool (vs code?), or have the same line width setting.

ultmaster: You might need to set line width to 140, which is defined in our pylint.

src/sdk/pynni/nni/nas/pytorch/mutables.py
ultmaster: Then delete it instead of commenting.

src/sdk/pynni/nni/nas/pytorch/modules.py
ultmaster: Where is this `rank/level` used?

squirrelsc: I thought to update on existing network for pdarts implementation, so added it. But finally, to get exact implementation, the pdarts recreate network each time. So it's not used in this time, but I believe it's a good mark for future implementation.

ultmaster: Looks "over-engineering" to me. I'm fine with it. Just make everything clear in the docstring. :)

src/sdk/pynni/nni/nas/pytorch/mutables.py
squirrelsc: removed

PR #1739, opened 2019-11-16
Extract base mutator/trainer and support ENAS micro search space

ultmaster:


examples/nas/enas/search.py
```diff
@@ -0,0 +1,47 @@
...
+                               metrics=accuracy,
+                               reward_function=reward_accuracy,
+                               optimizer=optimizer,
+                               callbacks=[LearningRateScheduler(lr_scheduler), ArchitectureCheckpoint("./checkpoints")],
+                               batch_size=args.batch_size,
+                               num_epochs=num_epochs,
+                               dataset_train=dataset_train,
+                               dataset_valid=dataset_valid,
+                               log_frequency=args.log_frequency)
+    trainer.train_and_validate()
```
QuanluZhang: train and validate?

src/sdk/pynni/nni/nas/pytorch/callbacks.py
```diff
@@ -0,0 +1,69 @@
+import json
+import logging
+import os
+
+import torch
+
+_logger = logging.getLogger(__name__)
+
+
+class Callback:
```
QuanluZhang: not a good name, better to make it clear that callback for what

src/sdk/pynni/nni/nas/pytorch/callbacks.py
```diff
@@ -0,0 +1,69 @@
...
+        pass
+
+    def on_batch_begin(self, epoch):
+        pass
+
+    def on_batch_end(self, epoch):
+        pass
+
+
+class LearningRateScheduler(Callback):
```
QuanluZhang: why put it here?

examples/nas/darts/model.py
```diff
@@ -0,0 +1,149 @@
...
+
+class Node(darts.DartsNode):
+    def __init__(self, node_id, num_prev_nodes, channels, num_downsample_connect, drop_path_prob=0.):
+        super().__init__(node_id, limitation=2)
+        self.ops = nn.ModuleList()
+        for i in range(num_prev_nodes):
+            stride = 2 if i < num_downsample_connect else 1
+            self.ops.append(
+                mutables.LayerChoice(
+                    [
```
squirrelsc: Where is the none - Zero op?

src/sdk/pynni/nni/nas/pytorch/callbacks.py
ultmaster: The design is from: https://github.com/keras-team/keras/blob/master/keras/callbacks/callbacks.py. Only a course version here. Will refactor later.

examples/nas/enas/search.py
ultmaster: You can `train()` or `validate()` or `train_and_validate()`.

src/sdk/pynni/nni/nas/pytorch/callbacks.py
ultmaster: https://github.com/keras-team/keras/blob/7a39b6c62d43c25472b2c2476bd2a8983ae4f682/keras/callbacks/callbacks.py#L910

examples/nas/darts/model.py
ultmaster: Because DARTS does not regard Zero as a valid op in final architecture, Zero can be thought of as something that is needed by search process. So zero is moved into mutator now.

PR #1744, opened 2019-11-18
NAS documentation initation

squirrelsc:


QuanluZhang:
looks great, better to briefly introduce the nas user interface here.

PR #1746, opened 2019-11-18
Fpgm algo implementation unit test

chicm-ms:
1.  Unit test for fpgm pruner
2.  Fix fpgm tensforflow 2.0 implementation bug found using the unit test.


src/sdk/pynni/tests/test_compressor.py
```diff
@@ -53,11 +53,6 @@ def test_torch_pruner(self):
         configure_list = [{'sparsity': 0.8, 'op_types': ['default']}]
         torch_compressor.LevelPruner(model, configure_list).compress()
 
-    def test_torch_fpgm_pruner(self):
```
xuehui1991: why remove from this file? why fgpm test need to put into a new file?

chicm-ms: Thanks for your question,  after rethinking, I think it looks better to put them together.
So the new file is removed.

PR #1759, opened 2019-11-20
modify docs

LongzeSong:
Added the option `searchSpacePath` in the yaml configuration file of the document. If this option is missing, it will cause nni to report an error.

### DIFF
--- a/docs/en_US/TrainingService/PaiMode.md
+++ b/docs/en_US/TrainingService/PaiMode.md
@@ -20,7 +20,7 @@ maxTrialNum: 100
 # choice: local, remote, pai
 trainingServicePlatform: pai
 # search space file
-searchSpacePath: nni_search_space.json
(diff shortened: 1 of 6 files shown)


docs/en_US/TrainingService/PaiMode.md
```diff
@@ -19,6 +19,8 @@ maxExecDuration: 3h
 maxTrialNum: 100
 # choice: local, remote, pai
 trainingServicePlatform: pai
+# search space file
+searchSpacePath: nni_search_space.json
```
xuehui1991: It seems all the name of search space file in nni examples all called "search_space.json".

So I suggest "nni_search_space.json" -> "search_space.json"

LongzeSong: ok~

PR #1760, opened 2019-11-21
Add MNIST example for TensorFlow v2.x

liuzhe-lz:


examples/trials/mnist-tfv2/mnist.py
```diff
@@ -0,0 +1,140 @@
...
+    optimizer = Adam(learning_rate=params['learning_rate'])
+    model.compile(optimizer=optimizer, loss='sparse_categorical_crossentropy', metrics=['accuracy'])
+
+    (x_train, y_train), (x_test, y_test) = load_dataset()
+
+    model.fit(
+        x_train,
+        y_train,
+        batch_size=params['batch_size'],
+        epochs=1,
```
liuzhe-lz: How many epochs should we use?

liuzhe-lz:
@chicm-ms Is `cli_test.py` hard-coded on testing machine? Please update the example path.

examples/trials/mnist-tfv2/mnist.py
chicm-ms: 10-50 should be fine for MNIST

chicm-ms:
The UT error in this PR has been fixed by https://github.com/microsoft/nni/pull/1746

chicm-ms:
> @chicm-ms Is `cli_test.py` hard-coded on testing machine? Please update the example path.

The example used by cli_test.py is defined in:
`test\config_test\examples\mnist.test.yml`

examples/trials/mnist-tfv2/mnist.py
```diff
@@ -0,0 +1,164 @@
...
+            _logger.info(
+                'Epoch: %s  Batch: %s  Intermediate accuracy: %s',
+                self._epoch, batch, logs['accuracy']
+            )
+            nni.report_intermediate_result(logs['accuracy'])  # send accuracy to NNI framework
+
+    def on_epoch_end(self, epoch, logs=None):
+        """To log current epoch"""
+        self._epoch += 1
+        self._steps = 0
```
chicm-ms: Normally we do not use step and epoch at the same time, for small dataset like MNIST, it is OK to just use epoch for simplicity.

liuzhe-lz: Because the batch size is generated by tuner, an epoch may have different number of mini-batches.
So if we report intermediate result every mini-batch, trials will have different length of intermediate results.
I'm trying to find a more elegant way...

ultmaster: It's more natural to report validation accuracy periodically instead of training accuracy, which requires doing validation periodically.

chicm-ms: you can report intermediate results on epoch end.

liuzhe-lz: Or should we use batch size to calculate epoch number instead?
It looks unfair to me if trials have different epochs, but this is what TF v1 example did (if I understood it correctly).

ultmaster: If the whole story is about steps (there is no such thing as epochs), then it's fair. If you are counting epochs, then perhaps you should forget about steps.

liuzhe-lz: I just found out that we can simply send intermediate result on epoch end...
Anyway we still need to decide whether to fix epoch number or to fix mini-batch number (`steps_per_epoch`).

chicm-ms: we do not need to fix epoch number,  it's fine to be unfair,  or we can simply remove epochs from hyper parameters.

liuzhe-lz: Sorry I missed some conversation because the page was not auto-refreshed...
Updating.

examples/trials/mnist-tfv2/mnist.py
```diff
@@ -0,0 +1,146 @@
...
+    nni.report_final_result(result[1])  # send final accuracy to NNI tuner and web UI
+    _logger.info('Final accuracy reported: %s', result[1])
+
+
+if __name__ == '__main__':
+    params = {
+        'dropout_rate': 0.5,
+        'conv_size': 5,
+        'hidden_size': 1024,
+        'batch_size': 100,
```
liuzhe-lz: For simplicity I don't want to write argparse in example code.
So use a large batch size here for faster UT...

chicm-ms:
Most likely this PR will break some integration tests,  submit to a dev branch and run integration test?

liuzhe-lz:
Merging to dev branch.

PR #1767, opened 2019-11-21
Fix fpgm

chicm-ms:
The current implementation calculates number of filters incorrectly, this PR fixes the problem.


QuanluZhang:
@tanglang96 please help review

src/sdk/pynni/nni/compression/tensorflow/builtin_pruners.py
```diff
@@ -155,40 +155,35 @@ def calc_mask(self, layer, config):
             return self.mask_dict.get(layer.name)
 
         try:
-            weight = tf.stop_gradient(tf.transpose(weight, [2, 3, 0, 1]))
-            masks = np.ones(weight.shape)
-
-            num_kernels = weight.shape[0] * weight.shape[1]
+            w = tf.stop_gradient(tf.transpose(tf.reshape(weight, (-1, weight.shape[-1])), [1, 0]))
+            masks = np.ones(w.shape)
+            num_kernels = w.shape[0]
```
tanglang96: as fpgm is short for Filter Pruning via Geometric Median, maybe rename `num_kernels`->`num_filters` could be more clear

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
```diff
@@ -257,22 +256,16 @@ def _get_distance_sum(self, weight, in_idx, out_idx):
...
-            w = weight.view(-1, weight.size(-1))
-            anchor_w = weight[out_idx, in_idx].unsqueeze(0).expand(w.size(0), w.size(1))
-        else:
-            raise RuntimeError('unsupported layer type')
+        assert len(weight.size()) in [3, 4], 'unsupported weight shape'
+
+        w = weight.view(weight.size(0), -1)
+        anchor_w = w[out_idx].unsqueeze(0).expand(w.size(0), w.size(1))
         x = w - anchor_w
         x = (x * x).sum((-2, -1))
```
tanglang96: why here is `x = (x * x).sum((-2, -1))`? x should be a matrix with shape `c_out * (c_in * k * k)`, to get the distance sum, maybe it should be `x = (x * x).sum(-1)` and then sqrt and sum.

chicm-ms: Thanks, I forgot to update here after reshape.

src/sdk/pynni/nni/compression/tensorflow/builtin_pruners.py
chicm-ms: ok

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
chicm-ms: It still can pass UT because for the UT case, the mistake changed the value of distance, but the sort order of distances are still correct.   We can see that the UT case are not strong enough.  Should be improved later.

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
```diff
@@ -215,7 +215,7 @@ def calc_mask(self, layer, config):
         masks = torch.ones(weight.size()).type_as(weight)
 
         try:
-            num_kernels = weight.size(0) * weight.size(1)
+            num_kernels = weight.size(0)
```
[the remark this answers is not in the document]
chicm-ms: thanks, updated.

Issue #1740, opened 2019-11-18
output log on stdout in "standalone" mode


QuanluZhang:
**What would you like to be added**: model compression and one-shot NAS in NNI SDK can run with a single trial without starting an NNI experiment. This can be called "standalone" mode, in this mode, better to output some log content to stdout.

**Why is this needed**: more user friendly

**Without this feature, how does current nni work**：check the log

**Components that may involve changes**: log component

PR #1768, opened 2019-11-21
Improve logging for standalone mode

liuzhe-lz:
This is a improved version of PR #1761 , which fixes issue #1740

### DIFF
--- a/src/sdk/pynni/nni/trial.py
+++ b/src/sdk/pynni/nni/trial.py
@@ -126,9 +126,10 @@ def report_intermediate_result(metric):
         serializable object.
     """
     global _intermediate_seq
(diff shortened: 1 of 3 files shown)


src/sdk/pynni/nni/trial.py
```diff
@@ -147,9 +148,10 @@ def report_final_result(metric):
     metric:
         serializable object.
     """
-    assert _params is not None, 'nni.get_next_parameter() needs to be called before report_final_result'
+    assert _params or trial_env_vars.NNI_PLATFORM is None, \
```
ultmaster: Why `_params is not None` is now `_params`?

liuzhe-lz: I believe `_params` must not be `{}` either.

PR #1771, opened 2019-11-22
add pruner unit test

tanglang96:
add l1filter_pruner and slim_pruner unit test

### DIFF
--- a/docs/en_US/Compressor/SlimPruner.md
+++ b/docs/en_US/Compressor/SlimPruner.md
@@ -34,6 +34,6 @@ We implemented one of the experiments in ['Learning Efficient Convolutional Netw
 | Model         | Error(paper/ours) | Parameters | Pruned    |
 | ------------- | ----------------- | ---------- | --------- |
 | VGGNet        | 6.34/6.40     | 20.04M   |           |
-| Pruned-VGGNet | 6.20/6.39     | 2.03M    | 88.5% |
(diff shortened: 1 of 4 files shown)


docs/en_US/Compressor/SlimPruner.md
```diff
@@ -34,6 +34,6 @@ We implemented one of the experiments in ['Learning Efficient Convolutional Netw
 | Model         | Error(paper/ours) | Parameters | Pruned    |
 | ------------- | ----------------- | ---------- | --------- |
 | VGGNet        | 6.34/6.40     | 20.04M   |           |
-| Pruned-VGGNet | 6.20/6.39     | 2.03M    | 88.5% |
+| Pruned-VGGNet | 6.20/6.26     | 2.03M    | 88.5% |
```
xuehui1991: could you provide some settings in those experiments? like hyper-parameters.

examples/model_compress/slim_pruner_torch_vgg19.py
```diff
@@ -169,7 +169,7 @@ def main():
     new_model.to(device)
     new_model.load_state_dict(torch.load('pruned_vgg19_cifar10.pth'))
     test(new_model, device, test_loader)
-    # top1 = 93.61%
+    # top1 = 93.74%
```
xuehui1991: does this value will change by some random seed? If will, we should not assume the top1 value is a specific value.

Issue #1764, opened 2019-11-21
cannot sort by Default metric in the Trial jobs tab


kepsail:
In the Trial jobs tab, when I click the (sort by) Default metric button, the results are always sorted by the column Trial No.

kepsail:
In Edge, the results are actually sorted by Trial No.
In Chrome, the results are confused.

lvybriage:
@kepsail Thanks for reporting this bug. I'll fix it soon.

lvybriage:
> In Edge, the results are actually sorted by Trial No.
> In Chrome, the results are confused.

The results are not sorted by `Trial No.` when you sorted by `default metric` column. Because webui sorted `default metric` column by trial's final metric but some trial hadn't final metric. It caused the confuse. I'll sort `default metric` by `final/Iatest intermediate` metric.

PR #1773, opened 2019-11-22
Fix table sort issue

lvybriage:
Fix issue #1764

If a trial haven't any metric, it will always in the final row when we sort these trials.


liuzhe-lz:
What's the root cause of this problem?
Is `a.accuracy` or `b.accuracy` undefined when it's expected to have value?

lvybriage:
> What's the root cause of this problem?
> Is `a.accuracy` or `b.accuracy` undefined when it's expected to have value?

a.accuracy means trial's final metric. but default metric column also has intermediate result.

lvybriage:
> What's the root cause of this problem?
> Is `a.accuracy` or `b.accuracy` undefined when it's expected to have value?

a.accuracy means trial's final metric. but default metric column also has intermediate result.

lvybriage:
> What's the root cause of this problem?
> Is `a.accuracy` or `b.accuracy` undefined when it's expected to have value?

a.accuracy means trial's final metric. but default metric column also has intermediate result.

PR #1786, opened 2019-11-26
Fix img

tanglang96:


chicm-ms:
Is there any other places which using these file names need to be updated?

tanglang96:
@chicm-ms No, docs use `.png`, previous files are `.PNG`, so I modified them all to `.png`

PR #1790, opened 2019-11-26
TensorFlow 2.0 MNIST example, without IT

liuzhe-lz:
Some example trials depend on `tensorflow.examples`, which is not available in `tf.compat.v1`.
So let integration test environment stick to TF 1.x for now.


examples/trials/mnist-tfv2/mnist.py
```diff
@@ -0,0 +1,146 @@
...
+        y_train,
+        batch_size=params['batch_size'],
+        epochs=10,
+        verbose=0,
+        callbacks=[ReportIntermediates()],
+        validation_data=(x_test, y_test)
+    )
+    _logger.info('Training completed')
+
+    result = model.evaluate(x_test, y_test, verbose=0)
```
ultmaster: write `loss, acc = model.evaluate...` here?

examples/trials/mnist-tfv2/config_pai.yml
```diff
@@ -0,0 +1,32 @@
...
+    #choice: maximize, minimize
+    optimize_mode: maximize
+trial:
+  command: python3 mnist.py
+  codeDir: .
+  gpuNum: 0
+  cpuNum: 1
+  memoryMB: 8196
+  #The docker image to run nni job on pai
+  image: msranni/nni:latest
```
ultmaster: Tensorflow 1.x is installed in msranni/nni right?

liuzhe-lz: I think so. Because nobody has upgraded it.

ultmaster: So is it appropriate for us to put an example here that doesn't even work?

liuzhe-lz: I removed docker-based config files. Now this example does work, but only supports limited training services.

liuzhe-lz: Because distributed platforms are not supported, I moved this example down in docs...

PR #1797, opened 2019-11-27
[NAS] fix bug on pdarts

squirrelsc:
After last refactoring, the network cannot be reduced correctly before training. With this fix, it can remove low performance ops correctly.


ultmaster:
Please merge into v1.2

src/sdk/pynni/nni/nas/pytorch/pdarts/mutator.py
```diff
@@ -25,13 +27,13 @@ def __init__(self, model, pdarts_epoch_index, pdarts_num_to_drop, switches={}):
         for mutable in self.mutables:
             if isinstance(mutable, LayerChoice):
 
-                switches = self.switches.get(mutable.key, [True for j in range(mutable.length)])
+                switches = self.switches.get(mutable.key, [True for j in range(mutable.length+1)])
+                switches.requires_grad = False
+                choices = self.choices[mutable.key]
 
-                for index in range(len(switches)-1, -1, -1):
+                for index in range(len(switches)-2, -1, -1):
```
QuanluZhang: better to add comment to explain why `-2`

squirrelsc: added

squirrelsc: new logic is simpler, so no -2 is here.

PR #1804, opened 2019-11-28
refine model compression examples

chicm-ms:


examples/model_compress/main_tf_quantizer.py
```diff
@@ -1,119 +0,0 @@
-from nni.compression.tensorflow import QAT_Quantizer
```
xuehui1991: why remove it ?

chicm-ms: This example has been moved to file QAT_torch_pruner.py

PR #1807, opened 2019-12-02
add sort for detail table column and style

lvybriage:
add sort fucntion for `start time` | `end time`  |  `intermediate result` column;

表格的column自适应，padding-left and padding-right are 15px(出现滚动条时)


ultmaster:
image

Default metrics becomes left align (others center align). By design?

lvybriage:
> image
> 
> Default metrics becomes left align (others center align). By design?

嗯

PR #1814, opened 2019-12-03
Added support for kubeflow 0.7

skyser2003:
Kubeflow 0.7 supports apiVersion v1 for tf-operator

### DIFF
--- a/src/nni_manager/config/kubeflow/tfjob-crd-v1.json
+++ b/src/nni_manager/config/kubeflow/tfjob-crd-v1.json
@@ -1,0 +1,17 @@
+{
+  "kind": "CustomResourceDefinition", 
+  "spec": {
+    "scope": "Namespaced", 
+    "version": "v1", 
+    "group": "kubeflow.org", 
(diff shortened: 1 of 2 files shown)


src/nni_manager/config/kubeflow/tfjob-crd-v1.json
```diff
@@ -0,0 +1,17 @@
...
+    "scope": "Namespaced", 
+    "version": "v1", 
+    "group": "kubeflow.org", 
+    "names": {
+      "kind": "TFJob", 
+      "plural": "tfjobs", 
+      "singular": "tfjob"
+    }
+  }, 
+  "apiVersion": "apiextensions.k8s.io/v1beta1",
```
SparkSnail: v1beta1?

skyser2003: https://github.com/kubeflow/tf-operator/blob/master/examples/crd/crd-v1.yaml
They use it so here.

PR #1815, opened 2019-12-04
change dir of auto-feature-engineering

xuehui1991:


chicm-ms:
There is a reference of the path in README.md

xuehui1991:
> There is a reference of the path in README.md

Thanks, fixed it.

PR #1820, opened 2019-12-04
Pruners refactor

tanglang96:
merge similar filter pruners(L1_norm, L2_norm, FPGM) into a single one


src/sdk/pynni/nni/compression/torch/builtin_pruners.py
```diff
@@ -320,61 +215,65 @@ def calc_mask(self, layer, config):
         weight = layer.module.weight.data
         op_name = layer.name
         op_type = layer.type
-        assert op_type == 'Conv2d', 'L1FilterPruner only supports 2d convolution layer pruning'
+        assert op_type == 'BatchNorm2d', 'SlimPruner only supports 2d batch normalization layer pruning'
         if op_name in self.mask_calculated_ops:
             assert op_name in self.mask_dict
             return self.mask_dict.get(op_name)
         mask = torch.ones(weight.size()).type_as(weight)
         try:
```
chicm-ms: It's unnecessary  to use 'try finally' here.

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
```diff
@@ -5,7 +5,7 @@
 import torch
 from .compressor import Pruner
 
-__all__ = ['LevelPruner', 'AGP_Pruner', 'FPGMPruner', 'L1FilterPruner', 'SlimPruner']
```
chicm-ms: Since L1 pruner and FPGM pruner are merged into rank pruner.
Documents and examples need to be updated as well.

examples/model_compress/rank_pruner_torch_vgg16.py
```diff
@@ -0,0 +1,176 @@
...
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torchvision import datasets, transforms
+from nni.compression.torch import RankFilterPruner
+
+criterion = 'GeometricMedian'
+
+
+class vgg(nn.Module):
```
chicm-ms: It is unnecessary to define vgg architecture by yourself.  Suggest to use pytorch provided vgg model structure to simplify the example.

tanglang96: Pytorch model usually towards ImageNet dataset which is extremely large and I'll put all the models used in the paper and example into a model folder which should be cleaner.

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
tanglang96: Thanks for review, I'll update them later

examples/model_compress/rank_pruner_torch_vgg16.py
chicm-ms: the provided model structure is the same as you defined, you don't have to download the weights.

chicm-ms: https://github.com/pytorch/vision/blob/master/torchvision/models/vgg.py
set pretrained = False, it will be the same as your code.

tanglang96: Pytorch models are pretrained for ImageNet and our implementations/papers are for CIFAR10, they have different weights and different architecture at the fully-connected layers.

chicm-ms: typically you can replace the classifier

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
tanglang96: resolved

src/sdk/pynni/nni/compression/torch/builtin_pruners.py
```diff
@@ -5,7 +5,8 @@
...
+__all__ = ['LevelPruner', 'AGP_Pruner', 'SlimPruner', 'RankFilterPruner', 'L1FilterPruner', 'L2FilterPruner',
```
QuanluZhang: do we still need `RankFilterPruner`?

tanglang96: No it should be `FPGMPruner`, it's my mistake...

QuanluZhang: not updated yet?

tanglang96: updated.

QuanluZhang:
@chicm-ms please help double check the refactor of FPGMPruner

PR #1824, opened 2019-12-05
Switch to eslint

chicm-ms:
Switch from tslint to eslint.
Rules and pipelines will be discussed and submitted in another PR.


azure-pipelines.yml
```diff
@@ -38,8 +38,8 @@ jobs:
     displayName: 'Run pylint'
   - script: |
       python3 -m pip install flake8 --user
-      IGNORE=./tools/nni_annotation/testcase/*:F821,./examples/trials/mnist-nas/*/mnist*.py:F821,./examples/trials/nas_cifar10/src/cifar10/general_child.py:F821
-      python3 -m flake8 . --count --per-file-ignores=$IGNORE --select=E9,F63,F72,F82 --show-source --statistics
+      EXCLUDES=./src/nni_manager/,./tools/nni_annotation/testcase/,./examples/trials/mnist-nas/*/mnist*.py,./examples/trials/nas_cifar10/src/cifar10/general_child.py
```
QuanluZhang: could you briefly explain this change?

chicm-ms: there are some python dependencies in src/nni_manager/node_modules introduced by eslint which make flake8 check fails.
--per-file-ignores can be used to ignore the combination of pattern and error number, but it is inconvenient to ignore all errors in specified directories,  we should use --exclude to replace --per-file-ignores

PR #1828, opened 2019-12-09
For support different python versions

xuehui1991:
This change could make the gradient feature selector could run in multiply python versions.


src/sdk/pynni/nni/feature_engineering/gradient_selector/fginitialize.py
```diff
@@ -33,8 +33,8 @@
 # pylint: disable=E0611
 from torch.utils.data.dataloader import _DataLoaderIter, _utils
 
-import nni.feature_engineering.gradient_selector.constants as constants
-import nni.feature_engineering.gradient_selector.syssettings as syssettings
+from . import constants as constants
```
chicm-ms: remove the `as constants` since they are the same?

xuehui1991: Good suggestion.

PR #1840, opened 2019-12-10
support trial config deduplication in SMAC tuner

QuanluZhang:
when `config_dedup` is True, a configuration will not be generated more than once

### DIFF
--- a/src/sdk/pynni/nni/smac_tuner/smac_tuner.py
+++ b/src/sdk/pynni/nni/smac_tuner/smac_tuner.py
@@ -231,7 +231,9 @@ class SMACTuner(Tuner):
             return self.param_postprocess(init_challenger.get_dictionary())
         else:
             challengers = self.smbo_solver.nni_smac_request_challengers()
+            challengers_empty = True
             for challenger in challengers:
+                challengers_empty = False
                 if self.dedup:
                     match = [v for k, v in self.total_data.items() \
                              if v.get_dictionary() == challenger.get_dictionary()]
@@ -239,7 +241,7 @@ class SMACTuner(Tuner):
                         continue
                 self.total_data[parameter_id] = challenger
                 return self.param_postprocess(challenger.get_dictionary())
-            assert self.dedup is True
+            assert challengers_empty is False, 'The case that challengers is empty is not handled.'
             self.logger.info('In generate_parameters: No more new parameters.')
             raise nni.NoMoreTrialError('No more new parameters.')
 
--- a/src/sdk/pynni/nni/smac_tuner/smac_tuner.py
+++ b/src/sdk/pynni/nni/smac_tuner/smac_tuner.py
@@ -24,6 +24,8 @@ from nni.utils import OptimizeMode, extract_scalar_reward
 
 from .convert_ss_to_scenario import generate_scenario
 
+logger = logging.getLogger('smac_AutoML')
+
 class SMACTuner(Tuner):
     """
     This is a wrapper of [SMAC](https://github.com/automl/SMAC3) following NNI tuner interface.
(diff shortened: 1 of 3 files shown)


src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -226,8 +228,16 @@ def generate_parameters(self, parameter_id, **kwargs):
...
             challengers = self.smbo_solver.nni_smac_request_challengers()
             for challenger in challengers:
+                if self.dedup:
+                    match = [v for k, v in self.total_data.items() \
+                             if v.get_dictionary() == challenger.get_dictionary()]
+                    if match:
+                        continue
                 self.total_data[parameter_id] = challenger
                 return self.param_postprocess(challenger.get_dictionary())
+            assert self.dedup is True
```
chicm-ms: In theory, if challengers is empty,  this assertion could fails if self.dedup is false, but in this situation, we want to raise NoMoreTrialError?

QuanluZhang: according to my understanding, `challengers` would not be empty. if it turns out `challengers` is empty, I want the experiment to crash.

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -18,26 +18,28 @@
...
     This is a wrapper of [SMAC](https://github.com/automl/SMAC3) following NNI tuner interface.
     It only supports ``SMAC`` mode, and does not support the multiple instances of SMAC3 (i.e.,
     the same configuration is run multiple times).
     """
-    def __init__(self, optimize_mode="maximize"):
+    def __init__(self, optimize_mode="maximize", config_dedup=False):
         """
         Parameters
         ----------
         optimize_mode : str
```
xuehui1991: update the doc string?

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
```diff
@@ -226,8 +229,16 @@ def generate_parameters(self, parameter_id, **kwargs):
...
             challengers = self.smbo_solver.nni_smac_request_challengers()
             for challenger in challengers:
+                if self.dedup:
+                    match = [v for k, v in self.total_data.items() \
+                             if v.get_dictionary() == challenger.get_dictionary()]
+                    if match:
+                        continue
                 self.total_data[parameter_id] = challenger
                 return self.param_postprocess(challenger.get_dictionary())
+            assert self.dedup is True
```
xuehui1991: I think we don't need the assert here

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
QuanluZhang: good catch, thanks

src/sdk/pynni/nni/smac_tuner/smac_tuner.py
QuanluZhang: i want it to have a quick fail

xuehui1991: check your logic and you could find it won't happen.

QuanluZhang: what if `challengers` is empty?

chicm-ms: assert challengers not empty before the for loop will make it more straightforward

QuanluZhang: https://github.com/automl/SMAC3/blob/81eba541e67c09aa307f2ba23d583f94c60b6125/smac/optimizer/ei_optimization.py#L699 this is the implementation of challengers, I am not sure whether it is good to use `len(challengers.challengers)`, seems very ugly.

chicm-ms: Got it, I agree that not to use challengers.chanllengers since it is intenal implementation. I am OK with this, or add another boolean variable to record and check whether challengers is empty.

QuanluZhang: updated, thanks.

PR #1841, opened 2019-12-10
Remove unused configuration in doc

SparkSnail:
dataDir and outputDir is not supported in PAI mode.


chicm-ms:
@squirrelsc  I rember we should not update Chinese document, right?

docs/en_US/TrialExample/SquadEvolutionExamples.md
```diff
@@ -112,10 +112,6 @@ trial:
   memoryMB: 32869
   #The docker image to run nni job on OpenPAI
   image: msranni/nni:latest
```
xuehui1991: So how to set our dataDir and outputDir?

SparkSnail: > So how to set our dataDir and outputDir?

PAI does not support setting outputDir from configuration anymore, the dataDir path NNI used is `hdfs://host:port/{username}/nni/{experiments}/{experimentId}/trials/{trialId}/nnioutput`, NNI does not support user setting the path. [refer](https://github.com/microsoft/nni/blob/master/docs/en_US/TrainingService/PaiMode.md#data-management)

SparkSnail:
> @squirrelsc I rember we should not update Chinese document, right?

fixed, removed change on chinese doc.

docs/en_US/TrialExample/SquadEvolutionExamples.md
xuehui1991: @SparkSnail Thanks. But the current usage in PAI seems complicated and troublesome. Can we make it easier?

SparkSnail: > @SparkSnail Thanks. But the current usage in PAI seems complicated and troublesome. Can we make it easier?

Sure, we are support using NFS for NNI to upload data to PAI, it will be easier to use in the future.

PR #1856, opened 2019-12-16
Fix local system as remote machine issue #1852

chicm-ms:


src/nni_manager/training_service/remote_machine/sshClientUtility.ts
```diff
@@ -114,9 +112,6 @@ export namespace SSHClientUtility {
         // Decompress the remote compressed file in and delete it
         await remoteExeCommand(`tar -oxzf ${remoteTarPath} -C ${remoteDirectory}`, sshClient);
         await remoteExeCommand(`rm ${remoteTarPath}`, sshClient);
```
QuanluZhang: why not `return Promise.resolve()`?

chicm-ms: no return value (undefined) is automatically wrapped as Promise<void> since `async` is used.

QuanluZhang: i see

src/nni_manager/training_service/remote_machine/sshClientUtility.ts
```diff
@@ -101,10 +101,8 @@ export namespace SSHClientUtility {
      * @param sshClient SSH client
      */
     export async function copyDirectoryToRemote(localDirectory: string, remoteDirectory: string, sshClient: Client, remoteOS: string): Promise<void> {
-        const deferred: Deferred<void> = new Deferred<void>();
-        const tmpTarName: string = `${uniqueString(10)}.tar.gz`;
-        const localTarPath: string = path.join(os.tmpdir(), tmpTarName);
-        const remoteTarPath: string = unixPathJoin(getRemoteTmpDir(remoteOS), tmpTarName);
+        const localTarPath: string = path.join(os.tmpdir(), `L${uniqueString(10)}.tar.gz`);
```
leckie-chn: let's add a more meaningful name for `tmpTarName` (from @yds05 )

PR #1860, opened 2019-12-17
Update README.md graph

lvybriage:
image


lvybriage:
> the image seems not right

Hi, Scarlett  This picture had been replaced.

PR #1862, opened 2019-12-18
Refactor dispatcher cmdline

chicm-ms:
Passthrough experiment settings to dispatcher, remove the unnecessary configuration parsing in nni manager.

Also fix following issue:
recognize customized tuner as builtin tuner if class name is same as one of builtin tuners.

### DIFF
--- a/src/sdk/pynni/nni/__main__.py
+++ b/src/sdk/pynni/nni/__main__.py
@@ -29,76 +29,55 @@ def augment_classargs(input_class_args, classname):
                 input_class_args[key] = value
     return input_class_args
 
-def create_builtin_class_instance(classname, jsonstr_args, is_advisor=False):
+
+def create_builtin_class_instance(class_name, class_args, is_advisor=False):
     if is_advisor:
-        if classname not in AdvisorModuleName or \
-            importlib.util.find_spec(AdvisorModuleName[classname]) is None:
-            raise RuntimeError('Advisor module is not found: {}'.format(classname))
-        class_module = importlib.import_module(AdvisorModuleName[classname])
-        class_constructor = getattr(class_module, AdvisorClassName[classname])
+        if class_name not in AdvisorModuleName or \
+            importlib.util.find_spec(AdvisorModuleName[class_name]) is None:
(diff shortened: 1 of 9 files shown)


src/sdk/pynni/nni/__main__.py
```diff
@@ -139,17 +105,14 @@ def main():
             raise
 
 
-def _run_advisor(args):
-    if args.advisor_class_name in AdvisorModuleName:
+def _run_advisor(exp_params):
+    if exp_params.get('advisor').get('builtinAdvisorName') in AdvisorModuleName:
```
ultmaster: If `exp_params.get('advisor')` can be None, I think `get` won't help. Should write `exp_params['advisor']` anyway.

chicm-ms: Here the key 'advisor' must exist, it is OK to use either `exp_params['advisor']` or `exp_params.get('advisor')`, just want to be consistent with other places where key may not exists.

src/sdk/pynni/nni/platform/standalone.py
```diff
@@ -14,7 +14,9 @@
     'send_metric',
 ]
 
-init_standalone_logger()
+if trial_env_vars.NNI_PLATFORM is None:
+    init_standalone_logger()
```
QuanluZhang: this file is for standalone mode, `trial_env_vars.NNI_PLATFORM` must be `None`, right?

chicm-ms: My understanding is yes,  we can see `trial_env_vars.NNI_PLATFORM is None` is used to check standalone mode at other places:
https://github.com/microsoft/nni/blob/master/src/sdk/pynni/nni/trial.py#L111
https://github.com/microsoft/nni/blob/master/src/sdk/pynni/nni/platform/__init__.py#L6

chicm-ms: Good catch, the problem is on environment variables, NNI_PLATFORM is an env var for trial not for dispatcher,  I will fix it in another PR.

PR #1868, opened 2019-12-23
TextNAS for SST

ultmaster:
Still debugging. Files to be ejected before merge:

1. searched.json
2. final_arc.json
3. retrain.py if accuracy of retraining is not aligned with paper.

Documentation not ready yet.


examples/nas/textnas/dataloader.py
```diff
@@ -0,0 +1,334 @@
+import logging
```
xuehui1991: do we need a README.md?

ultmaster: Of course. Not ready yet.

examples/nas/textnas/dataloader.py
```diff
@@ -0,0 +1,334 @@
...
+import pickle
+from collections import Counter
+
+import numpy as np
+import torch
+from torch.utils import data
+
+logger = logging.getLogger("nni.textnas")
+
+word_to_word_mapping = {
```
xuehui1991: using upper case?

examples/nas/textnas/dataloader.py
```diff
@@ -0,0 +1,334 @@
...
+
+    def __len__(self):
+        return len(self.sents)
+
+
+def sst_get_id_input(content, word_id_dict, max_input_length):
+    words = content.split(" ")
+    sentence = [word_id_dict["<pad>"]] * max_input_length
+    mask = [0] * max_input_length
+    unknown = word_id_dict["unknown"]
```
xuehui1991: can you unify punctuation?

examples/nas/textnas/dataloader.py
```diff
@@ -0,0 +1,334 @@
...
+    for sentence, _ in phrases:
+        words = sentence.split(' ')
+        for cur_word in words:
+            word = cur_word.strip()
+            word_num_dict[word] += 1
+    return word_num_dict
+
+
+def init_trainable_embedding(embedding_path, word_id_dict):
+    embed_dim = 300
```
xuehui1991: extract `embed_dim` as a parameter?

examples/nas/textnas/dataloader.py
ultmaster: There is indeed a mistake here. But I'm not sure it's related to punctuation. There are punctuations in glove.

examples/nas/textnas/dataloader.py
ultmaster: fixed

examples/nas/textnas/dataloader.py
ultmaster: I don't think the original code supports different wordvec. One solution is to use functions from torchtext, as done in `utils.py`. But I'm having a hard time aligning torchtext with the original dataloader provided by the authors. Thus give up on that one.

ultmaster: Good idea to extract `embed_dim`, but don't support customization here.

xuehui1991: really? Because of even the glove, they have different dimensions version. You don't have to hard code the embed_dim here.

ultmaster:
@pkuyym Style fix is partially done in #1890. Will merge back and finish all when that PR is ready.

PLoic:
Hi, thanks for the contribution. I just have a question about the [ `init_conv` ](https://github.com/microsoft/nni/blob/master/examples/nas/textnas/model.py#L68). If my understanding is right in the paper the result of `init_conv` can be used as input by the 5 first layers. Like this : 
image
 But here it used only by the first layer [L98](https://github.com/ultmaster/nni/blob/nas-textnas/examples/nas/textnas/model.py#L98), is it normal or am I missing something?
Thanks for your help

ultmaster:
Hi @PLoic, thanks for reaching out. As I'm only a helper guy who integrates TextNAS into NNI, I'm not sure I'm qualified to answer your question. /cc @pkuyym as he's one of the authors.

My wild guess is that the conv1x1 in your picture is not `first_conv` in the code. `first_conv` is just a preprocessing layer, which is not shown in paper figures. The data flow should be input -> first_conv -> conv1x1 -> ...

I hope you can open an issue for further discussion. Posting questions in a closed issue might not be such a good idea. :)

PLoic:
Thank you for your fast response ! 
I will continue my research so and open an new issue mentioning the authors :)

PR #1869, opened 2019-12-23
Fix dispatcher log

chicm-ms:


src/sdk/pynni/nni/__init__.py
```diff
@@ -1,14 +1,17 @@
...
+            super().__init__(self)
+            self.errorinfo = ErrorInfo
 
-    def __str__(self):
-        return self.errorinfo
+        def __str__(self):
+            return self.errorinfo
+else:
+    from .trial import *
+    from .smartparam import *
```
ultmaster: I think tuners might want to use functions in smartparam? e.g., `nni.randint`.

chicm-ms: If tuner need to use those distributions, it can use parameter_expressions.py

PR #1874, opened 2019-12-25
Update ant design version to avoid button cloud style

lvybriage:
It's ready to review.
* antd 3.10.10

* @types/react: version 16.8.4 could avoid two questions
   * `Generic type 'ReactElement<P>' requires 1 type argument(s)`
   * more high version is not compatible with `Button` components. [issue](https://github.com/ant-design/ant-design/issues/15700)

* update `yarn.lock` file

### DIFF
--- a/src/webui/package.json
+++ b/src/webui/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "antd": "3.10.9",
+    "antd": "3.10.10",
     "autoprefixer": "^9.6.1",
     "axios": "^0.19.0",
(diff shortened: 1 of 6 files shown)


lvybriage:
Need to fix eslint... This PR is not ready.

chicm-ms:
The diff on yarn.lock is too large, suggest to regenerate it base on the current one.

chicm-ms:
I prefer not to upgrade antd in v1.3.

liuzhe-lz:
Don't think it's necessary.
We are unlikely to using antd in next Christmas.

leckie-chn:
close this PR.

PR #1876, opened 2019-12-25
Fix a few issues related to fixed arc and from-tuner arc

ultmaster:
1. Fixed architecture doesn't handle device any more.
2. Add support for MutableScope in classic NAS.
3. Fix support for InputChoice (validation) in classic NAS.

### DIFF
--- a/src/sdk/pynni/nni/nas/pytorch/fixed.py
+++ b/src/sdk/pynni/nni/nas/pytorch/fixed.py
@@ -41,18 +41,18 @@ class FixedArchitecture(Mutator):
         return self._fixed_arc
 
 
-def _encode_tensor(data, device):
+def _encode_tensor(data):
     if isinstance(data, list):
         if all(map(lambda o: isinstance(o, bool), data)):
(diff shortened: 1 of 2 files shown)


src/sdk/pynni/nni/nas/pytorch/fixed.py
```diff
@@ -41,18 +41,18 @@ def sample_final(self):
         return self._fixed_arc
 
 
-def _encode_tensor(data, device):
+def _encode_tensor(data):
     if isinstance(data, list):
         if all(map(lambda o: isinstance(o, bool), data)):
-            return torch.tensor(data, dtype=torch.bool, device=device)  # pylint: disable=not-callable
+            return torch.tensor(data, dtype=torch.bool)  # pylint: disable=not-callable
```
xuehui1991: why remove all device?

ultmaster: NNI no longer handles device for fixed architecture. Actually in some cases, moving architectures to GPU might cause issues. (see ProxylessNAS)

PR #1882, opened 2019-12-27
refactor code structure of pruning algorithms

QuanluZhang:


src/sdk/pynni/nni/compression/torch/activation_rank_filter_pruners.py
```diff
@@ -0,0 +1,257 @@
...
+
+logger = logging.getLogger('torch activation rank filter pruners')
+
+class ActivationRankFilterPruner(Pruner):
+    """
+    A structured pruning base class that prunes the filters with the smallest
+    importance criterion in convolution layers to achieve a preset level of network sparsity.
+    Hengyuan Hu, Rui Peng, Yu-Wing Tai and Chi-Keung Tang,
+    "Network Trimming: A Data-Driven Neuron Pruning Approach towards Efficient Deep Architectures", ICLR 2016.
+    https://arxiv.org/abs/1607.03250
```
chicm-ms: Can we remove the paper description and reference as this base class does not implement the algos ?

QuanluZhang: good point, thanks.

src/sdk/pynni/nni/compression/torch/activation_rank_filter_pruners.py
```diff
@@ -0,0 +1,257 @@
...
+
+            def _hook(module_, input_, output, name=layer.name):
+                if len(self.collected_activation[name]) < self.statistics_batch_num:
+                    self.collected_activation[name].append(self.activation(output.detach().cpu()))
+
+            layer.module.register_forward_hook(_hook)
+        return self.bound_model
+
+    def _get_mask(self, base_mask, activations, num_prune):
+        return {'weight': None, 'bias': None}
```
chicm-ms: Can we raise NotImplementedError here instead?

QuanluZhang: good point!

QuanluZhang: which func name do you think is better? `_get_mask` or `get_mask`?

chicm-ms: `get_mask` is consistent with calc_mask,   they are both from base class.

src/sdk/pynni/nni/compression/torch/weight_rank_filter_pruners.py
```diff
@@ -0,0 +1,262 @@
...
+
+import logging
+import torch
+from .compressor import Pruner
+
+__all__ = ['L1FilterPruner', 'L2FilterPruner', 'FPGMPruner']
+
+logger = logging.getLogger('torch weight rank filter pruners')
+
+class WeightRankFilterPruner(Pruner):
```
chicm-ms: `WeightRankFilterPruner` is a little bit complex,  maybe `FilterPruner`?

src/sdk/pynni/nni/compression/torch/activation_rank_filter_pruners.py
QuanluZhang: updated, please help review again

src/sdk/pynni/nni/compression/torch/weight_rank_filter_pruners.py
QuanluZhang: I think `WeightRank` is the emphasis.

chicm-ms: Ok, that's fine for me.

PR #1898, opened 2019-12-30
Fix pai http format error, add protocol

SparkSnail:


src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -52,6 +52,7 @@ abstract class PAITrainingService implements TrainingService {
     protected authFileHdfsPath: string | undefined = undefined;
     protected portList?: string | undefined;
     protected paiJobRestServer?: PAIJobRestServer;
+    protected protocol: string = 'http://';
```
chicm-ms: protocol should be `http` instead of `http://`

SparkSnail: fixed

Issue #1802, opened 2019-11-28
Confuse about windows doc.


xuehui1991:
**Short summary about the issue/question**:
Can we run nni manager in one windows machine and remote trials to other windows machines?
The [doc](https://github.com/microsoft/nni/blob/master/docs/en_US/Tutorial/NniOnWindows.md) only mention "Currently we support local, remote and pai mode on Windows. Windows 10.1809 is well tested and recommended." which confuse me.

**Brief what process you are following**: 

**How to reproduce it**: 

**nni Environment**:
- nni version:
- nni mode(local|pai|remote):
- OS:
- python version:
- is conda or virtualenv used?: 
- is running in docker?:

**need to update document(yes/no)**:

**Anything else we need to know**:

scarlett2018:
https://github.com/microsoft/nni/pull/1901

scarlett2018:
fixed in another PR.

PR #1901, opened 2019-12-30
#1802 Update NniOnWindows.md

scarlett2018:
fix problem raised in #1802

### DIFF
--- a/docs/en_US/Tutorial/Installation.md
+++ b/docs/en_US/Tutorial/Installation.md
@@ -1,6 +1,18 @@
-# Installation of NNI
+# Plan your topology
+For different computation needs, NNI supports various out-of-the-box training modes, including: local machine, remote servers, pai, kubeflow and frameworklauncher. User can also customize their own training service in case of need.
 
-Currently we support installation on Linux, Mac and Windows.
+Before starting the installation, you should firstly plan for the topology. For new users, we would always recommend you to start with `local` mode, as it is the simplest topology, all the scripts and computation are happen on one single machine. You only need to install NNI on the machine you will run the training and execute the nni cmdline on that machine. 
+
+But in most cases, you might need more powerful computes on remote servers or the cloud. In the case of using remote servers, you will need to consider `remote` mode or the other training modes NNI supports. As illustrated below, to run trainings on remote servers, you will need to set up a **NNI Manager Node** and a set of **NNI Training Nodes**. The set up are all through the following `pip install` command in the following section.
+
+Note: For Windows, NNI only supports it work as **NNI Manager Node**. Anyone who has interest on support windows as training nodes as well could comment or even provide contributes to the following issue: Support Windows as Remote Training Node [#1973](https://github.com/microsoft/nni/issues/1973).
+
+
+![image](https://user-images.githubusercontent.com/39592018/72716629-82ea5e00-3bad-11ea-8536-b7a21fa22417.png)
+
+# Installation
+
+We support installation on Linux, Mac and Windows.
 
 ## **Installation on Linux & Mac**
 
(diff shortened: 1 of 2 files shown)


docs/en_US/Tutorial/NniOnWindows.md
```diff
@@ -1,6 +1,8 @@
-# NNI on Windows (experimental feature)
+# NNI on Windows
+As Linux version, NNI on Windows supports local, remote as well as cloud integration mode for OpenPAI and other training sercies.
```
xuehui1991: 我觉得这里有个很困惑的点是， 假如我手里有几台windows的机器，我能用一台去启动剩下的remote模式的nni吗？（我觉得有必要在这里指出这一点）

scarlett2018: 明白，等倪总confirm一下@leckie-chn。。。

chicm-ms: @xuehui1991 目前remote machine training service  支持在Windows上跑nni manager去连接remote linux （在linux上跑trial）,   但是不支持Windows 被nni manager连接.    要想在windows系统上跑trial job，只能用local training service.

QuanluZhang: @leckie-chn

leckie-chn: @xuehui1991 NNI windows只支持local模式，换句话说不能把windows机器当作remote的training service :)

xuehui1991: @scarlett2018  Seems we need to change here.

scarlett2018: > @scarlett2018 Seems we need to change here.

Updated

docs/en_US/Tutorial/NniOnWindows.md
```diff
@@ -1,6 +1,8 @@
-# NNI on Windows (experimental feature)
+# NNI on Windows
+Differ from Linux version, NNI only supports local mode right now. Please comment in [NNI's backlog issue](https://github.com/microsoft/nni/issues/1917) if you have needs on other modes for NNI on Windows.
```
QuanluZhang: ... right now. For other modes, you could use Windows as a machine to start an NNI experiment. For example, in remote mode, you could use Windows to run `nnictl` commands and it connects several Linux machines for running trials. Similarly, you could start an NNI experiment on Windows to run trials in PAI mode. But note that Windows cannot act as a remote server for running trials.

scarlett2018: > But note that Windows cannot act as a remote server for running trials.
we might need one diagram to explain nni's runtime nodes, like: nnictl node + trials node(s).

QuanluZhang: agree

scarlett2018: Updated, @xuehui1991, @leckie-chn , @QuanluZhang - please help to review.

docs/en_US/Tutorial/Installation.md
```diff
@@ -1,6 +1,18 @@
-# Installation of NNI
+# Plan your topology
```
squirrelsc: The topology concept makes end users feeling complex, and NNI is not so complex actually. I suggest not to add this section. If people want to implement new trainning platform, they need to know this. So it's more like design related things, shouldn't be put in installation part.

QuanluZhang: agree with @squirrelsc

scarlett2018: @squirrelsc  @QuanluZhang - agree the concern. while still think we need to have this document in a lower exposure place. If you agree this, where would you like to recommend for putting it?

scarlett2018:
As discussed, @squirrelsc will refactor the remote md instead of making this change.

PR #1904, opened 2019-12-30
Create pythonpublish.yml

goodman-ops:


QuanluZhang:
Thanks for your contribution, we already used azure pipeline, no need to add another pipeline.

PR #1909, opened 2019-12-31
merge release branch v1.3 back to master

leckie-chn:


leckie-chn:
use https://github.com/microsoft/nni/pull/1911 instead

PR #1926, opened 2020-01-06
Fix a few issues in EfficientNet introduced by API change and rephrase documentation

ultmaster:


README.md
```diff
@@ -88,6 +88,7 @@ Within the following table, we summarized the current NNI capabilities, we are g
            <li><a href="docs/en_US/TrialExample/GbdtExample.md">Auto-gbdt</a></li>
            <li><a href="docs/en_US/TrialExample/Cifar10Examples.md">Cifar10-pytorch</li></a>
            <li><a href="docs/en_US/TrialExample/SklearnExamples.md">Scikit-learn</a></li>
+           <li><a href="examples/trials/efficientnet">EfficientNet</a></li>
```
xuehui1991: Why not put the doc under the  "docs/en_US/TrialExample/"?

examples/trials/efficientnet/README.md
```diff
@@ -2,18 +2,18 @@
...
 
 ## Instructions
 
 1. Set your working directory here in this directory.
 2. Run `git clone https://github.com/ultmaster/EfficientNet-PyTorch` to clone this modified version of [EfficientNet-PyTorch](https://github.com/lukemelas/EfficientNet-PyTorch). The modifications were done to adhere to the original [Tensorflow version](https://github.com/tensorflow/tpu/tree/master/models/official/efficientnet) as close as possible (including EMA, label smoothing and etc.); also added are the part which gets parameters from tuner and reports intermediate/final results. Clone it into `EfficientNet-PyTorch`; the files like `main.py`, `train_imagenet.sh` will appear inside, as specified in the configuration files.
-3. Run `nnictl create --config config_net.yml` to find the best EfficientNet-B1. Adjust the training service (PAI/local/remote), batch size in the config files according to the environment.
+3. Run `nnictl create --config config_local.yml` (use `config_pai.yml` for PAI) to find the best EfficientNet-B1. Adjust the training service (PAI/local/remote), batch size in the config files according to the environment.
```
xuehui1991: I am sure that we should use PAI or OpenPAI?

README.md
ultmaster: If the docs are to be put under `TrialExample`, they are intended to be read on readthedocs, which means every link on README should be a link to readthedocs website, instead of a link to a markdown on Github. Refactoring involves many changes and should be open to discussion.

xuehui1991: If you concern that you cannot support markdown and readthedocs at the same time, you could put in the path but do not link to readthedocs in [rst file](https://github.com/microsoft/nni/blob/master/docs/en_US/examples.rst).

QuanluZhang: @ultmaster please put it under docs/

PR #1933, opened 2020-01-07
Support monitor mode when creating or resuming a new experiment

SparkSnail:


docs/en_US/Tutorial/Nnictl.md
```diff
@@ -49,6 +49,7 @@ nnictl support commands:
   |--config, -c|  True| |YAML configure file of the experiment|
   |--port, -p|False| |the port of restful server|
   |--debug, -d|False||set debug mode|
+  |--monitor, -m|False|set monitor mode|
```
ultmaster: Missing vertical separator

tools/nni_cmd/nnictl_utils.py
```diff
@@ -623,23 +623,41 @@ def show_experiment_info():
...
+    '''set the experiment monitor engine'''
     while True:
         try:
             os.system('clear')
             update_experiment()
             show_experiment_info()
-            time.sleep(args.time)
+            if auto_exit:
+                status = get_experiment_status(port)
+                if status in ['DONE', 'ERROR', 'STOPPED']:
```
ultmaster: Maybe we can print dispatcher and nnimanager log here (if the status is error). Because if user is running it in a container, when the program exit, the container is destroyed too. There is no way to retrieve the error info. Another option is to disable auto_exit in case `--debug` is set.

tools/nni_cmd/nnictl_utils.py
```diff
@@ -623,23 +623,41 @@ def show_experiment_info():
...
-def monitor_experiment(args):
-    '''monitor the experiment'''
-    if args.time <= 0:
-        print_error('please input a positive integer as time interval, the unit is second.')
-        exit(1)
+def set_monitor(auto_exit, time_interval, port=None, pid=None):
+    '''set the experiment monitor engine'''
     while True:
         try:
             os.system('clear')
```
chicm-ms: Is this working on Windows?

tools/nni_cmd/nnictl_utils.py
SparkSnail: > Maybe we can print dispatcher and nnimanager log here (if the status is error). Because if user is running it in a container, when the program exit, the container is destroyed too. There is no way to retrieve the error info. Another option is to disable auto_exit in case `--debug` is set.

nniManager.log content maybe too long, maybe it's not suitable to show these content in screen.

ultmaster: I think the case is not for screen. For screen user, they will never seek for foreground. For container users, they don't care. Furthermore, I don't think `os.system("clear")` has any effect if they are using container. Printing a lot is expected.

tools/nni_cmd/nnictl_utils.py
SparkSnail: fixed.

docs/en_US/Tutorial/Nnictl.md
SparkSnail: fixed.

Issue #1827, opened 2019-12-07
Refactor the outline of NNI document in readthedocs


QuanluZhang:
Below is a proposal:

![image](https://user-images.githubusercontent.com/16907603/70371651-b3817b80-1910-11ea-931a-49bc9ba145ad.png)

QuanluZhang:
Done

PR #1940, opened 2020-01-10
refactor the index of readthedocs

QuanluZhang:
resolve the issue #1827


docs/en_US/hpo_advanced.rst
```diff
@@ -0,0 +1,9 @@
+Advanced Features
+=================
+
+..  toctree::
+    Enable Multi-phase <AdvancedFeature/MultiPhase>
```
xuehui1991: suggest putting the multi-phase in last.

docs/en_US/hyperparameter_tune.rst
```diff
@@ -0,0 +1,15 @@
+######################
+Hyper-parameter Tuning
+######################
+
+..  toctree::
+    :maxdepth: 2
+
+    Write Trial <TrialExample/Trials>
```
xuehui1991: some of them are not just for HPO. I think the original name "tutorial" would be better.

And some of the doc about HPO doesn't include in.

docs/en_US/reference.rst
```diff
@@ -2,12 +2,9 @@ References
 ==================
```
xuehui1991: suggest changing "reference" to other name

xuehui1991: And put it in the top.

docs/en_US/hyperparameter_tune.rst
QuanluZhang: There is no tutorial now. Currently, other features does not use training service and WebUI. it is not proper to put them in reference. Do you have concrete suggestions?

could you specify "some of the doc"?

docs/en_US/reference.rst
QuanluZhang: could you suggestion for the name? cannot get your idea

docs/en_US/hyperparameter_tune.rst
xuehui1991: 在一开始设计的时候 training service是一个和HPO平行的概念 而不是附属于HPO. 实时上我们network morphsim这种NAS的算法是支持跑在不同的training service上的。 即使新的NAS interface不一定支持跑在例如openpai这样的training service上，我们也不能说以后没有这种可能？

docs/en_US/reference.rst
xuehui1991: 如果是我 我觉得这些是用nni必须知道和了解的内容，所以他们的doc都放在了tutorial这个文件夹下面。 如果是我 我觉得可以把这个名字改成“tutorial.rst”

docs/en_US/hyperparameter_tune.rst
QuanluZhang: yes, you are right. it is not only considering this specific part, we should consider the whole structure. we have four parts, but HPO is the most similar to the previous tutorial, thus we decide to explicitly show these four capabilities like #1827

docs/en_US/hpo_advanced.rst
Ronniechase12: @jodiemurphy13

PR #1954, opened 2020-01-14
Add unit tests for NAS

ultmaster:


src/sdk/pynni/nni/nas/pytorch/darts/mutator.py
```diff
@@ -63,18 +63,21 @@ def sample_final(self):
...
+                    for src_key in mutable.choose_from:
+                        if src_key not in edges_max:
+                            _logger.warning("InputChoice.NO_KEY in '%s' is weighted 0 when selecting inputs.", mutable.key)
+                        weights.append(edges_max.get(src_key, 0.))
+                    weights = torch.tensor(weights)  # pylint: disable=not-callable
+                    _, topk_edge_indices = torch.topk(weights, mutable.n_chosen)
+                    selected_multihot = []
+                    for i, src_key in enumerate(mutable.choose_from):
+                        if i not in topk_edge_indices and src_key in result:
+                            result[src_key] = torch.zeros_like(result[src_key])  # clear this choice to optimize calc graph
```
QuanluZhang: what is the meaning of this line?

src/sdk/pynni/nni/nas/pytorch/utils.py
```diff
@@ -17,6 +17,14 @@ def global_mutable_counting():
     return _counter
 
 
+def reset_global_mutable_counting():
```
QuanluZhang: global counting is still here?

src/sdk/pynni/nni/nas/pytorch/darts/mutator.py
ultmaster: If an edge is never selected, there is no need to calculate any op on this edge. This is to eliminate redundant calculation. If there is any problem with this, open an issue.

src/sdk/pynni/nni/nas/pytorch/utils.py
ultmaster: Changed to private method for future refactor (remove global counting entirely)

QuanluZhang: ok

src/sdk/pynni/nni/nas/pytorch/darts/mutator.py
QuanluZhang: please add comments here to explain.

PR #1956, opened 2020-01-14
Add foreground mode in nnictl

SparkSnail:


liuzhe-lz:
I thought this could be done fully in `nnictl` side.

SparkSnail:
> I thought this could be done fully in `nnictl` side.

The log content of node process is written to log file, if it is operated in `nnictl` side fully, `nnictl` need to watch the log change and read the changed lines, which is not a easy way.

tools/nni_cmd/launcher.py
```diff
@@ -493,8 +501,14 @@ def launch_experiment(args, experiment_config, mode, config_file_name, experimen
                                             experiment_config['experimentName'])
 
     print_normal(EXPERIMENT_SUCCESS_INFO % (experiment_id, '   '.join(web_ui_url_list)))
-    if args.watch:
-        set_monitor(True, 3, args.port, rest_process.pid)
+    if args.foreground:
+        try:
+            while True:
+                log_content = rest_process.stdout.readline().strip().decode('utf-8')
```
QuanluZhang: how about atderr?

tools/nni_cmd/launcher.py
```diff
@@ -120,9 +122,15 @@ def start_rest_server(port, platform, mode, config_file_name, experiment_id=None
         stderr_file.write(log_header)
         if sys.platform == 'win32':
             from subprocess import CREATE_NEW_PROCESS_GROUP
-            process = Popen(cmds, cwd=entry_dir, stdout=stdout_file, stderr=stderr_file, creationflags=CREATE_NEW_PROCESS_GROUP)
+            if args.foreground:
+                process = Popen(cmds, cwd=entry_dir, stdout=PIPE, stderr=STDOUT, creationflags=CREATE_NEW_PROCESS_GROUP)
```
QuanluZhang: a little strange here, for win32, stderr is STDOUT, for linux, stderr is PIPE?

SparkSnail: https://stackoverflow.com/questions/10406532/python-subprocess-output-on-windows  
If I set `stderr=PIPE` in windows, the process will crash, so I redirect `stderr=STDOUT`.

tools/nni_cmd/launcher.py
SparkSnail: There is no stderr content, since we only use `console.log`, while stderr catch the content of `console.error()`.

src/nni_manager/common/log.ts
```diff
@@ -81,6 +82,7 @@ class Logger {
         }
 
         this.readonly = isReadonly();
+        this.foreground = foreground;
```
chicm-ms: Suggest to do a little bit refactor on log.ts:
currently fileName being undefined means to use nnimanager.log as output file. change it to:
1. undefined means send output to STDOUT. 
2. Move file name `nnimanager.log` into main.ts

Then we do not need a `foreground` flag for log.ts.  Because both log files and STDOUT are places where log class can send content to, we do not need an extract flag `foreground` for STDOUT from log class perspective.

SparkSnail: fixed.

PR #1989, opened 2020-02-03
update lottery ticket pruner based on refactored compression code

QuanluZhang:
1. update lottery ticket pruner based on refactored compression code
2. fix bug: allow users to specify device of the model when exporting mask and model, used to place dummy input tensor for exporting onnx model.

### DIFF
--- a/src/sdk/pynni/nni/compression/torch/compressor.py
+++ b/src/sdk/pynni/nni/compression/torch/compressor.py
@@ -263,7 +263,7 @@ class Pruner(Compressor):
     def __init__(self, model, config_list):
         super().__init__(model, config_list)
 
-    def calc_mask(self, layer, config):
+    def calc_mask(self, layer, config, **kwargs):
         """
         Pruners should overload this method to provide mask for weight tensors.
         The mask must have the same shape and type comparing to the weight.
--- a/src/sdk/pynni/nni/compression/torch/compressor.py
+++ b/src/sdk/pynni/nni/compression/torch/compressor.py
@@ -295,7 +295,7 @@ class Pruner(Compressor):
         """
         _logger.info("compressing module %s.", layer.name)
         wrapper = PrunerModuleWrapper(layer.module, layer.name, layer.type, config, self)
-        # TODO: check weight exists
+        assert hasattr(layer.module, 'weight')
         wrapper.to(layer.module.weight.device)
         return wrapper
 
--- a/src/sdk/pynni/nni/compression/torch/compressor.py
+++ b/src/sdk/pynni/nni/compression/torch/compressor.py
@@ -41,6 +41,7 @@ class Compressor:
         self.modules_to_compress = None
         self.modules_wrapper = None
         self.buffers = {}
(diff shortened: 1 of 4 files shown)


src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -341,6 +347,21 @@ def export_model(self, model_path, mask_path=None, onnx_path=None, input_shape=N
...
+        Parameters:
+        -----------
+        model_state : dict
+            state dict saved from unwrapped model
+        """
+        if self.is_wrapped:
+            self._unwrap_model()
+            self.bound_model.load_state_dict(model_state)
+            self._wrap_model()
+        else:
```
Cjkkkk: when will `is_wrapped`  be false ? Before calling `compress` method?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -291,7 +294,10 @@ def _wrap_modules(self, layer, config):
             the configuration for generating the mask
         """
         _logger.info("compressing module %s.", layer.name)
-        return PrunerModuleWrapper(layer.module, layer.name, layer.type, config, self)
+        wrapper = PrunerModuleWrapper(layer.module, layer.name, layer.type, config, self)
+        assert hasattr(layer.module, 'weight')
```
Cjkkkk: why not using `check_weight` method?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -291,7 +294,10 @@ def _wrap_modules(self, layer, config):
...
+        wrapper.to(layer.module.weight.device)
```
Cjkkkk: can you explain this?

src/sdk/pynni/nni/compression/torch/compressor.py
QuanluZhang: yes.
it is possible that users call `load_model_state` before `compress`, we need an approach to check whether the model has been wrapped or not. any suggestion?

src/sdk/pynni/nni/compression/torch/compressor.py
QuanluZhang: where is this method defined, and how to use it?

src/sdk/pynni/nni/compression/torch/compressor.py
QuanluZhang: if the module is placed on GPU, buffers in the wrapped module are put on CPU, thus, I need to move the whole wrapper to the device that the module is placed

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: https://github.com/microsoft/nni/blob/d452a1665f802bfbe75372b6c942272b252c70a2/src/sdk/pynni/nni/compression/torch/compressor.py#L563

PR #1996, opened 2020-02-04
Fix pai examples

SparkSnail:


ultmaster:
Has it passed IT?

SparkSnail:
> Has it passed IT?

There is no code change, it is doc and example change. IT does not test these example files.

examples/trials/nas_cifar10/config_paiYarn_ppo.yml
```diff
@@ -0,0 +1,31 @@
...
+  command: sh ./macro_cifar10_pai.sh
+  codeDir: ./
+  gpuNum: 1
+  cpuNum: 1
+  memoryMB: 8196
+  image: msranni/nni:latest
+  virtualCluster: nni
+paiYarnConfig:
+  userName: your_account
+  passwd: your_passwd
```
ultmaster: passwd or passWord or password?

examples/trials/mnist-advisor/config_pai.yml
```diff
@@ -27,10 +27,13 @@ trial:
...
+  nniManagerNFSMountPath: /home/user/mnt
+  containerNFSMountPath: /mnt/data/user
+  paiStoragePlugin: team_wise
 paiConfig:
   #The username to login pai
   userName: username
-  #The password to login pai
-  passWord: password
+  #The token to login pai
+  token: token
```
xuehui1991: do we have a description of "where could I get the token"?

ultmaster: It's like github access token. It's achievable from PAI user settings. I think users can figure that out.

examples/trials/nas_cifar10/config_paiYarn_ppo.yml
SparkSnail: fixed.

examples/trials/mnist-advisor/config_pai.yml
SparkSnail: the token tutorial is in https://github.com/microsoft/nni/pull/2011

PR #1999, opened 2020-02-05
Dev compression speedup

QuanluZhang:
TODO:
1. replace `print` with logging


src/sdk/pynni/nni/compression/speedup/torch/compress_modules.py
```diff
@@ -0,0 +1,133 @@
...
+        The linear module to be replace
+    mask : ModuleMasks
+        The masks of this module
+
+    Returns
+    -------
+    torch.nn.Linear
+        The new linear module
+    """
+    assert isinstance(mask, ModuleMasks)
```
Cjkkkk: maybe provide error message if assert fails?

src/sdk/pynni/nni/compression/torch/apply_compression.py
```diff
@@ -0,0 +1,70 @@
...
+    ----------
+    model : torch.nn.module
+        The model to be compressed
+    masks_file : str
+        The path of the mask file
+    """
+    apply_comp = ApplyCompression(model, masks_file)
+    apply_comp.compress()
+
+class ApplyCompression(Pruner):
```
Cjkkkk: why subclass Pruner?

src/sdk/pynni/nni/compression/speedup/torch/compress_modules.py
QuanluZhang: good suggestion

src/sdk/pynni/nni/compression/torch/apply_compression.py
QuanluZhang: This class is for validating the correctness of ModelSpeedup, i.e., compare inference result from ModelSpeedup and that from ApplyCompression. ApplyCompression simply applies the masks, from implementation, it can be seen as a simple pruner.

examples/model_compress/speedup.md
```diff
@@ -0,0 +1,63 @@
...
+
+## slim pruner example
+
+on one V100 GPU,
+input tensor: `torch.randn(64, 3, 32, 32)`
+
+|Times| Mask Latency| Speedup Latency |
+|---|---|---|
+| 1 | 0.011968851089477539 | 0.005106925964355469 |
+| 2 | 0.020199298858642578 | 0.008769512176513672 |
```
chicm-ms: 1. the times are small,  maybe 100, 1000, 10000, etc?
2. the accuracy is too high, 6 digits are enough.

examples/model_compress/fpgm_torch_mnist.py
```diff
@@ -16,7 +16,8 @@ def forward(self, x):
         x = F.max_pool2d(x, 2, 2)
         x = F.relu(self.conv2(x))
         x = F.max_pool2d(x, 2, 2)
-        x = x.view(-1, 4 * 4 * 50)
+        #x = x.view(-1, 4 * 4 * 50)
+        x = x.view(64, -1)
```
chicm-ms: typically change it to x.view(x.size(0), -1)

examples/model_compress/model_speedup.py
```diff
@@ -0,0 +1,212 @@
...
+        return
+
+def l1filter_speedup(masks_file, model_checkpoint):
+    device = torch.device('cuda')
+    train_loader = torch.utils.data.DataLoader(
+        datasets.CIFAR10('./data.cifar10', train=True, download=True,
+                         transform=transforms.Compose([
+                             transforms.Pad(4),
+                             transforms.RandomCrop(32),
+                             transforms.RandomHorizontalFlip(),
```
chicm-ms: Maybe extract this piece as a `get_data_loaders` function?

examples/model_compress/model_speedup.py
```diff
@@ -0,0 +1,212 @@
...
+            transforms.ToTensor(),
+            transforms.Normalize((0.4914, 0.4822, 0.4465), (0.2023, 0.1994, 0.2010))
+        ])),
+        batch_size=200, shuffle=False)
+
+    model = VGG(depth=16)
+    model.to(device)
+    model.eval()
+
+    dummy_input = torch.randn(64, 3, 32, 32)
```
chicm-ms: It seems loaders are not used.

examples/model_compress/speedup.md
QuanluZhang: good suggestion, will update doc

examples/model_compress/fpgm_torch_mnist.py
QuanluZhang: good suggestion, thanks!

examples/model_compress/model_speedup.py
QuanluZhang: good point, thanks

examples/model_compress/speedup.md
QuanluZhang: new experiment numbers will be added in another pr

examples/model_compress/model_speedup.py
```diff
@@ -0,0 +1,212 @@
...
+        return
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser("speedup")
+    parser.add_argument("--example_name", type=str, default="slim", help="the name of pruning example")
+    parser.add_argument("--masks_file", type=str, default=None, help="the path of the masks file")
+    parser.add_argument("--model_checkpoint", type=str, default=None, help="the path of checkpointed model")
+    args = parser.parse_args()
+    
+    if args.example_name == 'slim':
```
Cjkkkk: it seems *_speedup have a lot same code, may be they can be put in one method? use dynamic module import based on parser arguments?

src/sdk/pynni/nni/compression/torch/pruners.py
```diff
@@ -209,34 +209,30 @@ def calc_mask(self, layer, config):
...
+        if filters >= 2 and num_prune >= 1:
             w_abs = weight.abs()
             mask_weight = torch.gt(w_abs, self.global_threshold).type_as(weight)
             mask_bias = mask_weight.clone()
             mask = {'weight': mask_weight.detach(), 'bias': mask_bias.detach()}
-        finally:
-            self.mask_dict.update({layer.name: mask})
-            self.mask_calculated_ops.add(layer.name)
-
+        if_calculated.copy_(torch.tensor(True)) # pylint: disable=not-callable
```
Cjkkkk: maybe use tensor(0) instead of tensor(True)

src/sdk/pynni/nni/compression/torch/apply_compression.py
Cjkkkk: I think it can simply be a method to go through all masks and apply to the weight accordingly. Because there is no need to hook forward method like pruner does and all cal_mask does is a dict lookup. To sum up, it is not a pruner and subclassing pruner makes it look weird. A simple function would do exactly the same and make much more sense?

src/sdk/pynni/nni/compression/torch/pruners.py
```diff
@@ -198,8 +197,9 @@ def __init__(self, model, config_list):
         all_bn_weights = torch.cat(weight_list)
         k = int(all_bn_weights.shape[0] * config['sparsity'])
         self.global_threshold = torch.topk(all_bn_weights.view(-1), k, largest=False)[0].max()
+        self.register_buffer("if_calculated", torch.tensor(False)) # pylint: disable=not-callable
```
Cjkkkk: use 0 instead of False

src/sdk/pynni/nni/compression/torch/apply_compression.py
```diff
@@ -0,0 +1,70 @@
...
+        for module_name in self.masks:
+            print('module_name: ', module_name)
+        config_list = self._build_config()
+        super().__init__(model, config_list)
+
+    def _build_config(self):
+        op_names = []
+        for module_name in self.masks:
+            op_names.append(module_name)
+        return [{'sparsity': 1, 'op_types': ['default', 'BatchNorm2d'], 'op_names': op_names}]
```
Cjkkkk: op_names are already presented in mask file, so I don't think you need a config.  Just getattr(model, opname).weight.mul_(mask[opname]).

examples/model_compress/model_speedup.py
```diff
@@ -0,0 +1,212 @@
...
+        apply_compression_results(model, masks_file)
+        dummy_input = dummy_input.to(device)
+        start = time.time()
+        for _ in range(32):
+            out = model(dummy_input)
+        #print(out.size(), out)
+        print('mask elapsed time: ', time.time() - start)
+        return
+    else:
+        #print("model before: ", model)
```
Cjkkkk: may be removed this or replace it with logger.debug and use flag to set logging level.

src/sdk/pynni/nni/compression/torch/apply_compression.py
QuanluZhang: good suggestion, let me fix it in a folow-up pr.

src/sdk/pynni/nni/compression/torch/pruners.py
QuanluZhang: it has been fixed in another pr, will merge that pr then.

src/sdk/pynni/nni/compression/torch/pruners.py
QuanluZhang: it has been fixed in another pr, will merge that pr then.

src/sdk/pynni/nni/compression/torch/apply_compression.py
QuanluZhang: this is the bug that i reported in teams, the logic of processing `op_types` and `op_name` is not correct, that is why i have to write `op_types` here.

examples/model_compress/model_speedup.py
QuanluZhang: yes, will do it

QuanluZhang:
@Cjkkkk your comments are very helpful, will fix them in the next pr soon.

PR #2003, opened 2020-02-06
Fix pruners for DataParallel support

chicm-ms:


examples/model_compress/fpgm_torch_mnist.py
```diff
@@ -27,8 +28,14 @@ def _get_conv_weight_sparsity(self, conv_layer):
...
 
     def print_conv_filter_sparsity(self):
-        conv1_data = self._get_conv_weight_sparsity(self.conv1)
-        conv2_data = self._get_conv_weight_sparsity(self.conv2)
+        if isinstance(self.conv1, nn.Conv2d):
+            conv1_data = self._get_conv_weight_sparsity(self.conv1)
+            conv2_data = self._get_conv_weight_sparsity(self.conv2)
+        else:
+            # self.conv1 is wrapped as PrunerModuleWrapper
+            conv1_data = self._get_conv_weight_sparsity(self.conv1.module)
```
QuanluZhang: a little confused, you get sparsity from weight rather than mask?

src/sdk/pynni/tests/test_compressor.py
```diff
@@ -138,7 +138,6 @@ def test_torch_fpgm_pruner(self):
         masks = pruner.calc_mask(layer, config_list[0], if_calculated=torch.tensor(0))
         assert all(torch.sum(masks['weight'], (1, 2, 3)).numpy() == np.array([45., 45., 45., 45., 0., 0., 45., 45., 45., 45.]))
 
-        pruner.update_epoch(1)
```
QuanluZhang: what is the reason of removing this line?

src/sdk/pynni/nni/compression/torch/pruners.py
```diff
@@ -165,9 +170,8 @@ def update_epoch(self, epoch):
 
         if epoch > 0:
             self.now_epoch = epoch
-            for k in self.if_init_list.keys():
-                self.if_init_list[k] = True
-
+            for wrapper in self.get_modules_wrapper():
+                wrapper.registered_buffers['if_calculated'].copy_(torch.tensor(0)) # pylint: disable=not-callable
```
QuanluZhang: which one is better? `wrapper.registered_buffers['if_calculated']` or `wrapper.if_calculated`

examples/model_compress/fpgm_torch_mnist.py
chicm-ms: In this way, we can ensure the weight is actually pruned. It can also be calculated from mask.

src/sdk/pynni/nni/compression/torch/pruners.py
chicm-ms: Just tested, `wrapper.if_calculated` does not work,  there is no error reported, but the value is still `1` we use  `wrapper.if_calculated.copy_(torch.tensor(0))`

src/sdk/pynni/tests/test_compressor.py
chicm-ms: We circumvent wrapper to directly call cal_mask in UT.  FPGM pruner call wrapper in update epoch.

examples/model_compress/fpgm_torch_mnist.py
QuanluZhang: got it, thanks

PR #2017, opened 2020-02-08
Doc update index

QuanluZhang:


docs/en_US/autotune_ref.md
```diff
@@ -0,0 +1,80 @@
+# Python API Reference of Auto Tune
```
ultmaster: This file can be an rst.

QuanluZhang: just be consistent with NasReference.md, what is the reason of preferring rst?

ultmaster: Can be. Not preferred to be. rst would save a few lines.

PR #2026, opened 2020-02-10
Fixed Import Error

ryujaehun:
In the latest version, PyTorch or upper 1.0  _DataLoaderIter is no more exist. 

so `ImportError: cannot import name '_DataLoaderIter'` raise some Import error.(in gradient_selector/fginitialize.py)

I do some exception handling code the head of code. and replace `_DataLoaderIter` to `_BaseDataLoaderIter`.

Thanks for your works.

### DIFF
--- a/src/sdk/pynni/nni/feature_engineering/gradient_selector/fginitialize.py
+++ b/src/sdk/pynni/nni/feature_engineering/gradient_selector/fginitialize.py
@@ -31,7 +31,10 @@ from sklearn.datasets import load_svmlight_file
 import torch
 from torch.utils.data import DataLoader, Dataset
 # pylint: disable=E0611
-from torch.utils.data.dataloader import _DataLoaderIter, _utils
+try:
+    from torch.utils.data.dataloader import _DataLoaderIter
+except ImportError:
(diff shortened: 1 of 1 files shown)


xuehui1991:
Dear @ryujaehun seems the pipeline crash few days ago.  Could you push again to run the build pipeline again? Otherwise, we cannot merge the code.

src/sdk/pynni/nni/feature_engineering/gradient_selector/fginitialize.py
```diff
@@ -31,7 +31,10 @@
 import torch
 from torch.utils.data import DataLoader, Dataset
 # pylint: disable=E0611
-from torch.utils.data.dataloader import _DataLoaderIter, _utils
+try:
+    from torch.utils.data.dataloader import _DataLoaderIter
```
xuehui1991: why not import the _utils?

xuehui1991: the following the code using the _utils

ultmaster:
Fixed in #2690.

chicm-ms:
Fixed with #2690

PR #2031, opened 2020-02-11
Increase the limit of request body

Sundrops:
When search space is big(>1000) using BatchTunner, response will be HTTP:413 Request Entity Too Large
`launcher.py`
```python
......
response = set_experiment(experiment_config, mode, args.port, config_file_name)
print_error(response)
print_error(response.text)
'''
ERROR: <Response [413]>
ERROR: <!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>PayloadTooLargeError: request entity too large<br> &nbsp; &nbsp;at readStreamxxx
'''
```


src/nni_manager/rest_server/nniRestServer.ts
```diff
@@ -34,7 +34,8 @@ export class NNIRestServer extends RestServer {
      */
     protected registerRestHandler(): void {
         this.app.use(express.static('static'));
-        this.app.use(bodyParser.json());
+        this.app.use(bodyParser.json({limit: '50mb'}));
+        this.app.use(bodyParser.urlencoded({limit: '50mb', extended: true}));
```
liuzhe-lz: I'm not familiar with body parser. Does this mean the URL can be 50mb long?

Sundrops: It means the data of  sending post request can be 50mb long.
```python
response = rest_post(experiment_url(port), json.dumps(request_data), REST_TIME_OUT, show_error=True)
def rest_post(url, data, timeout, show_error=False):
    '''Call rest post method'''
    try:
        response = requests.post(url, headers={'Accept': 'application/json', 'Content-Type': 'application/json'},\
                                 data=data, timeout=timeout)
        return response
        if show_error:
            print_error(exception)
        return None
```

Sundrops: The search space(json file) will be so big(>100kb), when you use BatchTunner. If not increase the limit of request body, it will fail for HTTP:413 Request Entity Too Large
```python
def set_experiment(experiment_config, mode, port, config_file_name):
    '''Call startExperiment (rest POST /experiment) with yaml file content'''
    ......
    request_data['searchSpace'] = experiment_config.get('searchSpace')
    ......
    response = rest_post(experiment_url(port), json.dumps(request_data), REST_TIME_OUT, show_error=True)

```

QuanluZhang: @Sundrops , thanks for the fix. One question: does this problem exist if you only keep line37 and remove line38?

Sundrops: Only keeping line37 can fix this problem now. I add line38 for the future. Do I need to remove line38?

QuanluZhang: @Sundrops thanks, I think we can only keep line37 for now :)

Sundrops: @QuanluZhang Done.

QuanluZhang:
@Sundrops could you change this pr to merge to v1.4, we want to include this change in v1.4, thanks

Sundrops:
@QuanluZhang Is it ok now? Do I need to open a pr newly?

QuanluZhang:
> @QuanluZhang Is it ok now? Do I need to open a pr newly?

See there are 27 changed files, so you need to merge v1.4 to your branch, then there would be only 1 changed file.

Sundrops:
@QuanluZhang Done.

PR #2033, opened 2020-02-11
fix model compression config validation

Cjkkkk:
fix model compression config validation.
rules:
only check op_types and op_names if they are in config file, skip checking otherwise.

### DIFF
--- a/src/sdk/pynni/nni/compression/torch/compressor.py
+++ b/src/sdk/pynni/nni/compression/torch/compressor.py
@@ -151,7 +151,13 @@ class Compressor:
             config = config.copy()
             # expand config if key `default` is in config['op_types']
             if 'op_types' in config and 'default' in config['op_types']:
-                config['op_types'].extend(default_layers.weighted_modules)
+                expanded_op_types = []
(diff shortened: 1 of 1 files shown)


src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -149,13 +149,18 @@ def select_config(self, layer):
         ret = None
         for config in self.config_list:
             config = config.copy()
-            config['op_types'] = self._expand_config_op_types(config)
-            if layer.type not in config['op_types']:
+            # expand config if key `default` is in config['op_types']
+            if 'op_types' in config and 'default' in config['op_types']:
+                config['op_types'].extend(default_layers.weighted_modules)
```
chicm-ms: better to remove 'default' after it is expanded.

Cjkkkk: I think it is best to just drop 'default' in op_types.

Cjkkkk: fixed.

PR #2044, opened 2020-02-12
Fix nnictl view command

SparkSnail:


tools/nni_cmd/nnictl.py
```diff
@@ -63,10 +63,11 @@ def parse_args():
     parser_resume.set_defaults(func=resume_experiment)
 
     # parse view command
-    parser_resume = subparsers.add_parser('view', help='view a stopped experiment')
-    parser_resume.add_argument('id', nargs='?', help='The id of the experiment you want to view')
-    parser_resume.add_argument('--port', '-p', default=DEFAULT_REST_PORT, dest='port', help='the port of restful server')
-    parser_resume.set_defaults(func=view_experiment)
+    parser_view = subparsers.add_parser('view', help='view a stopped experiment')
+    parser_view.add_argument('id', nargs='?', help='The id of the experiment you want to view')
```
xuehui1991: does here need "watch" paramter?

SparkSnail: There is no log when using view mode, have removed the parameter.

PR #2050, opened 2020-02-12
Fix windows pipeline timeout

chicm-ms:


examples/trials/cifar10_pytorch/main.py
```diff
@@ -176,6 +179,7 @@ def test(epoch):
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
     parser.add_argument("--epochs", type=int, default=200)
+    parser.add_argument("--batches", type=int, default=-1)
```
QuanluZhang: better to add comments to tell this argument is for integration test

xuehui1991: comment in teams

chicm-ms: Thanks, comments added.

PR #2058, opened 2020-02-14
Patch optimizer for pruner

Cjkkkk:
1 add patch *single* optimizer for *pruner*.
2 move wrap model into `__init__` method instead of `compress` method so that subclass pruner `__init__` methods can have access to wrapper.
3 add support to register buffers that are not necessarily registered as buffers in module, depending on whether type is torch.Tensor.
4 update quantizer and pruner based on new framework.
5 update examples.
6 update tests based on new framework


src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -272,9 +268,27 @@ class Pruner(Compressor):
...
+        self.patch_optimizer()
+
+    def patch_optimizer(self):
+        if self.optimizer is not None:
+            def patch_step(old_step):
+                def new_step(_, *args, **kwargs):
+                    # cal_mask and update buffers accordingly
+                    wrappers = self.get_modules_wrapper()
+                    for wrapper in wrappers:
+                        self.calc_mask(wrapper)
```
QuanluZhang: `calc_mask` should be placed before optimizer.step or after?

docs/en_US/Compressor/Framework.md
```diff
@@ -0,0 +1,69 @@
...
+        # do something to calculate weight_mask
+        if wrapper.if_calculated:
+            pass
+        else:
+            wrapper.if_calculated = torch.tensor(1)
+```
+
+### Multi-GPU support
+On multi-GPU training, buffers and parameters are copied to multiple GPU every time the `forward` method runs on multiple GPU. If buffers and parameters are updated in the `forward` method, an `in-place` update is needed to ensure the update is effective.
+Since `cal_mask` is called in the `optimizer.step` method, which happens after the `forward` method and happens only on one GPU, it supports multi-GPU naturally.
```
QuanluZhang: better to give an example to talk about how to implement activation pruner in this framework. In general, if a pruner wants to collect information at runtime. what is a common practice for this scenario.

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -272,9 +268,27 @@ class Pruner(Compressor):
 
     """
 
-    def __init__(self, model, config_list):
-        super().__init__(model, config_list)
-
+    def __init__(self, model, config_list, optimizer=None):
```
QuanluZhang: i would suggest to add one more argument to indicate whether calculate mask within __init__.

src/sdk/pynni/nni/compression/torch/pruners.py
```diff
@@ -202,7 +202,7 @@ def __init__(self, model, config_list):
         self.global_threshold = torch.topk(all_bn_weights.view(-1), k, largest=False)[0].max()
         self.register_buffer("if_calculated", torch.tensor(0)) # pylint: disable=not-callable
```
chicm-ms: One of the benefits to call calc_mask on step instead of forward is that we do not need to register buffer for flag variables like `if_calculated`, here we can switch `if_calculated` to original version to use a dict or set as a member of pruner.

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -272,9 +268,27 @@ class Pruner(Compressor):
...
+        if self.optimizer is not None:
+            def patch_step(old_step):
+                def new_step(_, *args, **kwargs):
+                    # cal_mask and update buffers accordingly
+                    wrappers = self.get_modules_wrapper()
+                    for wrapper in wrappers:
+                        self.calc_mask(wrapper)
+                    
+                    # call step overrided by user to do additional update
+                    self.step()
```
chicm-ms: is this pruner's step method?

src/sdk/pynni/nni/compression/torch/pruners.py
Cjkkkk: Then these variable won't be saved by `torch.state_dict() `?

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: Yes

src/sdk/pynni/nni/compression/torch/pruners.py
chicm-ms: Do you mean to save `if_calculated` as a part of model's state_dict?
The `if_calculated` is an pruner's internal variable during pruning process, we could have a lot of this kind of variable, such as index of epoch, index of batch, etc.  We should not save these.

chicm-ms: If for resume, the entire pruner should be saved and `if_calcuated` is a part of pruner, we can consider this later when it is necessary.

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -38,24 +40,30 @@ def __init__(self, model, config_list):
...
         """
-        if self.modules_to_compress is None:
-            self.modules_to_compress = []
-            for name, module in self.bound_model.named_modules():
-                layer = LayerInfo(name, module)
-                config = self.select_config(layer)
-                if config is not None:
-                    self.modules_to_compress.append((layer, config))
-        return self.modules_to_compress
+        modules_to_compress = []
```
QuanluZhang: where will this function be called? if this function is called only in __init__, please make it private

QuanluZhang: and why remove the check `if self.modules_to_compress is None:` ?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -38,24 +40,30 @@ def __init__(self, model, config_list):
         """
```
QuanluZhang: update docstring for `config_list`

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -87,26 +95,18 @@ def compress(self):
...
-
-        modules_to_compress = self.detect_modules_to_compress()
-        for layer, config in modules_to_compress:
-            wrapper = self._wrap_modules(layer, config)
-            self.modules_wrapper.append(wrapper)
-
-        self._wrap_model()
         return self.bound_model
 
     def register_buffer(self, name, value):
```
QuanluZhang: update docstring here

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -180,11 +180,13 @@ def update_epoch(self, epoch):
         epoch : num
             the current epoch number
         """
```
QuanluZhang: update docstring

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -87,26 +95,18 @@ def compress(self):
...
-        else:
-            self.modules_wrapper = []
-
-        modules_to_compress = self.detect_modules_to_compress()
-        for layer, config in modules_to_compress:
-            wrapper = self._wrap_modules(layer, config)
-            self.modules_wrapper.append(wrapper)
-
-        self._wrap_model()
         return self.bound_model
```
QuanluZhang: how about using `raise NotImplementedError()` here?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -200,6 +202,25 @@ def _wrap_modules(self, layer, config):
         raise NotImplementedError()
 
 
+    def add_activation_collector(self, collector):
```
QuanluZhang: could you update the two activation pruning algorithm to show how to use this collector?

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: 1 only in `__init__`  
2 because self.modules_to_compress is initialized in ` __init__`, which means it must be a list

QuanluZhang: 1. please add underline at the beginning of this function name
2. what if this function is called at other places such as pruning algorithm? if this function is only allowed to be called once, suggest to add that check, in order to avoid some unexpected behavior when this function is accidentally called multiple times

Cjkkkk: good suggestion!

src/sdk/pynni/nni/compression/torch/quantizers.py
```diff
@@ -203,18 +207,18 @@ def quantize_output(self, output, config, op, **kwargs):
...
+        out = self._quantize(output_bits, module, output)
+        out = self._dequantize(module, out)
         return out
 
     def fold_bn(self, config, **kwargs):
         # TODO simulate folded weight
         pass
 
-    def step(self):
+    def update_step(self):
```
chicm-ms: Sometimes users reset their step number at epoch begin, sometimes they don't, suggest to add a `step` parameter to allow pruner's step index to be consistent with user training code.  Same for `update_epoch`

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: fixed, calling compress will calculate default mask.

src/sdk/pynni/nni/compression/torch/pruners.py
Cjkkkk: In new version pruner register_buffer API, only buffer with type torch.Tensor will be registered as buffer, otherwise it will be a regular variable in module. Though the name of function is misleading, we can work on a better name.

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: Quantizer does not override compress method

src/sdk/pynni/nni/compression/torch/quantizers.py
Cjkkkk: What should be the value of the step parameter be? I don't think user keeps track of step number ?

examples/model_compress/QAT_torch_quantizer.py
```diff
@@ -35,6 +35,7 @@ def train(model, quantizer, device, train_loader, optimizer):
         loss = F.nll_loss(output, target)
         loss.backward()
         optimizer.step()
+        quantizer.update_step()
```
QuanluZhang: i think it is a little strange to call both `optimizer.step()` and `quantizer.update_step()`.

If `update_step()` is called every time `optimizer.step()` is called, then i think the api `update_step()` can be deleted.

src/sdk/pynni/nni/compression/torch/activation_rank_filter_pruners.py
```diff
@@ -48,29 +48,20 @@ def compress(self):
...
-            self.modules_wrapper.append(wrapper)
-            self.collected_activation[layer.name] = []
-
-            def _hook(module_, input_, output, name=layer.name):
+        for wrapper in self.get_modules_wrapper():
+            self.collected_activation[wrapper.name] = []
+
+            def _hook(module_, input_, output, name=wrapper.name):
                 if len(self.collected_activation[name]) < self.statistics_batch_num:
                     self.collected_activation[name].append(self.activation(output.detach().cpu()))
```
QuanluZhang: is there data race here?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -200,6 +213,25 @@ def _wrap_modules(self, layer, config):
...
```
QuanluZhang: a little confused about this function, you define it for users to use it, but in your implementation activation pruners do not use it...

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -272,10 +286,19 @@ class Pruner(Compressor):
...
+        self.patch_optimizer(self.update_mask)
 
-    def calc_mask(self, layer, config, **kwargs):
+    def compress(self):
+        self.update_mask()
+        return self.bound_model
+
+    def update_mask(self):
+        for wrapper in self.get_modules_wrapper():
+            self.calc_mask(wrapper)
```
QuanluZhang: better to update masks after `calc_mask()`

src/sdk/pynni/nni/compression/torch/activation_rank_filter_pruners.py
QuanluZhang: I mean DataParallel

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -468,45 +466,45 @@ class Quantizer(Compressor):
     """
 
     def __init__(self, model, config_list):
-        super().__init__(model, config_list)
+        super().__init__(model, config_list, None)
```
QuanluZhang: better to give default value `None` for optimizer in `Compressor`

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -566,15 +564,20 @@ def quant_backward(tensor, grad_output, quant_type):
...
     @staticmethod
-    def forward(ctx, tensor, quant_type, quant_func, config, layer, **kwargs):
+    def forward(ctx, tensor, quant_type, wrapper, **kwargs):
         ctx.save_for_backward(tensor, torch.Tensor([quant_type]))
-        return quant_func(tensor, config, op=layer.module, op_type=layer.type, op_name=layer.name, **kwargs)
+        if quant_type == QuantType.QUANT_INPUT:
+            return wrapper.quantizer.quantize_input(tensor, wrapper, **kwargs)
+        elif quant_type == QuantType.QUANT_WEIGHT:
+            return wrapper.quantizer.quantize_weight(tensor, wrapper, **kwargs)
+        else:
```
QuanluZhang: suggest to also check in this branch, and raise error in `else`

src/sdk/pynni/nni/compression/torch/pruners.py
```diff
@@ -81,48 +80,45 @@ def __init__(self, model, config_list):
...
         # if we want to generate new mask, we should update weigth first
-        w_abs = weight.abs() * mask['weight']
+        w_abs = weight.abs() * mask['weight_mask']
         threshold = torch.topk(w_abs.view(-1), k, largest=False)[0].max()
-        new_mask = {'weight': torch.gt(w_abs, threshold).type_as(weight)}
-        if_calculated.copy_(torch.tensor(1)) # pylint: disable=not-callable
+        new_mask = {'weight_mask': torch.gt(w_abs, threshold).type_as(weight)}
+        wrapper.if_calculated = True
 
         return new_mask
```
QuanluZhang: I found inconsistency, some pruners' calc_mask return masks, while some others update masks within this method

src/sdk/pynni/tests/test_compressor.py
```diff
@@ -130,22 +131,24 @@ def test_torch_fpgm_pruner(self):
...
+        model.conv2.module.weight.data = torch.tensor(w).float()
+        model.conv2.if_calculated = False
+        model.conv2.config = {'sparsity': 0.6, 'op_types': ['Conv2d']}
+        masks = pruner.calc_mask(model.conv2)
+        assert all(torch.sum(masks['weight_mask'], (1, 2, 3)).numpy() == np.array([45., 45., 0., 0., 0., 0., 0., 0., 45., 45.]))
 
     @tf2
     def test_tf_fpgm_pruner(self):
         model = get_tf_model()
-        config_list = [{'sparsity': 0.2, 'op_types': ['Conv2D']}, {'sparsity': 0.6, 'op_types': ['Conv2D']}]
```
QuanluZhang: cannot remove the second dict, it is testing which `sparsity` takes effect

src/sdk/pynni/tests/test_compressor.py
```diff
@@ -158,11 +161,6 @@ def test_tf_fpgm_pruner(self):
 
         assert all(masks.sum((1)) == np.array([45., 45., 45., 45., 0., 0., 45., 45., 45., 45.]))
 
-        model.layers[2].set_weights([weights[0], weights[1].numpy()])
-        masks = pruner.calc_mask(layer, config_list[1]).numpy()
-        masks = masks.reshape((-1, masks.shape[-1])).transpose([1, 0])
-        assert all(masks.sum((1)) == np.array([45., 45., 0., 0., 0., 0., 0., 0., 45., 45.]))
-
```
QuanluZhang: why this part is removed?

src/sdk/pynni/tests/test_compressor.py
```diff
@@ -207,33 +204,32 @@ def test_torch_slim_pruner(self):
...
-        assert all(mask1['bias'].numpy() == np.array([0., 1., 1., 1., 1.]))
-        assert all(mask2['bias'].numpy() == np.array([0., 1., 1., 1., 1.]))
+        mask1 = pruner.calc_mask(model.bn1)
+        mask2 = pruner.calc_mask(model.bn2)
+        assert all(mask1['weight_mask'].numpy() == np.array([0., 1., 1., 1., 1.]))
+        assert all(mask2['weight_mask'].numpy() == np.array([0., 1., 1., 1., 1.]))
+        assert all(mask1['bias_mask'].numpy() == np.array([0., 1., 1., 1., 1.]))
+        assert all(mask2['bias_mask'].numpy() == np.array([0., 1., 1., 1., 1.]))
 
+        model = TorchModel()
```
QuanluZhang: line 206 has instantiated `model`

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -87,26 +98,26 @@ def compress(self):
...
+        ----------
+        name : str
+            name of the variable
+        value: any
+            value of the variable
         """
-        self.buffers[name] = value
+        for wrapper in self.get_modules_wrapper():
+            if isinstance(value, torch.Tensor):
+                wrapper.register_buffer(name, value.clone())
```
chicm-ms: 1. Since wrapper's masks already registered within wrapper, it seems there is no requirement for now to register a tensor buffer?
2. It is convenient to use one method to set an attribute to all wrappers, maybe change this api to another name?

src/sdk/pynni/nni/compression/torch/pruners.py
chicm-ms: maybe `set_wrappers_attribute` ?

src/sdk/pynni/nni/compression/torch/quantizers.py
chicm-ms: the step value is an integer.  Normally users do validation or update lr policy per epoch or step, when they do validation/update lr policy per step,  they need to keep track of step number.  the step parameter can be optional.

src/sdk/pynni/nni/compression/torch/pruners.py
Cjkkkk: I make all pruners return masks and update these masks accordingly.

src/sdk/pynni/tests/test_compressor.py
Cjkkkk: I think this test should be moved to additional test for config selection test.

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: updated

src/sdk/pynni/tests/test_compressor.py
QuanluZhang: good point. could you add this one to the config selection test? and we can add more unit tests to it in future.

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: rename it to register variable now.

src/sdk/pynni/tests/test_compressor.py
Cjkkkk: in init method global threshold is computed

src/sdk/pynni/tests/test_compressor.py
Cjkkkk: This test is duplicated.

src/sdk/pynni/tests/test_compressor.py
Cjkkkk: Of course.

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: Done.

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: Updated.

examples/model_compress/QAT_torch_quantizer.py
Cjkkkk: add patch_optimizer in quantizer.

src/sdk/pynni/nni/compression/torch/pruners.py
Cjkkkk: Done.

src/sdk/pynni/nni/compression/torch/activation_rank_filter_pruners.py
Cjkkkk: There is. I haven't come up with how to remove the data racing here yet.

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: rename it to set_wrappers_attribute

src/sdk/pynni/nni/compression/torch/compressor.py
Cjkkkk: updated.

docs/en_US/Compressor/Framework.md
Cjkkkk: updated

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -35,15 +37,27 @@ def __init__(self, model, config_list):
...
+
         self.modules_to_compress = None
-        self.modules_wrapper = None
-        self.buffers = {}
+        self.modules_wrapper = []
         self.is_wrapped = False
 
-    def detect_modules_to_compress(self):
+        self._fwd_hook_handles = {}
+        self._fwd_hook_id = 0
```
QuanluZhang: could you add comment to explain `_fwd_hook_id`?

src/sdk/pynni/nni/compression/torch/compressor.py
```diff
@@ -456,57 +465,59 @@ def forward(self, *inputs):
...
-    def __init__(self, model, config_list):
-        super().__init__(model, config_list)
+    def __init__(self, model, config_list, optimizer=None):
+        super().__init__(model, config_list, optimizer)
         self.quant_grad = QuantGrad
+        if self.optimizer is not None:
+            self.patch_optimizer(self.step_with_optimizer)
+            for wrapper in self.get_modules_wrapper():
+                if 'weight' in wrapper.config['quant_types']:
+                    self.optimizer.add_param_group({"params": wrapper.module.old_weight})
```
QuanluZhang: could you add comment to explain this line?

PR #2061, opened 2020-02-15
add support for pytorchjob crd version 'v1'

jiapinai:
As supported by kubeflow 0.7,  the new crd version ```v1``` has been added for ```TFJob``` about two month ago via commit #1814 . So the version ```v1```  should be added for ```PyTorchJob```

### DIFF
--- a/src/nni_manager/config/kubeflow/pytorchjob-crd-v1.json
+++ b/src/nni_manager/config/kubeflow/pytorchjob-crd-v1.json
@@ -10,7 +10,7 @@
             "singular": "pytorchjob"
         }
     },
-    "apiVersion": "apiextensions.k8s.io/v1beta1",
+    "apiVersion": "kubeflow.org/v1",
     "metadata": {
         "name": "pytorchjobs.kubeflow.org"
     }
--- a/src/nni_manager/config/kubeflow/pytorchjob-crd-v1.json
+++ b/src/nni_manager/config/kubeflow/pytorchjob-crd-v1.json
@@ -1,0 +1,17 @@
+{
+    "kind": "CustomResourceDefinition",
+    "spec": {
+        "scope": "Namespaced",
+        "version": "v1",
+        "group": "kubeflow.org",
+        "names": {
+            "kind": "PyTorchJob",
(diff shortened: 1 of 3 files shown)


src/nni_manager/config/kubeflow/pytorchjob-crd-v1.json
```diff
@@ -0,0 +1,17 @@
...
+        "scope": "Namespaced",
+        "version": "v1",
+        "group": "kubeflow.org",
+        "names": {
+            "kind": "PyTorchJob",
+            "plural": "pytorchjobs",
+            "singular": "pytorchjob"
+        }
+    },
+    "apiVersion": "apiextensions.k8s.io/v1beta1",
```
SparkSnail: v1beta1?

jiapinai: > https://github.com/kubeflow/tf-operator/blob/master/examples/crd/crd-v1.yaml
They use it so here.

jiapinai: Already have test it. it's working :)

jiapinai: Checked again, https://github.com/kubeflow/pytorch-operator/blob/master/examples/mnist/v1/pytorch_job_mnist_nccl.yaml using `apiVersion: "kubeflow.org/v1"` instead of `"apiVersion": "apiextensions.k8s.io/v1beta1"`. 

Will get back after doing more tests.

SparkSnail: Thanks for the test, better to use `apiVersion: "kubeflow.org/v1"` since it's more clear.

jiapinai: Tests have been done. Turns out both "`kubeflow.org/v1`" and "`apiextensions.k8s.io/v1beta1`" are working well. As you mentioned,  better to use apiVersion: "`kubeflow.org/v1`" since it's more clear. The code has been updated, please kindly have a look

PR #2064, opened 2020-02-15
Small typo： grpah -> graph

dhassault:
Just a small typo in the doc.


docs/en_US/Compressor/ModelSpeedup.md
```diff
@@ -14,7 +14,7 @@ There are two types of pruning. One is fine-grained pruning, it does not change
```
xuehui1991: I can't tell any change in this PR.

dhassault: Sorry, here is the documentation extract:
The first step requires topology (i.e., connections) of the model, we use `jit.trace` to obtain the model `grpah` for PyTorch.
`grpah` -> graph

Issue #2057, opened 2020-02-14
[webui-v1.4-bug-bash] bugs


lvybriage:
-  refresh interval selectors doesn't work
-  defatult metric sort is not correctly in details table,    it's disordered when latest | final | --
![image](https://user-images.githubusercontent.com/35484733/74497038-899f9300-4f17-11ea-93de-31bb00eea61b.png)
- search trials by id, tooltip "Unexpected search filter Id"
- 点击空白处可以关闭log modal

PR #2066, opened 2020-02-16
fix some bugs in v1.4 bug bash

lvybriage:
fix issue #2057 

- refresh interval selectors doesn't work
- defatult metric sort is not correctly in details table, it's disordered when latest | final | --
- search trials by id | parameters, tooltip "Unexpected search filter Id" | "...parameters"

this item `点击空白处可以关闭log modal`:
add `isLightDismiss={true}` for Panel , but it doesn't work in the test. will continue to fix it in another PR.


SparkSnail:
which line fix `refresh interval selectors doesn't work`?

lvybriage:
> which line fix `refresh interval selectors doesn't work`?

line 537 & 538

PR #2067, opened 2020-02-16
Update WebUI for metrics data changing to json string

lvybriage:
Update WebUI for metrics data changing to json string

### DIFF
--- a/src/webui/src/static/function.ts
+++ b/src/webui/src/static/function.ts
@@ -43,7 +43,7 @@ const getFinalResult = (final?: MetricDataRecord[]): number => {
     let acc;
     let showDefault = 0;
     if (final) {
-        acc = JSON.parse(final[final.length - 1].data);
+        acc = JSON5.parse(final[final.length - 1].data);
         if (typeof (acc) === 'object') {
             if (acc.default) {
                 showDefault = acc.default;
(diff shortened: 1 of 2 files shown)


src/webui/src/static/function.ts
```diff
@@ -43,7 +43,7 @@ const getFinalResult = (final?: MetricDataRecord[]): number => {
     let acc;
     let showDefault = 0;
     if (final) {
-        acc = JSON.parse(final[final.length - 1].data);
+        acc = JSON5.parse(final[final.length - 1].data);
```
chicm-ms: Is it possible to call parseMetrics here?  there maybe more updates in parseMetrics later.

src/webui/src/static/function.ts
```diff
@@ -182,8 +182,6 @@ function formatTimestamp(timestamp?: number, placeholder?: string): string {
 function parseMetrics(metricData: string): any {
     if (metricData.includes('NaN')) {
         return JSON5.parse(metricData)
-    } else {
-        return JSON.parse(metricData)
```
chicm-ms: why remove this?

src/webui/src/static/function.ts
lvybriage: It's OK.

src/webui/src/static/function.ts
lvybriage: I thought UI only use JSON5 from the teams message.
It's misleading, I'll change this back.

src/webui/src/static/function.ts
```diff
@@ -61,7 +61,7 @@ const getFinalResult = (final?: MetricDataRecord[]): number => {
 const getFinal = (final?: MetricDataRecord[]): FinalType | undefined => {
     let showDefault: FinalType;
     if (final) {
-        showDefault = JSON.parse(final[final.length - 1].data);
+        showDefault = JSON5.parse(final[final.length - 1].data);
```
chicm-ms: Can this be replaced by parseMetrics?

lvybriage: OK.

PR #2069, opened 2020-02-17
proxylessnas update

QuanluZhang:


src/sdk/pynni/nni/nas/pytorch/utils.py
```diff
@@ -116,12 +116,14 @@ def update(self, val, n=1):
         n : int
             The weight of the new value.
         """
-        if not isinstance(val, float) and not isinstance(val, int):
+        if isinstance(val, float) or isinstance(val, int) \
+            or (isinstance(val, torch.Tensor) and not val.size()):
```
ultmaster: `torch.is_tensor(val)`. Also I believe we don't have to limit usage of AverageMeter to number and single-value tensor. Arrays and tensors can also use `AverageMeter` if users want to. In fact, I don't think this `if` is necessary at all.

docs/en_US/NAS/Proxylessnas.md
```diff
@@ -60,4 +60,4 @@ ProxylessNasMutator also implements the forward logic of the mutables (i.e., Lay
 
 ## Reproduce Results
 
-Ongoing...
+To reproduce the result, we first run the search, we found that though it runs many epochs the chosen architecture converges at the first several epochs. This is probably induced by hyper-parameters or the implementation, we are working on it. The test accuracy of the found architecture is top1: 72.31, top5: 90.26.
```
ultmaster: What's the corresponding value in paper?

QuanluZhang: it is not comparable because the numbers in the paper are from the experiments which take hardware into consideration, but our current implementation has not supported hardware

PR #2077, opened 2020-02-18
Fix NaN support

QuanluZhang:


src/sdk/pynni/nni/bohb_advisor/bohb_advisor.py
```diff
@@ -557,7 +557,8 @@ def handle_report_metric_data(self, data):
             Data type not supported
         """
         logger.debug('handle report metric data = %s', data)
-
+        if 'value' in data:
+            data['value'] = json_tricks.loads(data['value'])
```
liuzhe-lz: Is there any documentation affected?

QuanluZhang: This is a key/value from `data`, which is changed from obj to string, I think this is not documented before. But we should BTW.

PR #2082, opened 2020-02-19
update examples

Cjkkkk:
1 add DoReFaQuantizer_torch_mnist example.
2 support cuda and cpu for examples.


examples/model_compress/DoReFaQuantizer_torch_mnist.py
```diff
@@ -0,0 +1,92 @@
...
+        batch_size=64, shuffle=True)
+    test_loader = torch.utils.data.DataLoader(
+        datasets.MNIST('data', train=False, transform=trans),
+        batch_size=1000, shuffle=True)
+
+    model = Mnist()
+    model = model.to(device)
+    '''you can change this to DoReFaQuantizer to implement it
+    DoReFaQuantizer(configure_list).compress(model)
+    '''
```
chicm-ms: is it already DoReFaQuantizer ? this comments seems useless.

examples/model_compress/MeanActivation_torch_cifar10.py
```diff
@@ -82,7 +83,8 @@ def main():
 
     # Test base model accuracy
     print('=' * 10 + 'Test on the original model' + '=' * 10)
-    model.load_state_dict(torch.load('vgg16_cifar10.pth'))
+    if os.path.isfile('vgg16_cifar10.pth'):
+        model.load_state_dict(torch.load('vgg16_cifar10.pth'))
```
chicm-ms: It seems the `load_state_dict` and the retrain part should swap order, current logic is strange.

Cjkkkk: I don't understand why is the logic strange. Could you be more specific?

chicm-ms: if args.retrain is True,  we do not need to load model here.
if args.retrain is False, we  got a model with random parameter before `load_state_dict`,  our pruner should not prune a random model, so we should assert the 'vgg16_cifar10.pth' must exist.

Cjkkkk: fixed

examples/model_compress/DoReFaQuantizer_torch_mnist.py
Cjkkkk: fixed

PR #2086, opened 2020-02-21
Handle exception and kill restserver in nnictl

SparkSnail:


tools/nni_cmd/launcher.py
```diff
@@ -525,7 +525,14 @@ def create_experiment(args):
...
     nni_config.set_config('restServerPort', args.port)
-    launch_experiment(args, experiment_config, 'new', config_file_name)
+    try:
+        launch_experiment(args, experiment_config, 'new', config_file_name)
+    except Exception as exception:
+        nni_config = Config(config_file_name)
+        restServerPid = nni_config.get_config('restServerPid')
+        if restServerPid:
+            kill_command(restServerPid)
+        print_error(exception)
```
QuanluZhang: suggest to also add `exit` here

tools/nni_cmd/launcher.py
```diff
@@ -553,7 +560,15 @@ def manage_stopped_experiment(args, mode):
...
+    try:
+        launch_experiment(args, experiment_config, mode, new_config_file_name, experiment_id)
+    except Exception as exception:
+        nni_config = Config(new_config_file_name)
+        restServerPid = nni_config.get_config('restServerPid')
+        if restServerPid:
+            kill_command(restServerPid)
+        print_error(exception)
+        exit(1)
     new_nni_config.set_config('restServerPort', args.port)
```
QuanluZhang: curious about this line. if users use `--foreground`, this line would not be executed, thus, new rest port would not be set, is this a problem?

SparkSnail: good point, moved `new_nni_config.set_config('restServerPort', args.port)` before `launch_experiment()` function.

tools/nni_cmd/launcher.py
SparkSnail: This function is a top level function, no more code will be executed after this function. If there is error here, it will exit the process automatically.

QuanluZhang: if there is an exception, the exit code should be non-zero.
if you catch here, even there is an exception, exit code is still 0.

SparkSnail: ok, fixed

QuanluZhang: how about raising the error here instead of print_error?

SparkSnail: The output format of raising error is not elegant, NNICTL use uniform `print_error()` method for error information.

PR #2099, opened 2020-02-25
Refactor pruner examples

chicm-ms:


QuanluZhang:
@chicm-ms could you update the corresponding doc, since you delete some example files, there might be some broken links.

chicm-ms:
> @chicm-ms could you update the corresponding doc, since you delete some example files, there might be some broken links.

Thanks, slim example updated, other examples in doc are linking to the examples code directory, no need to update.

PR #2103, opened 2020-02-26
update succeed table in time

Lijiaoa:


src/webui/src/components/overview/SuccessTable.tsx
```diff
@@ -113,6 +113,11 @@ class SuccessTable extends React.Component<SuccessTableProps, SuccessTableState>
         }
     ];
 
+    componentWillReceiveProps(nextProps: SuccessTableProps): void {
```
liuzhe-lz: Why not put this logic in `render()`?

Lijiaoa: because ui got `trialIds` by props, got the table source by `trialIds`
when props changed, need to restore the state value in `componentWillReceiveProps` to trigger `render`. I think.

Issue #2087, opened 2020-02-21
WebUI crash after refresh


xuehui1991:
![image](https://user-images.githubusercontent.com/6280746/75002969-31313e00-54a1-11ea-80da-57a2215ac515.png)

The experiment only have one trial and one final data.

![image](https://user-images.githubusercontent.com/6280746/75002996-49a15880-54a1-11ea-917d-3c221db2e7f7.png)

**Short summary about the issue/question**:

**Brief what process you are following**: 

**How to reproduce it**: 

**nni Environment**:
- nni version:
- nni mode(local|pai|remote):
- OS:
- python version:
- is conda or virtualenv used?: 
- is running in docker?:

**need to update document(yes/no)**:

**Anything else we need to know**:

lvybriage:
will design an error page to show this status.

Issue #1918, opened 2020-01-03
WebUI is broken when final result's type is tensor


lvybriage:
**Short summary about the issue/question**:
```
"finalMetricData":[
            {
                "data":"{\"__instance_type__\":[\"torch\",\"Tensor\"],\"attributes\":{}}",
                "parameterId":"3",
                "sequence":0,
                "timestamp":1578029488459,
                "trialJobId":"t9wYQ",
                "type":"FINAL"
            }
        ],
```

![image](https://user-images.githubusercontent.com/35484733/71709184-b187db00-2e30-11ea-817e-3f9a3c272877.png)

**Brief what process you are following**: 

**How to reproduce it**: 

**nni Environment**:
- nni version:
- nni mode(local|pai|remote):
- OS:
- python version:
- is conda or virtualenv used?: 
- is running in docker?:

**need to update document(yes/no)**:

**Anything else we need to know**:

lvybriage:
WebUI doesn't support the final result's type is `tensor` at present. Please return the final result type as `{default: string}` or `{default: dict}`.

scarlett2018:
> WebUI doesn't support the final result's type is `tensor` at present. Please return the final result type as `{default: string}` or `{default: dict}`.
> 
> And the broken page by final tensor type we'll fix soon.

@lvybriage - any work left for this issue? may we close it now?

PR #2105, opened 2020-02-27
fix bug of !(object) !(number) metris

Lijiaoa:
https://github.com/microsoft/nni/issues/1918
https://github.com/microsoft/nni/issues/2087

### DIFF
--- a/src/webui/src/static/function.ts
+++ b/src/webui/src/static/function.ts
@@ -46,12 +46,7 @@ function parseMetrics(metricData: string): any {
 }
 
 const isArrayType = (list: any): boolean | undefined => {
-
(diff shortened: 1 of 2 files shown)


src/webui/src/static/function.ts
```diff
@@ -45,19 +45,30 @@ function parseMetrics(metricData: string): any {
     }
 }
 
+const isArrayType = (list: any): boolean | undefined => {
+
+    if (!Array.isArray) {
```
liuzhe-lz: Looks like these are browser compatibility stuff. I thought TS or node should deal with them?

Lijiaoa: had updated.

PR #2111, opened 2020-03-02
Remove final val in the intermediate result graph and fix layout style bug(table | search space | config) on desktop

Lijiaoa:
Complete:
- remove final result from intermeidate result graph in detail table component
- fix `search space | config` layout on desktop in overview page
- fix table style on desktop


Lijiaoa:
I find that the table's style is also wrong in desktop.
I'll continue to commit code to this PR.

Lijiaoa:
> I find that the table's style is also wrong in desktop.
> I'll continue to commit code to this PR

Committed. Please help review.

PR #2113, opened 2020-03-03
Upgrade tf version to fix security warning

chicm-ms:
1. keras version need to be upgraded to work with tensorflow 1.15
2. keras accuracy key changed in 2.3.1


ultmaster:
Are there tests (IT/UT) for these tuners?

chicm-ms:
> Are there tests (IT/UT) for these tuners?

There is UT for network_morphism tuner, and ga_squad is a trial, no builtin tuner for ga_squad

PR #2116, opened 2020-03-03
Fix issue 2075

liuzhe-lz:


src/webui/src/components/Modals/CustomizedTrial.tsx
```diff
@@ -59,29 +59,34 @@ class Customize extends React.Component<CustomizeProps, CustomizeState> {
         });
 
         // true: parameters are wrong
-        let flag = false;
+        let parametersIllegal = false;
         Object.keys(customized).map(item => {
```
ultmaster: Why not forEach?

Lijiaoa: > Why not forEach?

I'll continue to fix this in next release, thanks for Zhe and Yuge.

PR #2121, opened 2020-03-05
make assessors support metric data in dict

QuanluZhang:
1. make assessors support metric data in dict
2. fix bug when `includeIntermediateResults` is true

### DIFF
--- a/src/sdk/pynni/nni/utils.py
+++ b/src/sdk/pynni/nni/utils.py
@@ -62,6 +62,13 @@ def extract_scalar_reward(value, scalar_key='default'):
     """
     Extract scalar reward from trial result.
 
+    Parameters
+    ----------
+    value : int, float, dict
+        the reported final metric data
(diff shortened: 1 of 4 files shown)


src/sdk/pynni/nni/utils.py
```diff
@@ -78,6 +85,26 @@ def extract_scalar_reward(value, scalar_key='default'):
...
+    scalar_key : str
+        the key name that indicates the numeric number
+
+    Raises
+    ------
+    RuntimeError
+        Incorrect final result: the final result should be float/int,
+        or a dict which has a key named "default" whose value is float/int.
+    """
+    return [extract_scalar_reward(ele) for ele in trial_history]
```
chicm-ms: Looks like the `scalar_key` should be passed to `extract_scalar_reward`

QuanluZhang: good point!

PR #2143, opened 2020-03-11
[BUG] import dict metric data in SMAC and BOHB

QuanluZhang:


src/sdk/pynni/tests/test_builtin_tuners.py
```diff
@@ -141,36 +141,94 @@ def search_space_test_all(self, tuner_factory, supported_types=None, ignore_type
...
                                    ignore_types=["uniform_equal", "qloguniform_equal", "loguniform_equal", "quniform_clip_2"])
         # NOTE: types are ignored because `tpe.py line 465, in adaptive_parzen_normal assert prior_sigma > 0`
+        self.import_data_test(lambda: HyperoptTuner("tpe"))
 
     def test_random_search(self):
         self.search_space_test_all(lambda: HyperoptTuner("random_search"))
+        self.import_data_test(lambda: HyperoptTuner("random_search"))
 
     def test_anneal(self):
         self.search_space_test_all(lambda: HyperoptTuner("anneal"))
```
ultmaster: I would write:

```python
tuner_fn = lambda: HyperoptTuner("anneal")
self.search_space_test_all(tuner_fn)
self.import_data_test(tuner_fn)
```

It's better-looking.

QuanluZhang: updated

src/sdk/pynni/tests/test_builtin_tuners.py
```diff
@@ -141,50 +141,119 @@ def search_space_test_all(self, tuner_factory, supported_types=None, ignore_type
             logger.info("Full supported search space: %s", full_supported_search_space)
             self.search_space_test_one(tuner_factory, full_supported_search_space)
 
+    def import_data_test(self, tuner_factory, stype="choice_str"):
+        # import data at the beginning with number value and dict value
```
xuehui1991: use doc string """ """ here.

QuanluZhang: updated

PR #2153, opened 2020-03-16
Fix pruners

chicm-ms:


src/sdk/pynni/tests/test_pruners.py
```diff
@@ -0,0 +1,162 @@
...
+        'pruner_class': AGP_Pruner,
+        'config_list': [{
+            'initial_sparsity': 0,
+            'final_sparsity': 0.8,
+            'start_epoch': 0,
+            'end_epoch': 10,
+            'frequency': 1,
+            'op_types': ['default']
+        }],
+        'validators': []
```
QuanluZhang: yes, for iterative pruner, we should think about how to test them.

PR #2159, opened 2020-03-17
Update pai yaml merge method

SparkSnail:
Use additional yaml to override original yaml file.


ultmaster:
What happens now with arrays?

SparkSnail:
> What happens now with arrays?

new array items will append in original array, does not override original one.

SparkSnail:
close this pr, will refactor in next release.

PR #2161, opened 2020-03-18
refactor of modelspeedup example

QuanluZhang:


src/sdk/pynni/nni/compression/torch/apply_compression.py
```diff
@@ -3,68 +3,27 @@
...
-            Auxiliary information
-
-        Returns
-        -------
-        dict
-            Mask of the layer
-        """
-        assert layer.name in self.masks
-        return self.masks[layer.name]
+    masks = torch.load(masks_file, map_location)
```
ultmaster: Do we need to wrap the whole function with `with torch.no_grad()`?

examples/model_compress/model_speedup.py
```diff
@@ -9,145 +10,89 @@
...
-        if args.masks_file is None:
-            args.masks_file = 'mask_vgg16_cifar10.pth'
-        apoz_speedup(args.masks_file, args.model_checkpoint)
+
+    if args.example_name != 'all':
+        if args.masks_file is not None:
+            config[args.example_name]['masks_file'] = args.masks_file
+        if not os.path.exists(config[args.example_name]['masks_file']):
+            msg = '{} does not exist! You should specify masks_file correctly, '
+            msg += 'or use default one which is generated by model_prune_torch.py'
```
ultmaster: Use multi-line string. `+=` is not necessary.

src/sdk/pynni/nni/compression/torch/apply_compression.py
QuanluZhang: it is in eval mode

examples/model_compress/model_speedup.py
QuanluZhang: fixed, thanks!

PR #2217, opened 2020-03-21
Support nas installation

SparkSnail:
test pypi:
python3 -m pip install nni==1.4.2003040926 --user


ultmaster:
Just had a look at the unpacked wheel. I'm not sure whether it's okay to include all these source maps (`.js.map`) in production build. Since they are not large in volume, I'm fine with both keeping and ejecting.

image

SparkSnail:
> Just had a look at the unpacked wheel. I'm not sure whether it's okay to include all these source maps (`.js.map`) in production build. Since they are not large in volume, I'm fine with both keeping and ejecting.
> 
> image

webui package also include the .js.map files, I think it ok to keep them.

PR #2219, opened 2020-03-22
Add model compression config validation

### LINKED ISSUE #2218: Pruner error when there is no  op_types
Pruner assumes every named_modules have a 'weight' field, when there is no op_types field in config_list, the model itself is also added into models_to_compress

```
    def _detect_modules_to_compress(self):
        """
        detect all modules should be compressed, and save the result in `self.modules_to_compress`.
        The model will be instrumented and user should never edit it after calling this method.
        """
        if self.modules_to_compress is None:
            self.modules_to_compress = []
            for name, module in self.bound_model.named_modules():
                layer = LayerInfo(name, module)
                config = self.select_config(layer)
                if config is not None:
                    self.modules_to_compress.append((layer, config))
        return self.modules_to_compress

```

When there is no op_types in config_list, got following error:

```
ERROR: test_torch_pruner_validation (test_compressor.CompressorTestCase)
----------------------------------------------------------------------
  File "/mnt/c/nni/src/sdk/pynni/tests/test_compressor.py", line 307, in test_torch_pruner_validation
    torch_compressor.FPGMPruner(model, config_list, optimizer)
  File "/mnt/c/nni/src/sdk/pynni/nni/compression/torch/weight_rank_filter_pruners.py", line 219, in __init__
  File "/mnt/c/nni/src/sdk/pynni/nni/compression/torch/weight_rank_filter_pruners.py", line 33, in __init__
  File "/mnt/c/nni/src/sdk/pynni/nni/compression/torch/compressor.py", line 306, in __init__
  File "/mnt/c/nni/src/sdk/pynni/nni/compression/torch/compressor.py", line 58, in __init__
    wrapper = self._wrap_modules(layer, config)
  File "/mnt/c/nni/src/sdk/pynni/nni/compression/torch/compressor.py", line 348, in _wrap_modules
    wrapper = PrunerModuleWrapper(layer.module, layer.name, layer.type, config, self)
  File "/mnt/c/nni/src/sdk/pynni/nni/compression/torch/compressor.py", line 280, in __init__
    self.register_buffer("weight_mask", torch.ones(self.module.weight.shape))
  File "/home/chicm/.local/lib/python3.6/site-packages/torch/nn/modules/module.py", line 591, in __getattr__
    type(self).__name__, name))
AttributeError: 'TorchModel' object has no attribute 'weight'
```

chicm-ms:
1. Add config validation for pruners/quantizers
2. fix #2218


src/sdk/pynni/nni/compression/torch/utils.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+def validate_op_names(model, op_names, logger):
+    found_names = set()
+    for name, _ in model.named_modules():
+        found_names.add(name)
+
+    not_found_op_names = list(set(op_names) - found_names)
```
QuanluZhang: inconsistent: there is `list` here, but no in line 20.

chicm-ms: Thanks,  list added for better print.

src/sdk/pynni/nni/compression/torch/utils.py
```diff
@@ -0,0 +1,53 @@
...
+        self.data_schema = data_schema
+        self.compressor_schema = Schema(self._modify_schema(data_schema, model, logger))
+
+    def _modify_schema(self, data_schema, model, logger):
+        if not data_schema:
+            return data_schema
+
+        for k in data_schema[0]:
+            old_schema = data_schema[0][k]
+            if k == 'op_types' or (isinstance(k, Schema) and k._schema == 'op_types'):
```
QuanluZhang: what is `(isinstance(k, Schema) and k._schema == 'op_types')` used for?

chicm-ms: We can define 'op_types' schema like this:
1.  `'op_types': [str]`,    here k == 'op_types'
2. `Optional('op_types'): [str]`,   here `Optional' is an instance of `Schema`

QuanluZhang: got it, thanks

src/sdk/pynni/tests/test_pruners.py
```diff
@@ -34,7 +34,7 @@ def validate_sparsity(wrapper, sparsity, bias=False):
     'agp': {
         'pruner_class': AGP_Pruner,
         'config_list': [{
-            'initial_sparsity': 0,
+            'initial_sparsity': 0.,
```
SparkSnail: why use `0.` ?

chicm-ms: we require the sparsity field to be a float.

PR #2220, opened 2020-03-22
Add evalutaion scripts for TextNAS

pkuyym:
Currently, the reproduction couldn't be guaranteed, will keep polishing this PR.

Since the dataloader changed, we need re-do the hyper-parameter search. The performance of current setting: sst-2: 90.12, sst-5: 53.3


examples/nas/textnas/macro_child.py
```diff
@@ -0,0 +1,223 @@
...
+        else:
+            branch_id = (layer_id) * (layer_id + 1) // 2
+
+        bn_flag = False
+        for i in range(layer_id):
+            if self.sample_arc[branch_id + 1 + i] == 1:
+                bn_flag = True
+        branch_id = self.sample_arc[branch_id]
+
+        for operation_id in [0, 1, 2, 3]:  # conv_opt
```
ultmaster: I notice that this is a LayerChoice but it's not using NNI interface. Theoretically, with NNI, you can share the same model class in search and retrain. Are you having any trouble with that?

pkuyym: Totally agree, it is better to adopt the NNI interface. I need more time to learn the NNI interface and do the adaption.

PR #2241, opened 2020-03-26
[BUG] finding leaf modules

QuanluZhang:


src/sdk/pynni/nni/compression/speedup/torch/compressor.py
```diff
@@ -229,42 +229,39 @@ def _extract_leaf_modules(self, graph):
         list
             a list of scope name of all the leaf modules
         """
-        pieces = [] # each element is a dict
+        class SNode:
+            def __init__(self, name):
+                self.sname = name
+                self.childs = {}
```
ultmaster: childs? children?

src/sdk/pynni/nni/compression/speedup/torch/compressor.py
```diff
@@ -229,42 +229,39 @@ def _extract_leaf_modules(self, graph):
...
+        def traverse_tree(node, scope_name):
+            if scope_name == '':
+                sn = node.sname
+            else:
+                sn = scope_name + '/' + node.sname
+            if not node.childs:
+                if node.sname[-1] == ']':
+                    leaf_nodes.append(sn)
+            else:
+                for key in node.childs:
```
ultmaster:
```python
for child in node.children.values():
    traverse_tree(child, sn)
```

src/sdk/pynni/nni/compression/speedup/torch/compressor.py
```diff
@@ -229,42 +229,39 @@ def _extract_leaf_modules(self, graph):
...
+                    traverse_tree(node.childs[key], sn)
+        traverse_tree(root, '')
+        return leaf_nodes
```
ultmaster: If I'm right, what this is doing is actually: finding all leaf nodes that ends with `]`. If so, there is a simpler way to do it:

```python
all_names = [node.scopeName() for node in graph.nodes()]
all_names = list(filter(lambda x: x, all_names))  # filter out non-empty strings
all_names.sort()
leaf_nodes = []
for i, name in enumerate(all_names):
    if (i + 1 >= len(all_names) or not all_names[i + 1].startswith(name)) and name.endswith("]"):
        leaf_nodes.append(name)
return leaf_nodes
```

Since I didn't get enough context, I might be wrong...

chicm-ms: I tested both of your functions, did not find problem on simple model.
but with a Bert model: 
1. Yuge's code generates 163 results with 20 unique values
2. Quanlu's code generate 18 values

the two extra values in Yuge's results are:
```
{'BertModel/BertEncoder[encoder]/BertLayer/BertAttention[attention]/BertSelfAttention[self]',
 'BertModel/BertEncoder[encoder]/BertLayer/BertIntermediate[intermediate]'} 
```
these 2 are not leaf nodes per Bert source code: https://github.com/huggingface/transformers/blob/e392ba6938f50655a195ea7ec8a260b1e9fc6058/src/transformers/modeling_bert.py#L321

chicm-ms: I test both your functions, they can generate same results.

src/sdk/pynni/nni/compression/speedup/torch/compressor.py
QuanluZhang: just childs for now...

src/sdk/pynni/nni/compression/speedup/torch/compressor.py
QuanluZhang: @ultmaster too complex, I am not sure whether the logic is correct or not...

PR #2268, opened 2020-04-03
Update pr test cases for windows (#2267)

chicm-ms:
The Memory/CPU resource of azure provided hosts are very limited, it seems 20 intermediate results of each trial is too much, and causing problems on Windows.  It appears changing to 2 intermediate result for each trial solved the problem.

There is no intermediate result in the PR test cases in v1.4 and before,  and 20 intermediate results were added into each trial job in recent test refactoring.

### DIFF
--- a/test/config/naive_trial/naive_trial.py
+++ b/test/config/naive_trial/naive_trial.py
@@ -1,0 +1,14 @@
+import time
+import nni
+
+if __name__ == '__main__':
+    print('trial start')
+    params = nni.get_next_parameter()
+    print('params:', params)
+    epochs = 2
+
+    for i in range(epochs):
+        nni.report_intermediate_result(0.1 * (i+1))
+        time.sleep(1)
+    nni.report_final_result(0.8)
+    print('trial done')
--- a/test/config/pr_tests.yml
+++ b/test/config/pr_tests.yml
@@ -70,12 +70,18 @@ testCases:
   config:
     maxTrialNum: 2
     trialConcurrency: 2
+    trial:
+      codeDir: ../naive_trial
+      command: python3 naive_trial.py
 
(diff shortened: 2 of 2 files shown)


test/config/naive_trial/naive_trial.py
```diff
@@ -0,0 +1,14 @@
+import time
```
QuanluZhang, line 1: as you added naive_trial.py, what is the original trial.py used for?

QuanluZhang:
> The Memory/CPU resource of azure provided hosts are very limited, it seems 20 intermediate results of each trial is too much, and causing problems on Windows. It appears changing to 2 intermediate result for each trial solved the problem.

a little curious about why this is not a problem before?

chicm-ms:
> > The Memory/CPU resource of azure provided hosts are very limited, it seems 20 intermediate results of each trial is too much, and causing problems on Windows. It appears changing to 2 intermediate result for each trial solved the problem.
> 
> a little curious about why this is not a problem before?

It was 0 intermediate result previously,  I changed it to 20 during refactoring.

chicm-ms:
> Exactly what kind of resource constraints makes it unable to handle 20 intermediate results? Can this, by any chance, be related to previous instability problem on Windows platform?

I ran the same test on our Windows with 56GB Memory 6vcpu for a lot of times, the problem can not be reproduced.  
The azure provided win2016 image is on ds2_v2 image: https://docs.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series  , which is 7GB Memory.

test/config/naive_trial/naive_trial.py
chicm-ms: It is still being used by integration test.

PR #2274, opened 2020-04-07
fix bug of refresh from disable refresh to refresh

Lijiaoa:


src/webui/src/App.tsx
```diff
@@ -66,13 +67,12 @@ class App extends React.Component<{}, AppState> {
...
-    
+
     changeInterval = (interval: number): void => {
         this.setState({ interval });
-        if (this.timerId === null && interval !== 0) {
-            window.setTimeout(this.refresh);
-        } else if (this.timerId !== null && interval === 0) {
-            window.clearTimeout(this.timerId);
+        if (this.flag === false) {
+            this.refresh();
```
liuzhe-lz, line 74: `refresh()` will setup a (sequence of) timer. If I spam changing interval, there will be bunch of timers.

src/webui/src/App.tsx
```diff
@@ -17,7 +17,8 @@ interface AppState {
 }
 
 class App extends React.Component<{}, AppState> {
-    private timerId!: number | null;
+    private timerId!: number | undefined;
+    private flag: boolean = true;
```
liuzhe-lz, line 21: Please use a meaningful name. I can't really understand what this flag does.

Lijiaoa: OK

src/webui/src/App.tsx
Lijiaoa: had updated.

PR #2276, opened 2020-04-07
Fix IPC command length limit

liuzhe-lz:
The protocol for non-trialkeeper services is changed so that when a command exceeds 999999 length, it will be splitted into multiple parts.
For example, an "initialize" (`IN`) command with 3000000 letter `x` will be splitted to:
  - `INEXCEED` + 999999 letter `x`
  - `_CEXCEED` + 999999 letter `x` (`_C` stands for "continue")
  - `_CEXCEED` + 999999 letter `x`
  - `_E000003xxx` (`_E` stands for "end")


src/nni_manager/core/ipcInterface.ts
```diff
@@ -22,13 +22,22 @@ const ipcIncomingFd: number = 4;
  * @returns binary command data
  */
 function encodeCommand(commandType: string, content: string): Buffer {
-    const contentBuffer: Buffer = Buffer.from(content);
-    if (contentBuffer.length >= 1_000_000) {
-        throw new RangeError('Command too long');
+    let contentBuffer: Buffer = Buffer.from(content);
```
QuanluZhang, line 25: suggest to add docstring to explain the protocol

src/sdk/pynni/nni/protocol.py
```diff
@@ -43,9 +47,16 @@ def send(command, data):
...
+        if len(data) <= _max_msg_len:
+            msg = b'%b%06d%b' % (command.value, len(data), data)
+        else:
+            msg = b'%b______%b' % (command.value, data[:_max_msg_len])
+            data = data[_max_msg_len:]
+            while len(data) > _max_msg_len:
+                msg += b'_C999999%b' % data[:_max_msg_len]
+                data = data[_max_msg_len:]
+            msg += b'_E%06d%b' % (len(data), data)
         _out_file.write(msg)
```
QuanluZhang, line 60: if we can write data with any large content length, why not simply increase the number of bits for representing data length?

src/nni_manager/.eslintrc
```diff
@@ -23,7 +23,8 @@
         "@typescript-eslint/consistent-type-assertions": 0,
         "@typescript-eslint/no-inferrable-types": 0,
         "no-inner-declarations": 0,
-        "@typescript-eslint/no-var-requires": 0
+        "@typescript-eslint/no-var-requires": 0,
+        "no-constant-condition": ["error", {"checkLoops": false}]
```
liuzhe-lz, line 27: This rule is to allow `while (true)`

src/nni_manager/core/ipcInterface.ts
liuzhe-lz: The doc will be added to wiki.

PR #2294, opened 2020-04-11
refactoring: early return 'if else' syntax -> 'if' syntax

marload:
I think 'if syntax' is better than 'if else syntax' in the early return pattern.


liuzhe-lz:
Personally I prefer `if else` syntax unless it's *logically* early return.
For example, I think "windows" and "linux" are just two different branchs, while something like "found result in cache, no need to calculate" can be considered early return.
Anyway, it's just personal taste and I'm okay with both.

QuanluZhang:
@marload thanks for your contribution, as @liuzhe-lz said, `if else` better expresses the logic in those four places.

Issue #1984, opened 2020-01-31
The image in doc is out of date


xuehui1991:
![image](https://user-images.githubusercontent.com/6280746/73514256-ef762000-442b-11ea-81ff-dce46c9e1fda.png)

 The picture of WebUI in PAI doc is out of date.

**Short summary about the issue/question**:

**Brief what process you are following**: 

**How to reproduce it**: 

**nni Environment**:
- nni version:
- nni mode(local|pai|remote):
- OS:
- python version:
- is conda or virtualenv used?: 
- is running in docker?:

**need to update document(yes/no)**:

**Anything else we need to know**:

lvybriage:
I'll fix soon. Thank you, @xuehui1991

scarlett2018:
> I'll fix soon. Thank you, @xuehui1991

@lvybriage - - any work left for this issue? may we close it now?

scarlett2018:
in 1.6.

PR #2302, opened 2020-04-13
Update outdated pai image

SparkSnail:
https://github.com/microsoft/nni/issues/1984


QuanluZhang:
@SparkSnail so by default, the log path is local path? log file is copy to local by default?

SparkSnail:
> @SparkSnail so by default, the log path is local path? log file is copy to local by default?

This path is a local path which mounted to NFS, the data generated in container will be  synchronized to this path autonomically.

PR #2303, opened 2020-04-14
fix trial export

leckie-chn:
fix `nnictl experiment export` with the following error:
```python
  File "c:\program files\python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\program files\python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Program Files\Python37\Scripts\nnictl.exe\__main__.py", line 9, in <module>
  File "c:\program files\python37\lib\site-packages\nni_cmd\nnictl.py", line 219, in parse_args
    args.func(args)
  File "c:\program files\python37\lib\site-packages\nni_cmd\nnictl_utils.py", line 694, in export_trials_data
    raise(e)
  File "c:\program files\python37\lib\site-packages\nni_cmd\nnictl_utils.py", line 689, in export_trials_data
    formated_record = {**record['parameter'], **record['value'], **{'id': record['id']}}
TypeError: 'str' object is not a mapping
```
also fix blank new line issue on exported csv file :)

### DIFF
--- a/test/nni_test/nnitest/validators.py
+++ b/test/nni_test/nnitest/validators.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 
 import os.path as osp
+import subprocess
 import json
 import requests
 import nnicli as nc
@@ -12,6 +13,13 @@ class ITValidator:
     def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
         pass
 
+class ExportValidator(ITValidator):
+    def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
+        exp_id = osp.split(experiment_dir)[-1]
+        proc1 = subprocess.run(["nnictl", "experiment", "export", exp_id, "-t", "csv", "-f", "report.csv"])
+        assert proc1.returncode == 0, '`nnictl experiment export -t csv` failed with code %d' % proc1.returncode
+        proc2 = subprocess.run(["nnictl", "experiment", "export", exp_id, "-t", "json", "-f", "report.json"])
+        assert proc2.returncode == 0, '`nnictl experiment export -t json` failed with code %d' % proc2.returncode
 
 class MetricsValidator(ITValidator):
     def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
--- a/test/nni_test/nnitest/validators.py
+++ b/test/nni_test/nnitest/validators.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 
 import os.path as osp
+from os import remove
 import subprocess
 import json
 import requests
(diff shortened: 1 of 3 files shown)


chicm-ms:
Does this fix work on both linux and windows for the blank new line problem?

QuanluZhang:
@chicm-ms @leckie-chn maybe we also should test `export` in UT or IT?

tools/nni_cmd/nnictl_utils.py
```diff
@@ -700,11 +700,11 @@ def export_trials_data(args):
                 trial_records = []
                 for record in content:
                     if not isinstance(record['value'], (float, int)):
-                        formated_record = {**record['parameter'], **record['value'], **{'id': record['id']}}
+                        formated_record = {**record['parameter'], **json.loads(record['value']), **{'id': record['id']}}
```
QuanluZhang, line 703: maybe we should use `json_tricks` instead of `json` in this file?

QuanluZhang: for example `NaN`

tools/nni_cmd/nnictl_utils.py
```diff
@@ -700,11 +700,11 @@ def export_trials_data(args):
...
                     else:
                         formated_record = {**record['parameter'], **{'reward': record['value'], 'id': record['id']}}
                     trial_records.append(formated_record)
-                with open(args.path, 'w') as file:
+                with open(args.path, 'w', newline='') as file:
```
QuanluZhang, line 707: why add `newline` here?

leckie-chn: without `newline=''`, the exporter will produce extra newline in the exported csv file

leckie-chn: python doc reference: https://docs.python.org/3/library/csv.html#csv.writer

chicm-ms:
> @chicm-ms @leckie-chn maybe we also should test `export` in UT or IT?

For now there is no UT for nnictl,   @leckie-chn would you an IT case for `nnictl export` ?

tools/nni_cmd/nnictl_utils.py
leckie-chn: @QuanluZhang there are still many other usage of `json` package in `nnictl_utils.py`. Do you want to replace all of them?

test/nni_test/nnitest/validators.py
```diff
@@ -12,6 +13,13 @@ class ITValidator:
     def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
         pass
 
+class ExportValidator(ITValidator):
+    def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
+        exp_id = osp.split(experiment_dir)[-1]
+        proc1 = subprocess.run(["nnictl", "experiment", "export", exp_id, "-t", "csv", "-f", "report.csv"])
+        assert proc1.returncode == 0, '`nnictl experiment export -t csv` failed with code %d' % proc1.returncode
```
chicm-ms, line 20: suggest to print out the content of csv and json files, so that we can check the result manually on pipeline when necessary.

tools/nni_cmd/nnictl_utils.py
QuanluZhang: I think it is okay to keep them as long as `json.load` is not used on `value`. Could you help give a quick check?

QuanluZhang: seem no drawback to replace them, decided on you :)

test/nni_test/nnitest/validators.py
leckie-chn: fixed in https://github.com/microsoft/nni/pull/2303/commits/d5951935b42303b2ca9512c62ec74c2c6af718f1

PR #2333, opened 2020-04-17
Overview concurrency tooltip

Lijiaoa:
- change waiting trial status color to `#FDC401`
- image

### DIFF
--- a/src/webui/src/static/const.ts
+++ b/src/webui/src/static/const.ts
@@ -35,8 +35,10 @@ const COLUMN = ['Trial No.', 'ID', 'Duration', 'Status', 'Default', OPERATION];
 // all choice column !dictory final
 const COLUMNPro = ['Trial No.', 'ID', 'Start Time', 'End Time', 'Duration', 'Status',
 'Intermediate result', 'Default', OPERATION];
+const  CONCURRENCYTOOLTIP = 'TrialConcurrency is the number of trials running concurrently, which is the number of GPUs you want to use if you are setting gpuNum to 1.';
+
 export {
     MANAGER_IP, DOWNLOAD_IP, trialJobStatus, COLUMNPro, WEBUIDOC,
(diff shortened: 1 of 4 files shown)


src/webui/src/static/const.ts
```diff
@@ -35,8 +35,10 @@ const COLUMN = ['Trial No.', 'ID', 'Duration', 'Status', 'Default', OPERATION];
 // all choice column !dictory final
 const COLUMNPro = ['Trial No.', 'ID', 'Start Time', 'End Time', 'Duration', 'Status',
 'Intermediate result', 'Default', OPERATION];
+const  CONCURRENCYTOOLTIP = 'TrialConcurrency is the number of trials running concurrently, which is the number of GPUs you want to use if you are setting gpuNum to 1.';
```
liuzhe-lz, line 38: Suggest: Trial concurrency is the number of trials running concurrently, which is the number of GPUs you want to use if you set gpuNum to 1.

Lijiaoa: OK!

QuanluZhang: `which` is a little misleading here.

I think we can just keep the former half: Trial concurrency is the number of trials running concurrently.

Lijiaoa: I think if only show `Trial concurrency is the number of trials running concurrently.` user still don't understand `concurrency` meaning. Depend on you!

Lijiaoa: Finally， this tooltip changed to `Trial concurrency is the number of trials running concurrently.`  by Quanlu's comments.

src/webui/src/static/style/tableStatus.css
```diff
@@ -15,7 +15,7 @@
     color:	#EB0716;
 }
 .WAITING{
-    color:	#008B8B;
+    color: yellow;
```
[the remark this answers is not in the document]
yangou1988: Please try #FDC401 to display status WAITTING

PR #2357, opened 2020-04-22
improve PBT tuner

QuanluZhang:
1. handle trial failure case
2. support data import
3. add unittest


src/sdk/pynni/nni/pbt_tuner/pbt_tuner.py
```diff
@@ -74,18 +74,16 @@ def exploit_and_explore(bot_trial_info, top_trial_info, factor, resample_probabi
     top_hyper_parameters = top_trial_info.hyper_parameters
     hyper_parameters = copy.deepcopy(top_hyper_parameters)
     random_state = np.random.RandomState()
+    hyper_parameters['load_checkpoint_dir'] = hyper_parameters['save_checkpoint_dir']
```
chicm-ms, line 77: just curious, why this tuner explicitly pass `checkpoint_dir` to trial? is it a part of the tuner algorithm?

chicm-ms: Got it, it is for weight sharing..

src/sdk/pynni/tests/test_builtin_tuners.py
```diff
@@ -159,6 +159,61 @@ def search_space_test_all(self, tuner_factory, supported_types=None, ignore_type
...
+        tuner = PBTTuner(
+            all_checkpoint_dir=all_checkpoint_dir,
+            population_size=population_size
+        )
+        self.assertIsInstance(tuner, Tuner)
+        tuner.update_search_space(search_space)
+        for i in range(population_size - 1):
+            save_dirs.append(os.path.join(all_checkpoint_dir, str(i), str(1)))
+        for save_dir in save_dirs:
+            os.makedirs(save_dir, exist_ok=True)
```
chicm-ms, line 205: better to delete these dirs again at the end?

QuanluZhang: good point, thanks

chicm-ms:
Looks good.

PR #2372, opened 2020-04-24
Fix `make dev-install`

liuzhe-lz:


src/webui/yarn.lock
```diff
@@ -5593,7 +5593,7 @@ load-json-file@^4.0.0:
     pify "^3.0.0"
     strip-bom "^3.0.0"
 
-loader-fs-cache@>=1.0.3, loader-fs-cache@^1.0.0:
+loader-fs-cache@^1.0.0:
```
liuzhe-lz, line 5596: This line was automatically modified on `make`.

Makefile
```diff
@@ -205,14 +205,14 @@ install-node-modules:
 .PHONY: dev-install-node-modules
 dev-install-node-modules:
 	#$(_INFO) Installing NNI Package $(_END)
-	rm -rf $(NNI_PKG_FOLDER)
```
squirrelsc, line 208: It may need to test what if `make install` and then `make dev-install`.

QuanluZhang:
@liuzhe-lz please solve conflict

Makefile
liuzhe-lz: You should never do that. Remove NNI first before switching to another installation setup.
We don't want to investigate what will happen if pip is requested to overwrite a normal package with an egg package.

squirrelsc: Suggest to keep this line. Don't rely on human to do right things, and the normal install also have this line.

liuzhe-lz: I don't agree. Removing this folder does NOT uninstall NNI from user's system, instead it leads to a mysteries "partially installed/removed" state.
If a DEVELOPER do something wrong, I prefer to fail early.

squirrelsc: Any ambiguous state should be fixed in the Makefile. This change makes dev/normal installs script inconsistent, and dig an unnecessary pit. Anyway, if you still prefer, go ahead merge it.

liuzhe-lz: If you `make install` twice, it will be okay both with or without this folder, since pip and yarn are allowed to run multiple times with same options. Maybe this folder is removed in case the last bulid was corrupted.
But if someone `make install` then `make dev-install`, I don't know what will happen. Because the options feeded to pip are changed, and it's somewhat difficult to figure out pip's exact behaviour in such scenario.
If we want to ensure a clean install, we should trigger `clean` and `uninstall` targets instead of removing one or two folders. But as far as I know this is not the common practice of Makefile.

PR #2405, opened 2020-05-06
refactor: update eslint rules to remove warnings during build

squirrelsc:
update nni manager eslint rules.
1. add rule to ignore unused args, and treat other unused as error.
2. add run to treat no explicit function return type as error.
3. fix all errors.

### DIFF
--- a/src/nni_manager/training_service/pai/paiYarn/paiYarnTrainingService.ts
+++ b/src/nni_manager/training_service/pai/paiYarn/paiYarnTrainingService.ts
@@ -8,24 +8,22 @@ import * as path from 'path';
 import * as request from 'request';
 import * as component from '../../../common/component';
 
-import { EventEmitter } from 'events';
 import { Deferred } from 'ts-deferred';
 import { String } from 'typescript-string-operations';
-import { getExperimentId } from '../../../common/experimentStartupInfo';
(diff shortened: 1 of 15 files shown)


src/nni_manager/training_service/pai/paiYarn/paiYarnTrainingService.ts
```diff
@@ -283,10 +281,8 @@ class PAIYarnTrainingService extends PAITrainingService {
                 Authorization: `Bearer ${this.paiToken}`
             }
         };
-        request(submitJobRequest, (error: Error, response: request.Response, body: any) => {
+        request(submitJobRequest, (error: Error, response: request.Response, _body: any) => {
             if ((error !== undefined && error !== null) || response.statusCode >= 400) {
-                const errorMessage: string = (error !== undefined && error !== null) ? error.message :
```
QuanluZhang, line 288: @SparkSnail This `errorMessage` is not used?

SparkSnail: Seems only have declare of `errorMessage`, and `this.log.error(errorMessage)` is missing for some reason. I think it's better to add `this.log.error(errorMessage)` instead of deleting it.

PR #2416, opened 2020-05-09
Support dev install on windows

squirrelsc:
1. Add scripts in install.ps1 to support install as devlopment.
2. Add nasui uninstall scripts in uninstall.ps1.
3. Use rmdir more, as it works with symbolic links better.
4. Simplify development document, and include this change in it.


uninstall.ps1
```diff
@@ -4,26 +4,29 @@ $env:PYTHONIOENCODING = "UTF-8"
...
 
-# clean
-Remove-Item "src/nni_manager/dist" -Recurse -Force
-Remove-Item "src/nni_manager/node_modules" -Recurse -Force
-Remove-Item "src/webui/build" -Recurse -Force
-Remove-Item "src/webui/node_modules" -Recurse -Force
-Remove-Item $NNI_YARN_FOLDER -Recurse -Force
-Remove-Item $NNI_NODE_FOLDER -Recurse -Force
+# clean up
+cmd /c rmdir /s /q "build"
```
liuzhe-lz, line 24: What's the drawback of `Remove-Item`?

squirrelsc: Remove-item doesn't work for symbol links. If use powershell to remove a symbol link, it needs to call like (Get-Item xxx).Delete(). But the Delete() doesn't work for non-empty physical folder.

PR #2418, opened 2020-05-11
Add Python SDK version string

liuzhe-lz:
Test PYPI: [linux](https://test.pypi.org/project/nni/0.0.9.2005180212/#files) [windows](https://test.pypi.org/project/nni/0.0.9.2005180241/#files)

I'm confused by the letter "v" problem.
The version string placeholder is "999.0.0-developing", without leading letter "v". But during installation, it's replaced by git tag, which is something like "v1.5".
If the leading letter "v" is desired, we should fix the placeholder; and if it should not be there, we should fix `NNI_VERSION_VALUE` in makefiles. Personally I prefer not to include leading "v" in `NNI_VERSION_VALUE`, because that's Python's common practice as I know. If a module wants the leading "v", it can append it before placeholder and the replacement should work fine.

### DIFF
--- a/deployment/pypi/Makefile
+++ b/deployment/pypi/Makefile
@@ -47,6 +47,7 @@ build:
 	cp $(CWD)../../src/nni_manager/package.json $(CWD)nni
 	sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' $(CWD)nni/package.json
 	cd $(CWD)nni && $(NNI_YARN) --prod
+	sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' $(CWD)../../src/sdk/pynni/nni/__init__.py
 	cd $(CWD) && sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' setup.py && python3 setup.py bdist_wheel -p $(WHEEL_SPEC)
 	cd $(CWD)
 
--- a/src/sdk/pynni/nni/__init__.py
+++ b/src/sdk/pynni/nni/__init__.py
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+__version__ = '999.0.0-developing'
+
 from .env_vars import dispatcher_env_vars
 
 if dispatcher_env_vars.SDK_PROCESS != 'dispatcher':
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,7 @@ install-dependencies: $(NNI_NODE_TARBALL) $(NNI_YARN_TARBALL)
 .PHONY: install-python-modules
 install-python-modules:
 	#$(_INFO) Installing Python SDK $(_END)
+	sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' src/sdk/pynni/nni/__init__.py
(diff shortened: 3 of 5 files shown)


src/sdk/pynni/nni/__init__.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+__version__ = '999.0.0-developing'
```
chicm-ms, line 4: where this string is replaced with correct version?  this string in `setup.py` is replaced in `Makefile`

deployment/pypi/Makefile
```diff
@@ -47,6 +47,7 @@ build:
 	cp $(CWD)../../src/nni_manager/package.json $(CWD)nni
 	sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' $(CWD)nni/package.json
 	cd $(CWD)nni && $(NNI_YARN) --prod
+	sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' $(CWD)../../src/sdk/pynni/nni/__init__.py
 	cd $(CWD) && sed -ie 's/$(NNI_VERSION_TEMPLATE)/$(NNI_VERSION_VALUE)/' setup.py && python3 setup.py bdist_wheel -p $(WHEEL_SPEC)
```
chicm-ms, line 51: this a `sed` replacement statement in nni/Makefile

liuzhe-lz: When checking the Makefile I found a problem. I think dev-install should not replace the version string in `setup.py`.
Firstly developers are highly likely tracking their source files with git, and therefore the files should keep in sync with our repo. If a developer dev-install source code, do some editting, and then `git commit -a`, the replaced version string will be added to patch.
And I think the "developing" placeholder is a good version string for dev-install, because the target of symbol links may be any version. For example if I dev-install on v1.5 branch, and checkout v1.6 code tree later, the version string becomes misleading.
So I suggest only replace version string on normal install, and leave the placeholder unchanged for dev install.

chicm-ms: Yes, I agree that replacement on source generates annoying git changes.
The IT pipelines install nnimanager with source install and use depoyment install with docker based training service,  and there is a version check in trial keeper...  
https://github.com/microsoft/nni/blob/master/tools/nni_trial_tool/trial_keeper.py#L122
If we replace setup.py version and not replace nni.`__version__`  with source install, then in IT pipeline, the two types of version (nni.`__version__` and pip distribution version) will be different, this may cause potential problem.

suggest to remove the existing replacement in source install and update trial keeper version search pattern.   https://github.com/microsoft/nni/blob/master/tools/nni_trial_tool/trial_keeper.py#L24

liuzhe-lz: I agree source install should keep in sync with pip install. But I think dev install does not necessarily need to be same as release version. I mean, `999.0-developing` is a good version number for dev install I think, since it's always the "latest".
If dev install is currently used in release build, I feel okay to replace version string for now. But I believe it should use a separate target like `fast-install`.

So should we simply replace the version string anywhere for v1.6 release, or delay this PR for a refactor?

chicm-ms: build release pipeline is using deployment/pypi/Makefile pip install.  Since `999.0-developing` is a good version for dev, maybe remove the current `sed` replacement on `setup.py` for source/dev install ? This also can be done later.

PR #2401, opened 2020-05-01
add function for calculating the real model size

marsggbo:
refer to #1947

PR #2420, opened 2020-05-11
Replace layer choice with selected module after applied fixed architecture

ultmaster:
This should solve the problem #2401 is trying to solve.

### DIFF
--- a/docs/en_US/NAS/NasGuide.md
+++ b/docs/en_US/NAS/NasGuide.md
@@ -176,7 +176,7 @@ For example,
 }
 ```
 
-After applying, the model is then fixed and ready for final training. The model works as a single model, although it might contain more parameters than expected. This comes with pros and cons. The good side is, you can directly load the checkpoint dumped from supernet during the search phase and start retraining from there. However, this is also a model with redundant parameters and this may cause problems when trying to count the number of parameters in the model. For deeper reasons and possible workarounds, see [Trainers](./NasReference.md).
+After applying, the model is then fixed and ready for final training. The model works as a single model, and unused parameters and modules will be pruned.
 
 Also, refer to [DARTS](./DARTS.md) for code exemplifying retraining.
 
--- a/docs/en_US/NAS/NasGuide.md
+++ b/docs/en_US/NAS/NasGuide.md
@@ -176,7 +176,7 @@ For example,
 }
 ```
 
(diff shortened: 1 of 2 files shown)


src/sdk/pynni/nni/nas/pytorch/fixed.py
```diff
@@ -73,6 +77,36 @@ def sample_final(self):
...
+        Parameters
+        ----------
+        module : nn.Module
+            Module to be processed.
+        prefix : str
+            Module name under global namespace.
+        """
+        if module is None:
+            module = self.model
+        for name, mutable in module.named_children():
```
liuzhe-lz, line 95: Why not use the util function to traverse here?

src/sdk/pynni/nni/nas/pytorch/fixed.py
```diff
@@ -73,6 +77,36 @@ def sample_final(self):
...
+            global_name = (prefix + "." if prefix else "") + name
+            if isinstance(mutable, LayerChoice):
+                chosen = self._fixed_arc[mutable.key]
+                if sum(chosen) == 1 and max(chosen) == 1 and not mutable.return_mask:
```
QuanluZhang, line 99: what is the meaning of `not mutable.return_mask`?

ultmaster: If true, layer choice will return the choice vector. If we replace the layer choice module with the selected module, choice vector will be no longer available, and users’ code will crash.

QuanluZhang:
please update doc at proper place.

docs/en_US/NAS/NasGuide.md
```diff
@@ -176,7 +176,7 @@ For example,
 }
 ```
 
-After applying, the model is then fixed and ready for final training. The model works as a single model, although it might contain more parameters than expected. This comes with pros and cons. The good side is, you can directly load the checkpoint dumped from supernet during the search phase and start retraining from there. However, this is also a model with redundant parameters and this may cause problems when trying to count the number of parameters in the model. For deeper reasons and possible workarounds, see [Trainers](./NasReference.md).
+After applying, the model is then fixed and ready for final training. The model works as a single model, and unused parameters and modules will be pruned.
```
QuanluZhang, line 179: "will be" -> "are"

PR #2421, opened 2020-05-12
Fix it local windows

leckie-chn:
* fix empty trial results issue on windows
* return error when `nnictl trial export` with no trial results, along with test case


src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -334,9 +334,11 @@ class LocalTrainingService implements TrainingService {
...
                 }
                 //Refer https://github.com/Juul/tail-stream/issues/20
-                stream.end(0);
-                stream.emit('end');
-                this.jobStreamMap.delete(trialJob.id);
+                setTimeout(() => {
+                    stream.end(0);
+                    stream.emit('end');
+                    this.jobStreamMap.delete(trialJob.id);
+                }, 5000);
```
chicm-ms, line 341: better to add a comment here to state the reason for delay ending stream.

Issue #1541, opened 2019-09-11
Show trial error message


lvybriage:
Currently, the storage full message is only able to be got from the trial-jobs RESTful API, would like to expose it to the WEBUI.

![image](https://user-images.githubusercontent.com/35484733/64665344-cfe25900-d484-11e9-8e10-c5333f0defaa.png)

scarlett2018:
@lvybriage @yangou1988 - seems this one is a good one to merge with the warning message. we can show error/warning message in one style somehow. cc @QuanluZhang

Lijiaoa:
This issue had fixed in PR https://github.com/microsoft/nni/pull/2424

WebUI will show /trial-jobs error message if /trial-jobs return error filed. And website will stop refreshing page.

scarlett2018:
in 1.6.

PR #2424, opened 2020-05-13
Show trial error message

Lijiaoa:
This PR is for issue https://github.com/microsoft/nni/issues/1541 .

WebUI will show /trial-jobs error message if /trial-jobs return error filed. And website will stop refreshing page.

image

image


liuzhe-lz:
I think this message should be used for all REST requests instead of `/trial-jobs` only.

PR #2425, opened 2020-05-13
Increase IPC message length to 10^14

liuzhe-lz:


src/nni_manager/core/test/ipcInterface.test.ts
```diff
@@ -88,21 +80,11 @@ describe('core/protocol', (): void => {
     });
 
     it('sendCommand() should work without content', (): void => {
-        assert.equal(sentCommands[0], '(\'IN\', \'\')');
+        assert.equal(sentCommands[0], "('IN', '')");
```
QuanluZhang, line 83: what is the reason to make this change?

PR #2435, opened 2020-05-15
Analysis utils

zheng-ningxin:
Add several analysis tools for the neural network: (1) sensitivity analysis, (2) topology analysis
### Sensitivity Analysis Tool
 1. 'sensitivity_analysis.py' automatically analyze the sensitivity to the prune ratio for each layer in the model. In addition, the sensitivity results can also be visualized for user observation.  

### Topology Analysis Tools
1 'shape_dependency.py' analyze the network architecture and find the layers that have shape dependencies.
2 'mask_conflict.py' this module can fix the mask conflict between the layers has channel dependency. This module should be called before the 'speedup' ,so that, the speedup module can handle the models with residual connections.


QuanluZhang:
Thanks @zheng-ningxin , could you add doc (e.g., how to use the sensitivity tool, the requirement of this tool) under `https://github.com/microsoft/nni/tree/master/docs/en_US/Compressor`?

setup.py
```diff
@@ -40,7 +40,9 @@ def read(fname):
         'schema',
         'PythonWebHDFS',
         'colorama',
-        'scikit-learn>=0.20,<0.22'
+        'scikit-learn>=0.20,<0.22',
+        'graphviz',
+        'matplotlib'
```
QuanluZhang, line 45: suggest to use `nnictl package install` to add these dependencies.

src/sdk/pynni/nni/analysis_utils/sensitivity/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) Microsoft Corporation.
```
QuanluZhang, line 1: if `analysis_utils` is only used by model compression, suggest to put this directory under `src/sdk/pynni/nni/compression`

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+
+class SensitivityAnalysis:
+    def __init__(self, model, val_func, sparsities=None, prune_type='l1', early_stop=1.0):
+        """
+        Perform sensitivity analysis for this model.
+        Parameters
+        ----------
+            model:
+                the model to perform sensitivity analysis
+            val_func:
```
QuanluZhang, line 37: please provide more description about how to write this function.

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+                validation function for the model. Due to
+                different models may need different dataset/criterion
+                , therefore the user need to cover this part by themselves.
+                val_func take the model as the first input parameter, and
+                return the accuracy as output.
+            sparsities:
+                The sparsity list provided by users.
```
QuanluZhang, line 44: what is the meaning of this list? each element is a layer's sparsity?

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+logger.setLevel(logging.INFO)
+
+
+class SensitivityAnalysis:
+    def __init__(self, model, val_func, sparsities=None, prune_type='l1', early_stop=1.0):
+        """
+        Perform sensitivity analysis for this model.
+        Parameters
+        ----------
+            model:
```
QuanluZhang, line 35: the docstring format of the input arguments is:
arg : type
you missed the type field.

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
QuanluZhang: what is the meaning of the three numbers?

zheng-ningxin: They are sparsity values which means how much weight the pruner should prune.  Take [0.25, 0.5, 0.75] for an example, the SensitivityAnalysis will prune 25% 50% 75% weights gradually for each layer(SensitivityAnalysis only prune a layer once a time, the other layers are set to their original weights), and record the model's accuracy at the same time.

QuanluZhang: got it, thanks. please make it clear in docstring and doc.

zheng-ningxin: Ok~ :)

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+            start:
+                Layer index of the sensitivity analysis start
+            end:
+                Layer index of the sensitivity analysis end
+            val_args:
+                args for the val_function
+            val_kwargs:
+                kwargs for the val_funtion
+                The val_funtion will be called as:
+                    val_function(*val_args, **val_kwargs)
```
QuanluZhang, line 113: the order should be consistent with the input arguments

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+            for op_type in SUPPORTED_OP_TYPE:
+                if isinstance(submodel, op_type):
+                    self.target_layer[name] = submodel
+                    self.already_pruned[name] = 0
+
+    def analysis(self, val_args=None, val_kwargs=None, start=0, end=None):
+        """
+        This function analyze the sensitivity to pruning for
+        each conv layer in the target model.
+        If %start and %end are not set, we analyze all the conv
```
QuanluZhang, line 97: %start -> ```start```

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+        layers by default. Users can specify several layers to
+        analyze or parallelize the analysis process easily through
+        the %start and %end parameter.
+
+        Parameters
+        ----------
+            start:
+                Layer index of the sensitivity analysis start
```
QuanluZhang, line 105: what is the meaning of "Layer index"?

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+        This function analyze the sensitivity to pruning for
+        each conv layer in the target model.
+        If %start and %end are not set, we analyze all the conv
+        layers by default. Users can specify several layers to
+        analyze or parallelize the analysis process easily through
+        the %start and %end parameter.
+
+        Parameters
+        ----------
+            start:
```
QuanluZhang, line 104: no indent here

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
QuanluZhang: this index is only used internally, right? not exposed to users? if it is also exposed to users, we should explain it in more detail.

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+                accuracy when the prune ratio changes
+        """
+        if not end:
+            end = self.layers_count
+        assert start >= 0 and end <= self.layers_count
+        assert start <= end
+        if val_args is None:
+            val_args = []
+        if val_kwargs is None:
+            val_kwargs = {}
```
QuanluZhang, line 128: could directly assign `[]` and `{}` in input arguments.

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
zheng-ningxin: Start and end is the range of the conv layers to be analyzed. For example, if start and end are set to 1 and 10 respectively, then we only analyze the sensitivity of the first ten conv layers of the model. These parameters are for users to analyze the layers in parallel.

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
zheng-ningxin: use empty list or dict as default values will trigger the pylint warning.
https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
QuanluZhang: how does user know which conv is index 2? my concern is this user interface is not very friendly.

zheng-ningxin: how about we use the conv names to identify the start point and end point?

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
QuanluZhang: got it, thanks a lot!

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
QuanluZhang: we can discuss this part in the meeting.

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+        for layerid in range(start, end):
+            name = namelist[layerid]
+            self.sensitivities[name] = {}
+            for sparsity in self.sparsities:
+                # Calculate the actual prune ratio based on the already pruned ratio
+                sparsity = (
+                    1.0 - self.already_pruned[name]) * sparsity + self.already_pruned[name]
+                # TODO In current L1/L2 Filter Pruner, the 'op_types' is still necessary
+                # I think the L1/L2 Pruner should specify the op_types automaticlly
+                # according to the op_names
```
QuanluZhang, line 142: good point, let's discuss in the meeting

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+
+                self.sensitivities[name][sparsity] = val_acc
+                pruner._unwrap_model()
+                del pruner
+                # if the accuracy drop already reach the 'early_stop'
+                if val_acc + self.early_stop < self.ori_acc:
+                    break
+
+            # reset the weights pruned by the pruner, because
+            # out sparsities is sorted, so we donnot need to reset
```
QuanluZhang, line 159: "out" sparsities?

zheng-ningxin: it's a typo. "out" => "input"
what I want to say is "the input sparsity list is sorted."
I'll fix that. Thanks~

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+            # weight of the layer when the sparsity changes, instead,
+            # we only need reset the weight when the pruning layer changes.
+            self.model.load_state_dict(self.ori_state_dict)
+
+        return self.sensitivities
+
+    def visualization(self, outdir, merge=False):
```
QuanluZhang, line 166: what is user interface for visualization?

zheng-ningxin: the 'visualization' is the function that visualizes the sensitivities of the model. This function should be called by user himself if he needs to visualize the sensitivity results. We already return the sensitivity results as a dict in function 'analysis', so the visualization is not set as a mandatory option.

For example:
s_analyzer = SensitivityAnalysis(model=net, val_func=val)
sensitivity = s_analyzer.analysis(val_args=[net])
s_analyzer.visualization(outdir, merge=True)

src/sdk/pynni/nni/analysis_utils/sensitivity/torch/sensitivity_analysis.py
```diff
@@ -0,0 +1,257 @@
...
+                plt.tight_layout()
+                filepath = os.path.join(outdir, 'all.jpg')
+                plt.savefig(filepath, dpi=1000, bbox_inches='tight')
+                styleid += 1
+            plt.close()
+
+    def export(self, filepath):
+        """
+        Export the results of the sensitivity analysis
+        to a csv file.
```
QuanluZhang, line 227: please also describe the data format in the exported file here.

src/sdk/pynni/nni/analysis_utils/topology/torch/graph_from_trace.py
```diff
@@ -0,0 +1,321 @@
+# Copyright (c) Microsoft Corporation.
```
QuanluZhang, line 1: what is the difference between this file and `https://github.com/microsoft/nni/blob/master/src/sdk/pynni/nni/_graph_utils.py`?

zheng-ningxin: graph_from_trace has made some modifications to the network structure for the clarity of the visual graph, such as unpack tensor tuple handly.  graph_from_trace also can build the graph according to the existing Traced graph, so there is no conflict between the two modules, because we only trace the model once. In addition, _graph_utils build the graph by indexing the layers' name which is dangerous(https://github.com/microsoft/nni/issues/2485). I'm not sure if I explain this clearly. if not, let's discuss this in the meeting.

zheng-ningxin:
This pr depends on https://github.com/microsoft/nni/pull/2524, currently, the 'test_analysis_utils' unit test fails due to the bug https://github.com/microsoft/nni/issues/2485. Once https://github.com/microsoft/nni/pull/2524 is merged, this test can pass.

docs/en_US/model_compression.rst
```diff
@@ -22,3 +22,4 @@ For details, please refer to the following tutorials:
     Model Speedup <Compressor/ModelSpeedup>
     Automatic Model Compression <Compressor/AutoCompression>
     Implementation <Compressor/Framework>
+    CompressionUtils <Compressor/CompressionUtils>
```
QuanluZhang, line 25: CompressionUtils -> Compression Utilities

docs/en_US/sdk_reference.rst
```diff
@@ -7,4 +7,5 @@ Python API Reference
     :maxdepth: 1
 
     Auto Tune <autotune_ref>
-    NAS <NAS/NasReference>
\ No newline at end of file
+    NAS <NAS/NasReference>
+    CompressionUtils <Compressor/CompressionUtils>
```
QuanluZhang, line 11: this should be reference not tutorial, could follow `autotune_ref` and `NAS/NasReference`

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -0,0 +1,102 @@
...
+        self.dummy_input = dummy_input
+        self.graph = graph
+        self.mask_file = mask_file
+        self.masks = torch.load(self.mask_file)
+
+    def fix_mask_conflict(self):
+        """
+        Fix the mask conflict before the mask inference for the layers that
+        has shape dependencies. This function should be called before the
+        mask inference of the 'speedup' module.
```
QuanluZhang, line 46: in some cases, if this function is not called, `speedup` will fail?

zheng-ningxin: Yes, the speedup module cannot handle mask conflict right now and will fail on the cases that have mask conflict(such as resnet, mobilenet, etc). Resolving conflicts before speedup will make it easier for speedup to support operations such as add and concat.

docs/en_US/sdk_reference.rst
zheng-ningxin: I have updated the docs, thanks~

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,100 @@
...
+
+s_analyzer = SensitivityAnalysis(model=net, val_func=val)
+sensitivity = s_analyzer.analysis(val_args=[net])
+os.makedir(outdir)
+s_analyzer.export(os.path.join(outdir, filename))
+s_analyzer.visualization(outdir, merge=True)
+s_analyzer.visualization(outdir, merge=False)
+```
+
+Two key parameters of SensitivityAnalysis are model, and val_func. 'model' is the neural network that to be analyzed and the 'val_func' is the validation function that returns the model accuracy on the validation dataset. Due to different scenarios may have different ways to calculate the loss/accuracy, so users should prepare a function that returns the model accuracy on the dataset and pass it to SensitivityAnalysis.
```
chicm-ms, line 34: Depending on the task, the validation metrics maybe other metrics, no necessary to be `accuracy`

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
QuanluZhang: then, maybe we should add resolve of mask conflict in this function: https://github.com/microsoft/nni/blob/master/src/sdk/pynni/nni/compression/torch/speedup/compressor.py#L160

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,100 @@
...
+
+Futhermore, users can specify the sparsities values used to prune for each layer by optinal parameter 'sparsities'.
+```python
+s_analyzer = SensitivityAnalysis(model=net, val_func=val, sparsities=[0.25, 0.5, 0.75])
+``` 
+the SensitivityAnalysis will prune 25% 50% 75% weights gradually for each layer, and record the model's accuracy at the same time (SensitivityAnalysis only prune a layer once a time, the other layers are set to their original weights). If the sparsities is not set, SensitivityAnalysis will use the numpy.arange(0.1, 1.0, 0.1) as the default sparsity values.
+
+Users can also speed up the progress of sensitivity analysis by the early_stop option. By default, the SensitivityAnalysis will test the accuracy under all sparsities for each layer. In contrast, when the early_stop is set, the sensitivity analysis for a layer will stop, when the accuracy drop already reaches the threshold set by early_stop.
+```python
+s_analyzer = SensitivityAnalysis(model=net, val_func=val, sparsities=[0.25, 0.5, 0.75], early_stop=0.1)
```
chicm-ms, line 45: If early_stop is supported,  does the tool support configuring `maximize` / `minimize` mode of metrics ?

src/sdk/pynni/nni/compression/torch/utils/sensitivity_analysis.py
```diff
@@ -0,0 +1,210 @@
...
+        Parameters
+        ----------
+        val_args : list
+            args for the val_function
+        val_kwargs : dict
+            kwargs for the val_funtion
+        start : int
+            Layer index of the sensitivity analysis start.
+        end : int
+            Layer index of the sensitivity analysis end.
```
QuanluZhang, line 114: I think you should define the meaning of index. How users know index of each layer?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
zheng-ningxin: Yes, you are right, and I'm currently working on the support of the add&concat&group for the speedup module. I will submit a pr when I finish. Thanks~

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: Currently, we only support the accuracy, because the stop condition for early_stop is to judge if the current metric(for example, accuracy) is lower than original metric - threshold. If we use loss as the metric, we should judge if the current metric is larger than the original metric + threshold, because the losses of the model are growing during the pruning. However, it only needs a negligible effort to support the loss at the same time. I can update the code if needed.

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: Yes, of course, supporting the maximize/minimize model also only needs a negligible effort. My only concern is whether too many choices(early_stop&minimize/maximize) will confuse users.

chicm-ms: It's reasonable to allow user to set optimization mode,  otherwise only supporting maximize is too limited.  Suggest to replace `accuracy` with general term such as `metrics` or other term, and support maximize/minimize.

zheng-ningxin: Yeah, I'll update the code. Thanks for the advice.

src/sdk/pynni/nni/compression/torch/utils/sensitivity_analysis.py
zheng-ningxin: I agree that leave the parallel options to users may confuse them, How about we not expose these options and parallelize the analysis automatically according to the available GPUs? We can have a quick discussion about this point.

src/sdk/pynni/nni/compression/torch/utils/sensitivity_analysis.py
```diff
@@ -0,0 +1,210 @@
...
+                    row.append(self.sensitivities[layername][sparsity])
+                csv_w.writerow(row)
+
+    def update_already_pruned(self, layername, ratio):
+        """
+        Set the already pruned ratio for the target layer.
+        """
+        self.already_pruned[layername] = ratio
+
+    def load_state_dict(self, state_dict):
```
chicm-ms, line 205: It seems this method is not used?

src/sdk/pynni/nni/compression/torch/utils/sensitivity_analysis.py
```diff
@@ -0,0 +1,210 @@
...
+            csv_w.writerow(header)
+            for layername in self.sensitivities:
+                row = []
+                row.append(layername)
+                for sparsity in sorted(self.sensitivities[layername].keys()):
+                    row.append(self.sensitivities[layername][sparsity])
+                csv_w.writerow(row)
+
+    def update_already_pruned(self, layername, ratio):
+        """
```
chicm-ms, line 200: where is this `update_already_pruned` method used?

zheng-ningxin: Yes, both update_already_pruned and load_state_dict are called in the sensitivity_pruner which may be submitted in another pr.

chicm-ms: OK, got it, thanks

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Topology
+We also provide several tools for the topology analysis during the model compression.
+
+### ChannelDependency
+Complicated models may has residual connection/concat operations in their models. When the user prune these models, they need to be careful about the channel-count dependencies between the convolution layers in the model. If the layers has channel dependency are assigned with different sparsities (here we only discuss the structured pruning by L1FilterPruner/L2FilterPruner), then even the pruned model with mask works fine. but the pruned model cannot be speedup to the final model that run on the devices, because there will be a shape conflict when the model try to add/concat the outputs of these layers. This model is to find the layers that has channel count dependencies to help user better prune their model.
+
+#### Usage
+```python
+from nni.compression.analysis_utils.topology.torch.shape_dependency import ChannelDependency
```
chicm-ms: This package name `nni.compression.analysis_utils.topology.torch.shape_dependency` needs to be updated.

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,100 @@
...
+features.6,0.55128,0.53566,0.4887,0.4167,0.31178,0.19152,0.08612,0.01258,0.00236
+features.8,0.55696,0.54194,0.48892,0.42986,0.33048,0.2266,0.09566,0.02348,0.0056
+features.10,0.55468,0.5394,0.49576,0.4291,0.3591,0.28138,0.14256,0.05446,0.01578
+```
+
+## Topology
+We also provide several tools for the topology analysis during the model compression.
+
+### ChannelDependency
+Complicated models may has residual connection/concat operations in their models. When the user prune these models, they need to be careful about the channel-count dependencies between the convolution layers in the model. If the layers has channel dependency are assigned with different sparsities (here we only discuss the structured pruning by L1FilterPruner/L2FilterPruner), then even the pruned model with mask works fine. but the pruned model cannot be speedup to the final model that run on the devices, because there will be a shape conflict when the model try to add/concat the outputs of these layers. This model is to find the layers that has channel count dependencies to help user better prune their model.
```
chicm-ms, line 62: `This model`:  do you mean this tool?

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: I have updated the doc just now, sorry to forget to tell you. Please review on the latest version. Thanks~

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -0,0 +1,191 @@
...
+            # merge the dependencies
+            for parent in parent_convs:
+                if parent in self.dependency:
+                    dependency_set.update(self.dependency[parent])
+            # save the dependencies
+            for _node in dependency_set:
+                self.dependency[_node] = dependency_set
+
+
+    def filter_prune_check(self, ratios):
```
QuanluZhang, line 105: how this function is supposed to be used? by pruner?

zheng-ningxin: I think the pruner may need such an interface, if not, we can delete this function.

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: Yes, I'ill fix it . Thank~

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -0,0 +1,191 @@
...
+                    if other not in ratios:
+                        return False
+                    elif ratios[other] != ratios[node.name]:
+                        return False
+        return True
+
+
+    def export(self, filepath):
+        """
+        export the channel dependencies as a csv file.
```
QuanluZhang, line 143: better to tell the data format of this csv file

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
QuanluZhang: yes, we can add it back when a pruner uses this function.

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -0,0 +1,102 @@
...
+                shape = w_mask.size()
+                count = np.prod(shape[1:])
+                all_ones = []
+                all_zeros = []
+                for i in range(w_mask.size(0)):
+                    _count = torch.sum(w_mask[i])
+                    if _count == count:
+                        all_ones.append(i)
+                    elif _count == 0:
+                        all_zeros.append(i)
```
chicm-ms, line 70: better use matrix op to replace for loop if possible for performance, such as:
`channel_pruned = (w_mask.flatten(1).sum(-1) >= 1).sum().item() < w_mask.size(0)`

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
colorjam: do we support linear layers?

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        # check if the input is legal
+        if traced_model is None:
+            # user should provide model & dummy_input to trace the model or a already traced model
+            assert model is not None and dummy_input is not None
+        self.graph = TorchModuleGraph(model, dummy_input, traced_model)
+        self.dependency = dict()
+        self.build_channel_dependency()
+
+    def get_parent_convs(self, node):
```
colorjam, line 40: do we support linear layers?

zheng-ningxin: Do you mean channel dependencies introduced by the linear layers?

colorjam: Yes.

zheng-ningxin: Linear usually will not introduce channel dependency, because it is used at the end of the model in most cases. By the way, does the linear layer have the concept of output channel？If so, we should add the support for the Linear layers.

colorjam: I think `output_features` in linear layers is similar of output channel. But you're right, linear layers usually not introduce channel dependency.

zheng-ningxin: It cost negligible effort to support the Linear at the same time, so, why not? Thanks for the advice!

src/sdk/pynni/nni/compression/torch/utils/sensitivity_analysis.py
```diff
@@ -0,0 +1,242 @@
...
+            and 'l2', 'fine-grained' is also supported.
+        early_stop : float
+            If this flag is set, the sensitivity analysis
+            for a conv layer will early stop when the validation metric(
+            for example, accurracy/loss) has alreay droped/raised the value
+            of early_stop (0.05 for example).
+            The default value is None, which means the analysis won't stop
+            until all given sparsities are tested.
+        min_threshold : float
+            If the validation metric returned by the val_func is lower
```
chicm-ms, line 57: Suggest to combine `min_threshold` and `max_threshold` into one parameter `early_stop_mode` (`min` or `max`, string type), then the logic will be simpler and more align with other parts of nni (such as tuner's optimization mode)

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,104 @@
...
+            correct += predicted.eq(label).sum().item()
+    return correct / total
+
+s_analyzer = SensitivityAnalysis(model=net, val_func=val)
+sensitivity = s_analyzer.analysis(val_args=[net])
+os.makedir(outdir)
+s_analyzer.export(os.path.join(outdir, filename))
+```
+
+Two key parameters of SensitivityAnalysis are model, and val_func. 'model' is the neural network that to be analyzed and the 'val_func' is the validation function that returns the model accuracy/loss/ or other metrics on the validation dataset. Due to different scenarios may have different ways to calculate the loss/accuracy, so users should prepare a function that returns the model accuracy/loss on the dataset and pass it to SensitivityAnalysis.
```
QuanluZhang, line 32: 'model' -> `model`
'val_func' -> `val_func`

src/sdk/pynni/nni/compression/torch/utils/sensitivity_analysis.py
zheng-ningxin: Yes, it sounds reasonable, I have updated the interface, please review on the latest version, thanks~

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: Thanks~ I have fixed it, please review on the latest version~

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,109 @@
...
+s_analyzer = SensitivityAnalysis(model=net, val_func=val)
+sensitivity = s_analyzer.analysis(val_args=[net])
+os.makedir(outdir)
+s_analyzer.export(os.path.join(outdir, filename))
+```
+
+Two key parameters of SensitivityAnalysis are model, and val_func. model is the neural network that to be analyzed and the val_func is the validation function that returns the model accuracy/loss/ or other metrics on the validation dataset. Due to different scenarios may have different ways to calculate the loss/accuracy, so users should prepare a function that returns the model accuracy/loss on the dataset and pass it to SensitivityAnalysis.
+SensitivityAnalysis can export the sensitivity results as a csv file usage is shown in the example above.
+
+Futhermore, users can specify the sparsities values used to prune for each layer by optinal parameter 'sparsities'.
```
QuanluZhang, line 35: optinal -> optional

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,109 @@
...
+
+Users can also speed up the progress of sensitivity analysis by the early_stop_mode and early_stop_value option. By default, the SensitivityAnalysis will test the accuracy under all sparsities for each layer. In contrast, when the early_stop_mode and early_stop_value are set, the sensitivity analysis for a layer will stop, when the accuracy/loss has already meet the threshold set by early_stop_value. We support four early stop modes:  minimize, maximize, dropped, raised.
+minimize: The analysis stops when the validation metric return by the val_func lower than early_stop_value.
+maximize: The analysis stops when the validation metric return by the val_func larger than early_stop_value.
+dropped: The analysis stops when the validation metric has dropped by early_stop_value.
+raised: The analysis stops when the validation metric has raised by early_stop_value.
```
QuanluZhang, line 45: better to itemize these four values. you can check the rendered version.

QuanluZhang: get confused about `dropped` and `raised`. what is the difference between `dropped`/`raised` and `minimize`/`maximize`?

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,109 @@
...
+      
+```python
+s_analyzer = SensitivityAnalysis(model=net, val_func=val, sparsities=[0.25, 0.5, 0.75], early_stop_mode='dropped', early_stop_value=0.1)
+```
+If users only want to analyze several specified convolutional layers, users can specify the target conv layers by the 'sepcified_layers' parameter in analysis function. For example
```
QuanluZhang, line 50: how to specify a conv layer, using module name or module type?

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,109 @@
...
+```
+layername,0.05,0.1,0.2,0.3,0.4,0.5,0.7,0.85,0.95
+features.0,0.54566,0.46308,0.06978,0.0374,0.03024,0.01512,0.00866,0.00492,0.00184
+features.3,0.54878,0.51184,0.37978,0.19814,0.07178,0.02114,0.00438,0.00442,0.00142
+features.6,0.55128,0.53566,0.4887,0.4167,0.31178,0.19152,0.08612,0.01258,0.00236
+features.8,0.55696,0.54194,0.48892,0.42986,0.33048,0.2266,0.09566,0.02348,0.0056
+features.10,0.55468,0.5394,0.49576,0.4291,0.3591,0.28138,0.14256,0.05446,0.01578
+```
+
+## Topology
```
QuanluZhang, line 73: -> Topology Analysis

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: I have fixed several grammar errors, please review on the latest version, thanks~

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -0,0 +1,109 @@
+# Analysis Utils for Model Compression
```
QuanluZhang, line 1: could add

```eval_rst
.. contents::
```

after this line

QuanluZhang: in order for users to easily get what is the content of this doc

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: Sensitivity analyisis only analyze the conv layers, so we use the module name to specify the layers.

docs/en_US/Compressor/CompressionUtils.md
zheng-ningxin: dropped is used in the cases that you want to stop when the accuracy has dropped by 5%, for example.
minimize is used in the cases that you want to keep the accuracy larger than a specified threshold during the analysis.

docs/en_US/Compressor/CompressionUtils.md
QuanluZhang: then better to make it clear that the layers are specified through PyTorch module name.

zheng-ningxin: Sure, I'll update the doc and make this clear, thanks.

PR #2451, opened 2020-05-18
Reduce trial numbers for PAI IT

SparkSnail:


test/config/training_service.yml
```diff
@@ -78,6 +78,8 @@ paiYarn:
 pai:
   nniManagerIp:
   maxExecDuration: 15m
+  maxTrialNum: 1
```
QuanluZhang, line 81: reduced to only 1?

chicm-ms: Please add comment here why change maxTrialNum.

SparkSnail: > reduced to only 1?

1 trial job per example is enough for testing the example case in PAI. This mastrialNum=1 only works for PAI, and other features like NNICTL, assessor etc. could be tested by other pipelines.

SparkSnail: > Please add comment here why change maxTrialNum.

updated.

test/config/training_service.yml
```diff
@@ -78,12 +78,15 @@ paiYarn:
...
+  # PAI has job submission limitation, set maxTrialNum=1 to control trial job numbers for PAI 
+  maxTrialNum: 1
+  trialConcurrency: 1
   paiConfig:
     host:
     userName:
   trainingServicePlatform: pai
   trial:
-    gpuNum: 1
+    gpuNum: 1
```
chicm-ms, line 89: it seems a space is added here

PR #2470, opened 2020-05-21
improve windows install scripts

squirrelsc:
1. Reduce error message if deleted folder not exists.
1. Reduce unneccessary messages.
1. Organize logic better for reinstall scenario.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -51,7 +51,7 @@ else {
 $PIP_INSTALL = """$NNI_PYTHON3\python"" -m pip install "
 
 if (!(Test-Path $NNI_DEPENDENCY_FOLDER)) {
-    New-Item $NNI_DEPENDENCY_FOLDER -ItemType Directory
+    $null = New-Item $NNI_DEPENDENCY_FOLDER -ItemType Directory
 }
 $NNI_NODE_ZIP = $NNI_DEPENDENCY_FOLDER + "\nni-node.zip"
 $NNI_NODE_FOLDER = $NNI_DEPENDENCY_FOLDER + "\nni-node"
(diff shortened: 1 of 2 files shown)


install.ps1
```diff
@@ -51,7 +51,7 @@ else {
 $PIP_INSTALL = """$NNI_PYTHON3\python"" -m pip install "
 
 if (!(Test-Path $NNI_DEPENDENCY_FOLDER)) {
-    New-Item $NNI_DEPENDENCY_FOLDER -ItemType Directory
+    $null = New-Item $NNI_DEPENDENCY_FOLDER -ItemType Directory
```
SparkSnail, line 54: Why add `$null`?

install.ps1
```diff
@@ -160,31 +159,34 @@ cmd /c $NNI_YARN build
...
+cmd /c if exist "src\nni_manager\dist\static" rmdir /s /q src\nni_manager\dist\static
+cmd /c if exist "src\nni_manager\dist\nasui" rmdir /s /q src\nni_manager\dist\nasui
+
 if ($Development) {
-    New-Item -ItemType Junction -Path $($NNI_PKG_FOLDER) -Target "src\nni_manager\dist"
-    New-Item -ItemType Junction -Path "$($NNI_PKG_FOLDER)\node_modules" -Target "src\nni_manager\node_modules"
-    New-Item -ItemType Junction -Path $($NNI_PKG_FOLDER_STATIC) -Target "src\webui\build"
-    New-Item -ItemType Junction -Path $($NASUI_PKG_FOLDER) -Target "src\nasui\build"
+    $null = New-Item -ItemType Junction -Path $NNI_PKG_FOLDER -Target "src\nni_manager\dist"
+
```
SparkSnail, line 175: remove empty line.

install.ps1
squirrelsc: The original command, `New-Item $NNI_DEPENDENCY_FOLDER -ItemType Directory`, will output command result to console. The default output is like below, if there are a lot of this kind of information, it buries important information. Use $null can put stdout to null pipeline. If there is error, it can show on screen as red text also, not be send to null pipeline.

```
    Directory: D:\temp


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----          6/8/2020  11:30 AM                test1
```

SparkSnail: ok

PR #2483, opened 2020-05-25
homepage migration to readthedocs

leckie-chn:
https://nni.readthedocs.io/en/dev-homepage-1/


docs/en_US/_templates/index.html
```diff
@@ -0,0 +1,506 @@
+{% extends "!layout.html" %}
+
+{% set title = "Welcome To Neural Network Intelligence !!!"%}
+
+{% block document %}
+
```
squirrelsc, line 6: Add a link to Chinese? Though users can select zh in readthedocs menu, but it's not obvious.

docs/en_US/_templates/index.html
```diff
@@ -5,6 +5,7 @@
 {% block document %}
 
 <div>
+  <div class="chinese"><a href="https://github.com/microsoft/nni/blob/master/README_zh_CN.md">简体中文</a></div>
```
squirrelsc, line 8: It's great to have this link. Can it link to same place like en in readthedocs? for example, `https://nni.readthedocs.io/zh/latest/`?

leckie-chn: see https://github.com/microsoft/nni/pull/2483/commits/69a1c3212268c899897b349584402af293a1f21a

PR #2484, opened 2020-05-25
Remove duplicate data under /tmp folder, and other small changes.

squirrelsc:
Because there was concurrent issue on downloading pytorch mnist data, so there is trial id in dataset path. But it causes many copies of data on /tmp folder. This fix changes the folder to relative path to avoid duplicate data.

It may bring concurrent issue back on local platform, but not others, but it's mitigated already. First, if data is downloaded, pytorch will verify MD5, and not download it again. Second, pytorch examples run as single instance, and test cases is changed to single instance also.

Small improvements,
1. remove incorrect relative path in mnist-pbt-tuner-pytorch, it may be copied from mnist-pytorch.
1. let classic_nas example the same as above.


test/config/examples/mnist-pytorch.yml
```diff
@@ -1,8 +1,8 @@
 authorName: nni
 experimentName: default_test
 maxExecDuration: 15m
-maxTrialNum: 2
-trialConcurrency: 2
+maxTrialNum: 1
+trialConcurrency: 1
```
QuanluZhang, line 5: can we add download script at the beginning of each IT?

PR #2488, opened 2020-05-25
Fix storage logic

SparkSnail:
1. Revert local storage logic, all trials share a common folder in codeDir
2. User a `code` folder to store codeDir for trials in non-local platforms to avoid `run.sh` conflict

### DIFF
--- a/src/nni_manager/training_service/pai/paiK8S/paiK8SData.ts
+++ b/src/nni_manager/training_service/pai/paiK8S/paiK8SData.ts
@@ -31,6 +31,6 @@ fi`;
 
 export const PAI_K8S_TRIAL_COMMAND_FORMAT: string =
 `export NNI_PLATFORM=pai NNI_SYS_DIR={0} NNI_OUTPUT_DIR={1} NNI_TRIAL_JOB_ID={2} NNI_EXP_ID={3} NNI_TRIAL_SEQ_ID={4} MULTI_PHASE={5} \
-&& NNI_CODE_DIR={6} && cp -r $NNI_CODE_DIR/. $NNI_SYS_DIR && cd $NNI_SYS_DIR && sh install_nni.sh \
+&& NNI_CODE_DIR={6} && mkdir -p $NNI_SYS_DIR/code && cp -r $NNI_CODE_DIR/. $NNI_SYS_DIR/code && cd $NNI_SYS_DIR/code && sh install_nni.sh \
 && python3 -m nni_trial_tool.trial_keeper --trial_command '{7}' --nnimanager_ip '{8}' --nnimanager_port '{9}' \
 --nni_manager_version '{10}' --log_collection '{11}'`;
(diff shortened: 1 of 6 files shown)


src/nni_manager/training_service/pai/paiK8S/paiK8SData.ts
```diff
@@ -31,6 +31,6 @@ fi`;
 
 export const PAI_K8S_TRIAL_COMMAND_FORMAT: string =
 `export NNI_PLATFORM=pai NNI_SYS_DIR={0} NNI_OUTPUT_DIR={1} NNI_TRIAL_JOB_ID={2} NNI_EXP_ID={3} NNI_TRIAL_SEQ_ID={4} MULTI_PHASE={5} \
-&& NNI_CODE_DIR={6} && cp -r $NNI_CODE_DIR/. $NNI_SYS_DIR && cd $NNI_SYS_DIR && sh install_nni.sh \
-&& python3 -m nni_trial_tool.trial_keeper --trial_command '{7}' --nnimanager_ip '{8}' --nnimanager_port '{9}' \
+&& NNI_CODE_DIR={6} && mkdir -p $NNI_SYS_DIR/code && cp -r $NNI_CODE_DIR/. $NNI_SYS_DIR/code && sh $NNI_SYS_DIR/install_nni.sh \
```
squirrelsc, line 34: why need extra code folder?

src/nni_manager/training_service/remote_machine/extends/linuxCommands.ts
```diff
@@ -22,10 +22,10 @@ class LinuxCommands extends OsCommands {
             export NNI_PLATFORM=remote NNI_SYS_DIR=${workingDirectory} NNI_OUTPUT_DIR=${workingDirectory} NNI_TRIAL_JOB_ID=${trialJobId} \
             NNI_EXP_ID=${experimentId} NNI_TRIAL_SEQ_ID=${trialSequenceId} NNI_CODE_DIR=${codeDir}
             export MULTI_PHASE=${isMultiPhase}
-
-            cp -r $NNI_CODE_DIR/. $NNI_SYS_DIR
-            cd $NNI_SYS_DIR
-            sh install_nni.sh
+            mkdir -p $NNI_SYS_DIR/code
```
squirrelsc, line 25: It will break trial status check with extra `code` folder. As nnictl need check `.nni` folder, and it may break.
If it's important to keep code folder clean, the folders need to be changed throughtly.

src/nni_manager/training_service/pai/paiK8S/paiK8SData.ts
squirrelsc: If to avoid conflict, we can rename `run.sh` to `nni_run.sh` to avoid it.

SparkSnail: As discussed with @QuanluZhang , we use a `code` folder to store users' code files. In this way, NNI will not destroy users' folder structure.

squirrelsc: How about parameter file? It needs to manual run and verify it at least on remote, and pai.

src/nni_manager/training_service/remote_machine/extends/linuxCommands.ts
SparkSnail: The .nni folder is stroed in `NNI_SYS_DIR`, users' code is tored in `$NNI_SYS_DIR/code`, they are two paralled path, NNI SDK will write metics to `$NNI_SYS_DIR/.nni` folder, the path of code files does not effect this logic.

src/nni_manager/training_service/pai/paiK8S/paiK8SData.ts
squirrelsc: never mind, I had a check. It looks good, if test cases doesn't timeout.

PR #2521, opened 2020-06-02
Reuse OpenPAI jobs to run multiple trials

squirrelsc:
Designed new interface to support reusable training service, currently only applies to OpenPAI, and default disabled.

1. Replace trial_keeper.py to trial_runner.py, trial_runner holds an environment, and receives commands from nni manager to run or stop an trial, and return events to nni manager.
1. Add trial dispatcher, which inherits from original trianing service interface. It uses to share as many as possible code of all training service, and isolate with training services.
1. Add EnvironmentService interface to manage environment, including start/stop an environment, refresh status of environments.
1. Add command channel on both nni manager and trial runner parts, it supports different ways to pass messages between them. Current supported channels are file, web sockets. and supported commands from nni manager are start, kill trial, send new parameters; from runner are initialized(support some channel doesn't know which runner connected), trial end, stdout ((new type), including metric like before), version check (new type), gpu info (new type).
1. Add storage service to wrapper a storage to standard file operations, like NFS, azure storage and so on.
1. Partial support run multiple trials in parallel on runner side, but not supported by trial dispatcher side.

Other minor changes,
1. Add log_level to TS UT, so that UT can show debug level log.
1. Expose platform to start info.
1. Add RouterTrainingService to keep origianl OpenPAI training service, and support dynamic IOC binding.
1. Add more GPU info for future usage, including GPU mem total/free/used, gpu type.
1. Make some license information consistence.
1. Fix async/await problems on Array.forEach, this method doesn't support async actually.
1. Fix IT errors on download data, which causes by my #2484 .
1. Accelerate some run loop pattern by reducing sleep seconds.


src/nni_manager/training_service/reusable/channels/fileCommandChannel.ts
```diff
@@ -0,0 +1,152 @@
...
+ * MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and
+ * to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+ * The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING
+ * BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
```
chicm-ms, line 14: This license section is not consistent with other source files.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -411,20 +411,22 @@ class RemoteMachineTrainingService implements TrainingService {
...
             this.log.debug(`initializing ${executor.name}`);
-            await this.initRemoteMachineOnConnected(rmMeta, executor);
+            connectionPromises.push(this.initRemoteMachineOnConnected(rmMeta, executor));
             this.log.info(`connected to ${executor.name}`);
             if (++connectedRMNum === rmMetaList.length) {
                 deferred.resolve();
             }
-        });
+        }
+        Promise.all(connectionPromises);
```
chicm-ms, line 429: This new logic seems confusing when it works together with `++connectedRMNum` and `deferred.resolve()`.
Missing `await` before `Promise.all`

squirrelsc: fixed

src/nni_manager/training_service/reusable/channels/fileCommandChannel.ts
squirrelsc: I copied from other files. Which one is correct reference?

tools/nni_cmd/config_schema.py
```diff
@@ -463,7 +465,7 @@ def validate_pai_config_path(self, experiment_config):
                 if not taskRoles_dict:
                     raise SchemaError('Please set taskRoles in paiConfigPath config file!')
             else:
-                pai_trial_fields_required_list = ['image', 'gpuNum', 'cpuNum', 'memoryMB', 'paiStoragePlugin', 'command']
+                pai_trial_fields_required_list = ['image', 'gpuNum', 'cpuNum', 'memoryMB', 'command']
```
SparkSnail, line 468: paiStoragePlugin -> paiStorageConfigName

tools/nni_trial_tool/trial.py
```diff
@@ -0,0 +1,148 @@
...
+            self.name = "%s_%s" % (self.id, self.node_id)
+
+    def run(self):
+        # redirect trial's stdout and stderr to syslog
+        self.trial_syslogger_stdout = RemoteLogger(self.args.nnimanager_ip, self.args.nnimanager_port, 'trial', StdOutputType.Stdout,
+                                                   self.args.log_collection, self.id)
+
+        nni_log(LogType.Info, "%s: start to run trial" % self.name)
+
+        trial_working_dir = os.path.realpath(os.path.join(os.curdir, "..", "..", "trials", self.id))
```
SparkSnail, line 49: a little confused about the folder structure. trial_work_dir is in top level?

src/nni_manager/training_service/reusable/channels/fileCommandChannel.ts
chicm-ms: Reference nni repo, such as this one: https://github.com/microsoft/nni/blob/master/src/nni_manager/core/commands.ts

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -406,27 +406,23 @@ class RemoteMachineTrainingService implements TrainingService {
...
+            connectionPromises.push(this.initRemoteMachineOnConnected(rmMeta, executor));
             this.log.info(`connected to ${executor.name}`);
-            if (++connectedRMNum === rmMetaList.length) {
-                deferred.resolve();
-            }
-        });
+        }
 
-        return deferred.promise;
+        Promise.all(connectionPromises);
```
chicm-ms, line 425: missing `await` before Promise.all

src/nni_manager/training_service/reusable/channels/fileCommandChannel.ts
squirrelsc: fixed all incosistent ts files, but some python files are use this format.

tools/nni_cmd/config_schema.py
squirrelsc: fixed

tools/nni_trial_tool/trial.py
squirrelsc: trial_working_dir should be root dir of a trial, it has code, .nni folder inside.

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
squirrelsc: fixed

tools/nni_trial_tool/trial_runner.py
```diff
@@ -0,0 +1,264 @@
...
+    except SystemExit as se:
+        nni_log(LogType.Info, '{}: NNI trial runner exit with code {}'.format(args.node_id, se.code))
+        os._exit(se.code)
+    finally:
+        if trial_runner_syslogger is not None:
+            if trial_runner_syslogger.pipeReader is not None:
+                trial_runner_syslogger.pipeReader.set_process_exit()
+            trial_runner_syslogger.close()
+
+    # the process doesn't exit even main loop exit. So exit it explictly.
```
chicm-ms, line 263: why the process does not exit?

src/nni_manager/training_service/reusable/trialDispatcher.ts
```diff
@@ -0,0 +1,553 @@
...
+            case "RUNNING":
+            case "WAITING":
+            case "UNKNOWN":
+                {
+                    const environment = trial.environment;
+                    if (environment) {
+                        await this.commandChannel.sendCommand(environment, KILL_TRIAL_JOB, trial.id);
+                        trial.isEarlyStopped = isEarlyStopped;
+                        trial.status = trial.isEarlyStopped === true ?
+                            'EARLY_STOPPED' : 'USER_CANCELED';
```
chicm-ms, line 144: In current logic,  `trial.status` is not set here because the cancel could fail,   can we ensure the cancel succeed here?

tools/nni_trial_tool/trial_runner.py
squirrelsc: in some error cases, subthreads not exit. So it needs to force exit to prevent this kind of situation.

src/nni_manager/training_service/reusable/trialDispatcher.ts
squirrelsc: it cannot guarantee cancel succeed in any case, but it will try best... the runner will kill all child processes, which belows to trial code.

PR #2524, opened 2020-06-03
Bugfix issue2485

zheng-ningxin:
Fix the bug mentioned in issue 2485. 
In addition, to pass the merge check pipeline, I also fix several pylint warnings.

I visualize the resnet18, resnet34, mobilev2, googlenet, squeezenet1_1 by TorchModuleGraph for testing, and it works fine.


src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -8,31 +8,39 @@
 from collections import defaultdict
 import torch
 from torch.utils.tensorboard._pytorch_graph import NodePy, NodePyIO, NodePyOP, GraphPy
-
+from tensorboard.compat.proto.config_pb2 import RunMetadata
+from tensorboard.compat.proto.graph_pb2 import GraphDef
+from tensorboard.compat.proto.step_stats_pb2 import StepStats, DeviceStepStats
+from tensorboard.compat.proto.versions_pb2 import VersionDef
```
chicm-ms, line 14: tensorboard import only used by nas graph, we do not want to import tensorboard if it is not necessary, suggest to move these back.

chicm-ms:
would you add a test case with shared nn.Module in `test_model_speedup.py` ?

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -412,30 +424,42 @@ def _build_graph(self):
...
-            _logger.debug('node_group: %s', node_group)
-            nodes_py.nodes_op.append(node_group)
+            use_count = 0
+            for node in node_cpps:
+                if not node.kind().startswith('prim::'):
+                    # modules that have same scope name may have different locations in the
+                    # graph. Futhermore, there are also lots of prim:: nodes that in node_cpps,
+                    # so we also need to call the expand_non_prim_node.
+                    unique_name = module_name
+                    if use_count > 0:
```
chicm-ms, line 436: It seems use_count is applied on name except use_count == 0,  can we just always apply use_count in unique_name?

src/sdk/pynni/nni/_graph_utils.py
zheng-ningxin: I do agree with that, however, import modules in the function will trigger the pylint warnings which may fail the azure merge check pipeline, and block the merge progress. Is there any way to bypass the pylint check?
`_graph_utils.py:78:8: C0415: Import outside toplevel (tensorboard.compat.proto.config_pb2) (import-outside-toplevel)
_graph_utils.py:79:8: C0415: Import outside toplevel (tensorboard.compat.proto.graph_pb2) (import-outside-toplevel)
_graph_utils.py:80:8: C0415: Import outside toplevel (tensorboard.compat.proto.step_stats_pb2) (import-outside-toplevel)
_graph_utils.py:81:8: C0415: Import outside toplevel (tensorboard.compat.proto.versions_pb2) (import-outside-toplevel)`

src/sdk/pynni/nni/_graph_utils.py
zheng-ningxin: I suggest adding the suffix only when usecount>1, because most models will not be reused. Therefore, in most cases, the model will retain his original scopename. In this case, the name of the layers like conv can still be the same as the name in the name_modules() function which is more convenient. We can schedule a quick meeting to discuss this.

zheng-ningxin:
Fix the bug by adding a function called _expand_module_node without touching compressor.py. However, it does give us a reminder that you should be careful when using the name obtained by find_successor to get the corresponding module from the model.

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -444,47 +551,51 @@ def _build_graph(self):
...
         -------
         list
             a list of nodes who are the given node's successor
         """
         successors = []
-        for output in self.name_to_node[module_name].outputs:
-            assert output in self.input_to_node, "No node with input {}".format(output)
+        for output in self.name_to_node[unique_name].outputs:
+            # assert output in self.input_to_node, "No node with input {}(from {})".format(
+            #     output, module_name)
```
chicm-ms, line 594: the commented code can be removed.

zheng-ningxin: Sure~ I have updated.

PR #2535, opened 2020-06-07
Add flops and params counter

colorjam:
Support flops and params calculation with mask.

```python
from nni.counter import count_flops_params
flops, params = count_flops_params(model, (1, 1, 28, 28))
print(flops, params)

### CODE REVIEW

**Comment by ultmaster** [comment]
Please remove accidentally pushed pth files.

File: src/sdk/pynni/requirements.txt [hunk 1]
```diff
@@ -8,3 +8,6 @@ hyperopt==0.1.2
 
 # metis tuner
 scikit-learn==0.20
+
+# flops counter
+thop
```
ultmaster, line 13: I don't think adding a dependency is a good idea. You can try importing `thop` in `counter.py`. If it doesn't exist, post a warning message telling users to install it.

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,91 @@
...
+    device = next(model.parameters()).device
+    inputs = torch.randn(input_size).to(device)
+
+    m_list = list(model.modules())
+    hook_module_list = []
+    for idx, m in enumerate(m_list):
+        weight_mask = None
+        m_type = type(m)
+        if m_type in custom_ops:
+            if isinstance(m_list[idx-1], PrunerModuleWrapper):
```
ultmaster, line 44: What if idx == 0? What if PrunerModuleWrapper has multiple children?

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    assert input_size is not None
+
+    device = next(model.parameters()).device
+    inputs = torch.randn(input_size).to(device)
+
+    m_list = list(model.modules())
+    hook_module_list = []
+    for idx, m in enumerate(m_list):
```
ultmaster, line 40: Try not to store a copy of `model.modules()`. Use iterator instead.

examples/model_compress/run.sh
```diff
@@ -0,0 +1,2 @@
+# python model_prune_torch.py --pruner_name fpgm --pretrain_epochs 1 --prune_epochs 1
+python model_speedup.py --example_name fpgm
```
ultmaster, line 2: I think this change is also accidental.

src/sdk/pynni/nni/counter.py
colorjam: Thanks~ `PrunerModuleWrapper` is our customize wrapper and it directly wraps Conv or Linear module, so it won't have multiple children.

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    assert input_size is not None
+
+    device = next(model.parameters()).device
+    inputs = torch.randn(input_size).to(device)
+
+    hook_module_list = []
+    prev_m = None
+    for idx, m in enumerate(model.modules()):
```
ultmaster, line 48: idx is not used

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,100 @@
...
+        m_type = type(m)
+        if m_type in custom_ops:
+            if isinstance(prev_m, PrunerModuleWrapper):
+                weight_mask = prev_m.weight_mask
+
+            m.register_buffer('weight_mask', weight_mask)
+            hook_module_list.append(m)
+        prev_m = m
+
+    flops, params = profile(model, inputs=(inputs, ), custom_ops=custom_ops, verbose=False)
```
ultmaster, line 59: Expose verbose to user. Don't directly force it to be false.

QuanluZhang:
please add doc here https://nni.readthedocs.io/en/latest/Compressor/CompressionUtils.html, and put api reference here https://nni.readthedocs.io/en/latest/Compressor/CompressionReference.html

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,101 @@
...
+from torch.nn.modules.batchnorm import _BatchNorm
+from nni.compression.torch.compressor import PrunerModuleWrapper
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from thop import profile
+except ImportError:
+    _logger.warning('Please install thop first.')
```
QuanluZhang, line 16: it would be more friendly to provide the installation command.

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,101 @@
...
+    
+def count_flops_params(model: nn.Module, input_size=None, verbose=True):
+    """
+    Count FLOPs and Params of the given model. 
+    This function would identify the mask on the module 
+    and take the pruend shape into consideration.
```
QuanluZhang, line 22: pruend -> pruned

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,101 @@
...
+    from thop import profile
+except ImportError:
+    _logger.warning('Please install thop first.')
+    
+def count_flops_params(model: nn.Module, input_size=None, verbose=True):
+    """
+    Count FLOPs and Params of the given model. 
+    This function would identify the mask on the module 
+    and take the pruend shape into consideration. 
+    Note that, we only identify the output shape of the module,
```
QuanluZhang, line 23: this note is not clear, could you explain a little bit more?

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,101 @@
+# Copyright (c) Microsoft Corporation.
```
QuanluZhang, line 1: better to put this file under https://github.com/microsoft/nni/tree/master/src/sdk/pynni/nni/compression/torch/utils

src/sdk/pynni/nni/counter.py
```diff
@@ -0,0 +1,101 @@
...
+    hook_module_list = []
+    prev_m = None
+    for m in model.modules():
+        weight_mask = None
+        m_type = type(m)
+        if m_type in custom_ops:
+            if isinstance(prev_m, PrunerModuleWrapper):
+                weight_mask = prev_m.weight_mask
+
+            m.register_buffer('weight_mask', weight_mask)
```
QuanluZhang, line 56: no need to register buffer here if the model is not `PrunerModuleWrapper`?

colorjam: We do not directly register buffer on `PrunerModuleWrapper`. We only register buffer on `Conv` or `Linear` (please see custom_ops in Line 96) where the previous module was `PrunerModuleWrapper`.

src/sdk/pynni/nni/counter.py
colorjam: Sure. I add more explanation, please check it.

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -120,4 +120,17 @@ from nni.compression.torch.utils.mask_conflict import MaskConflict
 mc = MaskConflict('./resnet18_mask', net, data)
 mc.fix_mask_conflict()
 mc.export('./resnet18_fixed_mask')
-```
\ No newline at end of file
+```
+
+
+### ModelCounter
```
QuanluZhang, line 126: ModelCounter -> Model Flops/Parameters Calculator

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -120,4 +120,17 @@ from nni.compression.torch.utils.mask_conflict import MaskConflict
...
+We provide a model counter for calculating the model FLOPs and parameters. Users can easily get the model complexity during compression. Note that, for sturctured pruning, we only identify the remained filters according to its mask, which not taking the pruned input channels into consideration, so the calculated FLOPs will be larger than real number.
+
+### Usage
+```
+from nni.compression.torch.utils.counter import count_flops_params
+****
```
QuanluZhang, line 132: replace with blank line

src/sdk/pynni/nni/compression/torch/utils/counter.py
```diff
@@ -0,0 +1,103 @@
...
+        prev_m = m
+
+    flops, params = profile(model, inputs=(inputs, ), custom_ops=custom_ops, verbose=verbose)
+
+    for m in hook_module_list:
+        m._buffers.pop("weight_mask")
+
+    return flops, params
+
+def count_convNd_mask(m, x, y):
```
QuanluZhang, line 69: please add docstring to explain input arguments

src/sdk/pynni/nni/compression/torch/utils/counter.py
```diff
@@ -0,0 +1,103 @@
...
+    output_channel = y.size()[1]
+    output_size =  torch.zeros(y.size()[2:]).numel()
+    kernel_size = torch.zeros(m.weight.size()[2:]).numel()
+
+    bias_flops = 1 if m.bias is not None else 0
+
+    if m.weight_mask is not None:
+        output_channel = m.weight_mask.sum() // (m.in_channels * kernel_size)
+
+    total_ops = output_channel * output_size * (m.in_channels // m.groups * kernel_size + bias_flops)
```
QuanluZhang, line 79: does `total_ops` mean total flops?

src/sdk/pynni/nni/compression/torch/utils/counter.py
```diff
@@ -0,0 +1,103 @@
...
+import logging
+
+_logger = logging.getLogger(__name__)
+
+try:
+    from thop import profile
+except ImportError:
+    _logger.warning('Please install thop using command: pip install thop')
+    
+def count_flops_params(model: nn.Module, input_size=None, verbose=True):
```
suiguoxin, line 18: suggest to remove default value of input_size

src/sdk/pynni/nni/compression/torch/utils/counter.py
colorjam: Yes. Because [thop#L73](https://github.com/Lyken17/pytorch-OpCounter/blob/dbc052ec2d914e950b3e0fa24d3eea753a4e0bb7/thop/profile.py#L73) register the initial state with `total_ops`.

src/sdk/pynni/nni/compression/torch/utils/counter.py
colorjam: fix it. please review it on the latest version~

docs/en_US/Compressor/CompressionUtils.md
colorjam: Which one do you think is better? (`Calculator` or `Counter`). I think we should keep consistent of `counter.py` or `calculater.py`

QuanluZhang: let's use Counter then.

src/sdk/pynni/nni/compression/torch/utils/counter.py
QuanluZhang: got it

docs/en_US/Compressor/CompressionUtils.md
```diff
@@ -120,4 +120,16 @@ from nni.compression.torch.utils.mask_conflict import MaskConflict
 mc = MaskConflict('./resnet18_mask', net, data)
 mc.fix_mask_conflict()
 mc.export('./resnet18_fixed_mask')
-```
\ No newline at end of file
+```
+
+### Model FLOPs/Parameters Counter
+We provide a model counter for calculating the model FLOPs and parameters. Users can easily get the model complexity during compression. Note that, for sturctured pruning, we only identify the remained filters according to its mask, which not taking the pruned input channels into consideration, so the calculated FLOPs will be larger than real number.
```
QuanluZhang, line 126: => We provide a model counter for calculating the model FLOPs and parameters. This counter supports calculating FLOPs/parameters of a normal model without masks, it can also calculates FLOPs/parameters of a model with mask wrappers, which helps users easily check model complexity during model compression on NNI. Note that, for sturctured pruning, we only identify the remained filters according to its mask, which not taking the pruned input channels into consideration, so the calculated FLOPs will be larger than real number (i.e., the number calculated after Model Speedup).

chicm-ms:
Please remove the pth files.

PR #2548, opened 2020-06-11
Exit when installation failed on pipeline

ultmaster:
https://github.com/microsoft/azure-pipelines-tasks/issues/10125

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -116,7 +116,7 @@ jobs:
       python3 -m pip install torchvision==0.6.0 torch==1.5.0 --user
       python3 -m pip install tensorflow==1.15.2 --user
       brew install swig@3
-      if [ -e /usr/local/bin/swig ]; then rm /usr/local/bin/swig; fi
+      rm -f /usr/local/bin/swig
       ln -s /usr/local/opt/swig\@3/bin/swig /usr/local/bin/swig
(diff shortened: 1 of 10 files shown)


liuzhe-lz:
This will not fix the problem I suppose.
Bash script returns the return code of last command.

Following script returns 0:
```bash
[[ 1 == 2 ]]
echo x
```

While following returns 1:
```bash
[[ 1 == 2 ]]
```

azure-pipelines.yml
```diff
@@ -98,18 +103,20 @@ jobs:
...
-      echo "##vso[task.setvariable variable=PATH]${HOME}/Library/Python/3.7/bin:${PATH}"
     displayName: 'Install nni toolkit via source code'
   - script: |
+      set -e
       # pytorch Mac binary does not support CUDA, default is cpu version
       python3 -m pip install torchvision==0.6.0 torch==1.5.0 --user
       python3 -m pip install tensorflow==1.15.2 --user
       brew install swig@3
-      rm /usr/local/bin/swig
+      if [ -e /usr/local/bin/swig ]; then rm /usr/local/bin/swig; fi
```
liuzhe-lz, line 119: Does this equal to `rm -f`?

ultmaster: Yes, thanks.

PR #2552, opened 2020-06-12
Fix bugs in NAS UI (deferred from previous releases)

ultmaster:
1. Add tooltips for buttons
2. Fix issue when an edge is shared by multiple mutables
3. Auto-fit graph at initialization
4. Add unit-tests


src/nasui/src/graphUtils.test.ts
```diff
@@ -0,0 +1,26 @@
+import { Graph } from './graphUtils';
```
chicm-ms, line 1: Suggest to move all unit test stuff into `test` directory like nnimanager.

ultmaster: Here is the reference.

Jest will look for test files with any of the following popular naming conventions:

* Files with .js suffix in `__tests__` folders.
* Files with .test.js suffix.
* Files with .spec.js suffix.
The .test.js / .spec.js files (or the __tests__ folders) can be located at any depth under the src top level folder.

We recommend to put the test files (or __tests__ folders) next to the code they are testing so that relative imports appear shorter. For example, if App.test.js and App.js are in the same folder, the test only needs to import App from './App' instead of a long relative path. Collocation also helps find tests more quickly in larger projects.

ultmaster: I'll move them to `__tests__` folder. I checked nni manager conversion (`test` folder) and jest doesn't recognize them automatically.

chicm-ms: OK.

chicm-ms: btw, nnimanager is using mocha for test, what is the reason that nasui select jest instead of mocha?

ultmaster: I used the default settings provided by react-scripts.

PR #2556, opened 2020-06-15
Pipeline install fastfail

chicm-ms:
Fastfail for install dependencies step.


ultmaster:
We should make step `Install nni toolkit via source code` fast fail too.

PR #2565, opened 2020-06-17
Improve TF NAS performance

liuzhe-lz:


src/sdk/pynni/nni/nas/tensorflow/mutator.py
```diff
@@ -25,6 +27,26 @@ def sample_final(self):
     def reset(self):
         self._cache = self.sample_search()
 
+        # TensorFlow cannot detect unused `Mutable` branches.
+        # We need to fool the graph tracer for reasonable performance.
+        for mutable in self.mutables:
+            if isinstance(mutable, LayerChoice) and mutable._built:
+                mutable._backup_call = mutable.__call__
```
ultmaster, line 34: Where is this `_backup_call` used?

src/sdk/pynni/nni/nas/tensorflow/graph_utils.py
```diff
@@ -0,0 +1,249 @@
...
+
+import tensorflow as tf
+from tensorflow.python.eager import function as function_lib
+from tensorflow.python.framework import func_graph, ops
+from tensorflow.python.ops import array_ops, control_flow_ops, math_ops, resource_variable_ops
+from tensorflow.python.util import object_identity, tf_decorator
+
+from . import lift_to_graph
+
+class Function:
```
ultmaster, line 12: Should we add disclaimer to this code and comments on why this snippet of code should be copied and modified?

PR #2566, opened 2020-06-17
Check eth0 in nnictl

SparkSnail:


tools/nni_cmd/launcher_utils.py
```diff
@@ -297,10 +298,19 @@ def validate_pai_trial_conifg(experiment_config):
             print_warning(warning_information.format('outputDir'))
         validate_pai_config_path(experiment_config)
 
+def validate_eth0_device(experiment_config):
```
QuanluZhang, line 301: do we need to remove the previous validation code?

SparkSnail: no, the previous code is used in nniManager, they are two independent logic. In nniManager's code, it is normal to throw error when can not get os.networkInterfaces().eth0 as an independent function.

PR #2579, opened 2020-06-19
Support the Resnet/Squeezenet/Mobilenet for speedup

zheng-ningxin:
In this pr, the speedup module will support the add/cat operations and the convolution layers that have more than 1 group. I have tested the speedup module on the resnet18, squeezenet1_1, and mobilenetv_2 and it works fine.


zheng-ningxin:
I find another problem in the TorchModuleGraph (https://github.com/microsoft/nni/issues/2581). It may be too late to fix this problem before this release(code freeze at 6.22), but fortunately, there are not many models with this problem. I'll try to fix it in the next release. Thanks~

src/sdk/pynni/tests/test_compression_utils.py
```diff
@@ -11,13 +11,13 @@
 
 from nni.compression.torch import L1FilterPruner
```
chicm-ms, line 12: Would you please add test cases to verify the model speedup correctness for resnet, squeezenet and mobilenet like this test case?
https://github.com/microsoft/nni/blob/master/src/sdk/pynni/tests/test_model_speedup.py#L106

zheng-ningxin: Sure~

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -364,6 +366,50 @@ def _expand_module_node(self, node, node_name, unique_name, op_type, nodes,
...
+        input_tensors = list(list_construct_cpp.inputs())
+        for _tensor in input_tensors:
+            debug_name = _tensor.debugName()
+            input_order.append(self.output_to_node[debug_name].unique_name)
+        cat_info['in_order'] = input_order
+        input_shapes = [t.type().sizes() for t in input_tensors]
+        cat_info['in_shape'] = input_shapes
+        return cat_info
+
     def _extract_shape_info(self, node):
```
QuanluZhang, line 413: this function is written by me, it is only for `view` (pretty limited). maybe we can generalize this function to extract different module's shape if needed.

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -364,6 +366,50 @@ def _expand_module_node(self, node, node_name, unique_name, op_type, nodes,
...
+        Parameters
+        ----------
+        node_group : NodePyGroup
+        cpp_node: torch._C.Node
+            It should be ```aten::cat``` node
+
+        Returns
+        -------
+        dict
+            Include auxiliary information for the cat operation.
```
QuanluZhang, line 384: might be better to explain the content of the dict

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -364,6 +366,50 @@ def _expand_module_node(self, node, node_name, unique_name, op_type, nodes,
...
+        cat_dim = list(inputs)[1].toIValue()
+        cat_info['cat_dim'] = cat_dim
+        # get the order of the input tensors
+        # To get the order of the input tensors, we need
+        # to be aware of the topology of the model, which
+        # means we should extract the auxiliary information
+        # after the build_index function.
+        input_order = []
+        list_construct_cpp = list(cpp_node.inputs())[0].node()
+        input_tensors = list(list_construct_cpp.inputs())
```
QuanluZhang, line 404: so the order of tensors returned by `.inputs()` is the order of input arguments?

zheng-ningxin: According to my observation and experimental results, yes it is.
However, because jit itself lacks documentation, I have no documentation to support this point. 
I will read the source code of jit and double check it.

src/sdk/pynni/nni/compression/torch/speedup/compressor.py
```diff
@@ -53,9 +54,10 @@ def __init__(self, model, dummy_input, masks_file, map_location=None):
         self.bound_model = model
         self.masks = torch.load(masks_file, map_location)
         self.inferred_masks = dict() # key: module_name, value: ModuleMasks
+        self.dummy_input = dummy_input
         self.torch_graph = build_module_graph(model, dummy_input)
 
-    def infer_module_mask(self, module_name, mask=None, in_shape=None, out_shape=None):
+    def infer_module_mask(self, module_name, last_module, mask=None, in_shape=None, out_shape=None):
         """
```
QuanluZhang, line 61: please update docstring accordingly

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -1,16 +1,187 @@
...
-class MaskConflict:
-    def __init__(self, mask_file, model=None, dummy_input=None, graph=None):
+def fix_mask_conflict(masks, model=None, dummy_input=None, traced=None):
+    """
+    MaskConflict fix the mask conflict for the channel dependencies
+    and group dependency.
+
+    Parameters
+    ----------
+    model : torch.nn.Module
```
QuanluZhang, line 18: please use consistent order with input arguments

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -1,16 +1,187 @@
...
+    def fix_mask(self):
+        raise NotImplementedError
+
+    def export(self, path):
+        """
+        Export the masks after fixing the conflict to file.
+        """
+        torch.save(self.masks, path)
+
+class CatMaskPadding(MaskFix):
```
QuanluZhang, line 74: could you explain the logic of cat mask padding in docstring to deliver the high level idea of how the conflict is resolved?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -1,16 +1,187 @@
...
+            for layer in layers:
+                if layer in self.masks:
+                    count += 1
+                    if device is None:
+                        device = self.masks[layer]['weight'].device
+            if count == 0:
+                # no layer is pruned
+                continue
+            elif count == len(layers):
+                # all the layers have been pruned
```
QuanluZhang, line 96: even all the layers have been pruned, is it possible their masks are still not consistent?

QuanluZhang: got it.

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -1,16 +1,187 @@
...
+                continue
+            # pad the mask for the non-pruned layers
+            for layer in layers:
+                module = name_to_module[layer]
+                w_shape = module.weight.data.size()
+                w_mask = torch.ones(w_shape).to(device)
```
QuanluZhang, line 102: so the mask is all ones?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -19,45 +190,40 @@ def __init__(self, mask_file, model=None, dummy_input=None, graph=None):
             model to fix the mask conflict
         dummy_input : torch.Tensor
             input example to trace the model
-        mask_file : str
-            the path of the original mask file
-        graph : torch._C.Graph
+        masks : dict
+            a dict object that stores the masks
+        graph : torch._C.torch.jit.TopLevelTracedModule
```
QuanluZhang, line 195: inconsistent order with input arguments

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
zheng-ningxin: Cat concatenates the input masks as the output mask, so when part of the input layers are not pruned, we still need to pass the masks of these not-pruned layers(all ones) to the cat operation to ensure the  shape  of the final output mask is right.

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
```diff
@@ -111,28 +115,48 @@ def replace_conv2d(conv, mask):
...
+    input_step = int(conv.in_channels / conv.groups)
+    filter_step = int(out_channels / conv.groups)
     if mask.input_mask is not None:
-        tmp_weight_data = torch.index_select(conv.weight.data if tmp_weight_data is None else tmp_weight_data,
-                                             1, in_channels_index)
-    assert tmp_weight_data is not None, "Conv2d weight should be updated based on masks"
-    new_conv.weight.data.copy_(tmp_weight_data)
+        for groupid in range(conv.groups):
+            start = groupid * input_step
+            end = (groupid + 1) * input_step
```
chicm-ms, line 148: If the input channel is pruned, are we assuming that the input channels are pruned evenly for each group?   Do we print some warning if the input channels are not pruned evenly?

zheng-ningxin: Yes, the `GroupMaskConflict` makes sure that each group is pruned evenly before the masks are passed into the `infer_modules_masks`.  I will add several asserts to check if the input channel mask is legal.

PR #2584, opened 2020-06-22
Add pipeline for tfv2

chicm-ms:


azure-pipelines.yml
```diff
@@ -28,8 +28,8 @@ jobs:
   - script: |
       set -e
       python3 -m pip install torch==1.5.0+cpu torchvision==0.6.0+cpu -f https://download.pytorch.org/whl/torch_stable.html --user
-      python3 -m pip install tensorflow==1.15.2 --user
-      python3 -m pip install keras==2.1.6 --user
+      python3 -m pip install tensorflow==2.2.0 --user
```
QuanluZhang, line 31: this part is changed, does this mean tf 1.x is not tested any more?

chicm-ms: tf1.x is still tested in the job `ubuntu_1604_python35_legacy_torch_tf`

QuanluZhang: got it

PR #2601, opened 2020-06-28
Dev graphnas

JunweiSUN:
Add the GraphNAS example to the nni examples (https://github.com/microsoft/nni/tree/master/examples/nas). This example is modified by the [original implementation](https://github.com/GraphNAS/GraphNAS).


scarlett2018:
please add copy right for source codes

Issue #2581, opened 2020-06-19
Should Support the TupleUnpack/ListUnpack in TorchModuleGraph


zheng-ningxin:
**Environment**:
- NNI version: master(latest)
- NNI mode (local|remote|pai): local
- Server OS (for remote mode only): Ubuntu16.04
- Python version: 3.7
- PyTorch/TensorFlow version: 1.4.0
- Is conda/virtualenv/venv used?: yes
- Is running in Docker?: no

**What issue meet, what's expected?**:
Because NNI does not have special treatment for TupleUnpack and ListUnpack operations, TorchModuleGraph will have a graph construction error in some models with TupleUnpack and ListUnpack operations(such as retina face/shufflenet). I want to support these operations in the next release and fix this problem.

**How to reproduce it?**: 
Take shufflenet as an example. The TorchModuleGraph will build the graph in the wrong way(as shown in the following img). 
![image](https://user-images.githubusercontent.com/49771382/85117689-d2fb4880-b251-11ea-8b75-df9543a3874d.png)

In contrast, the right way to build the graph of the shufflenet is shown in the following graph.
![image](https://user-images.githubusercontent.com/49771382/85117727-e3132800-b251-11ea-8851-ab10fdaa91ba.png)

**Additional information**:

zheng-ningxin:
https://github.com/microsoft/nni/pull/2609

PR #2609, opened 2020-06-29
Support the List/Tuple Construct/Unpack operation for TorchModuleGraph

zheng-ningxin:
Support the list/tuple construct/unpack operation for the TorchModuleGraph.
Fix the bug mentioned in https://github.com/microsoft/nni/issues/2581.

In the original version, we take the list/tuple construct/unpack operation nodes as unimportant nodes and merge them with the adjacent important nodes. However, merging the unpack nodes will lead to a graph construct error.


src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -199,6 +203,8 @@ def __init__(self, name, unique_name, node_type, op_type, node_cpps, inputs=None
             All the inputs of this node, each element is debugName of one input
         outputs: list of str
             All the outputs of this node, each element is debugName of one output
+        key_node: torch._C.Node
+            The key node of this NodePyGroup.
```
QuanluZhang, line 207: what is the meaning of key node?

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -510,6 +518,65 @@ def _build_index(self, nodes_op):
...
+        """
+        Judge if a cpp node is a key function node.
+        If so, we should not merge this node into the
+        adjacent node.
+        """
+        if node_cpp.kind().startswith('aten::'):
+            # the nodes that start with 'aten' are key function
+            # nodes
+            return True
+        if node_cpp.kind() in [LIST_UNPACK_KIND, TUPLE_UNPACK_KIND]:
```
QuanluZhang, line 531: not construct type here?

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -510,6 +518,65 @@ def _build_index(self, nodes_op):
...
+            # We cannot merge the List/Tuple
+            # Construct/Unpack func into other nodes, else it
+            # may lead to a graph construction error.
```
QuanluZhang, line 534: what is this error like?

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -510,6 +518,65 @@ def _build_index(self, nodes_op):
...
+        the graph. Note: this function will change the
+        graph structure.
+        """
+        if hasattr(self, 'unpacked'):
+            # if already unpacked the tuple/list manually
+            return
+        for node in self.nodes_py.nodes_op:
+            if node.op_type in [TUPLE_UNPACK_KIND, LIST_UNPACK_KIND]:
+                unpack_cpp = node.key_node
+                last_cpp = list(unpack_cpp.inputs())[0].node()
```
QuanluZhang, line 551: index is 0, why call it `last`?

src/sdk/pynni/nni/_graph_utils.py
```diff
@@ -510,6 +518,65 @@ def _build_index(self, nodes_op):
...
+                if last_cpp.kind() in [TUPLE_CONSTRUCT_KIND, LIST_CONSTRUCT_KIND]:
+                    # we need check if the tensor tuple or tensor list is produced
+                    # by a list/tuple construct node. If so, we can unpack the tuple
+                    # or list manunally.
+                    _logger.debug('List/Tuple Construct Node(cpp) %s', str(last_cpp))
+                    _logger.debug('List/Tuple Unpack Node(cpp) %s', str(unpack_cpp))
+                    assert len(list(unpack_cpp.outputs())) == len(list(last_cpp.inputs()))
```
QuanluZhang, line 558: a little confused about this assert, i think the main reason is i don't understand what is `last_cpp`

src/sdk/pynni/nni/_graph_utils.py
zheng-ningxin: Key nodes are the nodes that should not be merged into other nodes. In the past, we only take the `aten::` nodes as the important(key) nodes. In this pr, we also take the list/tuple unpack nodes as the key nodes.

src/sdk/pynni/nni/_graph_utils.py
zheng-ningxin: Take the shufflenet as an example:
https://github.com/microsoft/nni/issues/2581

src/sdk/pynni/nni/_graph_utils.py
zheng-ningxin: The `last` node actually refers to the previous(last) visited node.  In most scenarios, this `last_cpp` is the corresponding construct node of the tuple/list.

PR #2614, opened 2020-06-30
Fix test case

chicm-ms:
The download command before `nnictl` works only for `local` training service.


QuanluZhang:
btw, is there other options to test `concurrency > 1`?

PR #2627, opened 2020-07-03
Reusable environment support GPU scheduler, add test cases and refactoring.

squirrelsc:
**GPU scheduler experience**

1. in paiConfig, add gpuNum. If it uses to submit template. So that, if gpuNum is 1 in trial, and 4 in pai config. One job can run 4 trials in parallel.
1. in paiConfig add maxTrialNumPerGpu, useActiveGpu. The pattern refers to remote machine, but all openpai job is the same settings.
1. in paiConfig add cpuNum, memoryMB. The same settings are in trial config also. either of them is ok. But it makes sense that they are settings of openpai, should be in pai config.
1. It doesn't support to use gpu scheduler on multi-node environment. Firstly, it's hard to control in this case, Secondly, multi-node is for distrubuted training of one trial, but gpu scheduler means one node serve multiple trials. It doesn't need both in most cases.
1. GPU scheduler is enabled only enableGpuCollectorgpuNum in trial config is specified. The GPU count of environment is collected in runtime.

**GPU scheduler code changes**
1. Copied GPU Scheduler logic from remote machine, as it supports multiple environments not like local.
1. Add TrialGpuSummary to save assignment states with gpu info together.
1. Add runningTrialCount in environment, so that to support non-gpu scheduling well.
1. Add usableGpus in environment, so it can support specify which GPUs are assigned to NNI. It's not exposed to end user, but covered by UT already.
1. Though multi-node doesn't support GPU scheduler, but GPU information of all nodes are collected. defaultGpuSummary uses to get GPU information for GPU scheduler.

**Refactoring**
1. Add isRunnerReady flag to make environment.status only save env status.
1. Add reuseEnvironment in trial config, but not expose to user yet. it can uses to support single trial in future. The assignedTrialCount in environment uses to support this feature.
1. Make channel more stable in runner, when env exit due to timeout.
1. Make environment loop interval configurable for different environment service.
1. Add flag hasMoreEnvironments, so that environment service can stop create more, if there is any error. Remote services can stop create more, if all servers are used.
1. Add flag shouldUpdateTrials in TrialDispatcher, so that the loop can run faster. It saves time not only for UT, also for production.
1. Fix some bugs in storage service, which are found by UT.
1. Fix TrialDispatcher bug, if metric send too fast, some may lost.
1. Update getIPV4Address to let user know how to fix it.
1. Remove useless code in AmlCommandChannel, AMLEnvironmentService.
1. Move some shared code outside remote machine gpu scheduler.
1. Add some flags to prevent too many duplicate logs.

**Test updates**
1. Update test code to support IT.
2. Add UT for trialDispatcher and mounted storage service.


src/nni_manager/training_service/common/gpuData.ts
```diff
@@ -3,6 +3,17 @@
...
 'use strict';
 
+export enum ScheduleResultType {
+    // Schedule succeeded
+    SUCCEED,
+
+    // Temporarily, no enough available GPU right now
+    TMP_NO_AVAILABLE_GPU,
+
+    // Cannot match requirement even if all GPU are a
```
SparkSnail, line 13: a => available?

squirrelsc: Copied from previous code, should be like below,

No environment can match the hard requirement, like the GPU number is smaller than trial asked.

src/nni_manager/training_service/pai/paiConfig.ts
```diff
@@ -20,12 +27,16 @@ export class PAIClusterConfig {
...
+    constructor(userName: string, host: string, passWord?: string, token?: string, reuse?: boolean,
+        cpuNum?: number, memoryMB?: number, gpuNum?: number) {
         this.userName = userName;
         this.passWord = passWord;
         this.host = host;
         this.token = token;
         this.reuse = reuse;
+        this.cpuNum = cpuNum;
+        this.memoryMB = memoryMB;
+        this.gpuNum = gpuNum;
```
SparkSnail, line 39: in other platforms like pai, remote, kubeflow, we added gpuNum, cpuNum setting in trialConfig, not cluster config field. I think we better to unify them.

src/nni_manager/package.json
```diff
@@ -39,6 +39,7 @@
     "@types/express": "^4.16.0",
     "@types/glob": "^7.1.1",
     "@types/js-base64": "^2.3.1",
+    "@types/js-yaml": "^3.12.5",
```
SparkSnail, line 42: already has this `js-yaml` package? refer line 72

src/nni_manager/training_service/pai/paiConfig.ts
squirrelsc: There are three things about gpuNum,

1. Settings are here to tell OpenPAI what kind of environment is needed. It's unique to OpenPAI, not to remote, aml. I'm not sure about kueflow.
1. Environment capacity. It's general to all environments. It doesn't need config like this, since gpu collector (nvdia-smi) dectects environment capacity in runtime. 
1. The requirement of a `single` trial. It's in trialConfig. With above environment capacity, we can schedule multiple trials on one environment. BTW, cpuNum and memoryMB in trial config is not used today, it may be useful in future.

src/nni_manager/package.json
squirrelsc: It uses to provide a TypeScript type declaration, so that we can import js lib like `import * as yaml from 'js-yaml';`, instead of `const yaml = require('js-yaml');`.

src/nni_manager/training_service/reusable/gpuScheduler.ts
```diff
@@ -0,0 +1,235 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
```
chicm-ms, line 2: It seems the main logic of this scheduler is very similar with remote machine gpu scheduler,  suggest to refactor remote machine gpu scheduler and reuse it for `EnvironmentInformation`

src/nni_manager/training_service/reusable/test/utCommandChannel.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
```
chicm-ms, line 1: why the naming convention for this file is different than others (xxx.test.ts)?

src/nni_manager/training_service/reusable/gpuScheduler.ts
squirrelsc: Let's minimum changes each time. The original GpuScheduler has many dependencies on remote machine schema, and new GpuScheduler changed schema of gpu info also.

The long term goal is to migrate remote machine to trialdispatcher. It can improve remote machine performance, and more stable by UT on trialDispatcher level.

tools/nni_cmd/config_schema.py
```diff
@@ -200,21 +206,21 @@ def validate(self, data):
...
-    'paiConfig': Or({
-        'userName': setType('userName', str),
-        'passWord': setType('passWord', str),
-        'host': setType('host', str),
-        Optional('reuse'): setType('reuse', bool)
-    }, {
+    'paiConfig': {
         'userName': setType('userName', str),
-        'token': setType('token', str),
+        Or('passWord', 'token', only_one=True): str,
```
SparkSnail, line 211: what is `only_one` used?

squirrelsc:
@SparkSnail Shinai, please help merge the code, if you think it's ok.

SparkSnail:
> @SparkSnail Shinai, please help merge the code, if you think it's ok.

ok, will merge it after the pipeline pass.

SparkSnail:
> > > @SparkSnail Shinai, please help merge the code, if you think it's ok.
> > 
> > 
> > ok, will merge it after the pipeline pass.
> 
> How about aml? does it support nvidia-smi?

image

Can not detect `nvidia-smi` in my test case, not sure if it is related to aml cluster, I'll double confirming ,due to lack of resources, need to queue long long time for one experiment, waiting.

SparkSnail:
> > > > @SparkSnail Shinai, please help merge the code, if you think it's ok.
> > > 
> > > 
> > > ok, will merge it after the pipeline pass.
> > 
> > 
> > How about aml? does it support nvidia-smi?
> 
> image
> 
> Can not detect `nvidia-smi` in my test case, not sure if it is related to aml cluster, I'll double confirming ,due to lack of resources, need to queue long long time for one experiment, waiting. Merge this pr first, aml will have another pr after the confirmation.

Double confirmed, aml can get `nvidia-smi` command in other cluster.
image

squirrelsc:
> Double confirmed, aml can get `nvidia-smi` command in other cluster.

Can you try if current trial.py can get it also?

PR #2669, opened 2020-07-09
azure-pipelines.yml: Upgrade pylint and astroid

cclauss:
* https://github.com/PyCQA/pylint/blob/master/ChangeLog
* https://github.com/PyCQA/astroid/blob/master/ChangeLog


chicm-ms:
astroid 2.5.0 has net yet been released.

chicm-ms (approved):

PR #2689, opened 2020-07-15
fix chmod bug

JunweiSUN:
The original allowPermission can not handle strings with wildcard, remove single quotes from the original function.

### DIFF
--- a/src/nni_manager/training_service/remote_machine/extends/linuxCommands.ts
+++ b/src/nni_manager/training_service/remote_machine/extends/linuxCommands.ts
@@ -48,13 +48,13 @@ class LinuxCommands extends OsCommands {
     }
 
     public allowPermission(isRecursive: boolean = false, ...folders: string[]): string {
-        const folderString = folders.join("' '");
+        const folderString = folders.join(" ");
         let command;
 
         if (isRecursive) {
-            command = `chmod 777 -R '${folderString}'`;
+            command = `chmod 777 -R ${folderString}`;
         } else {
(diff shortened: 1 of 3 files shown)


src/nni_manager/training_service/remote_machine/extends/linuxCommands.ts
```diff
@@ -48,13 +48,13 @@ class LinuxCommands extends OsCommands {
     }
 
     public allowPermission(isRecursive: boolean = false, ...folders: string[]): string {
-        const folderString = folders.join("' '");
+        const folderString = folders.join(" ");
         let command;
 
         if (isRecursive) {
-            command = `chmod 777 -R '${folderString}'`;
+            command = `chmod 777 -R ${folderString}`;
```
SparkSnail, line 55: Why remove the `''`?

JunweiSUN: because shell can not expand wildcard in single quotes. For example, this line can not be executed correctly:
```
chmod 777 '/tmp/nni/*'
```
This pr will change the command to
```
chmod 777 /tmp/nni/*
```

liuzhe-lz: The quote mark is required when the folder name contains place. Please add it back.

liuzhe-lz:
The ideal solution should be setting shared permission on creation. Though we might need to keep `chmod` for at least one release to fix old files on our customer's computers.
But if we are releasing a hotfix version in recent days I accept this patch.

PR #2690, opened 2020-07-15
update gradient_selector dataloader iterator import

chicm-ms:
This PR upgrade gradient_selector torch version from v1.1 to v1.5.

Differences between torch v1.1 and v1.5 related to this PR:
1. `_DataLoaderIter` in v1.1 is replaced by `_SingleProcessDataLoaderIter` and `_MultiProcessingDataLoaderIter` in v1.5
2. In v1.5,  `torch.autograd._make_grads` requires shapes of outputs and grads must be consistent,  there is no this checking in v1.1: 
https://github.com/pytorch/pytorch/blob/v1.5.0/torch/autograd/__init__.py#L25
3. `timeout` needs to be `0` for _SingleProcessDataLoaderIter


JunweiSUN:
Maybe we need to tell users in the docs that they need to use torch >= 1.5 if they use this selector?

PR #2692, opened 2020-07-16
Hotfix for v1.7.1

SparkSnail:


src/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
```diff
@@ -122,7 +122,7 @@ export class OpenPaiEnvironmentService extends EnvironmentService {
                                     break;
                                 case 'SUCCEEDED':
                                 case 'FAILED':
```
squirrelsc, line 124: add a break above Failed, so that succeeded env won't fail experiment.

src/nni_manager/training_service/reusable/environments/amlEnvironmentService.ts
```diff
@@ -83,8 +83,8 @@ export class AMLEnvironmentService extends EnvironmentService {
     public async refreshEnvironmentsStatus(environments: EnvironmentInformation[]): Promise<void> {
         environments.forEach(async (environment) => {
             const amlClient = (environment as AMLEnvironmentInformation).amlClient;
-                    if (!amlClient) {
-            throw new Error('AML client not initialized!');
+            if (!amlClient) {
+                return Promise.reject('AML client not initialized!');
```
squirrelsc, line 87: BTW, why throw error doesn't work?

src/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
SparkSnail: updated.

src/nni_manager/training_service/reusable/environments/amlEnvironmentService.ts
SparkSnail: There is no try catch outside the function, throw this error directly will not be shown in web ui.

squirrelsc: Did you try it? there are a lot of throw error.

SparkSnail: yes, I tried it. If we only use throw Error without catch, the error information will only shown in `nnictl log stderr` instead of webui.

src/nni_manager/training_service/reusable/environments/amlEnvironmentService.ts
```diff
@@ -98,8 +98,7 @@ export class AMLEnvironmentService extends EnvironmentService {
                     environment.setFinalStatus('SUCCEEDED');
                     break;
                 case 'FAILED':
-                    environment.setFinalStatus('FAILED');
-                    break;
+                    return Promise.reject(`AML: job ${environment.jobId} is failed!`);
```
QuanluZhang, line 101: besides returning reject, do we need to set final status here?

SparkSnail: don't need to set status for environment here. when return Promise.reject, the experiment become `error` status, and nniManager stops. this final status is not useful anymore.

QuanluZhang: got it

src/nni_manager/training_service/reusable/environments/amlEnvironmentService.ts
squirrelsc: well, the status is not visible so far. But if we expose it to webui in future, it would be a hole.

src/nni_manager/training_service/reusable/environments/amlEnvironmentService.ts
squirrelsc: well, the status is not visible so far. But if we expose it to webui in future, it would be a hole.

QuanluZhang: agree

SparkSnail: updated.

src/nni_manager/training_service/reusable/environments/amlEnvironmentService.ts
chicm-ms: The `throw new Error` here should work without `try catch` because there is a `catch`  in nnimanager and this promise is called from nnimanager.  It works in other training service. Would you please check why it is not working?

PR #2694, opened 2020-07-16
add sub-directory for testing pai resume

linbinskn:
image
There is no need to change resume test. To test the sub-directory, we can add a new file in the sklearn-regression.yml' s codedir. I create a sub-directory in the codedir and put a config file in it.


test/config/integration_tests.yml
```diff
@@ -157,6 +157,8 @@ testCases:
 - name: nnictl-resume-1
   configFile: test/config/examples/sklearn-regression.yml
   setExperimentIdtoVar: $resumeExpId
+  # for subfolder in codedir test
+  launchCommand: mkdir ../examples/trials/sklearn/regression/subfolder && nnictl create --config $configFile --debug
```
chicm-ms, line 161: would you also create a file in the subfolder?

PR #2695, opened 2020-07-17
Fix evolution tuner RuntimeError and IndexError

colorjam:
The current version of evolution tuner has several problems:
1. `trialConcurrency` should be larger than `population_size`, otherwise, the experiment will `raise RuntimeError('The population is empty')`. 
   > Add callback function

2. Do not deal with trial failure. If the number of failed trails larger than `population_size`, the experiment will `raise RuntimeError('The population is empty')`.
   > Resume the failed trial.

3. The end time of each trial may not be the same, so, the population length maybe 1. It can not make a comparison between two randomly selected individuals.


src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
```diff
@@ -140,7 +175,12 @@ def generate_parameters(self, parameter_id, **kwargs):
             A group of candaidte parameters that evolution tuner generated.
         """
         if not self.population:
-            raise RuntimeError('The population is empty')
+            is_rand = dict()
+            for item in self.space:
+                is_rand[item] = True
+
+            config = json2parameter(self.searchspace_json, is_rand, self.random_state)
```
tabVersion, line 182: Is it possible to generate duplicate parameters?

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
```diff
@@ -115,20 +101,69 @@ def update_search_space(self, search_space):
...
+        self.num_running_trials -= 1
+        logger.info('trial (%d) end')
+
+        if not success:
+            indv = self.running_trials[parameter_id]
+            self.population.append(indv)
+        if self.credit > 1:
+            param_id = self.param_ids.popleft()
+            config = self._generate_individual(param_id)
+            logger.debug('Send new trial (%d, %s) for reducing credit', param_id, config)
```
chicm-ms, line 126: should this block be  moved to `receive_trial_result` ?

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
```diff
@@ -115,20 +101,69 @@ def update_search_space(self, search_space):
...
-            is_rand[item] = True
-
-        for _ in range(self.population_size):
-            config = json2parameter(
-                self.searchspace_json, is_rand, self.random_state)
-            self.population.append(Individual(config=config))
 
+    def trial_end(self, parameter_id, success, **kwargs):
+        """
+        To deal with trial failure. If a trial fails, readd the trial parameters into population.
```
QuanluZhang, line 107: "readd"?

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
```diff
@@ -115,20 +101,69 @@ def update_search_space(self, search_space):
...
+            True if the trial successfully completed; False if failed or terminated.
+        **kwargs
+            Not used
+        """
+        self.num_running_trials -= 1
+        logger.info('trial (%d) end')
+
+        if not success:
+            indv = self.running_trials[parameter_id]
+            self.population.append(indv)
```
QuanluZhang, line 122: this config is failed, why add it back to `self.population`?

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
```diff
@@ -168,17 +207,41 @@ def generate_parameters(self, parameter_id, **kwargs):
                 is_rand[self.space[i]] = (self.space[i] == mutation_pos)
             config = json2parameter(
                 self.searchspace_json, is_rand, self.random_state, self.population[0].config)
-            self.population.pop(1)
-            # remove "_index" from config and save params-id
 
-            total_config = config
+            if not self.population:
+                self.population.pop(1)
```
QuanluZhang, line 212: what if `self.population` only has one element?

QuanluZhang:
@colorjam UT is failed might because `trail_end` is not handled in UT. @ultmaster FYI

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
colorjam: Yes, the first generation is random generated.

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
colorjam: Based on the discussion of the meeting, a failed trial may due to`out of memory`, so add the config back to the population.

QuanluZhang: if the reason is out of memory, you run it again it will also fail due to out of memory

colorjam: If not deal with failed trials, the population would become smaller. Do you think generating a new random individual would be better?

QuanluZhang: yes, randomly generating one is a possible option.

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
colorjam: Thanks~ I have fixed it, please review on the latest version.

QuanluZhang:
@colorjam please double check whether the generated configs are the same as the previous implementation when there is no failed trial.

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
colorjam: I modify the docstring for clarity, please review on the latest version, thanks~

colorjam:
@QuanluZhang Thanks, I have checked it. I use the previous implementation to generate individuals.

src/sdk/pynni/nni/evolution_tuner/evolution_tuner.py
colorjam: Thanks~ fix it. Please check the latest version.

chicm-ms (approved):

QuanluZhang (approved):

PR #2696, opened 2020-07-17
Show 'NaN' and 'Infinity' metrics as 0

Lijiaoa:
image


liuzhe-lz:
We should not "fix" data during loading. It should happen when plotting.

And I suggest to use a different color for NaN/Infinity points. It may look like this example: [https://echarts.apache.org/examples/en/editor.html?c=bubble-gradient]
Changing points to crosses can be even better, but I don't know if echarts can do that.

src/webui/src/static/function.ts
```diff
@@ -51,8 +50,9 @@ const convertDuration = (num: number): string => {
 };
 
 function parseMetrics(metricData: string): any {
-    if (metricData.includes('NaN')) {
-        return JSON5.parse(JSON5.parse(metricData));
+    if (metricData.includes('NaN') || metricData.includes('Infinity')) {
```
QuanluZhang, line 53: this `includes` method does string match?

Lijiaoa: 是的

PR #2718, opened 2020-07-23
support display trial log on local mode

JunweiSUN:


Lijiaoa:
Approve for webui changes.

src/nni_manager/core/nnimanager.ts
```diff
@@ -325,6 +325,10 @@ class NNIManager implements Manager {
         // FIXME: unit test
     }
 
+    public async getTrialLog(trialJobId: string, logType: LogType): Promise<string> {
+        return this.trainingService.getTrialLog(trialJobId, logType);
```
QuanluZhang, line 329: it might be better to catch `MethodNotImplementedError` here, then return log message like `XXX training service does not support retrieving log`.

src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -184,6 +184,18 @@ class LocalTrainingService implements TrainingService {
         return trialJob;
     }
 
+    public async getTrialLog(trialJobId: string, logType: LogType): Promise<string> {
```
QuanluZhang, line 187: better to add unittest for this function

src/webui/src/components/public-child/OpenRow.tsx
```diff
@@ -105,7 +111,23 @@ class OpenRow extends React.Component<OpenRowProps, OpenRowState> {
...
+                                                    text="View trial log"
+                                                />
+                                                <PrimaryButton
+                                                    onClick={this.openTrialLog.bind(this, 'TRIAL_STDERR')}
+                                                    text="View stderr log"
+                                                    styles={{ root: { marginLeft: 15 } }}
+                                                />
+                                            </div>
+                                        </div>
+                                    </div>
```
chicm-ms, line 130: It seems the `MethodNotImplementedError` not handled by UI,  consider following options:
1.  the open log buttons only enabled for `local`
2.  show a proper error message when MethodNotImplementedError is encountered.

src/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -184,6 +184,18 @@ class LocalTrainingService implements TrainingService {
...
+        let logPath: string;
+        if (logType === 'TRIAL_LOG') {
+            logPath = path.join(this.rootDir, 'trials', trialJobId, 'trial.log');
+        } else if (logType === 'TRIAL_STDERR') {
+            logPath = path.join(this.rootDir, 'trials', trialJobId, 'stderr');
+        } else {
+            throw new Error('unexpected log type');
+        }
+        return fs.promises.readFile(logPath, 'utf8');
```
SparkSnail, line 196: does this readFile() function need to catch exception?

src/nni_manager/common/trainingService.ts
```diff
@@ -8,6 +8,8 @@
  */
 type TrialJobStatus = 'UNKNOWN' | 'WAITING' | 'RUNNING' | 'SUCCEEDED' | 'FAILED' | 'USER_CANCELED' | 'SYS_CANCELED' | 'EARLY_STOPPED';
 
+type LogType = 'TRIAL_LOG' | 'TRIAL_STDERR';
```
SparkSnail, line 11: Does TRIAL_LOG means stdout log?

src/nni_manager/core/nnimanager.ts
JunweiSUN: Currently only local mode will display buttons, I think maybe it'll be better to throw this error to remind developer when developing this feature on other training services.

src/nni_manager/training_service/local/localTrainingService.ts
JunweiSUN: Sure. On the way learning how to add nni_manager's ut...

src/nni_manager/training_service/local/localTrainingService.ts
JunweiSUN: This exception should be catched by `getTrialLog` in `restHandler.ts`

src/nni_manager/common/trainingService.ts
JunweiSUN: Yes, corresponding to `trial.log`.

src/nni_manager/rest_server/restHandler.ts
```diff
@@ -268,6 +269,19 @@ class NNIRestHandler {
         });
     }
 
+    private getTrialLog(router: Router): void {
+        router.get('/trial-log/:id/:type', async(req: Request, res: Response) => {
+            this.nniManager.getTrialLog(req.params.id, req.params.type).then((log: string) => {
+                if (log === '') {
```
liuzhe-lz, line 275: I don't suggest sending unstructured infomation through REST API.
Either leave it empty and let web UI to check it, or encode it as a JSON object.

The idea behind is that REST API is designed for sending "raw data" which can be easily understood by program, rather than "UI string".

src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -172,6 +172,15 @@ class RemoteMachineTrainingService implements TrainingService {
         }
     }
 
+    /**
+     * Get trial job log
+     * @param _trialJobId ID of trial job
+     * @param _logType 'TRIAL_LOG' | 'TRIAL_STDERR'
+     */
+    public async getTrialLog(_trialJobId: string, _logType: LogType): Promise<string> {
```
liuzhe-lz, line 180: The doc string should be written in base class.

JunweiSUN: Agreed. I write doc string in this class because I found other functions like `getTrialJob` also have a doc string in this file. Should I remove them all?

src/nni_manager/rest_server/restHandler.ts
JunweiSUN: I think here `"No logs available"` is also a "raw data", since this logic only check if the return value of `getTrialLog` is an empty string and if so, change it to `No logs available`.

src/nni_manager/core/nnimanager.ts
QuanluZhang: could you explain how did you fix this one?

JunweiSUN: In the latest commit, I revert this to the original version (i.e., not handle the error here) since the front end will check the training service now. The error here could also be an file operation error, so I let the restHander to catch the error throwed here.

src/webui/src/components/public-child/OpenRow.tsx
```diff
@@ -105,7 +111,23 @@ class OpenRow extends React.Component<OpenRowProps, OpenRowState> {
...
                                     />
                                     :
-                                    <TrialLog logStr={logPathRow} id={trialId} />
+                                    <div>
+                                        <TrialLog logStr={logPathRow} id={trialId} />
+                                        {/* view each trial log in drawer*/}
+                                        <div id="trialog">
+                                            <div className="copy" style={{ marginTop: 15 }}>
+                                                <PrimaryButton
+                                                    onClick={this.openTrialLog.bind(this, 'TRIAL_LOG')}
```
QuanluZhang, line 120: do you hide this button if the training service is not local?

JunweiSUN: Yes, in https://github.com/microsoft/nni/blob/73989e06459120f9e954eb3edf55898ec4ecd2ae/src/webui/src/components/public-child/OpenRow.tsx#L106

src/nni_manager/common/trainingService.ts
ultmaster: I think we discussed and decided that it should be ERROR_MESSAGE

src/webui/src/components/public-child/OpenRow.tsx
QuanluZhang: got it

src/nni_manager/training_service/test/localTrainingService.test.ts
```diff
@@ -72,6 +72,31 @@ describe('Unit Test for LocalTrainingService', () => {
...
+                index: 0
+            }
+        };
+        const jobDetail: TrialJobDetail = await localTrainingService.submitTrialJob(form);
+
+        chai.expect(jobDetail.status).to.be.equals('WAITING');
+        localTrainingService.run()
+        await delay(5000);
+        //chai.expect(jobDetail.status).to.be.equals('RUNNING');
+        chai.expect(await localTrainingService.getTrialLog(jobDetail.id, 'TRIAL_LOG')).to.be.a('string');
```
QuanluZhang, line 94: why not compare with 'This is trial log'?

JunweiSUN: This part need refactor, since the submitted jobs are not actually "running". The `run` function doesn't take effect. After sync with liuzhe, it may be better to add this test in IT. Maybe I should arrange an meeting to discuss issues about this pr.

QuanluZhang: in line 91, you already run the training service, why it does not take effect? @liuzhe-lz @chicm-ms

JunweiSUN: I add a `generate_logfile` function in mockedTrial.py to generate mock logfiles. However, these files seems not generated after `run()`.

QuanluZhang: why?

src/nni_manager/common/trainingService.ts
JunweiSUN: change to `TRIAL_LOG`  and `TRIAL_ERROR`

src/nni_manager/training_service/local/localTrainingService.ts
JunweiSUN: Added.

src/nni_manager/training_service/test/localTrainingService.test.ts
```diff
@@ -72,6 +74,38 @@ describe('Unit Test for LocalTrainingService', () => {
...
+        fs.writeFileSync(path.join(jobDetail.workingDirectory, 'stderr'), 'trial stderr')
+        chai.expect(await localTrainingService.getTrialLog(jobDetail.id, 'TRIAL_LOG')).to.be.equals('trial log');
+        chai.expect(await localTrainingService.getTrialLog(jobDetail.id, 'TRIAL_ERROR')).to.be.equals('trial stderr');
+        fs.unlinkSync(path.join(jobDetail.workingDirectory, 'trial.log'))
+        fs.unlinkSync(path.join(jobDetail.workingDirectory, 'stderr'))
+        fs.rmdirSync(jobDetail.workingDirectory)
+        fs.rmdirSync(path.join(rootDir, 'trials'))
+
+        await localTrainingService.cancelTrialJob(jobDetail.id);
+    }).timeout(20000);
```
QuanluZhang, line 107: this is testing training service, do we need to also test nnimanager or restserver?

chicm-ms: This UT for `getTrialLog` looks OK for me, we only want to test getTrialLog here, we do not care about other parts.

QuanluZhang: ok

src/nni_manager/training_service/test/localTrainingService.test.ts
```diff
@@ -72,6 +74,38 @@ describe('Unit Test for LocalTrainingService', () => {
         chai.expect(jobDetail.status).to.be.equals('USER_CANCELED');
     }).timeout(20000);
 
+    it('Get trial log', async () => {
+        // set meta data
+        const trialConfig: string = `{\"command\":\"python3 mockedTrial.py\", \"codeDir\":\"${localCodeDir}\",\"gpuNum\":0}`
```
chicm-ms, line 79: I am wondering whether this `mockedTrial.py` will cause some conflicts regarding log content.  It may generate log which make the log file content is not as expected.

chicm-ms: How about change the command to `sleep 5` ?

JunweiSUN: fixed. now use the original trialConifig in line 23

PR #2722, opened 2020-07-23
Upgrade pai restful API

SparkSnail:
1. upgrade restserver v1 to v2
2. remove password configuration in PAI mode, since in v2 the password related API is not supported.
3. show error information when get unvalid token or unvalid http host.


chicm-ms:
would you add this change to fix mac pipeline?  https://github.com/microsoft/nni/pull/2715/files

src/nni_manager/rest_server/restValidationSchemas.ts
```diff
@@ -103,7 +103,6 @@ export namespace ValidationSchemas {
             }),
             pai_config: joi.object({ // eslint-disable-line @typescript-eslint/camelcase
                 userName: joi.string().min(1).required(),
-                passWord: joi.string().min(1),
```
QuanluZhang, line 106: please also update doc accordingly

src/nni_manager/training_service/pai/paiJobInfoCollector.ts
```diff
@@ -63,8 +63,11 @@ export class PAIJobInfoCollector {
 
         //TODO : pass in request timeout param?
         request(getJobInfoRequest, (error: Error, response: request.Response, _body: any) => {
-            if ((error !== undefined && error !== null) || response.statusCode >= 500) {
-                this.log.error(`PAI Training service: get job info for trial ${paiTrialJob.id} from PAI Cluster failed!`);
+            // Status code 200 for success
+            if ((error !== undefined && error !== null) || response.statusCode !== 200) {
+                const errorMessage: string = (error !== undefined && error !== null) ? error.message :
+                    `PAI Training service: get job info for trial ${paiTrialJob.id} from PAI Cluster failed!, http code:${response.statusCode}, http body: ${JSON.stringify(_body)}`;
```
QuanluZhang, line 69: why `_body` needs `JSON.stringify`?

src/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
```diff
@@ -248,19 +244,6 @@ export class OpenPaiEnvironmentService extends EnvironmentService {
         return deferred.promise;
     }
 
-    private async refreshPlatform(): Promise<void> {
```
QuanluZhang, line 251: why remove the refresh logic?

src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -162,8 +162,7 @@ abstract class PAITrainingService implements TrainingService {
         }
 
         const stopJobRequest: request.Options = {
-            uri: `${this.protocol}://${this.paiClusterConfig.host}/rest-server/api/v1/user/${this.paiClusterConfig.userName}\
-/jobs/${trialJobDetail.paiJobName}/executionType`,
+            uri: `${this.protocol}://${this.paiClusterConfig.host}/rest-server/api/v2/jobs/${this.paiClusterConfig.userName}~${trialJobDetail.paiJobName}/executionType`,
```
squirrelsc, line 165: remove updatePaiToken function in this file also?

src/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
squirrelsc: It's ok to remove, as it only refresh pai token. And it's a pattern for future platform, which need regular refresh credential.

src/nni_manager/training_service/pai/paiTrainingService.ts
SparkSnail: This function here is for paiYarnTrainingService, which still supports password login.

src/nni_manager/training_service/pai/paiJobInfoCollector.ts
SparkSnail: I've tested the response data of PAI, seems this list API in v2 returns a object, and need to use JSON.stringify to decode the content, while the submit API returns a string directly, does not need to do this.

src/nni_manager/rest_server/restValidationSchemas.ts
SparkSnail: the PAIMode doc has already removed the password configuration, while it keeps in our code. Only need to remove code this time.

SparkSnail:
> would you add this change to fix mac pipeline? https://github.com/microsoft/nni/pull/2715/files

updated.

src/nni_manager/training_service/pai/paiTrainingService.ts
```diff
@@ -178,7 +177,8 @@ abstract class PAITrainingService implements TrainingService {
         const deferred: Deferred<void> = new Deferred<void>();
 
         request(stopJobRequest, (error: Error, response: request.Response, _body: any) => {
-            if ((error !== undefined && error !== null) || response.statusCode >= 400) {
+            // Status code 202 for success.
+            if ((error !== undefined && error !== null) || response.statusCode !== 202) {
```
squirrelsc, line 181: It's better keep original status code. The HTTP status code is standard. < 400 is normal state. 4xx is permission issue, and  5xx are server errros. Don't hard code to 202 only.

SparkSnail: There is a case that when submit job to http request, pai returns a 3xx response, but the job is not submitted correctly.

squirrelsc: 3xx means redirection, and it should be handled in most HTTP sdk internally. If it happens here, you can check if the sdk is not used correctly. For example, disabled auto redirect.

SparkSnail: updated, add `followAllRedirects:true` to handle redirect.

src/nni_manager/training_service/pai/paiJobInfoCollector.ts
```diff
@@ -63,8 +63,9 @@ export class PAIJobInfoCollector {
 
         //TODO : pass in request timeout param?
         request(getJobInfoRequest, (error: Error, response: request.Response, _body: any) => {
-            if ((error !== undefined && error !== null) || response.statusCode >= 500) {
-                this.log.error(`PAI Training service: get job info for trial ${paiTrialJob.id} from PAI Cluster failed!`);
+            // Status code 200 for success
+            if ((error !== undefined && error !== null) || response.statusCode >= 400) {
+                // The job refresh time could be ealier than job submission, so it might return 404 error code, need refactor
```
QuanluZhang, line 68: why there is no log here?

src/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
QuanluZhang: @squirrelsc  you mean we will add dedicated token refresh logic in the future?

squirrelsc: I mean, if there are similar requirements, here is a reference.

QuanluZhang: then, why remove pai token refresh? no need to refresh token any more on pai?

src/nni_manager/training_service/pai/paiJobInfoCollector.ts
SparkSnail: The 404 error is always shown, it is caused by NNI's job refresh logic, not expected to shown. The original logic use `statusCode>=500` to avoid log 404 error, it is not reasonable. We should refactor the logic to avoid refresh job while it is not submitted. If add a error log here, will cause confuse for users currently, will add a log here after fixing the logic.

QuanluZhang: got it, thanks

PR #2725, opened 2020-07-23
Fix remote machine connection logic

SparkSnail:


src/nni_manager/training_service/remote_machine/remoteMachineTrainingService.ts
```diff
@@ -82,6 +84,12 @@ class RemoteMachineTrainingService implements TrainingService {
         restServer.setEnableVersionCheck = this.versionCheck;
         this.log.info('Run remote machine training service.');
         while (!this.stopping) {
+            if (this.sshConnectionPromises.length) {
```
squirrelsc, line 87: To improve readabilty, and avoid tricks of JS, it's bettter to write like `this.sshConnectionPromises.length > 0`

squirrelsc, line 87: It looks should be put outside of while loop, as it's one time work.

SparkSnail: fixed

SparkSnail: fixed

PR #2728, opened 2020-07-24
Add API to query intermediate results in NAS benchmark

### LINKED ISSUE #2721: nasbench201 api bug
Changing the order of arch dictionary causes `query_nb201_trial_stats` returns nothing.

**Environment**:
- NNI version: v1.7
- Client OS: Windows10 WSL
- Python version: 3.7

code:
```python
arch = {"0_1": "conv_3x3", "0_2": "none", "1_2": "avg_pool_3x3", "0_3": "none", "1_3": "none", "2_3": "conv_3x3"}
print(len([t for t in query_nb201_trial_stats(arch, 200, 'cifar100')]))
arch = {"0_1": "conv_3x3", "0_2": "none", "0_3": "none", "1_2": "avg_pool_3x3", "1_3": "none", "2_3": "conv_3x3"}
print(len([t for t in query_nb201_trial_stats(arch, 200, 'cifar100')]))
```
result:
```
3
0
```

ultmaster:
* Fix #2721 
* Add API to return intermediate results

### DIFF
--- a/src/sdk/pynni/nni/nas/benchmarks/nds/model.py
+++ b/src/sdk/pynni/nni/nas/benchmarks/nds/model.py
@@ -4,6 +4,7 @@ from peewee import CharField, FloatField, ForeignKeyField, IntegerField, Model
 from playhouse.sqlite_ext import JSONField, SqliteExtDatabase
 
 from nni.nas.benchmarks.constants import DATABASE_DIR
+from nni.nas.benchmarks.utils import json_dumps
 
 db = SqliteExtDatabase(os.path.join(DATABASE_DIR, 'nds.db'), autoconnect=True)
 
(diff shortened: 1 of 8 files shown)


JunweiSUN:
What is the reason that caused #2721 ? It seems that if user set include_intermediates=False, this issue still exists.

src/sdk/pynni/nni/nas/benchmarks/nds/model.py
```diff
@@ -52,8 +53,8 @@ class NdsTrialConfig(Model):
         'residual_basic',
         'vanilla',
     ])
-    model_spec = JSONField(index=True)
-    cell_spec = JSONField(index=True, null=True)
+    model_spec = JSONField(json_dumps=json_dumps, index=True)
```
QuanluZhang, line 56: why this modification can resolve the dict element sequence issue?

ultmaster: The comparison of JSON field is based on string comparison, and to compare two dumped JSON of unordered dict, they need to be sorted ahead.

PR #2731, opened 2020-07-24
[webui] could view trial log | stderr on webui(only support local mode)

Lijiaoa:
联调中。
This PR is similar with PR https://github.com/microsoft/nni/pull/1738


ultmaster:
Close as it's already done in #2718.

PR #2736, opened 2020-07-24
fix SimulatedAnnealingPruner export mask issue

suiguoxin:
self.modules_wrapper is not properly updated, this cause an error when 'export_model' is called


QuanluZhang:
> self.modules_wrapper is not properly updated, this cause an error when 'export_model' is called

could you describe the error?

suiguoxin:
> > self.modules_wrapper is not properly updated, this cause an error when 'export_model' is called
> 
> could you describe the error?

A all-one mask will be exported because `self.modules_wrapper` is not updated, `self.modules_wrapper` is needed in the function [`export_model()`](https://github.com/microsoft/nni/blob/master/src/sdk/pynni/nni/compression/torch/compressor.py#L378).

`Pruner.compress()` works just fine before this fix because `self.bound_model` is correctly updated.

PR #2744, opened 2020-07-28
[webui] format code by eslint and prettier

Lijiaoa:
- [x] [stylelint](https://stylelint.io/user-guide/rules/about) reference
    - `yarn stylelint --fix` could almost format css/scss code

- [x] [prettier](https://prettier.io/docs/en/index.html)
- [x] add some eslint rule
        
```
 // no use `arrow-parens: [2, "as-needed", { "requireForBlockBody": true }]`,
 // because `"requireForBlockBody": true`  is conflict with prettier `arrowParens: 'avoid',`
 "arrow-parens": [2, "as-needed"],
 "no-empty": 2, // block is null will error
 "no-multiple-empty-lines": [2, { "max": 1 }], // max empty line is 1
```


src/webui/src/components/Modals/ChangeColumnComponent.tsx
```diff
@@ -41,7 +47,7 @@ class ChangeColumnComponent extends React.Component<ChangeColumnProps, ChangeCol
         } else {
             if (source.includes(label)) {
                 // remove from source
-                const result = source.filter((item) => item !== label);
+                const result = source.filter(item => item !== label);
```
liuzhe-lz, line 50: I think this is less clear to human. It's hard to figure out associative order at first glance to me.

Lijiaoa: alwaysParens: `avoid | always`;
default val is avoid, not add () for arrow function.
So we change it to `always`?

src/webui/.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
```
liuzhe-lz, line 1: Is this file used by a specific text editor? If yes it should not be committed to code tree.

src/webui/.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
```
liuzhe-lz, line 8: And I wonder why your files do not have trailing new line even if this is set to `true`.

src/webui/.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+insert_final_newline = true
```
liuzhe-lz, line 9: And I wonder why your files do not have trailing new line even this is set to `true`...

src/webui/src/App.tsx
```diff
@@ -117,39 +123,43 @@ class App extends React.Component<{}, AppState> {
...
+                <Stack className='contentBox'>
+                    <Stack className='content'>
                         {/* if api has error field, show error message */}
-                        {
-                            errorList.map((item, key) => {
-                                return (
-                                    item.errorWhere && <div key={key} className="warning">
-                                        <MessageInfo info={item.errorMessage} typeInfo="error" />
+                        {errorList.map((item, key) => {
+                            return (
```
liuzhe-lz, line 150: Is `return` necessary here?

liuzhe-lz:
I think we should discuss the rules in a meeting, just like what we have done for pylint rules.
For pylint, we listed all default rules violated by our code, and discussed them one by one.

src/webui/.editorconfig
Lijiaoa: but `.tsx .scss` these files all have new line in last line.

src/webui/src/App.tsx
Lijiaoa: I also think so. We can use arrow function rules to standardize it.

ultmaster (approved):

liuzhe-lz (approved):

PR #2752, opened 2020-07-31
Adding sorting function based on metric to get_job_metrics of nnicli

prismformore:
Adding sorting function (both maximize and minimize) to get_job_metrics of nnicli.
This will be helpful for those who can not access webui.

### DIFF
--- a/src/sdk/pycli/nnicli/nni_client.py
+++ b/src/sdk/pycli/nnicli/nni_client.py
@@ -136,8 +136,8 @@ def get_job_metrics(trial_job_id=None, sort=False):
     api_path = METRICS_PATH if trial_job_id is None else os.path.join(METRICS_PATH, trial_job_id)
     job_metrics = _nni_rest_get(api_path)
     
-    if sort == True and trial_job_id != None:
-        return sorted(metrics, key=cmp_to_key(lambda x, y: -1 if float(x['data'].replace('"', '')) < float(y['data'].replace('"', '')) else 1, reverse=True)
+    if sort == True and trial_job_id == None:
+        return sorted(job_metrics, key=cmp_to_key(lambda x, y: -1 if float(x['data'].replace('"', '')) < float(y['data'].replace('"', '')) else 1, reverse=True)
     else:  
         return job_metrics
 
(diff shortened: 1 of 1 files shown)


prismformore:
fail

src/sdk/pycli/nnicli/nni_client.py
```diff
@@ -129,10 +130,20 @@ def get_job_statistics():
...
 
-def get_job_metrics(trial_job_id=None):
+def get_job_metrics(trial_job_id=None, sort=None):
     """return trial job metrics"""
     api_path = METRICS_PATH if trial_job_id is None else os.path.join(METRICS_PATH, trial_job_id)
-    return _nni_rest_get(api_path)
+    job_metrics = _nni_rest_get(api_path)
+    if sort == 'max' and trial_job_id == None:
+        return sorted(job_metrics, key=cmp_to_key(lambda x, y: -1 \
+                                                  if float(x['data'].replace('"', '')) < float(y['data'].replace('"', '')) \
```
tabVersion, line 139: Maybe you can use `eval()` to handle serialized data.

tabVersion:
How your code handles dict metric?
Does your code always sort by the default key?

src/sdk/pycli/nnicli/nni_client.py
```diff
@@ -129,10 +130,20 @@ def get_job_statistics():
     """return trial job statistics information as a dict"""
     return _nni_rest_get(JOB_STATISTICS_PATH)
 
-def get_job_metrics(trial_job_id=None):
+def get_job_metrics(trial_job_id=None, sort=None):
```
chicm-ms, line 133: `nni_client.py` is only a wrapper of rest api, this is not a good place to add additional features to rest api.
Suggest to:
1.  discuss with team whether to add sort feature into the datastore API.
OR
2.  let end user of nnicli to do the sorting.

QuanluZhang:
@prismformore thanks for your contribution. This function is useful. But since it is relatively simple, we believe users can do it by themselves after getting the results from nnictl

QuanluZhang:
@prismformore feel free to reopen if you have more questions about this.

PR #2753, opened 2020-07-31
update experiment dir

JunweiSUN:
update `nni/experiments` to `nni-experiments` in docs, rest server and nnictl commands.

### DIFF
--- a/src/nni_manager/training_service/remote_machine/shellExecutor.ts
+++ b/src/nni_manager/training_service/remote_machine/shellExecutor.ts
@@ -175,7 +175,7 @@ class ShellExecutor {
     }
 
     public getRemoteExperimentRootDir(experimentId: string): string {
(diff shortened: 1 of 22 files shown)


src/nni_manager/training_service/remote_machine/shellExecutor.ts
```diff
@@ -175,7 +175,7 @@ class ShellExecutor {
     }
 
     public getRemoteExperimentRootDir(experimentId: string): string {
-        return this.joinPath(this.tempPath, 'nni', 'experiments', experimentId);
+        return this.joinPath(this.tempPath, 'nni-experiments', experimentId);
```
SparkSnail, line 178: This is not NNI's local home path, need to confirm if there are others folder under /tmp/nni

PR #2754, opened 2020-07-31
Update config.yml

xinshouke:
I run the sample with the below error ,but I used Pthon ,it works!
cmd.exe : Python was not found but can be installed from the Microsoft Store: https://go.microsoft.com/fwlink?linkID=2082640
所在位置 C:\Users\xinsh\nni\experiments\sVPyAII6\trials\Ik1oq\run.ps1:11 字符: 1
+ cmd.exe /c python3 main.py 2>"C:\Users\xinsh\nni\experiments\sVPyAII6 ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (Python was not ...?linkID=2082640:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

### DIFF
--- a/examples/trials/auto-gbdt/config.yml
+++ b/examples/trials/auto-gbdt/config.yml
@@ -16,6 +16,6 @@ tuner:
     #choice: maximize, minimize
     optimize_mode: minimize
 trial:
-  command: python3 main.py
+  command: Python main.py
   codeDir: .
   gpuNum: 0


JunweiSUN:
Hi @xinshouke, the reason why your cmd.exe can not find `python` maybe is that `python` is not in your `Path` env. When you type `python` in cmd or powershell, do you see `>>>` or "cannot found python" ?

xinshouke:
I suspect you are correct!

PR #2396, opened 2020-04-30
Remove multiphase from WebUI

squirrelsc:
As requirement of #2390 , removed multiphase from WebUI, document, and put all trials flatten on WebUI.

1. The major change is to split trial-jobs to trials in webui, and use parameter id as part of trial id.
2. If multiphase is enabled, the limitation of trial controls job count only, not trials actually.
3. When multiphase enabled, the trial status may not be right. Previous trials in a job will be marked as success, only last trial presents the job's status.
4. multiphase related documents and UX are removed.

minor changes,
1. Update dev document for webui development.

Issue #2560, opened 2020-06-16
duplicated Trial No. on webui when an experiment is resumed


QuanluZhang:
**Environment**:
- NNI version: 1.6
- NNI mode (local|remote|pai): local
- Client OS: ubuntu 16.04
- Server OS (for remote mode only):
- Python version: 3.5
- PyTorch/TensorFlow version: --
- Is conda/virtualenv/venv used?: No
- Is running in Docker?: No

 

**What issue meet, what's expected?**: duplicated Trial No. on webui when an experiment is resumed

![image](https://user-images.githubusercontent.com/16907603/84742220-c02f0c80-afe2-11ea-9337-fa469fc4abd7.png)

**How to reproduce it?**: resume an experiment

**Additional information**: --

squirrelsc:
It needs to fix manual created trial get null name too.

Lijiaoa:
This issue had been fixed in v1.8 
PR #https://github.com/microsoft/nni/pull/2760

PR #2760, opened 2020-08-04
delete multiphase in webui

Lijiaoa:
#2396


Lijiaoa:
related issue: https://github.com/microsoft/nni/issues/2560

PR #2767, opened 2020-08-05
deal with all type metrics

Lijiaoa:
still to fix
- [ ] show NaN but had final metric
image

- [ ] hyper-parameter default yAxis
image


Lijiaoa:
close for update PR in #2782

PR #2769, opened 2020-08-06
Enable gpu scheduler in AML mode

SparkSnail:


src/nni_manager/rest_server/restValidationSchemas.ts
```diff
@@ -160,7 +159,10 @@ export namespace ValidationSchemas {
             aml_config: joi.object({ // eslint-disable-line @typescript-eslint/camelcase
                 subscriptionId: joi.string().min(1),
                 resourceGroup: joi.string().min(1),
-                workspaceName: joi.string().min(1)
+                workspaceName: joi.string().min(1),
+                computeTarget: joi.string().min(1),
+                maxTrialNumPerGpu: joi.number(),
```
squirrelsc, line 164: It looks a base config is necessary to put shared parameters.

SparkSnail: This validate format follows the body schema passed from client, these shared parameters is in different fields, like `aml_config`, `pai_config`, unless we add a `base_config` in nnictl side, other wise this validation format can not change.

PR #2676, opened 2020-07-10
Compression doc structure refactor

suiguoxin:
- Reuse class doc string in "Read the Docs" as in NAS
- remove redundant parameter 'optimizer ' from Pruners ('level', 'slim', 'fpgm', 'l1', 'l2') and change related unit test accordingly;
- fix some typos

PR #2772, opened 2020-08-07
fix IT pruning example issue

suiguoxin:
Parameter 'optimizer '  was removed from Pruners ('level', 'slim', 'fpgm', 'l1', 'l2') in PR https://github.com/microsoft/nni/pull/2676.  This example should be changed accordingly.

### DIFF
--- a/examples/model_compress/model_prune_torch.py
+++ b/examples/model_compress/model_prune_torch.py
@@ -145,7 +145,10 @@ def create_model(model_name='naive'):
 def create_pruner(model, pruner_name, optimizer=None):
     pruner_class = prune_config[pruner_name]['pruner_class']
     config_list = prune_config[pruner_name]['config_list']
-    return pruner_class(model, config_list, optimizer)
+    if pruner_name in ['level', 'slim', 'fpgm', 'l1filter']:
+        return pruner_class(model, config_list)
+    else:
+        return pruner_class(model, config_list, optimizer)
 
 def train(model, device, train_loader, optimizer):
     model.train()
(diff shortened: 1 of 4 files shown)


examples/model_compress/model_prune_torch.py
```diff
@@ -145,7 +145,10 @@ def create_model(model_name='naive'):
 def create_pruner(model, pruner_name, optimizer=None):
     pruner_class = prune_config[pruner_name]['pruner_class']
     config_list = prune_config[pruner_name]['config_list']
-    return pruner_class(model, config_list, optimizer)
+    if pruner_name in ['level', 'slim', 'fpgm', 'l1filter']:
```
chicm-ms, line 148: why keep optimizer parameter for `ActivationMeanRankFilterPruner` and `ActivationAPoZRankFilterPruner`? they do not need optimizer hook either.  For all one-shot pruners, it seems only TaylorFOWeightFilterPruner needs optimizer ( to calculate gradient contribution).

suiguoxin: > why keep optimizer parameter for `ActivationMeanRankFilterPruner` and `ActivationAPoZRankFilterPruner`? they do not need optimizer hook either. For all one-shot pruners, it seems only TaylorFOWeightFilterPruner needs optimizer ( to calculate gradient contribution).

In the current implementation, `ActivationMeanRankFilterPruner` and `ActivationAPoZRankFilterPruner` need the optimizer hook.

QuanluZhang: as discussed in the meeting, let's not do this broken change?

PR #2774, opened 2020-08-07
add RESTful api doc

tabVersion:


docs/en_US/reference.rst
```diff
@@ -10,3 +10,4 @@ References
     NNI Annotation <Tutorial/AnnotationSpec>
     SDK API References <sdk_reference>
     Supported Framework Library <SupportedFramework_Library>
+    RESTful API <RestAPI>
```
liuzhe-lz, line 13: Oh I found REST API is not in the same topic of other pages.
SDK API is designed for _users_, while REST API is meant to be used by _developers.
I suggest to put the doc into a wiki page.

PR #2780, opened 2020-08-11
Fix problem: datastore encodes metrics data into json string again while saving

chicm-ms:
Previously datastore encodes metrics data into json string before saving it into database because the field is of string type.
Later the metrics data field is already encoded into json string by SDK, then datastore should remove the encoding code.
This problem causes the metric data look like this:
```
 "finalMetricData":[
            {
                "data":"\"{\\\"default\\\": 0.6, \\\"loss\\\": 0.66, \\\"other\\\": 0.666}\"",
                "parameterId":"0",
                "sequence":0,
                "timestamp":1597147814203,
                "trialJobId":"ABRiy",
                "type":"FINAL"
            }
        ],
```

The correct result should be this:
```
"finalMetricData":[
            {
                "data":"{\"default\": 0.6, \"loss\": 0.66, \"other\": 0.666}",
                "parameterId":"0",
                "sequence":0,
                "timestamp":1597148506323,
                "trialJobId":"epX8Z",
                "type":"FINAL"
            }
        ],
```

TODO:
1. nnictl, nnicli -- corresponding update


SparkSnail:
Tested on current nnictl, don't have affect for nnictl.

src/nni_manager/core/sqlDatabase.ts
```diff
@@ -203,7 +203,7 @@ class SqlDB implements Database {
     public storeMetricData(trialJobId: string, data: string): Promise<void> {
         const sql: string = 'insert into MetricData values (?,?,?,?,?,?)';
         const json: MetricDataRecord = JSON.parse(data);
-        const args: any[] = [Date.now(), json.trialJobId, json.parameterId, json.type, json.sequence, JSON.stringify(json.data)];
+        const args: any[] = [Date.now(), json.trialJobId, json.parameterId, json.type, json.sequence, json.data];
```
QuanluZhang, line 206: maybe we should remove JSON.parse in this line https://github.com/microsoft/nni/blob/bcefce6a814c2c1228e25fcd98c88fea9e8111a9/src/nni_manager/core/nniDataStore.ts#L170?

QuanluZhang: BTW, do we have unit test on data export and resume? maybe we should run IT for this pr

ultmaster:
This PR will break:

* Compatibility with experiments from previous releases
* WebUI mock data needs to be converted

I think it's too risky to merge this PR at this time before code freeze. Suggest merging into master after cutting branch.

src/nni_manager/core/sqlDatabase.ts
chicm-ms: Yes, we need data import/export IT test case.

PR #2781, opened 2020-08-12
TF NAS fix

liuzhe-lz:


src/sdk/pynni/nni/nas/tensorflow/mutables.py
```diff
@@ -94,21 +92,18 @@ def __init__(self, op_candidates, reduction='sum', return_mask=False, key=None):
...
-            self._built = True
         out, mask = self.mutator.on_forward_layer_choice(self, *inputs)
         if self.return_mask:
             return out, mask
         return out
 
+    def build(self, input_shape):
+        self._check_built()
+        for op in self.choices:
+            op.build(input_shape)
```
QuanluZhang, line 105: what is the meaning of `op.build`?

liuzhe-lz: TF layers do not require input shape to `__init__`, so they create weights when `build` is called.
Normally this happens implicitly at first forward pass. We trigger it manually because only one (or some) candidate layer is connected at first forward so other candidate layers cannot inference their input shape automatically.

QuanluZhang: got it, thanks.

PR #2782, opened 2020-08-12
deal with all type metrics

Lijiaoa:


src/webui/src/App.tsx
```diff
@@ -99,14 +102,22 @@ class App extends React.Component<{}, AppState> {
         this.setState({ bestTrialEntries: entries });
     }
 
+    shouldComponentUpdate(nextProps: any, nextState: AppState): boolean {
```
liuzhe-lz, line 105: Don't do this...

Lijiaoa: If don't use shouldComponentUpdate, parent App.tsx component will refresh three times when experiment done, and then child components will also refresh.  And App.tsx's shouldComponentUdate only watch a boolean value.

src/webui/src/App.tsx
```diff
@@ -177,8 +187,7 @@ class App extends React.Component<{}, AppState> {
         // experiment status and /trial-jobs api's status could decide website update
         if (['DONE', 'ERROR', 'STOPPED'].includes(EXPERIMENT.status) || TRIALS.jobListError()) {
             // experiment finished, refresh once more to ensure consistency
-            this.setState({ interval: 0 });
-            this.lastRefresh();
```
liuzhe-lz, line 181: I wrote `lastRefresh` because due to async nature when experiment status become `DONE`, some state may have not updated to their final value yet. For example maybe the last trial is still running at last `trial-jobs` REST call.

src/webui/src/static/function.ts
```diff
@@ -84,15 +84,18 @@ const getFinalResult = (final?: MetricDataRecord[]): number => {
     }
 };
 
+function isNaNorInfinity(val: number): boolean {
+    return Object.is(val, NaN) || Object.is(val, Infinity);
```
liuzhe-lz, line 88: I think `Number.isNaN` and `Number.isFinite` is more clear for this purpose.

src/webui/src/static/model/trial.ts
```diff
@@ -110,7 +111,15 @@ class Trial implements TableObj {
         const endTime = this.info.endTime || new Date().getTime();
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const duration = (endTime - this.info.startTime!) / 1000;
-
+        let accuracy;
+        if(this.acc !== undefined && this.acc.default !== undefined){
+            if(typeof this.acc.default === 'number'){
+                accuracy = JSON5.parse(this.acc.default);
```
liuzhe-lz, line 117: What's this for?

src/webui/src/static/model/trial.ts
```diff
@@ -190,10 +201,10 @@ class Trial implements TableObj {
     }
 
     public parameters(axes: MultipleAxes): Map<SingleAxis, any> {
-        const tempHyper = this.info.hyperParameters;
-        if (tempHyper === undefined) {
-            throw new Map([['error', 'This trial\'s parameters are not available.']]);
+        if (this.info === undefined || this.info.hyperParameters === undefined) {
+            throw new Map();
```
liuzhe-lz, line 205: Why throw a `Map`？

src/webui/src/static/model/trial.ts
Lijiaoa: to fix this state: when user give final result as  {tensor: xxx, data: xxx}

src/webui/src/static/function.ts
Lijiaoa: OK.

src/webui/src/static/model/trial.ts
liuzhe-lz: Why to parse a number?

PR #2791, opened 2020-08-13
Fix nnictl experiment delete

SparkSnail:


tools/nni_cmd/nnictl_utils.py
```diff
@@ -497,7 +497,8 @@ def experiment_clean(args):
         home = str(Path.home())
         local_dir = nni_config.get_config('experimentConfig').get('logDir')
         if not local_dir:
-            local_dir = os.path.join(home, 'nni', 'experiments', experiment_id)
+            local_dir = os.path.join(home, 'nni', 'experiments')
```
chicm-ms, line 500: default experiment parent dir has been changed to `nni-experiments`  in this release.

tools/nni_cmd/nnictl_utils.py
```diff
@@ -497,7 +497,8 @@ def experiment_clean(args):
...
+        local_dir = os.path.join(local_dir, experiment_id)
         local_clean(local_dir)
```
chicm-ms, line 502: we may want to clean `db`, `log`, `trials` directories one by one, to avoid user accidently specify a log dir containing other things while starting experiment.

tools/nni_cmd/nnictl_utils.py
SparkSnail: updated

SparkSnail:
> A safer way is to delete db,log,and trials one by one, then check if the exp dir is empty, delete it if it is empty

updated, although I think this way might be a little complicated, for different kind of trainingService might have different folders, and we may add more folders in logDir in the future, and need to maintain this nnictl command.

PR #2793, opened 2020-08-14
Hotfix installation problem on Windows without conda/venv

liuzhe-lz:
On Windows `site.getsitepackages()[0]` can either be the base directory like `C:\Python38`, or `site-packages`. So all items in that list need to be checked.
The ideal solution is to find a reliable resource packaging machanism. `data_files` is not designed for this use case.

### DIFF
--- a/src/sdk/pynni/nni/package_utils.py
+++ b/src/sdk/pynni/nni/package_utils.py
@@ -286,32 +286,9 @@ def create_customized_class_instance(class_params):
 
     return instance
 
-def get_python_dir(sitepackages_path):
-    if sys.platform == "win32":
-        return str(Path(sitepackages_path))
-    else:
-        return str(Path(sitepackages_path).parents[2])
-
 def get_nni_installation_parent_dir():
     ''' Find nni installation parent directory
     '''
-    def try_installation_path_sequentially(*sitepackages):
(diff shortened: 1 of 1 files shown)


src/sdk/pynni/nni/package_utils.py
```diff
@@ -321,12 +298,21 @@ def _generate_installation_path(sitepackages_path):
...
+            python_dir = _try_installation_path_sequentially(*site.getsitepackages(), site.getusersitepackages())
     return python_dir
 
+def _try_installation_path_sequentially(*sitepackages):
+    '''Try different installation path sequentially util nni is found.
+    Return None if nothing is found
+    '''
+    for sitepackage in sitepackages:
+        for path in [Path(sitepackage).parents[2], Path(sitepackage)]:
+            if (path / 'nni' / 'main.js').is_file():
```
SparkSnail, line 312: what is value of `path / 'nni' / 'main.js'`?

liuzhe-lz: A `Path` object referring to `main.js` file in nni folder in `path`.

chicm-ms (approved):

SparkSnail (approved):

PR #2808, opened 2020-08-20
[v1.8 bug-bash] fix webui error 

Lijiaoa:
@colorjam 

about table open row `log`
- change `HDFS` to `NFS`
- show `trial stdout` as `Trial stdout` and show Trial stdout as hyperlink on pai mode

@SparkSnail 

- show table default metric `FINAL | LATEST` corectly

about compare modal: 

- update legend style
- add scroll bar for table rather than Modal's scroll-x

some format changes.

### DIFF
--- a/src/webui/src/components/public-child/PaiTrialChild.tsx
+++ b/src/webui/src/components/public-child/PaiTrialChild.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { DOWNLOAD_IP } from '../../static/const';
+import LogPathChild from './LogPathChild';
 
 interface PaiTrialChildProps {
     logString: string;
@@ -36,7 +37,10 @@ class PaiTrialChild extends React.Component<PaiTrialChildProps, {}> {
                                         trial stdout
                                     </a>
                                     :
-                                    <span>trial stdout: {logString}</span>
+                                    <LogPathChild
+                                        eachLogpath={logString}
+                                        logName="Trial stdout:"
+                                    />
                             }
                         </div>
                 }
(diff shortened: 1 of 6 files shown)


src/webui/src/components/public-child/PaiTrialChild.tsx
```diff
@@ -36,7 +37,10 @@ class PaiTrialChild extends React.Component<PaiTrialChildProps, {}> {
                                         trial stdout
                                     </a>
                                     :
-                                    <span>trial stdout: {logString}</span>
+                                    <LogPathChild
+                                        eachLogpath={logString}
+                                        logName="Trial stdout:"
+                                    />
```
colorjam, line 43: why the log of pai mode use 2 different components? can they use the same `PaiTrialLog` component?

Lijiaoa: `logPathChild ` just confirm that log string is hyperlink. But `log` nested components will refactor next release.

src/webui/src/static/model/trial.ts
```diff
@@ -304,11 +304,16 @@ class Trial implements TableObj {
     }
 
     private renderNumber(val: any): string {
-        if(typeof val === 'number'){
+        if (typeof val === 'number') {
             if (isNaNorInfinity(val)) {
                 return `${val}`; // show 'NaN' or 'Infinity'
             } else {
-                return `${formatAccuracy(val)} (FINAL)`;
+                if (this.accuracy === undefined) {
```
SparkSnail, line 311: Is `this.accuracy` final result value?

SparkSnail (approved):

ultmaster (approved):

PR #2817, opened 2020-08-24
TF compression fix and UT

liuzhe-lz:
`tf.keras.Sequential` is too tricky to compress, removed from v1.8 release.

### DIFF
--- a/src/sdk/pynni/nni/compression/tensorflow/compressor.py
+++ b/src/sdk/pynni/nni/compression/tensorflow/compressor.py
@@ -6,7 +6,10 @@ Abstract base classes for TensorFlow model compression.
 """
 
 import logging
+
 import tensorflow as tf
+assert tf.__version__.startswith('2'), 'NNI model compression only supports TensorFlow v2.x'
+
 from . import default_layers
 
 _logger = logging.getLogger(__name__)
@@ -130,8 +133,6 @@ class Pruner(Compressor):
         tf.keras.Model
             The compressed model, for convenience. This is exactly the same object to constructor argument.
         """
-        print(self.bound_model.run_eagerly)
-        self.bound_model.run_eagerly = True
         self._update_mask()
         return self.bound_model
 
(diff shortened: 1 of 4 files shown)


src/sdk/pynni/tests/test_compressor_tf.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class CnnModel(Model):
+    def __init__(self):
+        super().__init__()
+        self.conv = Conv2D(filters=10, kernel_size=3, activation='relu')
+        self.pool = MaxPool2D(pool_size=2)
+        self.flatten = Flatten()
+        self.fc1 = Dense(units=10, activation='relu')
+        self.fc2 = Dense(units=1, activation='softmax')
```
chicm-ms, line 66: I suspect the output of `fc2` is always `1` since there must be a `1` if activation is` softmax`.

src/sdk/pynni/tests/test_compressor_tf.py
```diff
@@ -0,0 +1,103 @@
...
+tensor1x10 = tf.constant([[1.0] * 10])
+
+pruners = {
+    'level': (lambda model: LevelPruner(model, [{'sparsity': 0.9, 'op_types': ['default']}])),
+}
+
+
+class TfCompressorTestCase(TestCase):
+
+    def test_layer_detection(self):
```
chicm-ms, line 38: maybe to skip the tests if tf version is 1.x ? since we have pipelines with tf 1.x installed

src/sdk/pynni/tests/test_compressor_tf.py
liuzhe-lz: I'll fix it. But since the result of this model is never used, there should be no real difference.
The correctness of mask algorithm is checked by naive model.

chicm-ms: OK.

src/sdk/pynni/nni/compression/tensorflow/compressor.py
```diff
@@ -75,6 +78,8 @@ class Compressor:
 
     def __init__(self, LayerWrapperClass, model, config_list):
         assert isinstance(model, tf.keras.Model)
+        if isinstance(model, tf.keras.Sequential):
+            raise ValueError('NNI model compression does not support `Sequential` model for now')
```
QuanluZhang, line 82: why does not support `Sequential`?

chicm-ms (approved):

QuanluZhang (approved):

src/sdk/pynni/nni/compression/tensorflow/compressor.py
liuzhe-lz: Because it's too hard to modify it's attributes properly. `Sequence` abuses complex properties.

PR #2818, opened 2020-08-24
[bug bash] issue 2706

zheng-ningxin:
Fix a small bug.

### DIFF
--- a/src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
+++ b/src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
@@ -118,6 +118,8 @@ class CatMaskPadding(MaskFix):
                 continue
             # pad the mask for the non-pruned layers
             for layer in layers:
+                if layer in self.masks:
+                    continue
                 module = name_to_module[layer]
                 w_shape = module.weight.data.size()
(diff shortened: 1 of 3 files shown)


src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -122,7 +122,8 @@ def fix_mask(self):
                 w_shape = module.weight.data.size()
                 w_mask = torch.ones(w_shape).to(device)
                 b_mask = None
-                if hasattr(module, 'bias'):
+                if hasattr(module, 'bias') and module.bias is not None:
```
chicm-ms, line 125: why the config_list in issue 2709 can trigger this problem?

chicm-ms (approved):

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
zheng-ningxin: This bug will be triggered only when some input convolutional layers of the cat operation are pruned, and the others are not. In the unit test, we will randomly assign a sparsity for each conv layer, so all the conv layers are pruned in the model.

QuanluZhang (approved):

PR #2857, opened 2020-09-03
Update bash-completion

JunweiSUN:
Update tools/bash-completion, adding new commands and options, resolve previous bugs.
Also update some lines in docs/en_US/Nnictl.md and python version requirements.


deployment/pypi/setup.py
```diff
@@ -15,6 +17,34 @@
 else:
     raise NotImplementedError('current platform {} not supported'.format(os_type))
 
+class AutoCompletion(install):
+    def run(self):
+        COMP_URL = 'https://raw.githubusercontent.com/microsoft/nni/master/tools/bash-completion'
```
QuanluZhang, line 22: it is not good to use master branch here

deployment/pypi/setup.py
```diff
@@ -15,6 +17,34 @@
...
+        if os_type == 'Linux':
+            HOME = os.environ.get('HOME')
+            if not HOME:
+                install.run(self)
+                return # can not get $HOME, abort.
+            if not os.geteuid(): # run as root
+                BASH_COMP_PREFIX = '/usr/share/bash-completion/completions'
```
QuanluZhang, line 29: have you tried macos?

docs/en_US/Tutorial/Nnictl.md
```diff
@@ -704,6 +704,21 @@ Debug mode will disable version check function in Trialkeeper.
...
 
 * __nnictl webui url__
+  * Description
+
+    Show an experiment's webui url
+
+  * Usage
+
+    ```bash
+    nnictl webui url [options]
```
QuanluZhang, line 714: `options` -> `experiment_id`

deployment/pypi/setup.py
QuanluZhang: why not pack it into python package?

tools/bash-completion
```diff
@@ -1,38 +1,42 @@
 # list of commands/arguments
```
QuanluZhang, line 1: is it possible to automatically generate this file? @liuzhe-lz

deployment/pypi/setup.py
JunweiSUN: Sure, will change to a specific release branch.
If we pack it into python package, it is hard to locate the root of the python package, so I choose to directly download it.

tools/bash-completion
JunweiSUN: Maybe we can just write the whole file's content to the correct location.

docs/en_US/Tutorial/Nnictl.md
JunweiSUN: I think here we should follow the same format as above.

deployment/pypi/setup.py
QuanluZhang: sure, it is ok to not support macOS, but what is the consequence of installing nni in macOS? this directory will be created?

docs/en_US/Tutorial/Nnictl.md
QuanluZhang: i agree, that's why i suggest changing `options` to `experiment_id`...

deployment/pypi/setup.py
JunweiSUN: Only created in Linux. Line 23 do os version check.

QuanluZhang: got it, thanks

deployment/pypi/setup.py
```diff
@@ -48,7 +79,7 @@
         'nnicli': '../../src/sdk/pycli/nnicli'
     },
     package_data = {'nni': ['**/requirements.txt']},
-    python_requires = '>=3.5',
+    python_requires = '>=3.6',
```
SparkSnail, line 82: why upgrade python?

deployment/pypi/setup.py
```diff
@@ -15,6 +17,35 @@
...
     raise NotImplementedError('current platform {} not supported'.format(os_type))
 
+class AutoCompletion(install):
+    def run(self):
+        RELEASE_VERSION = os.popen('git describe --tags --abbrev=0').read().strip()
+        COMP_URL = 'https://raw.githubusercontent.com/microsoft/nni/{}/tools/bash-completion'.format(RELEASE_VERSION)
+        if os_type == 'Linux':
+            HOME = os.environ.get('HOME')
+            if not HOME:
+                install.run(self)
```
SparkSnail, line 27: what is `install.run(self)` used for here?

JunweiSUN: Not quiet sure, seems that this line actually do the install action. Every example about how to use cmdclass has this line, like this [link](https://stackoverflow.com/questions/21915469/python-setuptools-install-requires-is-ignored-when-overriding-cmdclass)

deployment/pypi/setup.py
JunweiSUN: v1.8 no longer support python 3.5, see #2778

deployment/pypi/setup.py
```diff
@@ -15,6 +17,35 @@
 else:
     raise NotImplementedError('current platform {} not supported'.format(os_type))
 
+class AutoCompletion(install):
```
scarlett2018, line 20: better to remove installation, and update doc to explain how to enable bash completion, or output message after nni installation to tell users to enable bash completion

JunweiSUN: Installation removed. An [tutorial documentation](https://github.com/microsoft/nni/blob/0153b3f6eff4cce807857bcee30c872fab63caf2/docs/en_US/CommunitySharings/AutoCompletion.md) has been added.

JunweiSUN:
@QuanluZhang @SparkSnail @scarlett2018 @liuzhe-lz Since this feature will be refactored in v2.0, now I provide a tutorial to guide users how to install bash-completion instead of installing it with `pip`. Please review again when you have time.

liuzhe-lz (approved):

docs/en_US/CommunitySharings/AutoCompletion.md
```diff
@@ -0,0 +1,39 @@
+# Auto Completion
```
QuanluZhang, line 1: "Auto Completion" -> "Auto Completion of nnictl Commands"

QuanluZhang (approved):

SparkSnail (approved):

docs/en_US/CommunitySharings/AutoCompletion.md
JunweiSUN: Updated.

docs/en_US/CommunitySharings/AutoCompletion.md
```diff
@@ -0,0 +1,39 @@
+# Auto Completion for nnictl Commands
```
QuanluZhang, line 1: please mention it after installation doc to let users know there is auto completion support. for example, after QuickStart installation

JunweiSUN: Sure, has added.

PR #2863, opened 2020-09-04
Pretty JSON formatter

liuzhe-lz:
This util formats JSON basing on expected width.

Take following object as example:
```JSON
{
    "batch_size": {"_type": "choice", "_value": [16, 32, 64, 128]},
    "hidden_size": {"_type": "choice", "_value": [128, 256, 512, 1024]},
    "lr": {"_type": "choice", "_value": [0.0001, 0.001, 0.01, 0.1]},
    "momentum": {"_type": "uniform", "_value": [0, 1]}
}
```

When the width is 60, it will be formatted as:

```JSON
{
    "batch_size": {
        "_type": "choice",
        "_value": [16, 32, 64, 128]
    },
    "hidden_size": {
        "_type": "choice",
        "_value": [128, 256, 512, 1024]
    },
    "lr": {
        "_type": "choice",
        "_value": [0.0001, 0.001, 0.01, 0.1]
    },
    "momentum": {
        "_type": "uniform",
        "_value": [0, 1]
    }
}
```

When the width is 40, it will be:

```JSON
{
    "batch_size": {
        "_type": "choice",
        "_value": [
            16,
            32,
            64,
            128
        ]
    },
    "hidden_size": {
        "_type": "choice",
        "_value": [
            128,
            256,
            512,
            1024
        ]
    },
    "lr": {
        "_type": "choice",
        "_value": [
            0.0001,
            0.001,
            0.01,
            0.1
        ]
    },
    "momentum": {
        "_type": "uniform",
        "_value": [
            0,
            1
        ]
    }
}
```

The `_value` arrays will always get collapsed or expanded together, since the util thinks they are "similar".


src/webui/src/static/json_util.ts
```diff
@@ -0,0 +1,177 @@
...
+
+        } else {
+            // these objects look like class instances, so we will try to group their fields
+            const uniqueKeys = new Set(childrenKeys);
+            const iters = new Map();
+            for (let key of uniqueKeys) {
+                const fields = nonNull.map(obj => obj[key]).filter(v => v !== undefined);
+                let elements;
+                if (detectBatch(fields)) {  // look like same field of class instances
+                    elements = batchFormat(fields, indent + '    ', width, key);
```
ultmaster, line 83: Can we expose the interface of indent width?

ultmaster:
I thought it is to call `JSON.stringify` and remove some blanks and newlines. Turned out it was a whole new stringifier.

QuanluZhang:
@Lijiaoa please help review

src/webui/src/static/json_util.ts
liuzhe-lz: As long as it's not configurable in web UI, I think we can make life easier.

ultmaster (approved):

src/webui/yarn.lock
```diff
@@ -3671,7 +3671,7 @@ debug@^4.0.1, debug@^4.1.0, debug@^4.1.1:
   dependencies:
     ms "^2.1.1"
 
-debuglog@*, debuglog@^1.0.1:
```
Lijiaoa, line 3674: could you remove yarn.lock changes? Because PR #2866 had same changes.

Lijiaoa:
PR https://github.com/microsoft/nni/pull/2880 had modifiy the rule `no-use-before-define`

src/webui/yarn.lock
liuzhe-lz: `yarn.lock` is automatically modified on build.
I wonder why it's not committed when `package.json` changed...

Lijiaoa (approved):

src/webui/yarn.lock
Lijiaoa: emmm, the changes of `yarn.lock` should be committed by PR which changes `package.json`. but PR #2866 had done this with other dev update. I think #2866 should be merged into master. so I suggest not upload yarn.lock changes in this PR.

chicm-ms (approved):

PR #2864, opened 2020-09-04
Silence IPC environment warning in nnictl config validator

liuzhe-lz:


chicm-ms (approved):

SparkSnail (approved):

tools/nni_cmd/config_schema.py
```diff
@@ -1,8 +1,9 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-import os
 import json
+import logging
+import os
```
SparkSnail, line 6: what is the order of imported object? `os` should be after `netifaces` in character order?

PR #2865, opened 2020-09-04
Add the input channel dependency pruning.

zheng-ningxin:
Add the input channel pruning dependency analysis. 
AMC pruner prunes the input channel of the convolutional layers, so we also need to provide an input channel dependency analysis tool.


src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -185,6 +185,49 @@ def dependency_sets(self):
             d_sets.append(tmp_set)
         return d_sets
 
+class InputChannelDependency(ChannelDependency):
```
QuanluZhang, line 188: where is this class used?

zheng-ningxin: Supposed to be used in AMC pruner.

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -185,6 +185,49 @@ def dependency_sets(self):
...
+        super(InputChannelDependency, self).__init__(model, dummy_input, traced_model)
+
+    def _get_following_convs(self, tensor):
+        queue = []
+        following_convs = []
+        queue.extend(self.graph.input_to_node[tensor])
+        while queue:
+            curnode = queue.pop(0)
+            if curnode.op_type == 'Conv2d' or curnode.op_type == 'Linear':
+                # find the first met conv
```
QuanluZhang, line 199: `Linear` is not included into `following_convs`?

zheng-ningxin: Yes, in case the pruner may also want to prune the input channel of the linear layer.

QuanluZhang: might be better to not use `folllowing_convs` then

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -185,6 +185,49 @@ def dependency_sets(self):
...
+                following_convs.append(curnode.name)
+                continue
+            successors = self.graph.find_successors(curnode.unique_name)
+            successors = [self.graph.name_to_node[name] for name in successors]
```
QuanluZhang, line 203: seems you only handle `Conv2d` and `Linear`. Is it possible that there is an op type between `tensor` and one following op, and this op type blocks the spread of shape inference?

zheng-ningxin: Yes, reshape operations (such as `view` ) will change the shape of the tensor and may truncate the dependency relation.

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
zheng-ningxin: fixed~ thanks.

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
zheng-ningxin: updated~

src/sdk/pynni/nni/compression/torch/utils/shape_dependency.py
```diff
@@ -185,6 +185,52 @@ def dependency_sets(self):
...
+    def __init__(self, model, dummy_input=None, traced_model=None):
+        super(InputChannelDependency, self).__init__(model, dummy_input, traced_model)
+
+    def _get_following_convs(self, tensor):
+        queue = []
+        key_layers = []
+        queue.extend(self.graph.input_to_node[tensor])
+        while queue:
+            curnode = queue.pop(0)
+            if curnode.op_type == 'Conv2d' or curnode.op_type == 'Linear':
```
chicm-ms, line 198: It seems linear is not added into the set.   I tested it, for resnet 18:  `layer4.1.conv1` and `fc` should be in the same set, but  `fc` is not added into the set.

zheng-ningxin: There is a flatten operation between the `fc` layer and the `layer4.1.conv1`, in the current version, the reshape version may cut the dependency.
image

chicm-ms: since we are pruning feature map on dim=1,  can we make flatten not break dependencies if second params is `1` ? `torch.flatten(x, 1)`

zheng-ningxin: Yes, you are right, we should check whether the parameters will break dependency~ I'll update, thanks~

zheng-ningxin: Updated~

chicm-ms (approved):

QuanluZhang (approved):
still need a clear rule to explain how to build the input dependency group

PR #2876, opened 2020-09-08
AMC supports resnet

chicm-ms:
TODO items in future PR:
1. refactor / unify  amc/utils flops counters
2. dependency aware refactoring


src/sdk/pynni/nni/compression/torch/pruning/amc/amc_pruner.py
```diff
@@ -182,47 +169,42 @@ def __init__(
...
-                tau=tau,
-                init_delta=init_delta,
-                delta_decay=delta_decay,
-                max_episode_length=max_episode_length,
-                debug=debug,
-                train_episode=train_episode,
-                epsilon=epsilon
-            )
-            self.agent = DDPG(nb_states, nb_actions, self.ddpg_args)
+        print('=> Saving logs to {}'.format(self.output_dir))
```
zheng-ningxin, line 174: suggest to replace `print` with `logging`

examples/model_compress/amc/amc_search.py
```diff
@@ -28,27 +29,29 @@ def parse_args():
     parser.add_argument('--train_episode', default=800, type=int, help='number of training episode')
     parser.add_argument('--n_gpu', default=1, type=int, help='number of gpu to use')
     parser.add_argument('--n_worker', default=16, type=int, help='number of data loader worker')
-    parser.add_argument('--job', default='train_export', type=str, choices=['train_export', 'export_only'],
-                        help='search best pruning policy and export or just export model with searched policy')
-    parser.add_argument('--export_path', default=None, type=str, help='path for exporting models')
-    parser.add_argument('--searched_model_path', default=None, type=str, help='path for searched best wrapped model')
+    parser.add_argument('--suffix', default=None, type=str, help='path for searched best wrapped model')
```
QuanluZhang, line 32: this path is for saving or loading the wrapped model?

src/sdk/pynni/nni/compression/torch/pruning/amc/channel_pruning_env.py
```diff
@@ -200,7 +200,9 @@ def step(self, action):
                 self.best_reward = reward
                 self.best_strategy = self.strategy.copy()
                 self.best_d_prime_list = self.d_prime_list.copy()
-                torch.save(self.model.state_dict(), os.path.join(self.args.output, 'best_wrapped_model.pth'))
+                best_model = os.path.join(self.args.output, 'best_model.pth')
+                best_mask = os.path.join(self.args.output, 'best_mask.pth')
+                self.pruner.export_model(model_path=best_model, mask_path=best_mask)
```
QuanluZhang, line 205: do we have ut or it to test `export_model`?

QuanluZhang:
do we have experiment number for AMC in doc?

src/sdk/pynni/nni/compression/torch/pruning/amc/amc_pruner.py
chicm-ms: Thanks, fixed.

src/sdk/pynni/nni/compression/torch/pruning/amc/lib/net_measure.py
```diff
@@ -85,11 +85,11 @@ def measure_layer(layer, x):
     return
 
 
-def measure_model(model, H, W):
+def measure_model(model, H, W, device):
```
QuanluZhang, line 88: where is this function used? you modified its definition but there is no corresponding modification in the calls of this function

chicm-ms: Thanks for good catch, fixed.

src/sdk/pynni/nni/compression/torch/pruning/amc/channel_pruning_env.py
chicm-ms: Thanks, ut added.

examples/model_compress/amc/amc_search.py
chicm-ms: yes, the `searched_model_path` was used to save/load the temporary wrapped module before export, it is not necessary any more after changing to nni speedup.

chicm-ms:
> do we have experiment number for AMC in doc?

do you mean benchmark result for AMC?

QuanluZhang:
> > do we have experiment number for AMC in doc?
> 
> do you mean benchmark result for AMC?

i mean reproducing the result in the paper

examples/model_compress/amc/amc_search.py
QuanluZhang: this is a path, but the name is "suffix". it is a little misleading

chicm-ms: Thanks,  the help string was wrong, fixed.

chicm-ms:
> > > do we have experiment number for AMC in doc?
> > 
> > 
> > do you mean benchmark result for AMC?
> 
> i mean reproducing the result in the paper

reproduce number added.

QuanluZhang (approved):

examples/model_compress/amc/amc_search.py
```diff
@@ -28,27 +29,29 @@ def parse_args():
...
-    elif model == 'mobilenetv2' and dataset == 'cifar10':
+        net = MobileNet(n_class=n_class)
+    elif model == 'mobilenetv2':
         from mobilenet_v2 import MobileNetV2
-        net = MobileNetV2(n_class=10)
+        net = MobileNetV2(n_class=n_class)
+    elif model.startswith('resnet'):
+        net = resnet.__dict__[model](pretrained=True)
+        in_features = net.fc.in_features
+        net.fc = nn.Linear(in_features, n_class)
```
zheng-ningxin, line 54: why the `resnet` models need to modify the last linear layer in such a way and the `mobilenet` models don't ？Can they change the number of the classes in the same way?

chicm-ms: the mobilenet code is copied from amc repo, it does not load imagenet pretrained weight because the repo provides an amc pretrained weight.
for torchvision resnet implementation, if we want to load pretrained weight, we need to set num_classes = 1000, then replace  fc after loading pretrained weight.  Loading pretrained weight should be helpful.

zheng-ningxin: Got it!

src/sdk/pynni/nni/compression/torch/pruning/amc/lib/net_measure.py
```diff
@@ -85,11 +85,11 @@ def measure_layer(layer, x):
...
     global count_ops, count_params
     count_ops = 0
     count_params = 0
-    data = torch.zeros(2, 3, H, W).cuda()
+    data = torch.zeros(2, 3, H, W).to(device)
```
zheng-ningxin, line 92: The input channel is limited to 3 which may limit the usage scenario of AMC pruner (only suitable for the models whose input channel equals 3)? 
By the way, `measure_model` looks duplicated with `count_flops_params` in `nni/compression/torch/utils/counter.py`, shall we reuse `count_flops_params` here?

chicm-ms: Thanks for good catch, AMC uses 2 flops counters,  one for each layer and another for entire model, but count_flops_params currently can not  count flops for each layer.  So we can only replace one counter with `count_flops_params`.   We are planning to refactor the flops counters in AMC in another PR.

zheng-ningxin: Got it~ Thanks~

zheng-ningxin (approved):

PR #2878, opened 2020-09-09
fix experiment import bug and add it cases: experiment import

J-shang:
In docs, experiment import data format example below, `value` can be set as float, but `json_tricks.loads()` only accept string, so fix it.
```
[
  {"parameter": {"x": 0.5, "y": 0.9}, "value": 0.03},
  {"parameter": {"x": 0.4, "y": 0.8}, "value": 0.05},
  {"parameter": {"x": 0.3, "y": 0.7}, "value": 0.04}
]
```

Add IT cases about experiment import and export.


src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -114,7 +114,7 @@ def handle_import_data(self, data):
         data: a list of dictionaries, each of which has at least two keys, 'parameter' and 'value'
         """
         for entry in data:
-            entry['value'] = json_tricks.loads(entry['value'])
+            entry['value'] = json_tricks.loads(str(entry['value']))
```
QuanluZhang, line 117: what is the type of `entry['value']`? a `dict` object?

J-shang: In this simple case like `[{"parameter": {...}, "value": 0.03}]`, it is `<class 'float'>`

QuanluZhang: if it is float, why we invoke `json_tricks.loads`?

J-shang: maybe in some complex scene, user will set value as other types? like numpy array? This should invoke `json_tricks.loads` to converted back.

QuanluZhang: i think `entry['value']` should be a string, otherwise, there is a bug somewhere else.

src/sdk/pynni/nni/msg_dispatcher.py
```diff
@@ -114,6 +114,7 @@ def handle_import_data(self, data):
         data: a list of dictionaries, each of which has at least two keys, 'parameter' and 'value'
         """
         for entry in data:
+            entry['value'] = entry['value']  if type(entry['value']) is str else json_tricks.dumps(entry['value'])
```
QuanluZhang, line 117: better to use `isinstance` instead of `type() is`

J-shang: In this case `type() is` may better, this because we really want to know is whether the type of entry['value'] is `<class 'str'>`. If the type of entry['value'] is a subclass of `str`, `isinstance(entry['value'], str)` will return `true` and  `type(entry['value']) is str` will return `false`. Although `json_tricks.dumps()` will treat subclass of `str` as `str`, I think it is necessary to distinguish subclasses of `str` from `str` at the beginning.

test/config/nnictl_experiment/test_import.json
```diff
@@ -0,0 +1,4 @@
+[
+    {"parameter": {"C": 0.15940134774738896, "kernel": "sigmoid", "degree": 3, "gamma": 0.07295826917955316, "coef0 ": 0.0978204758732429}, "value": 0.6},
```
ultmaster, line 2: `"coef0 "`?

src/sdk/pynni/nni/msg_dispatcher.py
scarlett2018: got it

test/config/nnictl_experiment/test_import.json
J-shang: Because in `search_space.json` is `"coef0 "`, should I modify them all?
https://github.com/microsoft/nni/blob/1d8b8e48e2f3b5ffe8336cfb5d5e3e6b06b24ad2/examples/trials/sklearn/classification/search_space.json#L6

test/config/integration_tests.yml
```diff
@@ -171,6 +171,20 @@ testCases:
...
+  setExperimentIdtoVar: $imoprtExpId
+  stopCommand: nnictl experiment import $imoprtExpId -f config/nnictl_experiment/test_import.json
+  onExitCommand: nnictl stop
+
+# Experiment export test
+- name: nnictl-experiment-export
+  configFile: test/config/nnictl_experiment/sklearn-classification.yml
+  launchCommand: nnictl resume $imoprtExpId
+  stopCommand: nnictl experiment export --filename config/nnictl_experiment/test_export.json --type json --intermediate
+  onExitCommand: nnictl stop
```
chicm-ms, line 186: This is not the proper way to use `stopCommand` and `onExitCommand`.
Here the result of `nnictl experiment export`  should be checked,  you can use a validator to run `export` command and validate the results.

QuanluZhang (approved):

test/config/integration_tests.yml
J-shang: I will try to modify them and add a validator.

ultmaster (approved):

test/nni_test/nnitest/validators.py
```diff
@@ -33,6 +33,14 @@ def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
             print('\n\n')
         remove('report.json')
 
+class ImportValidator(ITValidator):
+    def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
+        exp_id = osp.split(experiment_dir)[-1]
+        import_data_file_path = kwargs.get('import_data_file_path')
+        proc = subprocess.run(['nnictl', 'experiment', 'import', exp_id, '-f', import_data_file_path])
+        assert proc.returncode == 0, \
+            '`nnictl experiment import {0} -f {1}` failed with code {2}'.format(exp_id, import_data_file_path, proc.returncode)
```
chicm-ms, line 42: Any other means to verify the data is actually imported successfully besides checking the return code?

J-shang: I'll view the code again to find if there is a way to check the result.

J-shang: In order to check the imported data, add an api `api/v1/nni/experiment/imported-data` to get the imported data. I don't know if it is appropriate to do so. Maybe users also have demand to view the data they have imported?

chicm-ms: It is OK to call `api/v1/nni/experiment/imported-data` to verify

chicm-ms:
@SparkSnail   Please help review `nni_cmd/updater.py `

SparkSnail (approved):

chicm-ms (approved):

PR #2887, opened 2020-09-11
Compression support for tf.keras.Sequential

liuzhe-lz:
"A little bit breaking" change: `Compressor.compress()` no longer guarantees to return the original model.

This is because `tf.keras.Sequential` is designed for read-only usage and therefore is too tricky to compress in place. So we create a new `Sequential` with instrumented layers instead.

### DIFF
--- a/src/sdk/pynni/nni/compression/tensorflow/compressor.py
+++ b/src/sdk/pynni/nni/compression/tensorflow/compressor.py
@@ -55,7 +55,7 @@ class Compressor:
         for wrapper in self.wrappers:
             setattr(wrapper, name, value)
 
-    def validate_config(model, config_list):
+    def validate_config(self, model, config_list):
         """
         Compression algorithm should overload this function to validate configuration.
         """
--- a/src/sdk/pynni/nni/compression/tensorflow/compressor.py
+++ b/src/sdk/pynni/nni/compression/tensorflow/compressor.py
@@ -15,41 +15,6 @@ from . import default_layers
 _logger = logging.getLogger(__name__)
 
 
-class LayerInfo:
-    """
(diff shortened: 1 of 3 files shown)


src/sdk/pynni/nni/compression/tensorflow/compressor.py
```diff
@@ -228,80 +189,61 @@ def call(self, *inputs):
...
-    assert isinstance(model, tf.keras.Model)
-    if isinstance(model, tf.keras.Sequential):
-        _logger.warning('`Sequential` model is not supported yet, ignored.')
-    ret = {}
-    for key, value in model.__dict__.items():
-        if isinstance(value, tf.keras.Model):
-            ret.update(_locate_layers(value, cur_path + [key]))
-        elif isinstance(value, tf.keras.layers.Layer):
-            ret[id(value)] = LayerInfo(value, cur_path + [key])
+def _instrument(layer, config_list, LayerWrapperClass, compressor):
```
chicm-ms, line 192: Is it possible to move `_instrument` and other functions into `Compressor` class?

QuanluZhang, line 192: better to add docstring here

QuanluZhang (approved):

src/sdk/pynni/nni/compression/tensorflow/compressor.py
liuzhe-lz: Yes, of course. I just personally don't like heavy methods which are used in (and can only be used in) constructor. Because I think "partially constructed" objects can be confusing.

chicm-ms (approved):

PR #2891, opened 2020-09-11
FBv3 tuner

tabVersion:
implementation for FBNetV3

this version cannot run because:

* unified search space for parameters and architecture (ss_gen not support)
* the algorithm returns a set of model architectures and no more trials in stage 2 (to be discussed)


QuanluZhang:
@J-shang please check whether it is helpful.
closed for now

PR #2894, opened 2020-09-14
Upgrade node and yarn

ultmaster:
~Makefile in deployment folder still needs to be updated correspondingly.~

~Yarn will use the latest version per @liuzhe-lz 's suggestion.~

Node.js 10.13 -> 10.22

yarn 1.12.1 -> 1.22.5

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -14,7 +14,7 @@ else {
     $nodeUrl = "https://nodejs.org/download/release/v10.22.0/node-v10.22.0-win-x86.zip"
 }
 # nodejs
-$yarnUrl = "https://yarnpkg.com/latest.tar.gz"
+$yarnUrl = "https://github.com/yarnpkg/yarn/releases/download/v1.22.5/yarn-v1.22.5.tar.gz"
 $unzipNodeDir = "node-v*"
 $unzipYarnDir = "yarn-v*"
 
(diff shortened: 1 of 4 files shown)


liuzhe-lz:
I'm not suggesting to use latest yarn. I just said it's acceptable...

install.ps1
```diff
@@ -7,12 +7,13 @@ $install_yarn = $true
 
 if ([Environment]::Is64BitOperatingSystem) {
     $OS_VERSION = 'win64'
+    $nodeUrl = "https://nodejs.org/download/release/v10.22.0/node-v10.22.0-win-x64.zip"
 }
 else {
     $OS_VERSION = 'win32'
```
liuzhe-lz, line 13: Wait, does NNI supports 32 bit Windows?

liuzhe-lz (approved):

Lijiaoa (approved):

install.ps1
ultmaster: IDK. I'm sure NNI doesn't support 32-bit Linux.

chicm-ms:
> I'm not suggesting to use latest yarn. I just said it's acceptable...

Agree, suggest to use a fixed version.

chicm-ms:
Suggest to run IT pipelines: windows/linux local and remote,  and build pipeline.

chicm-ms (approved):

SparkSnail (approved):

Issue #2883, opened 2020-09-11
The BatchNorm affine problem in SPOS


CuriousCat-7:
According to [this](https://github.com/megvii-model/SinglePathOneShot/issues/3#issuecomment-557450100) issue, the BatchNorm's affine parameter should be set to `False` in the training phase and according to the [official retrain code](https://github.com/megvii-model/SinglePathOneShot/blob/master/src/Evaluation/template/network.py), should be set back to `True` in the **retrain phase**. 

But NNI [still use the SPOSOneShot](https://github.com/microsoft/nni/blob/0a21a901227d156cf65ce6330d5fcfc0205c1ffa/examples/nas/spos/scratch.py#L99), which means BatchNorm is set `affine == False`.

I think that's why the NNI [can not reproduce the official result but very close to it](https://nni.readthedocs.io/zh/latest/NAS/SPOS.html#id9).

ultmaster:
Could you share you reproduction result? If your modification improves the reproduction result, please raise a pull request.

Thanks very much.

CuriousCat-7:
> Could you share you reproduction result? If your modification improves the reproduction result, please raise a pull request.
> 
> Thanks very much.

Ok, I will try, with my eight TITAN Xp cards......

CuriousCat-7:
> Could you share you reproduction result? If your modification improves the reproduction result, please raise a pull request.
> 
> Thanks very much.

Could you tell me your GPU configuration? As far as I know, different kinds of GPUs and different numbers of GPUs will get different results in DL training tasks.  I am using 8 TITAN XP cards with CUDA 10.1, and I think your device configuration may be 4 V100 cards, am I right?

CuriousCat-7:
I have got: 
`top1 = 0.728800, top5 = 0.911251`
where "72.88%" is better than the NNI official "72.14%" but still lower than Megvii official "73.61%".
I think the accuracy can still raise if I increase the epoch num, but I don't what to cheat QAQ.

Thanks

ultmaster:
Good news that you've improved the accuracy by ~0.7%.

You are right that my experiments are done with 4 v100 GPUs. Sadly I don't have the log right now.

240 epochs is just a number for fair comparison. I'm sure you can get a better number with, e.g., 1200 epochs, or whatever tricks like learning rate warmup, or cosine learning rate decay.

CuriousCat-7:
[...]

PR #2902, opened 2020-09-17
Increase SPOS accuracy

CuriousCat-7:
fix SPOS always affine=False bug; change workers to 4 to avoid DALI breaking out; add dump_checkpoint to save the result searched by scratch.py.
I have increased the top1 accuracy from 72.14% to 72.88%, reported in [issue](https://github.com/microsoft/nni/issues/2883#issuecomment-693779116)

### DIFF
--- a/examples/nas/spos/blocks.py
+++ b/examples/nas/spos/blocks.py
@@ -10,7 +10,7 @@ class ShuffleNetBlock(nn.Module):
     When stride = 1, the block receives input with 2 * inp channels. Otherwise inp channels.
     """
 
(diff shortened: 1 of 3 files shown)


ultmaster (approved):

CuriousCat-7:
It looks like fail  #20200917.3  is caused by an installation problem on macOS which has been solved by [#2904]. Need to run the auto unit test program again.

QuanluZhang:
@ultmaster could you help fix the pipeline error?

QuanluZhang (approved):

PR #2906, opened 2020-09-17
Speedup supports channel pruning

chicm-ms:


src/sdk/pynni/nni/compression/torch/pruning/one_shot.py
```diff
@@ -134,9 +134,10 @@ def __init__(self, model, config_list, pruning_algorithm, optimizer=None, **algo
 
     def validate_config(self, model, config_list):
         schema = CompressorSchema([{
-            'sparsity': And(float, lambda n: 0 < n < 1),
-            'op_types': ['Conv2d'],
-            Optional('op_names'): [str]
+            Optional('sparsity'): And(float, lambda n: 0 < n < 1),
```
QuanluZhang, line 137: sparsity becomes optional?

src/sdk/pynni/nni/compression/torch/speedup/compressor.py
```diff
@@ -28,16 +28,22 @@ def get_module_by_name(model, module_name):
...
-    return model, leaf_module
+        if hasattr(model, name):
+            model = getattr(model, name)
+        else:
+            return None, None
+    if hasattr(model, name_list[-1]):
+        leaf_module = getattr(model, name_list[-1])
+        return model, leaf_module
+    else:
+        return None, None
```
QuanluZhang, line 39: curious about what kind of `module_name` returns `None, None`?

src/sdk/pynni/nni/compression/torch/speedup/compressor.py
```diff
@@ -28,16 +28,22 @@ def get_module_by_name(model, module_name):
...
 
 class ModelSpeedup:
     """
     This class is to speedup the model with provided weight mask
     """
 
-    def __init__(self, model, dummy_input, masks_file, map_location=None):
+    def __init__(self, model, dummy_input, masks_file, map_location=None, conv_prune_dim=0):
```
QuanluZhang, line 46: seems strange to put a conv specific argument here

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -221,73 +221,81 @@ def __init__(self, masks, model=None, dummy_input=None, traced=None):
...
-                    break
-                channel_remain.update(all_ones)
-                _logger.debug('Layer: %s ', name)
-                _logger.debug('Original pruned filters: %s', str(all_zeros))
-            # Update the masks for the layers in the dependency set
-            if fine_grained or out_channels is None:
-                # if use the fine-grained pruner or all the layers in
-                # this dependency set are not pruned
+                if name in self.masks:
+                    if len(self.masks[name]['weight'].size()) == 4: # Conv
```
QuanluZhang, line 248: better to use op type to identify different ops? rather than using `.size()`?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -221,73 +221,81 @@ def __init__(self, masks, model=None, dummy_input=None, traced=None):
...
+                        channel_masks.append((self.masks[name]['weight'].abs().sum(sum_idx) != 0).int())
```
QuanluZhang, line 249: is it possible to automatically check which dim is pruned rather than using input argument `conv_prune_dim`?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -221,73 +221,81 @@ def __init__(self, masks, model=None, dummy_input=None, traced=None):
...
                 continue
-            if not all_pruned:
-                # if some layer are not pruned at all
-                # then all the layers in this dependency set
-                # cannot be pruned due to the shape dependency.
-                channel_remain.update(range(out_channels))
-            ori_channels = 0
+            num_channels_list = [len(x) for x in channel_masks if x is not None]
+            # number of channels in same set should be identical
+            assert len(set(num_channels_list)) == 1
```
QuanluZhang, line 264: it is possible that the numbers of kept channels of multiple convs are different? because each layer's sparsity is specified by users

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
chicm-ms: Yes, this can be done. will update.

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
chicm-ms: Good suggestion.

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
chicm-ms: This is total number of channels on specified dimension.  should be same.

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
```diff
@@ -115,15 +115,18 @@ def replace_conv2d(conv, mask):
     else:
         out_channels_index = mask.output_mask.mask_index[1]
         out_channels = out_channels_index.size()[0]
-
-    _logger.debug("replace conv2d with in_channels: %d, out_channels: %d", in_channels, out_channels)
+    groups = conv.groups
+    if conv.in_channels == conv.out_channels == conv.groups:
+        assert in_channels == out_channels
+        groups = in_channels
```
zheng-ningxin, line 121: Why do we need such a specific judgment?

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
```diff
@@ -141,13 +144,16 @@ def replace_conv2d(conv, mask):
     # channal is also divided into serveral groups and each group
     # filter may have different input channel indexes.
     input_step = int(conv.in_channels / conv.groups)
-    in_channels_group = int(in_channels / conv.groups)
-    filter_step = int(out_channels / conv.groups)
+    in_channels_group = int(in_channels / groups)
+    filter_step = int(conv.out_channels / conv.groups)
```
zheng-ningxin, line 148: What's the difference between the `groups` and `conv.groups`?

src/sdk/pynni/nni/compression/torch/pruning/one_shot.py
chicm-ms: If  `exclude` is specified, sparsity is not necessary.

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
chicm-ms: for depthwise conv layers,  such as  in_channel == out_channel == groups == 32,   if we prune any of the three, for example prune output to 16 channels,  we need to prune input channels and groups to 16 too, otherwise it does not work.

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
chicm-ms: conv.groups is the number of groups for original  conv layer.    groups is the  number of groups for new conv layer.    For depthwise conv layer,  the new groups is different than original one.

src/sdk/pynni/nni/compression/torch/speedup/compressor.py
chicm-ms: Those functional node which  is not an attribute of model object,  such as `relu`, `add`

src/sdk/pynni/nni/compression/torch/pruning/one_shot.py
QuanluZhang: got it

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
QuanluZhang: this is specifically for depthwise conv, suggest to add comment to mention it.

QuanluZhang: maybe we can make it general?

QuanluZhang: seems it is indeed specific for depthwise conv in which case the number of groups is reduced, in other cases the number is not changed

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
QuanluZhang, line 148: `conv.out_channels` -> `out_channels`?

src/sdk/pynni/nni/compression/torch/utils/utils.py
```diff
@@ -0,0 +1,87 @@
...
+        How the masks of convolutional layers are pruned, this depends on pruning algorithms, it should
+        return 1 for masks generated by AMCPruner, and returns 0 for masks generated by the rest
+        NNI builtin pruners.
+        0: filter pruning, prune filters of weights which causes channels of output feature maps are pruned.
+        1: channel pruning, prune kernels corresponding to each input channels which causes channels of
+           input feature maps are pruned.
+    """
+    dim0_preserved, dim1_preserved = 0., 0.
+    dim0_num, dim1_num = 0., 0.
+    for module_name in masks:
```
QuanluZhang, line 56: better to add `if continue` to only deal with conv module

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
zheng-ningxin: Maybe we can generalize this into two categories:
(1) keep the number of the filters of each group equal, (don't change the number of the group)
(2) just remove the whole group filters(not only for the depth-wise conv)
I'm not sure if I express this clearly, we can set up a quick meeting and discuss this

src/sdk/pynni/nni/compression/torch/speedup/compressor.py
```diff
@@ -187,8 +169,11 @@ def speedup_model(self):
         """
         training = self.bound_model.training
         _logger.info("start to speed up the model")
+        conv_prune_dim = detect_mask_prune_dim(self.masks)
+        set_conv_prune_dim(conv_prune_dim)
+
         _logger.info("fix the mask conflict of the interdependent layers")
-        fix_mask_conflict(self.masks, self.bound_model, self.dummy_input)
+        fix_mask_conflict(self.masks, self.bound_model, self.dummy_input, conv_prune_dim=conv_prune_dim)
```
QuanluZhang, line 176: is it better to put `detect_mask_prune_dim` in to `fix_mask_conflict`?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -6,8 +6,25 @@
...
+    """
+    Parameters:
+    dim: int
+        0: filter pruning
+        1: channel pruning
+    """
+    global conv_prune_dim
+    conv_prune_dim = dim
+
+def get_conv_prune_dim():
```
QuanluZhang, line 26: seems no one calls this function?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -276,7 +322,15 @@ def dropout_inshape(module_masks, mask):
...
+    if module_masks.output_mask is None:
+        module_masks.set_output_mask(mask)
+        module_masks.set_input_mask(mask)
+        return module_masks.input_mask
+    # if alreay visited
+    assert all(module_masks.output_mask.mask_index[1] == mask.mask_index[1])
 
+    return module_masks.output_mask
 
 def cat_inshape(module_masks, mask, cat_info, last_visited):
```
zheng-ningxin, line 335: Shall we also support the `cat_outshape` in the future?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -552,25 +718,44 @@ def relu_inshape(module_masks, mask):
     """
     assert isinstance(mask, CoarseMask)
     if module_masks.input_mask is not None:
-        # check if has a mask conflict
-        assert module_masks.input_mask <= mask
+        # mask conflict should be solved before speedup
+        assert module_masks.input_mask == mask
```
zheng-ningxin, line 722: From a semantic point of view, I think `<=` maybe better. 
Since currently we pass the mask is by reference, here, `== `and `<=` have no difference, because `mask` and `input_mask` actually point to the same tensor. Both of these assertions are right. But semantically speaking, I recommend using `<=`. Especially when we intend to replace the `mask passing by reference` to `passing by value`, the cat scenario will fail on this assertion if we use `==`:

For example, relu(cat(conv1, conv2)), 
The out_mask of the first visit to `cat` from `conv1` may be {1: 1, 2, 3}, and after the second visit to `cat` from `conv1`, the output mask may have some incremental update, let's say it will be {1: 1, 2, 3, 6,7,8}. and it will fail on the assert.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -612,12 +826,15 @@ def conv2d_mask(module_masks, mask):
     CoarseMask, CoarseMask
         The mask of its input tensor, the mask of its output tensor
     """
-    def convert_to_coarse_mask(mask):
+    def convert_to_coarse_mask(mask, dim=0):
```
zheng-ningxin, line 829: Maybe we can generalize this function in the future?
`convert_to_coarse_mask` is a useful utility, given a mask, it should be able to convert this mask to an coarse_mask automatically without the parameter `dim`.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -376,6 +430,20 @@ def add_inshape(module_masks, mask):
...
+    """
+    Inference the input mask of the add operation from the
+    output mask.
+    """
+    assert isinstance(mask, CoarseMask)
+
+    if module_masks.output_mask is None:
+        module_masks.set_output_mask(mask)
+        module_masks.set_input_mask(mask)
+        return mask
```
QuanluZhang, line 443: how to deal with multiple inputs?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -626,64 +843,69 @@ def convert_to_coarse_mask(mask):
...
+    index, weight_cmask, bias_cmask = convert_to_coarse_mask(mask, dim=conv_prune_dim)
+
     if index is None:
         # TODO: fine grained mask speedup
         return None, None
     # deal with coarse grain mask
+    # mask conflict should be solved by fix_mask_conflict before speedup
     if 'weight' in module_masks.param_masks:
-        module_masks.param_masks['weight'].merge(weight_cmask)
-        module_masks.param_masks['bias'].merge(bias_cmask)
```
zheng-ningxin, line 663: I think we should merge the mask here, but I'm not very clearly, we can discuss later.

For example, 
conv1 and conv2 are two adjacent layers. conv2 prunes the input channel(4,5,6) and conv1 prunes the output channel(1,2,3).

src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py
chicm-ms: Good catch,  updated and add multi groups test case.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -478,6 +574,33 @@ def view_inshape(module_masks, mask, shape):
...
+    assert len(shape['in_shape']) == 4
+    assert len(shape['out_shape']) == 2
+    assert shape['out_shape'][1] == shape['in_shape'][1] * \
+        shape['in_shape'][2]*shape['in_shape'][3]
+
+    assert isinstance(mask, CoarseMask)
+    assert mask.mask_index[1] is not None
+    assert mask.mask_index[0] is None
+
+    return mask
```
QuanluZhang, line 603: do we need to increase `len` of `mask_index` to 4?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -478,6 +574,33 @@ def view_inshape(module_masks, mask, shape):
...
+    assert shape['in_shape'][0] == shape['out_shape'][0]
+    assert len(shape['in_shape']) == 4
+    assert len(shape['out_shape']) == 2
+    assert shape['out_shape'][1] == shape['in_shape'][1] * \
+        shape['in_shape'][2]*shape['in_shape'][3]
+
+    assert isinstance(mask, CoarseMask)
+    assert mask.mask_index[1] is not None
+    assert mask.mask_index[0] is None
+
```
QuanluZhang, line 602: `module_masks.set_output_mask(mask)`?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Sure.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Thanks for good catch, updated.

src/sdk/pynni/nni/compression/torch/utils/utils.py
chicm-ms: Thanks, updated.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Fixed.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Fixed.

src/sdk/pynni/tests/test_model_speedup.py
```diff
@@ -104,6 +105,60 @@ def zero_bn_bias(model):
                 shape = module.running_mean.data.size()
                 module.running_mean = torch.zeros(shape).to(device)
 
+class L1ChannelMasker(WeightMasker):
```
zheng-ningxin, line 108: Shall we put `L1ChannelMasker` into the structure_pruning.py?  What is the reason for putting this class in the test module?

chicm-ms: This is just a simple channel pruning Masker to test channel pruning speedup.  Maybe move to structure_pruning.py when it is needed.

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -221,74 +222,85 @@ def __init__(self, masks, model=None, dummy_input=None, traced=None):
...
-                    out_channels = w_mask.size(0)
-                shape = w_mask.size()
-                count = np.prod(shape[1:])
-                all_ones = (w_mask.flatten(1).sum(-1) == count).nonzero().squeeze(1).tolist()
-                all_zeros = (w_mask.flatten(1).sum(-1) == 0).nonzero().squeeze(1).tolist()
-                if len(all_ones) + len(all_zeros) < w_mask.size(0):
-                    # In fine-grained pruning, there is no need to check
-                    # the shape conflict
-                    _logger.info('Layers %s using fine-grained pruning', ','.join(dset))
-                    fine_grained = True
```
zheng-ningxin, line 259: Shall we check if the mask is fine-grained in the new version?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
```diff
@@ -221,74 +222,85 @@ def __init__(self, masks, model=None, dummy_input=None, traced=None):
...
-                _logger.debug('Original pruned filters: %s', str(all_zeros))
-            # Update the masks for the layers in the dependency set
-            if fine_grained or out_channels is None:
-                # if use the fine-grained pruner or all the layers in
-                # this dependency set are not pruned
+                if name in self.masks:
+                    _, m = get_module_by_name(self.model, name)
+                    assert m is not None
+                    if type(m).__name__ == 'Conv2d':
+                        channel_masks.append((self.masks[name]['weight'].abs().sum(sum_idx) != 0).int())
```
zheng-ningxin, line 252: what if the mask is fine-grained?

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
chicm-ms: Thanks, fine-grained check added.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: as we discussed, we should not allow both input and output being pruned, it is checked in detect_prune_dim.

src/sdk/pynni/nni/compression/torch/utils/mask_conflict.py
chicm-ms: Thanks, fine-grained checked added back.

src/sdk/pynni/nni/compression/torch/speedup/compressor.py
chicm-ms: as we discussed, conv_prune_dim need to be used by infer_shape.py

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -601,6 +792,35 @@ def batchnorm2d_mask(module_masks, mask):
...
     return input_cmask, output_cmask
 
+def linear_mask(module_masks, mask):
+    """
+    Infer input and output shape from weight mask
+
+    Parameters
+    ----------
+    module_masks : ModuleMasks
+        The ModuleMasks instance of the batchnorm2d
```
QuanluZhang, line 802: `batchnorm2d` -> `linear`

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -601,6 +792,35 @@ def batchnorm2d_mask(module_masks, mask):
...
+        The mask of its weights, from the user provided mask file
+
+    Returns
+    -------
+    CoarseMask, CoarseMask
+        The mask of its input tensor, the mask of its output tensor
+    """
+    assert 'weight' in mask
+
+    nonzero_index = torch.nonzero(mask['weight'].sum(0), as_tuple=True)[0]
```
QuanluZhang, line 813: better to make it clear that this function only deals with 0th dimension, using for example asserts

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -601,6 +792,35 @@ def batchnorm2d_mask(module_masks, mask):
...
+
+    # infer shape of input tensor
+    input_cmask = CoarseMask(num_dim=2)
+    input_cmask.add_index_mask(dim=1, index=nonzero_index)
```
QuanluZhang, line 817: linear layer's input could have more than 2 dimensions, `nonzero_index` should be assigned to the last dimension

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -601,6 +792,35 @@ def batchnorm2d_mask(module_masks, mask):
...
+
+    if conv_prune_dim == 0:
+        return None, None
```
QuanluZhang, line 820: it is a little strange to have conv related information in linear op.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -632,64 +855,69 @@ def convert_to_coarse_mask(mask):
...
-        module_masks.set_param_masks('bias', bias_cmask)
-    output_cmask = CoarseMask(num_dim=4)
-    output_cmask.add_index_mask(dim=1, index=index)
-    if module_masks.output_mask is None:
-        module_masks.set_output_mask(output_cmask)
-    else:
-        module_masks.output_mask.merge(output_cmask)
-    return None, module_masks.output_mask
+        if conv_prune_dim == 0:
+            module_masks.set_param_masks('bias', bias_cmask)
```
QuanluZhang, line 894: that dim is 0 means output channel is pruned, right? then bias should also be pruned

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -725,21 +960,19 @@ def conv2d_outshape(module_masks, mask):
     assert isinstance(mask, CoarseMask)
     assert mask.mask_index[1] is not None
     assert mask.mask_index[0] is None
-    assert mask.mask_index[2] is None
-    assert mask.mask_index[3] is None
```
QuanluZhang, line 729: why remove these two lines?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
```diff
@@ -725,21 +960,19 @@ def conv2d_outshape(module_masks, mask):
...
+    if m.in_channels == m.out_channels == m.groups:
         module_masks.output_mask = mask
-    # infer shape of parameters
-    weight_cmask = CoarseMask(num_dim=4)
-    weight_cmask.add_index_mask(dim=0, index=mask.mask_index[1])
-    bias_cmask = CoarseMask(num_dim=1)
-    bias_cmask.add_index_mask(dim=0, index=mask.mask_index[1])
-    module_masks.set_param_masks('weight', weight_cmask)
-    module_masks.set_param_masks('bias', bias_cmask)
-    # input shape is not changed
```
QuanluZhang, line 744: output mask affects weight mask, why remove this part?

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Thanks, fixed.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Thanks, fixed.

src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py
chicm-ms: Thanks, fixed.

QuanluZhang (approved):

zheng-ningxin (approved):

ultmaster (approved):

PR #2911, opened 2020-09-20
Fix eslint for json_utils.ts

ultmaster:
`json_utils.ts` somehow doesn't pass the pipeline now.

I don't know how that happened. Maybe we are testing HEAD instead of MERGE?

### DIFF
--- a/src/webui/src/static/json_util.ts
+++ b/src/webui/src/static/json_util.ts
@@ -29,8 +29,7 @@ function batchFormat(
     width: number,
     keyOrKeys?: string | string[]
 ): string[] {
-
-    let keys: string[];  // dict key as prefix string
+    let keys: string[]; // dict key as prefix string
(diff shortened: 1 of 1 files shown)


Lijiaoa (approved):

src/webui/src/static/json_util.ts
```diff
@@ -50,7 +49,7 @@ function batchFormat(
 
     const hasNested = nonNull.some(obj => detectNested(obj));
 
-    if (!hasNested && lines.every(line => (line.length + curIndent.length < width))) {
+    if (!hasNested && lines.every(line => line.length + curIndent.length < width)) {
```
liuzhe-lz, line 52: I personally think this expression is really hard to read...

liuzhe-lz (approved):

chicm-ms (approved):

PR #2930, opened 2020-09-29
Upgrade sphinx to 3.2.1

tabVersion:
Due to Recommonmark stopping updates, a new dependency is introduced (myst-parser).
But our doc heavily relies on `eval_rst` and we still need Recommonmark to handle this.


tabVersion:
done in a new pr

PR #2947, opened 2020-10-12
fix speedup with CUDA

chicm-ms:


zheng-ningxin:
If the mask tensor is on the wrong device, it should fail on the speedup unit test? 
Why the previous code can pass the speedup unit test

chicm-ms:
> If the mask tensor is on the wrong device, it should fail on the speedup unit test?
> Why the previous code can pass the speedup unit test

in v1.8 the only usage of the created mask in `view_inshape` is here:
https://github.com/microsoft/nni/blob/v1.8/src/sdk/pynni/nni/compression/torch/speedup/compress_modules.py#L56
it calls `to(device)` to make the two tensor on same device before use it.

in v1.9
there are more usage like here, the two tensors must be on same device before comparing:
https://github.com/microsoft/nni/blob/master/src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py#L757

QuanluZhang (approved):

zheng-ningxin (approved):

Tudor33 (commented):
src/sdk/pynni/nni/compression/torch/speedup/infer_shape.py

PR #2949, opened 2020-10-13
upgrade macos

QuanluZhang:


test/pipelines/pipelines-it-installation.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
 - job: 'pip_install_macOS_python36'
   pool:
-    vmImage: 'macOS-10.13'
+    vmImage: 'macOS-latest'
   strategy:
     matrix:
```
chicm-ms, line 21: the `strategy` section can be removed,  it does not take effect as we expected.
The actual default python version depends on the image.

ultmaster (approved):

test/pipelines/pipelines-it-installation.yml
QuanluZhang: updated, thanks!

chicm-ms (approved):

test/pipelines/pipelines-it-installation.yml
```diff
@@ -16,11 +16,7 @@ jobs:
```
chicm-ms, line 16: better to remove the `python36` in job name

QuanluZhang: updated

PR #2951, opened 2020-10-13
Add remote reuse pipeline & fix reuse mode bug

SparkSnail:
1. add remote reuse mode pipeline test files
2. when kill trial in reuse mode, the trial process might has finished, and kill process failed. Need to catch exception.


test/config/training_service.yml
```diff
@@ -95,6 +95,8 @@ pai:
     containerNFSMountPath: 
     paiStorageConfigName: 
 remote:
+  remoteConfig:
+    reuse: false
```
QuanluZhang, line 99: so there is no test for "reuse" is true?

SparkSnail: no, the default value is false, and this value will set to true when pipeline set `$(remote_reuse)`.

chicm-ms:
why do we add a pipeline for reuse?  can we just add a few `reuse` mode test cases instead?

SparkSnail:
> why do we add a pipeline for reuse? can we just add a few `reuse` mode test cases instead?

1. To keep align with pai-reuse mode pipeline. 2. Add a pipeline can test all kinds of cases, only add a few reuse test cases might miss some scenario, for example this reuse pipeline tested that early stopped job might cause exception in reuse mode.

PR #2960, opened 2020-10-14
Cifar update

TairuiWang:
updated the cifar config file


TairuiWang:
Oops, wrong repo. sorry

PR #2963, opened 2020-10-15
Bug Fix: Fix break markdown tables in Chinese documentation

JunweiSUN:
Fix break markdown tables in Chinese documentation by running the following codes in nni's root dir:

```python
import os

def strip_blank_lines(filepath):
    f1 = open(filepath, 'r', encoding='utf-8')
    tmp = filepath + '.tmp'
    f2 = open(tmp, 'w', encoding='utf-8')
    for line in f1.readlines():
        if not line.strip():
            if not line.endswith("\n"):
                f2.write(line.strip())
            else:
                f2.write(line.strip() + "\n")
        else:
            f2.write(line)
    f1.close()
    f2.close()
    os.remove(filepath)
    os.rename(tmp, filepath)


for root, dirs, files in os.walk("."):
    for filename in files:
        if filename.endswith("_zh_CN.md"):
            strip_blank_lines(os.path.join(root, filename))

for root, dirs, files in os.walk("docs/zh_CN"):
    for filename in files:
        if filename.endswith(".md"):
            strip_blank_lines(os.path.join(root, filename))
              
```
The reason why the chinese table is broken is that when you translate strings in crowdin, it may add blank lines (lines only contain space or tab) to the chinese version markdown files. So before merge the branch to master, we can use the code above to strip all blank lines in chinese markdown docs.

This PR also removes the docs/zh_CN/Compressor dir which no longer exists in doc/en_US.

Since some strings have not been translated in crowdin, it may be better to merge this pr after all strings have been translated.


squirrelsc:
No other comment, just make sure all changes are automated, without any manual trigger.

docs/zh_CN/Compression/CompressionReference.md
```diff
@@ -1,18 +1,18 @@
-# 模型压缩 Python API 参考
+# Python API Reference of Compression Utilities
```
squirrelsc, line 1: There are fallback's to English. Make sure they are translated before merging.

PR #2981, opened 2020-10-19
Fix remote reuse bugs

SparkSnail:


tools/nni_trial_tool/trial.py
```diff
@@ -137,10 +137,13 @@ def is_running(self):
...
-                nni_log(LogType.Info, "%s: killing trial" % self.name)
-                for child in psutil.Process(self.process.pid).children(True):
-                    child.kill()
-                self.process.kill()
+                try:
+                    nni_log(LogType.Info, "%s: killing trial" % self.name)
+                    for child in psutil.Process(self.process.pid).children(True):
+                        child.kill()
+                    self.process.kill()
+                except Exception as ex:
```
liuzhe-lz, line 145: Can we catch a more specific bug?

tools/nni_trial_tool/trial.py
```diff
@@ -137,10 +137,13 @@ def is_running(self):
...
         if trial_id == self.id or trial_id is None:
             if self.process is not None:
-                nni_log(LogType.Info, "%s: killing trial" % self.name)
-                for child in psutil.Process(self.process.pid).children(True):
-                    child.kill()
-                self.process.kill()
+                try:
+                    nni_log(LogType.Info, "%s: killing trial" % self.name)
+                    for child in psutil.Process(self.process.pid).children(True):
+                        child.kill()
```
squirrelsc, line 143: If there is try catch, it's better to try catch for each kill. So that one fail won't effect others.

tools/nni_trial_tool/trial.py
```diff
@@ -137,10 +137,13 @@ def is_running(self):
...
+                    self.process.kill()
+                except Exception as ex:
+                    nni_log(LogType.Error, "kill trial %s failed: %s " % (trial_id, str(ex)))
```
squirrelsc, line 146: It likes a clean up, don't need error level log. debug or info is enough.

src/nni_manager/training_service/reusable/environments/remoteEnvironmentService.ts
```diff
@@ -137,40 +137,37 @@ export class RemoteEnvironmentService extends EnvironmentService {
...
+                            environment.setStatus('FAILED');
                         }
+                        this.releaseEnvironmentResource(environment);
                     }
-                } catch (error) {
-                    this.releaseEnvironmentResource(environment);
-                    this.log.error(`Update job status exception, error is ${error.message}`);
                 }
             }
+        } catch (error) {
```
QuanluZhang, line 168: no need to release environment resource here, why?

QuanluZhang:
@SparkSnail could you briefly explain why your changes fix the problem?

src/nni_manager/training_service/reusable/environments/remoteEnvironmentService.ts
SparkSnail: There is a case that environment is submitted, but it starts slowly and hasn't start process and create pid file, the system call refresh function to read pid file, it will cause `no such file` exception.

squirrelsc: I see, it looks you don't check this file. Check if environment.isRunnerReady, then check the file. It will depend on first initialized message. And you set env status to running is too earlier. In remote, it's better to wait isRunnerReady first, then check file status, and set to running, success or failed.

tools/nni_trial_tool/trial.py
SparkSnail: This error is used to catch psutil.Process(), not child kill. In some kind of scene, trial has already exited, and kill() command is sent later. Will throw process not exist error.

tools/nni_trial_tool/trial.py
SparkSnail: ok, add a NoSuchProcess.

tools/nni_trial_tool/trial.py
SparkSnail: psutil.NoSuchProcess is expected exit issue, use info level. For other kinds of unexpected issue, I think use error level is better.

src/nni_manager/training_service/reusable/environments/remoteEnvironmentService.ts
SparkSnail: no, if the environment is failed to start, the isRunnerReady will always be false, but we need to refresh env status to failed here.

squirrelsc: How do you know it failed to start? You may can wait the pid file when initializing, instead of set env to running directly.

SparkSnail: Check process return code to detect if env is failed to start. Added detecting logic for pid file exist.

src/nni_manager/training_service/reusable/trialDispatcher.ts
```diff
@@ -664,17 +664,16 @@ class TrialDispatcher implements TrainingService {
     }
 
     private releaseEnvironment(trial: TrialDetail): void {
-        if (undefined === trial.environment) {
-            throw new Error(`TrialDispatcher: environment is not assigned to trial ${trial.id}, and cannot be released!`);
```
squirrelsc, line 668: What's reason to remove the check? it helps to find unexpected behavior

SparkSnail: IT find a case that assessor report two kill command continuous for a trial, a trial released environment, then release again will throw exception.

QuanluZhang (approved):

chicm-ms (approved):

src/nni_manager/training_service/remote_machine/osCommands.ts
```diff
@@ -29,6 +29,7 @@ abstract class OsCommands {
     public abstract extractFile(tarFileName: string, targetFolder: string): string;
     public abstract executeScript(script: string, isFile: boolean): string;
     public abstract addPreCommand(preCommand: string | undefined, command: string | undefined): string | undefined;
+    public abstract fileExistCommand(filePath: string): string | undefined;
```
squirrelsc, line 32: fileExists is enough. Command is for command related, but this one is not related.

squirrelsc (approved):

PR #2992, opened 2020-10-20
delete PR 2982 changes

Lijiaoa:
PR#2982, PR#2900  should be merge into v1.9, but merge into master.


ultmaster (changes requested):
This will break all the commit history in WebUI. Suggest revert to last commit and force push.

PR #3011, opened 2020-10-21
no need to learn mask when requires_grad is False

twmht:
some networks may set `requires_grad` to false at some layers.

When `requires_grad` of weight or bias  is set to false, we don't need to mask them out.


zheng-ningxin:
I think whether the parameter needs a gradient or not will not affect whether the value should be masked. If this parameter has a corresponding mask tensor, then it should be masked.

chicm-ms:
`requires_grad` controls weight update, masks are configured through `op_names` and `op_type`.
We do not assume that `requires_grad` can affect masks configuration.

twmht:
agree

PR #3034, opened 2020-10-27
Support single machine scheduler policy in AML mode

SparkSnail:


docs/en_US/TrainingService/AMLMode.md
```diff
@@ -81,6 +81,13 @@ amlConfig:
     * optional key, default 1. Used to specify the max concurrency trial number on a GPU device.
 * useActiveGpu
     * optional key, default false. Used to specify whether to use a GPU if there is another process. By default, NNI will use the GPU only if there is no other active process in the GPU.
+* machineScheduler
+    * optional key, default round-robin. Used to set the machine scheduler policy for trial job, values choose from `random, round-robin, single`.
+    ```
+    random: trial jobs will be scheduled into different nodes randomly.
+    round-robin: trial jobs will be scheduled into different node sequentially.
+    single: trial jobs will always be scheduled into same node.
```
QuanluZhang, line 89: this description is a little confusing. this schedule always uses one node?

docs/en_US/TrainingService/AMLMode.md
```diff
@@ -81,6 +81,13 @@ amlConfig:
     * optional key, default 1. Used to specify the max concurrency trial number on a GPU device.
 * useActiveGpu
     * optional key, default false. Used to specify whether to use a GPU if there is another process. By default, NNI will use the GPU only if there is no other active process in the GPU.
+* machineScheduler
```
QuanluZhang, line 84: not suggest to expose this option in configuration. we can simply use the new scheduler for aml training service.

ts/nni_manager/common/manager.ts
```diff
@@ -26,6 +26,7 @@ interface ExperimentParams {
     multiThread?: boolean;
     versionCheck?: boolean;
     logCollection?: string;
+    machineScheduler?: string;
```
QuanluZhang, line 29: yes, we can put the scheduler here internally.

docs/en_US/TrainingService/AMLMode.md
squirrelsc, line 84: It shouldn't be a AML only feature, may be useful for other platforms.

docs/en_US/TrainingService/AMLMode.md
squirrelsc, line 89: It needs to explain what's value to user to schedule on same node.

ts/nni_manager/training_service/reusable/gpuScheduler.ts
```diff
@@ -30,7 +30,7 @@ export class GpuScheduler {
 
     // private readonly machineExecutorMap: Set<TrialDetail>;
     private readonly log: Logger = getLogger();
-    private readonly policyName: SCHEDULE_POLICY_NAME = 'round-robin';
+    private policyName: SCHEDULE_POLICY_NAME = 'round-robin';
```
squirrelsc, line 33: low case name?

docs/en_US/TrainingService/AMLMode.md
SparkSnail: @QuanluZhang @squirrelsc  we need to discuss the scenario of this feature in the meeting, to set this single schuduler only for aml or for all of platforms. To expose the configuration or set it as internally.

ts/nni_manager/training_service/reusable/gpuScheduler.ts
SparkSnail: policyname is not a valid word.

SparkSnail:
close this pr, will refactor logic.

PR #3037, opened 2020-10-27
Refactor code hierarchy part 3: Unit test

liuzhe-lz:
This PR fixes most unit test cases broken by code hierarchy refactor, on Linux and macOS.

All Python UT are moved to `test/ut` folder. Run them with:

```bash
$ cd test
$ pytest ut
```

The new UT pipeline is placed in `pipelines` directory, and I will put future v2.0 pipelines there too.
One important reason that I changed pipeline location is to make the transition smooth.
Once pipeline porting is done, we may either move them back to old location or just remove the old ones. This can be discussed later.

There are several broken test cases which need further investigation. You can search `FIXME` in `test/ut` folder to find them.
UT on Windows is more problematic and is supposed to be fixed later. In fact it was not fully tested in v1.x versions.


.gitignore
```diff
@@ -5,6 +5,10 @@
 /test/model_path/
 /test/temp.json
 /test/ut/sdk/*.pth
+/test/ut/tools/annotation/_generated/
+/ts/nni_manager/exp_profile.json
+/ts/nni_manager/metrics.json
```
chicm-ms, line 10: why add this 3 json files?

chicm-ms (approved):

pipelines/fast-test.yml
```diff
@@ -0,0 +1,216 @@
+# To reduce debug cost, steps are sorted differently on each platform,
```
SparkSnail, line 1: can't get the point, are these test cases in this file duplicated with other pipelines?

test/ut/tools/trial_tool/test_file_channel.py
```diff
@@ -0,0 +1,133 @@
...
+
+    #        self.assertTrue(os.path.exists(runner_file_name))
+    #        with open(runner_file_name, "rb") as runner:
+    #            lines = runner.readlines()
+    #        self.assertListEqual(lines, [b'GI00000000000010"command1"\n', b'GI00000000000010"command2"\n'])
+    #    finally:
+    #        if fc is not None:
+    #            fc.close()
+
+    #def test_send_multi_node(self):
```
SparkSnail, line 56: why annotate these functions?

.gitignore
liuzhe-lz: They are generated during unit test.

pipelines/fast-test.yml
liuzhe-lz: This file is basically `azure-pipelines.yml`. I'm trying to do some refactor and don't want to block other PRs during development. When refactor is done, I will replace the old pipeline with this one.

test/ut/tools/trial_tool/test_file_channel.py
liuzhe-lz: Because this test case failed and I can't fix it.
As stated in FIXME, [`BaseChannel.send()`](https://github.com/microsoft/nni/blob/master/nni/tools/trial_tool/base_channel.py#L72) will raise if the second argument is a string, contradicting to it's doc, and this test case calls it with `fc1.send(CommandType.ReportGpuInfo, "command1")`.
I think this test case was skipped in old pipeline.

SparkSnail (approved):

PR #3052, opened 2020-11-01
[Retiarii] pytorch code converter

QuanluZhang:


test/convert_test/mnasnet.py
```diff
@@ -0,0 +1,243 @@
+import warnings
```
liuzhe-lz, line 1: Please put UT at `test/ut/<package_name>`

PR #3054, opened 2020-11-02
Send SIGINT instead of SIGKILL

Sundrops:
the process of trialJob cannot exit gracely because sigkill cannot be caught

### DIFF
--- a/ts/nni_manager/training_service/local/localTrainingService.ts
+++ b/ts/nni_manager/training_service/local/localTrainingService.ts
@@ -253,7 +253,8 @@ class LocalTrainingService implements TrainingService {
 
             return Promise.resolve();
         }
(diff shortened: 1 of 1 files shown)


liuzhe-lz:
If the trial process is stuck and cannot handle SIGINT, it may be there forever.
I think it will be nice to send SIGINT first, and then send SIGKILL on timeout.

Sundrops:
It's better to add a field into the configuration file and the default value is 'SIGKILL'.

liuzhe-lz:
> It's better to add a field into the configuration file and the default value is 'SIGKILL'.

I think this problem can be handled gracefully without user's attention, so I prefer not to change user interface.
If you have problem on implementation, I can do this later.

PR #3059, opened 2020-11-03
[Retiarii] Base execution engine, codegen and trainer

ultmaster:
~I had a painful time trying to installing NNI on my machine. Let's fix the NNI installation first because testing this PR.~

Test: `nnictl create --config nni.yaml`


nni/retiarii/integration.py
```diff
@@ -0,0 +1,150 @@
...
+
+    The conversion advisor provides are minimum. It is only a send/receive module, and execution engine
+    needs to handle all the rest.
+
+    FIXME
+        How does advisor exit when strategy exists?
+
+    Attributes
+    ----------
+    send_trial_callback
```
liuzhe-lz, line 37: Why `send_trial()` has a callback?

ultmaster: IDK. Maybe execution engine wants to use it somehow?

Let's keep that in mind and maybe remove it in future...

PR #2987, opened 2020-10-19
[v2.0] Refactor code hierarchy (part 2)

liuzhe-lz:
Second part of code hierarchy refactor: Python package hierarchy and installation scripts.  (last part: #2962 )

To install NNI from this branch:
```bash
    $ python setup.py develop
```

## Moved Directories ##

| Content | Source | Destination |
|-|-|-|
| HPO algorithms | `nni.XXX_{tuner,assessor,advisor}` | `nni.algorithms.hpo.XXX_{tuner,assessor,advisor}` |
| NAS algorithms | `nni.nas.{pytorch,tensorflow}.XXX` | `nni.algorithms.nas.{pytorch,tensorflow}.XXX` |
| Compression algorithms | `nni.compression.{torch,tensorflow}.{pruning,quantization}` | `nni.algorithms.compression.{pytorch,tensorflow}.{pruning,quantization}` |
| Feature engineering algorithms | `nni.feature_engineering.XXX_selector` | `nni.algorithms.feature_engineering.XXX_selector` |
| Internal SDK modules | `nni.{msg_dispatcher,protocol,platform,...}` | `nni.runtime.{msg_dispatcher,protocol,platform,...}` |
| Package utility | `nni.package_utils` | `nni.tools.package_utils` |
| Tool packages | `nni_{annotation,cmd,gpu_tool,trial_tool}` | `nni.tools.{annotation,cmd,gpu_tool,trial_tool}` |
| "nnicli" | `nnicli` | `nni.experiment` |

## Discussion ##

0. This is a coarse grained refactor. There are packages I don't know where to place (e.g. `nni.compression.torch.utils`). I don't want to make everything perfect in PR, as long as they do not block future developing.
1. Algorithms are moved to a separate folder. Because I think if it's hard to isolate algorithm packages, it will also be hard for users to implement custom algorithms.
2. Public SDK APIs are not moved because I found it's hard to keep compatibility for codes like `from nni.tuner import Tuner`. So alternatively I moved all internal stuff to `runtime`.
3. "NNI client" is renamed to "experiment", because it's mainly for experiment management and I'm looking forward to add non-client features like experiment creating (from dispatcher process).
4. ~~PyTorch is called `torch` in model compression and `pytorch` in NAS. I don't know which one is better.~~ (Mao chose `pytorch`)

## Summary ##

What part 2 includes:
  + Refactor Python code hierarchy
  + Port install scripts to python
  + Fix release build (building wheel)

What part 2 breaks:
[...]

PR #3060, opened 2020-11-03
Fix develop install

ultmaster:


setup_ts.py
```diff
@@ -132,6 +133,7 @@ def prepare_nni_node():
     node_src = Path('toolchain/node', node_executable_in_tarball)
     node_dst = Path('nni_node', node_executable)
     shutil.copyfile(node_src, node_dst)
+    os.chmod(node_dst, os.stat(node_dst).st_mode | stat.S_IEXEC)
```
liuzhe-lz, line 136: `shutil.copymode()` should be better.

liuzhe-lz:
Empty `__init__.py` should be unnecessary since Python 3.3.

setup.py
```diff
@@ -164,7 +164,7 @@ def run(self):
 
 class Develop(develop):
     def finalize_options(self):
-        self.user = True  # always use `develop --user`
+        self.user = False  # always use `develop --user`
```
liuzhe-lz, line 167: I prefer to explicitly check conda/venv/virtualenv.
Pip will not fallback to `--user` when python directory is protected, and the error message will not suggest to do so either.

If you don't want to do the check anyway, just remove this function. Forcing to `False` will make it impossible to install as normal user without venv.

ultmaster:
> Empty `__init__.py` should be unnecessary since Python 3.3. (github cannot comment on empty file)

@liuzhe-lz I added them because on my machine, nnictl cannot be installed at all because `tools` does not have an `__init__.py` and all the files within are ignored. I'm using Python 3.7.5 with pip 19.3.1 under conda. Not sure why.

ultmaster:
@liuzhe-lz will open a separate PR to resolve comments (and some other issues, possibly). 👍

liuzhe-lz:
> > Empty `__init__.py` should be unnecessary since Python 3.3. (github cannot comment on empty file)
> 
> @liuzhe-lz I added them because on my machine, nnictl cannot be installed at all because `tools` does not have an `__init__.py` and all the files within are ignored. I'm using Python 3.7.5 with pip 19.3.1 under conda. Not sure why.

How did you install it? As mentioned in the [PR](https://github.com/microsoft/nni/pull/2987) you should use `python setup.py develop`, which will not involve pip.

ultmaster:
The issue has been fixed in #3063.

PR #3070, opened 2020-11-06
fix bug of advisor and assesor written backwards

Lijiaoa:
- [x] `advisor.builtinAdvisorName` is written on webui, but when user use other name instead of the field `builtinAdvisorName`, webui will miss this value

### DIFF
--- a/ts/webui/src/components/overview/command/Command1.tsx
+++ b/ts/webui/src/components/overview/command/Command1.tsx
@@ -9,12 +9,12 @@ export const Command1 = (): any => {
     let title = '';
     let builtinName = '';
     if (tuner !== undefined) {
-        title = title.concat('Tuner');
+        title = 'Tuner';
         if (tuner.builtinTunerName !== undefined) {
             builtinName = builtinName.concat(tuner.builtinTunerName);
         }
     }
-    
+
     if (assessor !== undefined) {
         title = title.concat('/ Assessor');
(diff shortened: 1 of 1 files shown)


liuzhe-lz:
I think the logic is kinda "too specific".
I suggest to scan for all advisor/tuner/assessors and then `.join()`.

liuzhe-lz (approved):

ts/webui/src/components/overview/command/Command1.tsx
```diff
@@ -32,8 +37,8 @@ export const Command1 = (): any => {
             <div>
                 <p className='command'>Training platform</p>
                 <div className='nowrap'>{EXPERIMENT.profile.params.trainingServicePlatform}</div>
-                <p className='lineMargin'>{title}</p>
-                <div className='nowrap'>{builtinName}</div>
+                <p className='lineMargin'>{title.join('/')}</p>
```
liuzhe-lz, line 40: .join(' / ')

ultmaster (approved):

ts/webui/src/components/overview/command/Command1.tsx
Lijiaoa: seems like `Tuner/Assessor/Advisor` is better than `Tuner / Assessor / Advisor`, It's also similar with the website style.
image

Issue #3056, opened 2020-11-02
Accept log_dir in Trainer as an input rather than hard coding it


savan77:
**Environment**: Linux
- NNI version: 1.9
- NNI mode (local|remote|pai):
- Client OS: Ubuntu
- Server OS (for remote mode only):
- Python version: 3.6.11
- PyTorch/TensorFlow version: PyTorch 1.4.0
- Is conda/virtualenv/venv used?: No
- Is running in Docker?: No

**Log message**:
 - nnimanager.log: 
 - dispatcher.log:
 - nnictl stdout and stderr:
 

**What issue meet, what's expected?**:
This is not an issue per se but a request to change Trainer class to accept log_dir as an input rather than setting it to `logs`/<time-stamp>.

**How this can help?**
We run our applications on Kubernetes clusters. So, we need to move output and log files to specific folders to persist the data. We can set the paths for ModelCheckpoints in `examples/nas/enas/search.py`. But we cannot set the path for log_dir as we also want to be able to persist log files. And since these scripts use pip package, I can't even just change the code. I need to re-install it. I think it will be more useful and flexible to accept log_dir as an input in `nni/nas/pytorch/trainer.py`. Same goes for other scripts.

**How to reproduce it?**: 

**Additional information**:
I am happy to send a PR, if you want.

ultmaster:
Excellent! Please send a PR. Thanks.

savan77:
Thanks @ultmaster

Right now, ENAS and other NAS algorithms only support CIFAR10 and the scripts don't accept an argument to select a dataset. We are changing those scripts to allow users to use any classification data. Would you also be wanted to add this here? If so, can you provide some guidance around this such that we can make it more generic? Right now, we have placed data loaders in `examples/nas/enas/datasets.py`.

I can send a PR with both changes.

ultmaster:
@savan77 Basically you can use any dataset you want by setting `train_loader` and `valid_loader`. If you want to use other datasets, you can customize the examples by changing the code. Examples are just for demo purposes. We cannot provide the usage of any dataset on earth. :)

savan77:
@ultmaster Thanks for the response. I should have been more clearer. I meant any classification data with specific directory structure such as `class1` -> `all_images_for_class_1`, `class_2` -> `all_images_for_class_2`, etc. This way we can use `ImageFolder` from PyTorch to load any classification data and similarly in TF. Anyway, I will send a PR just for the log_dir.

[...]

PR #3072, opened 2020-11-07
Fix node path in linux

Yqiyuan:
Fix "[Errno 2] No such file or directory: 'node'" when launching Restful server.
The path of 'node' should be absolute path, otherwise it will point to the system 'node'.


liuzhe-lz:
Will be fixed by #3056

PR #3081, opened 2020-11-11
experiment management backend

J-shang:
- [x] nnictl generate experimentId & merge folder
- [x] Modify .experiment structure
- [x] Experiment manager
  - [x] add experiment manager
  - [x] add test
- [x] Refactor
- [ ] (stretch) Merge .config and db.experimentProfile.param

TODO:
~- If we can add a timestamp in `NNIManager.status`, like `NNIManager.status.timestamp`?~

Konwn bug:
- [x] If an error occurs in `ExperimentsManager.withLock`, `.experiment.lock` will not be del. This will block nnictl r&d `.experiment`. Need implement check expiration function in python filelock.
- [x] (#3106 ) It seems the experiment can not receive `SIGTERM` in windows. `psutil.Process(pid).terminate()`?


liuzhe-lz (approved):

nni/tools/nnictl/common_utils.py
```diff
@@ -95,3 +97,23 @@ def generate_folder_name():
         temp_dir = generate_folder_name()
     os.makedirs(temp_dir)
     return temp_dir
+
+def get_file_lock(path: string, timeout=-1, stale=10):
+    class SimpleFileLock(filelock.SoftFileLock):
```
liuzhe-lz, line 102: This creates a new class every time the function gets called, which is unnecessary.

ts/nni_manager/core/experimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+    experimentId: string;
+    isCrashed: boolean;
+}
+
+interface FileInfo {
+    buffer: Buffer;
+    mtime: number;
+}
+
+class ExperimentsManager implements ExpManager {
```
SparkSnail, line 25: I would prefer `NNIExperimentsManager || ExperimentsManagerImpl`  implements `ExperimentsManager` instead of `ExperimentsManager` implements `ExpManager`

ts/nni_manager/core/experimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+        } else {
+            const experimentsInformation = JSON.parse(fs.readFileSync(this.experimentsPath).toString());
+            updateList.forEach((expId: string) => {
+                if (experimentsInformation[expId]) {
+                    experimentsInformation[expId]['status'] = 'STOPPED';
+                } else {
+                    this.log.error(`Experiment Manager: Experiment Id ${expId} not found, this should not happen`);
+                }
+            });
+            fs.writeFileSync(this.experimentsPath, JSON.stringify(experimentsInformation));
```
SparkSnail, line 124: This updateAllStatus() function logic does not need lock?

ts/nni_manager/core/experimentsManager.ts
J-shang: Good suggestion, I will modify them.

ts/nni_manager/core/experimentsManager.ts
J-shang: Due to the cleanliness of the code and it is actually a sync function, lock it at line 46. If it is necessary to unify the lock position of the function, I will transferred the lock in `readExperimentsInfo()` in line 98.

SparkSnail (approved):

nni/tools/nnictl/common_utils.py
```diff
@@ -95,3 +97,23 @@ def generate_folder_name():
         temp_dir = generate_folder_name()
     os.makedirs(temp_dir)
     return temp_dir
+
+class SimpleFileLock(filelock.SoftFileLock):
+    def __init__(self, lock_file, timeout=-1, stale=10):
```
chicm-ms, line 102: better to add a doc string for this class

nni/tools/nnictl/common_utils.py
```diff
@@ -95,3 +97,23 @@ def generate_folder_name():
...
+                if os.path.isfile(self._lock_file) and time.time() - os.stat(self._lock_file).st_mtime > self._stale:
+                    os.remove(self._lock_file)
+                self.acquire()
+                return self
+            except Exception:
+                print_warning('[{}] fail lock file, auto try again!'.format(count))
+                count += 1
+
+def get_file_lock(path: string, timeout=-1, stale=10):
+    return SimpleFileLock(path + '.lock', timeout=timeout, stale=stale)
```
chicm-ms, line 119: what is the purpose to set stale = 10 seconds?

nni/tools/nnictl/config_utils.py
```diff
@@ -54,39 +56,53 @@ class Experiments:
...
         self.experiment_file = os.path.join(home_dir, '.experiment')
-        self.experiments = self.read_file()
+        self.lock = get_file_lock(self.experiment_file, timeout=1, stale=2)
+        with self.lock:
+            self.experiments = self.read_file()
 
-    def add_experiment(self, expId, port, startTime, file_name, platform, experiment_name, endTime='N/A', status='INITIALIZED'):
+    def add_experiment(self, expId, port, startTime, platform, experiment_name, endTime='N/A', status='INITIALIZED',
+                       tag=[], pid=None, webuiUrl=[], logDir=[]):
         '''set {key:value} paris to self.experiment'''
```
chicm-ms, line 65: typo: `paris`

nni/tools/nnictl/config_utils.py
```diff
@@ -54,39 +56,53 @@ class Experiments:
...
-            return False
-        self.experiments[expId][key] = value
-        self.write_file()
-        return True
+        with self.lock:
+            self.experiments = self.read_file()
+            if expId not in self.experiments:
+                return False
+            self.experiments[expId][key] = value
+            self.write_file()
```
chicm-ms, line 89: suggest to add indent to dump the json file to make the file more readable.

ts/nni_manager/core/nniExperimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+        });
+        const updateList: Array<CrashedInfo> = (await Promise.all(expIdList.map((expId) => {
+            return this.checkCrashed(expId, experimentsInformation[expId]['pid']);
+        }))).filter(crashedInfo => crashedInfo.isCrashed);
+        if (updateList.length > 0){
+            const result = await this.withLock(this.updateAllStatus, updateList.map(crashedInfo => crashedInfo.experimentId), fileInfo.mtime);
+            if (result !== undefined) {
+                return result;
+            } else {
+                return this.getExperimentsInfo();
```
chicm-ms, line 50: do we need `await` here?

chicm-ms: and do we need a `delay` here ?

ts/nni_manager/core/nniExperimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+                clearTimeout(this.profileUpdateTimer[key]);
+                this.profileUpdateTimer[key] = undefined;
+            }
+            this.withLockSync(() => {
+                const experimentsInformation = JSON.parse(fs.readFileSync(this.experimentsPath).toString());
+                if (experimentsInformation[experimentId]) {
+                    experimentsInformation[experimentId][key] = value;
+                    fs.writeFileSync(this.experimentsPath, JSON.stringify(experimentsInformation));
+                } else {
+                    this.log.error(`Experiment Manager: Experiment Id ${experimentId} not found, this should not happen`);
```
chicm-ms, line 80: if this should not happen,  maybe better to throw error and crash with `assert`

ts/nni_manager/core/nniExperimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+                }
+            });
+        } catch (err) {
+            this.log.error(err);
```
chicm-ms, line 84: is this err recoverable?

ts/nni_manager/core/nniExperimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+            newPath = path.resolve(newPath);
+        }
+        this.log.info(`Set new experiment information path: ${newPath}`);
+        this.experimentsPath = newPath;
+    }
+
+    public setExperimentInfo(experimentId: string, key: string, value: any): void {
+        try {
+            if (this.profileUpdateTimer[key] !== undefined) {
+                clearTimeout(this.profileUpdateTimer[key]);
```
QuanluZhang, line 71: better to add comment to explain why clear timer here

ts/nni_manager/core/nniExperimentsManager.ts
```diff
@@ -0,0 +1,160 @@
...
+        return this.withLock((path: string) => {
+                const buffer: Buffer = fs.readFileSync(path);
+                const mtime: number = fs.statSync(path).mtimeMs;
+                return {buffer: buffer, mtime: mtime};
+            },
+            this.experimentsPath);
+    }
+
+    private async checkCrashed(expId: string, pid: number): Promise<CrashedInfo> {
+        const alive: boolean = await isAlive(pid);
```
QuanluZhang, line 108: you said different experiments may use the same pid, is there any plan to fix this issue?

nni/tools/nnictl/common_utils.py
J-shang: Setting `stale=10` is unreasonable. Maybe modify it to `stale=-1` is better? means it will seize the lock immediately.

nni/tools/nnictl/config_utils.py
J-shang: Indeed, I will add indent.

ts/nni_manager/core/nniExperimentsManager.ts
J-shang: yes, we need `await` and a `delay` is better, will add them.

ts/nni_manager/core/nniExperimentsManager.ts
J-shang: not different experiments may use the same pid. Some other process may use the crashed experiment pid recorded in `.experiment`, so when we check if pid is alive to check if experiment is alive, it may make some mistake. An easy way to fix this issue is to check if the `experiment id` running on the port matches the `experiment id` recorded in `.experiment`.

ts/nni_manager/core/nniExperimentsManager.ts
J-shang: This error is due to other processes lock the file, I add some details to distinguish this kind of error.

QuanluZhang (approved):

nni/tools/nnictl/config_utils.py
```diff
@@ -54,39 +56,53 @@ class Experiments:
     def __init__(self, home_dir=NNICTL_HOME_DIR):
         os.makedirs(home_dir, exist_ok=True)
         self.experiment_file = os.path.join(home_dir, '.experiment')
-        self.experiments = self.read_file()
+        self.lock = get_file_lock(self.experiment_file, timeout=1, stale=2)
```
chicm-ms, line 59: what is the problem that we try to solve with `timeout=1, stale=2` ?

chicm-ms: the name `timeout` parameter of  `get_file_lock` and `SimpleFileLock` is a little bit misleading, it is used in an inner loop, not the overall timeout.

J-shang: We will check if the lock file is modified over two seconds to determine whether the current lock need to be forced released. 

This because we assume that all operations on the `.experiment` file should be completed within `2` seconds. If the lock file has not been modified for more than `2` seconds, we think the process that generated the lock may have crashed without releasing the lock. 

The `async lock` implementation in TS code has the same logic. The only difference is that TS only retries to lock 100 times, with an interval of 0.1s, then it will throw error. In python code lock will non-stop retry and print warning with an interval of `timeout` because we think users has the ability to judge when `nnictl` should be forced to end.

`timeout=1` means we will check if lockfile is expired with an interval of `1` second.

J-shang: Indeed, `timeout` -> `check_interval`, `SimpleFileLock` -> `ExpiredFileLock` maybe better?

chicm-ms: Got it, then there should be error or at least warning message in the log while removing `self._lock_file`

chicm-ms: We could solve this problem  with special handling the timeout exception, no need to introduce `stale`
`check_interval` is good,  `SimpleFileLock` -> `SimplePreemptiveLock`
```
def __enter__()
    while True:
        try:
            self.aquire()
            return self
        except Timeout:
            # here remove self._lock_file
```

ts/nni_manager/main.ts
```diff
@@ -27,11 +29,10 @@ import { PAIYarnTrainingService } from './training_service/pai/paiYarn/paiYarnTr
 import { DLTSTrainingService } from './training_service/dlts/dltsTrainingService';
 
 function initStartupInfo(
-    startExpMode: string, resumeExperimentId: string, basePort: number, platform: string,
+    startExpMode: string, ExperimentId: string, basePort: number, platform: string,
```
liuzhe-lz, line 32: It should be "camelCase".

nni/tools/nnictl/common_utils.py
```diff
@@ -95,3 +98,43 @@ def generate_folder_name():
...
+            except TimeoutError:
+                print_warning('fail lock file, auto try again!')
+
+    def _acquire(self):
+        open_mode = os.O_WRONLY | os.O_CREAT | os.O_EXCL | os.O_TRUNC
+        try:
+            lock_file_names = glob.glob(self._lock_file + '.*')
+            for file_name in lock_file_names:
+                if os.path.exists(file_name) and time.time() - os.stat(file_name).st_mtime < self._timeout:
+                    raise TimeoutError()
```
chicm-ms, line 122: return `None` here is more consistent with the logic of  base class

nni/tools/nnictl/common_utils.py
```diff
@@ -95,3 +98,43 @@ def generate_folder_name():
...
+    '''this is a lock support check lock expiration, if you do not need check expiration, you can use SoftFileLock'''
+    def __init__(self, lock_file, check_interval=-1):
+        super(__class__, self).__init__(lock_file, check_interval)
+        self._lock_file_name = '{}.{}'.format(self._lock_file, os.getpid())
+
+    def __enter__(self):
+        while True:
+            try:
+                self.acquire()
+                return self
```
chicm-ms, line 112: base class already has a loop, we can set `timeout=-1` for base class,  then we do not need this `__enter__`

chicm-ms: we can rename the `check_interval` back to `stale` with this new design.

J-shang: Indeed, we do not need `__enter__` any more, I will remove it.

nni/tools/nnictl/common_utils.py
```diff
@@ -95,3 +98,36 @@ def generate_folder_name():
...
+        super(__class__, self).__init__(lock_file, timeout=-1)
+        self._lock_file_name = '{}.{}'.format(self._lock_file, os.getpid())
+        self._stale = stale
+
+    def _acquire(self):
+        open_mode = os.O_WRONLY | os.O_CREAT | os.O_EXCL | os.O_TRUNC
+        try:
+            lock_file_names = glob.glob(self._lock_file + '.*')
+            for file_name in lock_file_names:
+                if os.path.exists(file_name) and time.time() - os.stat(file_name).st_mtime < self._stale:
```
chicm-ms, line 114: we need special handling `-1` value for stale to make it means never expire:
```
if os.path.exists(file_name) and (self._stale < 0 or time.time() - os.stat(file_name).st_mtime < self._stale):
```

J-shang: Indeed, handle it.

chicm-ms (approved):

PR #3088, opened 2020-11-13
Integration test

liuzhe-lz:
Pending tasks: (in future PR)
  - port tf1 test cases to pytorch or tf2
  - windows for non-local training services
  - kubeflow
  - framework controller


Dockerfile
```diff
@@ -73,15 +75,15 @@ RUN python3 -m pip --no-cache-dir install pandas==0.23.4 lightgbm==2.2.2
 #
 # Install NNI
 #
-RUN python3 -m pip --no-cache-dir install nni
+COPY dist/nni-${NNI_RELEASE}-py3-none-manylinux1_x86_64.whl .
```
SparkSnail, line 78: why change the install method?

liuzhe-lz: The Dockerfile might be used by two category of users:
  1. Customers
  2. NNI devs

For customers, if they want NNI image with latest release version, they can simply pull from docker hub. So if a customer uses Dockerfile, they are most likely using a modified NNI package, in which case they won't want latest NNI from pypi.

For us, we use pipelines to build and upload docker images. The release pipeline always builds docker image together with wheel, so there is no benefit to push then pull. It can use the same wheel file for pypi and docker image.

chicm-ms (approved):

SparkSnail (approved):

PR #3096, opened 2020-11-17
quick fix unknow trial report to resume experiment

J-shang:
~It should be noted that this may still cause some errors. The fundamental solution is to send `trialId` or `parameter` together when `ReportMetricData`. Or let dispatcher generate `trialId`.~

~Involves four `JobRestServer`, dlts, kubernetes, pai, remote. Need check if `TrialJobId` exist before emit `metric` event.~

Add trial job id filtering function in `nnimanager`.

### DIFF
--- a/nni/runtime/msg_dispatcher.py
+++ b/nni/runtime/msg_dispatcher.py
@@ -39,7 +39,7 @@ def _sort_history(history):
 # Tuner global variables
 _next_parameter_id = 0
 _trial_params = {}
-'''key: trial job ID; value: parameters'''
+'''key: parameter ID; value: parameters'''
 _customized_parameter_ids = set()
 
 
@@ -114,7 +114,7 @@ class MsgDispatcher(MsgDispatcherBase):
         data: a list of dictionaries, each of which has at least two keys, 'parameter' and 'value'
         """
         for entry in data:
-            entry['value'] = entry['value']  if type(entry['value']) is str else json_tricks.dumps(entry['value'])
+            entry['value'] = entry['value'] if type(entry['value']) is str else json_tricks.dumps(entry['value'])
             entry['value'] = json_tricks.loads(entry['value'])
         self.tuner.import_data(data)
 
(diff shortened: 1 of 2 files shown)


liuzhe-lz (approved):

nni/runtime/msg_dispatcher.py
```diff
@@ -182,8 +182,12 @@ def _handle_final_metric_data(self, data):
             customized = True
         else:
             customized = False
+        if id_ in _trial_params:
             self.tuner.receive_trial_result(id_, _trial_params[id_], value, customized=customized,
                                             trial_job_id=data.get('trial_job_id'))
+        else:
+            _logger.info('Find unknown job parameter id %s, maybe something goes wrong.', _trial_params[id_])
```
QuanluZhang, line 189: better to use warning

nni/runtime/msg_dispatcher.py
```diff
@@ -182,8 +182,12 @@ def _handle_final_metric_data(self, data):
...
+            return
```
QuanluZhang, line 190: this return is useless

ts/nni_manager/core/nnimanager.ts
```diff
@@ -678,11 +678,15 @@ class NNIManager implements Manager {
...
-        if (this.dispatcher === undefined) {
-            throw new Error('Error: tuner has not been setup');
+        if (this.trialJobs.has(metric.id)){
+            await this.dataStore.storeMetricData(metric.id, metric.data);
+            if (this.dispatcher === undefined) {
+                throw new Error('Error: tuner has not been setup');
+            }
+            this.dispatcher.sendCommand(REPORT_METRIC_DATA, metric.data);
+        } else {
+            this.log.error(`NNIManager received non-existent trial job metrics: ${metric}`);
```
QuanluZhang, line 688: also warning might be better here

J-shang: fixed them

QuanluZhang (approved):

Issue #3119, opened 2020-11-24
checkpoint save and load under qat failed 


eedalong:

linbinskn:
After testing save function, we reproduced this problem as you say. Thank you for your contribution, we will review this PR as soon as possible.

eedalong:
problem fixed, so i will close this issue

PR #3120, opened 2020-11-24
fix buffer allocate and wrapper name confliction

eedalong:
#3119 

1. register zero_point and scale with a real value Tensor rather not None, cause latter value will cause state dict load failure 

2. avoid wrapper's name to be the same with module's name, cause it will cause some problems when load state dict


linbinskn:
You have to resolve the conflicts in unit test since have changes the name of wrap module. For example, modify 
the  **test_compressor.py** in path-to-nni/test/ut/sdk.

eedalong:
> You have to resolve the conflicts in unit test since you have changed the name of wrap module. For example, modify
> the **test_compressor.py** in path-to-nni/test/ut/sdk.

what's the test command?

PR #3124, opened 2020-11-25
fix checkpoint load error and stop updating paramters in evaluation stage

eedalong:
#3119 
 When testing QAT in NNI, users would run into these problems
1. checkpoint load fails, which will cause incremental training fail

2. in evaluation stage, parameters of weight quantization and output quantization should not be updated

3. when we load checkpoint and then evaluate immediately, we would see bad model performance because of the reset of inner variable :[steps]

this pr fix these problems

### DIFF
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -88,13 +88,14 @@ def update_quantization_param(bits, rmin, rmax):
     if rmin.is_cuda:
         rmin = torch.min(rmin, torch.Tensor([0]).cuda())
         rmax = torch.max(rmax, torch.Tensor([0]).cuda())
+        qmin = torch.Tensor([0]).cuda()
+        qmax = torch.Tensor([(1 << bits) - 1]).cuda()
     else:
         rmin = torch.min(rmin, torch.Tensor([0]))
         rmax = torch.max(rmax, torch.Tensor([0]))
+        qmin = torch.Tensor([0])
+        qmax = torch.Tensor([(1 << bits) - 1])
 
-    # the min and max quantized values, as floating-point values
-    qmin = 0
-    qmax = (1 << bits) - 1
     # First determine the scale.
     scale = (rmax - rmin) / (qmax - qmin)
 
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -143,11 +143,11 @@ class QAT_Quantizer(Quantizer):
                     types of nn.module you want to apply quantization, eg. 'Conv2d'
         """
         super().__init__(model, config_list, optimizer)
-        self.steps = 1
         modules_to_compress = self.get_modules_to_compress()
+        self.bound_model.register_buffer("steps", torch.Tensor([1]))
         for layer, config in modules_to_compress:
-            layer.module.register_buffer("zero_point", None)
-            layer.module.register_buffer("scale", None)
+            layer.module.register_buffer("zero_point", torch.Tensor([0.0]))
+            layer.module.register_buffer("scale", torch.Tensor([1.0]))
             if "output" in config.get("quant_types", []):
                 layer.module.register_buffer('ema_decay', torch.Tensor([0.99]))
                 layer.module.register_buffer('tracked_min_biased', torch.zeros(1))
@@ -229,7 +229,8 @@ class QAT_Quantizer(Quantizer):
         quant_start_step = config.get('quant_start_step', 0)
         assert weight_bits >= 1, "quant bits length should be at least 1"
 
-        if quant_start_step > self.steps:
+        # we dont update weight in evaluation stage
+        if quant_start_step > self.bound_model.steps or not wrapper.training:
             return weight
 
         # if bias exists, quantize bias to uint32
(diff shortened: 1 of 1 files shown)


eedalong:
Here scale and zero point are defined with a 1-dimensioon tensor, but might got size changed to 0-dimension tensor during calculation by function call update_quantization_param(bits, rmin, rmax). 

rmin and rmax  are calculated by torch.min() and torch.max(), they are 0-dimension tensor, and in update_quantization_param(bits, rmin, rmax), you calculated min value by rmin = min(rmin, 0),  after this step, rmin is still a 0-dimension tensor, such as Tensor(1.23) but not Tensor([1.23]). 

Then when we save checkpoint, we might save 0-dimension tensor for scale and zero_point, And now BOMM!

checkpoint load is failed due to paramter size dismatch, because scale and zero_point are defined as 1-dimension Tensor but we see 0-dimension Tensor in checkpoint. this is fixed in commit 1ed0163d8c0e37d0209ba073e7a30d863362bc01

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -258,15 +264,17 @@ def quantize_output(self, output, wrapper, **kwargs):
...
-                                                                   module.ema_decay, self.steps)
-        module.scale, module.zero_point = update_quantization_param(output_bits, module.tracked_min, module.tracked_max)
+        # we dont update output quantization parameters in evaluation stage
+        if wrapper.training:
+            current_min, current_max = torch.min(output), torch.max(output)
+            module.tracked_min_biased, module.tracked_min = update_ema(module.tracked_min_biased, current_min,
+                                                                       module.ema_decay, self.bound_model.steps)
+            module.tracked_max_biased, module.tracked_max = update_ema(module.tracked_max_biased, current_max,
+                                                                       module.ema_decay, self.bound_model.steps)
+            module.scale, module.zero_point = update_quantization_param(output_bits, module.tracked_min, module.tracked_max)
```
linbinskn, line 277: At present, we have to re-calculate the scale and zero_point of output and weight. Because the parameter of activation and weight all use the same scale and zero-point of such module. If quantize both activation and weight during testing without updating,  parameters would be only one of them, and the result could be wrong. But this design that scale and zero_point using the same parameter doesn't make sense, we plan to refactor it in the following release.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -143,11 +148,11 @@ def __init__(self, model, config_list, optimizer=None):
                     types of nn.module you want to apply quantization, eg. 'Conv2d'
         """
         super().__init__(model, config_list, optimizer)
-        self.steps = 1
         modules_to_compress = self.get_modules_to_compress()
+        self.bound_model.register_buffer("steps", torch.Tensor([1]))
         for layer, config in modules_to_compress:
-            layer.module.register_buffer("zero_point", None)
-            layer.module.register_buffer("scale", None)
+            layer.module.register_buffer("zero_point", torch.Tensor([0.0]))
```
linbinskn, line 154: If we just use 0-dimension definition here, would it be more simple and we don't need to add conversion?

nni/algorithms/compression/pytorch/quantization/quantizers.py
eedalong: um, Indeed activation and weight use the same field, but once a user defined to quantize output of the layer, module.zero_point  module.scale will always be activation's zero_point and scale , so it will not cause anything wrong

nni/algorithms/compression/pytorch/quantization/quantizers.py
eedalong: define 0-dimension does save some work,defining it to 1-dimension tensor is just for keeping consitent with other buffered variables

eedalong:
add cuda transformation in quantize calculation, because when using multi GPU training , we might have tensor type mismatch problem, i.e. we might have an error of adding a cpu tensor and a cuda tensor. Although I have no idea why this is happening ==.

linbinskn (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -85,12 +85,17 @@ def update_quantization_param(bits, rmin, rmax):
...
     # Otherwise, we would not meet the requirement that 0 be an exactly
     # representable value.
-    rmin = min(rmin, 0)
-    rmax = max(rmax, 0)
+    if rmin.is_cuda:
+        rmin = torch.min(rmin, torch.Tensor([0]).cuda())
+        rmax = torch.max(rmax, torch.Tensor([0]).cuda())
+        qmin = torch.Tensor([0]).cuda()
+        qmax = torch.Tensor([(1 << bits) - 1]).cuda()
+    else:
```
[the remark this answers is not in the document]
eedalong: if we claim scale and zero-point as a 1-dimension tensor in QAT, but here changes scale and zero_point to 0-dimension, then we load checkpoint after saving, we would have type mismatch problem, cuz scale and zero_point are claimed as 1-dimension tensor, but in checkpoint, they are saved as a 0-dimension tensor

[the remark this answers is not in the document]
eedalong: So what i do here is to keep the scale and zero_point's size consistent during all calculation

J-shang (approved):

linbinskn (approved):

liuzhe-lz (approved):

PR #3138, opened 2020-11-29
New experiment config

liuzhe-lz:
This PR defines the new experiment config schema.

## Example Usage ##

Following code snippet will generate a valid `config.yaml` file:

```python
from ruamel import yaml
from nni.experiment.config import ExperimentConfig, AlgorithmConfig, LocalConfig, convert

config = ExperimentConfig(
    experiment_name = 'MNIST',
    trial_concurrency = 1,
    max_experiment_duration = '10m',
    max_trial_number = 10,
    search_space_file = 'search_space.json',
    tuner = AlgorithmConfig(name='TPE', class_args={'optimize_mode': 'maximize'}),
    trial_command = 'python3 mnist.py',
    trial_code_directory = '.',
    training_service = LocalConfig(use_active_gpu=True)
)

config = convert.to_old_yaml(config)
yaml.dump(config, open('config.yaml', 'w'))
```
Or alternatively:

```python
from ruamel import yaml
from nni.experiment.config import ExperimentConfig, convert

config = ExperimentConfig('local')
config.trial_concurrency = 1
config.max_experiment_duration = '10m'
config.max_trial_number = 10
config.search_space = {
    "dropout_rate": {"_type": "uniform", "_value": [0.5, 0.9]},
    "conv_size": {"_type": "choice", "_value": [2, 3, 5, 7]}
}
config.tuner = AlgorithmConfig(name='TPE', class_args={'optimize_mode': 'maximize'})
config.trial_command = 'python3 mnist.py'
config.trial_code_directory = '.'
config.training_service.use_active_gpu = True

config = convert.to_old_yaml(config)
yaml.dump(config, open('config.yaml', 'w'))
```

## Design Notes ##

The schema should be expressible in both YAML/JSON object and Python dataclass.
In YAML/JSON the keys are expected to be camelCase and in Python they should be snake_case.
This technically is not a problem since `Config.load()` and `Config.json()` have routines to do the convert.

I think the consistency with old schema is not a case. Because consistence is to help migration, while we should not push exist users to migrate at all.
The most important goal of this refactor is to reduce users' learning cost. And for exist users, to not learn the cost is zero.
I believe we will continue to support old schema for very long time, maybe until NNI v3.0. So it basically benefits no body to push the migration unless a user is unsatisfied old schema and proactively want a new one.
And in the other hand, the more difference between two schema, the smaller possibility NNI may fail to detect schema version.

Taking Yuge's advice, the config is divided into platform-independent part and nested training service config.
Trial config is no longer nested because I can't find a reason to do so. While tuner/assessor/advisor config is still nested because the same format may repeats for more than one algorithms.
There is a problem that whether training service config should share the same key. Using same key is definitely better in Python, but using "localCofing" / "remoteConfig" may looks better in YAML.
The problem might affect heterogeneous training service. If training services use same key "training service config", it will be a object/dict for heterogeneous. And if they use different keys, there will be multiple training service config objects.


liuzhe-lz:
```
- experiment name: string | undefined
    Mnemonic name of the experiment. This will be shown in web UI and nnictl.

- training service: string
    Supported value: "local", "remote", "openpai"

- search space file: string | undefined
    Path to a JSON file containing the search space.
    If it is a relative path, it relates to the directory containing this experiment config.
    Search space format is determined by tuner. Common format for built-in tuners is documeted here <LINK>
    Mutually exclusive to `search space`.

- search space: Object | undefined
    Search space object.
    The format is determined by tuner. Common format for built-in tuners is documented here <LINK>
    Mutually exclusive to `search space file`.

- trial command: string | list<string>
    Command(s) to launch trial.
    Bash will be used for Linux and macOS, while PowerShell will be used for Windows.

- trial code directory: string = "."
    Path to the directory containing trial source files.
    All files in this directory will be sent to training machine, unless there is a `.nniignore` file <LINK>
    If it is a relative path, it relates to the directory containing this experiment config.

- trial concurrency: int
    Specify how many trials should be run concurrently.
    The real concurrency also depends on hardware resources and may be less than this number.

- trial gpu number: int | undefined
    Number of GPUs used by each trial.
    If set to zero, trials may not have access to any GPU.
    If not specified, trials will be created as if they require no GPU, but they can still access all GPUs on the training machine.

- max experiment duration: string | undefined
    Limit the duration of this experiment if specified.
    When the time runs out, the experiment will stop creating trials but continue to serve web UI.
    Format: number + "s"/"m"/"h"/"d" (stands for seconds/minuts/hours/days respectively)
    Examples: "10m", "0.5h"

- max trial number: int | undefined
    Limit the number of trials to create if specified.
    When the budget runs out, the experiment will stop creating trials but continue to serve web UI.

- nni manager ip: string | undefined
    IP of current machine, used by training machines to access NNI manager. Not used in local mode.
    If not specified, this will be the default IPv4 address of outgoing connection.

- use annotation: bool = false
    Enable annotation <LINK>
    When using annotation, search space should not be specified manually.

- reuse mode: bool = false
    Enable reuse mode <LINK>

- debug: bool = false
    Enable debug mode.
    Logging will be more verbose and some internal validation will be loosen.

- log level: string | undefined
    (might hidden)
    Supported value: "trace", "debug", "info", "warning", "error", "fatal"
    Defaults to "info" if `debug` is false, otherwise "debug".
    This option will affect all modules except trials, including tuner, NNI manager, training service, etc.
    For Python modules, "trace" acts as `logging.DEBUG` and "fatal" acts as `logging.CRITICAL`.

- experiment working directory: string | undefined
    (might hidden)
    Specify the directory to place log, checkpoint, metadata, and other run-time stuff.
    `<home>/nni-experiments` will be used by default.
    NNI will create a subdirectory named by experiment ID, so it is safe to use same directory for multiple experiments.

- tuner gpu indices: list<int> | string | undefined
    Limit the GPUs visible to tuner, assessor, and advisor.
    This will be the `CUDA_VISIBLE_DEVICES` environment variable of tuner process.
    Because tuner, assessor, and advisor run in same process, this option will affect them all.

* tuner / assessor / advisor: Algorithm

  - name: string | undefined
      Name of a built-in algorithm, or an algorithm registered with `nnictl package install` <NEED UPDATE>
      Mutually exclusive to `class`.

  - class: string | undefined
      Qualified class name.
      Mutually exclusive to `name`.
      Example: `hyperopt_tuner.HyperoptTuner`

  - code directory: string | undefined
      Path to directory containing the `class`.
      If not specified, the class will be looked up in Python module search path <https://docs.python.org/3/tutorial/modules.html#the-module-search-path>
      If it is a relative path, it relates to the directory containing this experiment config.

  - class arguments: Object = {}
      Arguments of the algorithm.
      See algorithm's document for supported value.

* local

  - use active gpu: bool
      Specify whether NNI should submit trials to GPUs occupied by other tasks.
      If your are using desktop system with GUI, set this to `true`.

  - max trial number per gpu: int = 1
      Specify how many trials can share one GPU.

  - gpu indices: list<int> | string | undefined
      Limit the GPUs visible to trial processes.
      This will be the `CUDA_VISIBLE_DEVICES` environment variable.

* remote

  - machine list: list<RemoteMachine>

    - host: string
        IP or hostname (domain name) of the remote machine.

    - port: int = 22
        SSH service port.

    - user: string
        Login user name.

    - password: string | undefined
        Login password.
        If not specified, `ssh key file` will be used instead.

    - ssh key file: string = "~/.ssh/id_rsa"
        Path to ssh key file (identity file).
        Only used when `password` is not specified.
        If it is a relative path, it relates to the directory containing this experiment config.

    - ssh passphrase: string | undefined
        Passphrase of SSH identity file.

    - use active gpu: bool
        Specify whether NNI should submit trials to GPUs occupied by other tasks.

    - max trial number per gpu: int = 1
        Specify how many trials can share one GPU.

    - gpu indices: list<int> | string | undefined
        Limit the GPUs visible to trial processes.
        This will be the `CUDA_VISIBLE_DEVICES` environment variable.

    - trial prepare command: string | list<string> | undefined
        Command(s) to run before launching each trial.
        This is useful if preparing steps vary for different machines.

* openpai

  - host: string
      Hostname of OpenPAI service.

  - user name: string
      OpenPAI user name.

  - token: string
      OpenPAI user token.
      This can be found in your OpenPAI user settings page.

  - trial cpu number: int = 1
      Number of CPUs used by each trial.

  - trial memory: string
      Memory used by each trial.
      Examples: "1gb", "512mb"

  - docker image: string = "msranni/nni"
      Label for docker image.

  - docker auth file: string | undefined
      Path to user authentication file. Used for private docker registry.
      See OpenPAI's document for auth file format. <LINK>
      If it is a relative path, it relates to the directory containing this experiment config.
      # FIXME: Seems OpenPAI does not recommend to use authFile now. Consider expand it to multiple fields.

* [temporarily removed]

  - multiThread
      reason: I think it should be customized in dispatcher subclass, not experiment config

  - versionCheck
      reason: hard to explain what it does ("trial keeper" should not be shown to users)

  - logCollection
      reason: seems not under maintaince

  - tuner.includeIntermediateResults
      reason: suggest to re-design the feature

  - trial.shmMB (openpai)
      reason: can't find in openpai's doc

  - trial.portList (openpai)
      reason: can't find in openpai's doc
```

SparkSnail:
Does this pr support heterogeneousConfig?

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+nni manager ip
+--------------
+
+IP of current machine, used by training machines to access NNI manager. Not used in local mode.
+
+type: ``Optional[str]``
+
+If not specified, this will be the default IPv4 address of outgoing connection.
+
+// there is an easy way to get the IP mentioned above, I will remove netifaces denendency
```
ultmaster, line 144: typo: dependency

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+
+
+use annotation
+--------------
+
+Enable `annotation <https://nni.readthedocs.io/en/stable/Tutorial/AnnotationSpec.html>`_.
```
ultmaster, line 150: Suggest relative path

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+
+This will be the ``CUDA_VISIBLE_DEVICES`` environment variable of tuner process.
+
+Because tuner, assessor, and advisor run in same process, this option will affect them all.
+
+
+tuner
+-----
+
+Specify the tuner [TODO:link]
```
ultmaster, line 216: Can we have the link now?

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+
+Specify the advisor [TODO:link]
+
+type: Optional `AlgorithmConfig`_
+
+
+training service
+----------------
+
+Specify `training service <https://nni.readthedocs.io/en/stable/TrainingService/Overview.html>`_.
```
ultmaster, line 240: Same here

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+-----------------------------
+
+Mount point of storage service (typically NFS) in docker container.
+
+type: ``str``
+
+This must be an absolute path.
+
+
+open pai config
```
ultmaster, line 564: I think openpai is a word (at least OpenPAI is).

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+use active gpu
+--------------
+
+Specify whether NNI should submit trials to GPUs occupied by other tasks.
+
+type: ``bool``
+
+If your are using desktop system with GUI, set this to ``True``.
+
+// need to discuss default value
```
ultmaster, line 318: What is the default value here?

docs/en_US/reference/experiment_config.rst
```diff
@@ -0,0 +1,575 @@
...
+trial gpu number
+----------------
+
+Number of GPUs used by each trial.
+
+type: ``Optional[int]``
+
+If set to zero, trials will have no access to any GPU. 
+
+If not specified, trials will be created and scheduled as if they do not use GPU,
```
ultmaster, line 107: What is the conclusion from last meeting? What does -1, 0, None mean respectively?

nni/experiment/config/convert.py
```diff
@@ -76,7 +76,7 @@ def to_v1_yaml(config: ExperimentConfig, skip_nnictl: bool = False) -> Dict[str,
         data['remoteConfig'] = {'reuse': ts['reuseMode']}
         data['machineList'] = []
         for machine in ts['machineList']:
-            machine = {
+            data['machineList'].append({
```
SparkSnail, line 79: how about port?

nni/experiment/config/convert.py
```diff
@@ -86,24 +86,84 @@ def to_v1_yaml(config: ExperimentConfig, skip_nnictl: bool = False) -> Dict[str,
...
     elif ts['platform'] == 'pai':
-        data['trial']['cpuNum'] = ts['trialCpuNumber']
-        data['trial']['memoryMB'] = util.parse_size(ts['trialMemorySize'])
         data['trial']['image'] = ts['docker_image']
+        data['trial']['nniManagerNFSMountPath'] = ts['local_storage_mount_point']
+        data['trial']['containerNFSMountPath'] = ts['container_storage_mount_point']
         data['paiConfig'] = {
             'userName': ts['username'],
             'token': ts['token'],
             'host': 'https://' + ts['host'],
```
SparkSnail, line 98: The host does not contains prefix? how about `http://`?

nni/experiment/config/convert.py
```diff
@@ -135,9 +195,19 @@ def to_cluster_metadata(config: ExperimentConfig) -> List[Dict[str, Any]]:
...
+        ret.append({'aml_config': experiment_config['amlConfig']})
+
+    elif config.training_service.platform == 'kubeflow':
+        ret.append({'kubeflow_config': experiment_config['kubeflowConfig']})
+
+    elif config.training_service.platform == 'frameworkcontroller':
+        ret.append({'frameworkcontroller_config': experiment_config['frameworkcontrollerConfig']})
+
+    elif config.training_service.platform == 'adl':
+        pass
```
SparkSnail, line 210: why pass `adl`?

nni/experiment/config/remote.py
```diff
@@ -0,0 +1,52 @@
...
+__all__ = ['RemoteConfig', 'RemoteMachineConfig']
+
+
+@dataclass(init=False)
+class RemoteMachineConfig(ConfigBase):
+    host: str
+    port: int = 22
+    user: str
+    password: Optional[str] = None
+    ssh_key_file: PathLike = '~/.ssh/id_rsa'
```
SparkSnail, line 21: The ssh_key is optional key, is used to connect remote machine, not local file. I think set it to None as default is better.

nni/experiment/config/convert.py
liuzhe-lz: https://github.com/microsoft/nni/blob/master/nni/tools/nnictl/launcher.py#L133
ADL does not use cluster metadata.

nni/experiment/config/convert.py
liuzhe-lz: mistake

nni/experiment/config/convert.py
liuzhe-lz: Do you mean there are cases that OpenPAI cluster supports only HTTP and not HTTPS?

nni/experiment/config/remote.py
liuzhe-lz: As said in doc, the behavior is:
  - when password is explicitly set, use password
  - when password is not set and identity file is explicitly set, use specified identify file
  - when both are not set, use default identity file

I expect this behavior because when using openssh directly, I don't need to specify anything when I put identify file at `~/.ssh/id_rsa`.
I know it can be other algorithms like `id_edXXX` though, but I believe they are rarely used.

docs/en_US/reference/experiment_config.rst
liuzhe-lz: I can't find a doc mention built-in tuner and custom tuner at same time.

docs/en_US/reference/experiment_config.rst
liuzhe-lz: So it will be `openpaiConfig` and `openpai_config` in yaml and python respectively. I think it looks weird. Discuss it in meeting.

docs/en_US/reference/experiment_config.rst
liuzhe-lz: No change from v1.

ultmaster (approved):

nni/experiment/config/convert.py
```diff
@@ -76,8 +78,9 @@ def to_v1_yaml(config: ExperimentConfig, skip_nnictl: bool = False) -> Dict[str,
         data['remoteConfig'] = {'reuse': ts['reuseMode']}
         data['machineList'] = []
         for machine in ts['machineList']:
-            machine = {
+            data['machineList'].append({
                 'ip': machine['host'],
+                'port': machine['port'],
```
SparkSnail, line 83: If users didn't set port, it will be `'port': None` here?

SparkSnail (approved):

PR #3134, opened 2020-11-28
Refactor builtin/customized tuner installation

chicm-ms:
register all builtin tuners/assessors/advisors into `deployment/registered_algorithms.yml` and copy it into nni config dir in `setup.py`
nni config dir is the following suggested path:

|  Environment   |     "data_files" Path      | Suggested Path |
| -------------- | -------------------------- | -------------- |
| Linux/global   | /usr/nni                   | ~/.config/nni  |
| Linux/user     | ~/.local/nni               | ~/.config/nni  |
| Linux/conda    | (env_root)/nni             | (env_root)/nni |
| Windows/global | Program Files/Python38/nni | %appdata%/nni  |
| Windows/user   | %appdata%/python/nni       | %appdata%/nni  |
| Windows/conda  | (env_root)/nni             | (env_root)/nni |

`nnictl package` command is replaced with `nnictl algo`

TODO:
- [x] `extras_require` for `SMAC`
- [x] SMAC/BOHB/PPOTuner dependencies install.
- [x] SMAC/BOHB/PPOTuner import check.
- [x] SMAC/BOHB/PPOTuner doc
- [x] nni config dir for windows.

PR #3139, opened 2020-11-30
Add algorithms importing shortcut

liuzhe-lz:


ultmaster:
What is the status of IDE/plugin support for shortcuts?

ultmaster:
Pending #3134 as the installation dependency of tuners are broken right now.

I managed to import `nni.nas.pytorch` with a few additional hacks, but failed to import `nni.nas.pytorch.darts`. It seems that the module that is injected with `setattr` cannot appear in the import part. Need @liuzhe-lz 's help.

PR #3148, opened 2020-12-02
Add quantized model export description.

linbinskn:
Add and example of saving and loading quantized model in QAT.

### DIFF
--- a/docs/en_US/Compression/QuickStart.md
+++ b/docs/en_US/Compression/QuickStart.md
@@ -197,16 +197,22 @@ pruner.export_model(model_path='model.pth', mask_path='mask.pth', onnx_path='mod
 
 #### Export Quantized Model
 
-You can export the quantized model directly by using ```torch.save``` api and the quantized model can be loaded by ```torch.load``` without any extra modification. The following example shows the normal proceduce of saving and loading quantized model in QAT.
+You can export the quantized model directly by using ```torch.save``` api and the quantized model can be loaded by ```torch.load``` without any extra modification. The following example shows the normal proceduce of saving, loading quantized model and get related parameters in QAT.
 ```
 # save quantized model
 torch.save(model.state_dict(), "quantized_model.pkt")
 
 # simulate model loading procedure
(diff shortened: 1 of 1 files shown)


docs/en_US/Compression/QuickStart.md
```diff
@@ -193,4 +195,18 @@ pruner.export_model(model_path='model.pth')
 pruner.export_model(model_path='model.pth', mask_path='mask.pth', onnx_path='model.onnx', input_shape=[1, 1, 28, 28])
 ```
 
+#### Export Quantized Model
+
+You can export the quantized model directly by using ```torch.save``` api and the quantized model can be loaded by ```torch.load``` without any extra modification. The following example shows the normal proceduce of saving and loading quantized model in QAT.
```
QuanluZhang, line 200: better to briefly explain where is quantization related data saved, for example, scale, zero point etc.

docs/en_US/Compression/QuickStart.md
```diff
@@ -193,4 +195,24 @@ pruner.export_model(model_path='model.pth')
 pruner.export_model(model_path='model.pth', mask_path='mask.pth', onnx_path='model.onnx', input_shape=[1, 1, 28, 28])
 ```
 
+#### Export Quantized Model
+
+You can export the quantized model directly by using ```torch.save``` api and the quantized model can be loaded by ```torch.load``` without any extra modification. The following example shows the normal proceduce of saving, loading quantized model and get related parameters in QAT.
+```
+# save quantized model
+torch.save(model.state_dict(), "quantized_model.pkt")
```
QuanluZhang, line 203: it is not clear what the relation between `model` and `model_1` below...

linbinskn: Have changed model_1 to qmodel_load and add related annotation.

docs/en_US/Compression/QuickStart.md
linbinskn: Agree.

PR #3156, opened 2020-12-04
Fix numpy version for Python 3.6

ultmaster:


J-shang (approved):

setup.py
```diff
@@ -77,6 +77,8 @@
     'prettytable'
 ]
 
+if sys.version_info < (3, 7):
+    dependencies[dependencies.index('numpy')] = 'numpy<1.20'
```
J-shang, line 81: If we need a limited numpy version anyway, maybe `'numpy<1.19.4'` is better? Due to `'numpy >= 1.19.4'` have compatibility issue in windows. https://tinyurl.com/y3dm3h86

SparkSnail (approved):

setup.py
```diff
@@ -77,6 +77,10 @@
     'prettytable'
 ]
 
+if sys.platform == 'win32':
```
liuzhe-lz, line 80: This is not the right way. It should depend on the machine installing the wheel, not the machine building the wheel.
It should use [environment marker](https://www.python.org/dev/peps/pep-0508/#environment-markers)

PR #3157, opened 2020-12-04
fix /version not handle error

J-shang:
previous:
```
(node:6506) UnhandledPromiseRejectionWarning: Error: Cannot find module '/home/xxx/nni/ts/nni_manager/dist/package.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Promise.resolve.then (/home/xxx/nni/ts/nni_manager/dist/common/utils.js:276:34)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:6506) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:6506) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

now:
```
[12/4/2020, 6:38:07 PM] ERROR [ { Error: Cannot find module '/home/xxx/nni/ts/nni_manager/dist/package.json'
      at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
      at Function.Module._load (internal/modules/cjs/loader.js:562:25)
      at Module.require (internal/modules/cjs/loader.js:692:17)
      at require (internal/modules/cjs/helpers.js:25:18)
      at Promise.resolve.then (/home/xxx/nni/ts/nni_manager/dist/common/utils.js:276:34) code: 'MODULE_NOT_FOUND' } ]
```

```
async function getVersion(): Promise<string> {
    const deferred: Deferred<string> = new Deferred<string>();
    import(path.join(__dirname, '..', 'package.json')).then((pkg) => {
        deferred.resolve(pkg.version);
    }).catch((error) => {
        deferred.reject(error);
    });
    return deferred.promise;
}
```

**Need Discuss**
`develop` in Linux use `symlink`, and use `__dirname` in `getVersion()`.
`package.json` is in `/home/xxx/nni/ts/nni_manager` not `/home/xxx/nni/ts/nni_manager/dist`. Do we need make a copy of `package.json` in `dist`?

### DIFF
--- a/ts/nni_manager/rest_server/restHandler.ts
+++ b/ts/nni_manager/rest_server/restHandler.ts
@@ -98,8 +98,11 @@ class NNIRestHandler {
 
     private version(router: Router): void {
         router.get('/version', async (req: Request, res: Response) => {
-            const version = await getVersion();
-            res.send(version);
+            getVersion().then((version) => {
+                res.send(version)
+            }).catch((err: Error) => {
+                this.handleError(err, res);
+            });
         });
     }
 


ts/nni_manager/rest_server/restHandler.ts
```diff
@@ -98,8 +98,11 @@ class NNIRestHandler {
 
     private version(router: Router): void {
         router.get('/version', async (req: Request, res: Response) => {
-            const version = await getVersion();
-            res.send(version);
+            getVersion().then((version) => {
+                res.send(version)
+            }).catch((err: Error) => {
```
SparkSnail, line 103: This getVersion() function is used in trainingService, should fix package.json issue to get correct version number, not just catch error here.

qfyin: A note from the discussion:
Find a standard error handler as the fallback for all REST APIs.
Specific errors can still be handled at first place.

setup_ts.py
```diff
@@ -144,6 +144,7 @@ def compile_ts():
     # todo: I don't think these should be here
     shutil.rmtree('ts/nni_manager/dist/config', ignore_errors=True)
     shutil.copytree('ts/nni_manager/config', 'ts/nni_manager/dist/config')
+    shutil.copy('ts/nni_manager/package.json', 'ts/nni_manager/dist/package.json')
```
liuzhe-lz, line 147: I don't think this should be here. Maybe we can do in `copy_nni_node()`.
Web UI and NAS UI does not copy `package.json` to there compiling output.
And in release mode this `package.json` is not used directly. The version string needs modification.

J-shang: `copy_nni_node()` already has copied `package.json`, should we do this in `symlink_nni_node()`? or use `copy_nni_node()` for all situation in `build()`?

liuzhe-lz: Oh, so the problem is node.js only accepts regular file package.json, not symlinked package.json?

J-shang: the problem is `__dirname` is the original location of the file, not symlinked location.

Issue #3159, opened 2020-12-06
QuantGrad computation  in quantization is not correct


eedalong:
QuantGrad use STE for grad back propagation. STE, according to google's whitepaper, is computed by ![](http://119.3.255.23:8001/issue.jpg), but in current nni's implementation, its just quantize_grad = output_grad.

eedalong:
I'll fix this in next pr

linbinskn:
Thank you for your issue and PR.
This change may not be necessary because the grad_output is the grad of quantized weight(**W_out**) and all of the **W_out** have been in the range of **Wmin and Wmax** due to clamping during quantizing. Have you tested whether this change will bring extra advantages such as improving final accuracy or speeding up the model convergence during training or not?

eedalong:
um

eedalong:
The problem is solved, and pr is merged. So I will close this issue. Cheers Guys!

PR #3160, opened 2020-12-06
fix quant grad function calculation error

eedalong:
#3159 

Quant grad calculation is not correct in current nni's implementation according to [google's whitepaper](https://arxiv.org/pdf/1806.08342.pdf), it should be

### DIFF
--- a/nni/compression/pytorch/compressor.py
+++ b/nni/compression/pytorch/compressor.py
@@ -590,10 +590,19 @@ class QuantGrad(torch.autograd.Function):
     Base class for overriding backward function of quantization operation.
     """
     @classmethod
-    def _quantize(cls, x, scale, zp):
-        r"""Reference function for quantizing x -- non-clamped.
+    def _quantize(cls, x, scale, zero_point):
         """
-        return ((x / scale) + zp).round()
+        Reference function for quantizing x -- non-clamped.
+        Parameters
+        ----------
+        x : Tensor
+            tensor to be quantized
+        scale : Tensor
+            scale for quantizing x
+        zero_point : Tensor
+            zero_point for quantizing x
+        """
+        return ((x / scale) + zero_point).round()
     @classmethod
     def get_bits_length(cls, config, quant_type):
         """
@@ -618,9 +627,15 @@ class QuantGrad(torch.autograd.Function):
             input of quantization operation
         grad_output : Tensor
             gradient of the output of quantization operation
-        quant_type : QuantType
+        scale : Tensor
             the type of quantization, it can be `QuantType.QUANT_INPUT`, `QuantType.QUANT_WEIGHT`, `QuantType.QUANT_OUTPUT`,
             you can define different behavior for different types.
+        zero_point : Tensor
+            zero_point for quantizing tensor
+        qmin : Tensor
+            quant_min for quantizing tensor
+        qmax : Tensor
+            quant_max for quantizng tensor
         Returns
         -------
         tensor
--- a/nni/compression/pytorch/compressor.py
+++ b/nni/compression/pytorch/compressor.py
@@ -589,8 +589,26 @@ class QuantGrad(torch.autograd.Function):
     """
     Base class for overriding backward function of quantization operation.
     """
+    @classmethod
+    def _quantize(cls, x, scale, zp):
+        r"""Reference function for quantizing x -- non-clamped.
+        """
+        return ((x / scale) + zp).round()
+    @classmethod
+    def get_bits_length(cls, config, quant_type):
+        """
+        get bit for quantize config
+        :param config:
+        :param quant_type:
+        :return:
+        """
+        if isinstance(config["quant_bits"], int):
+            return config["quant_bits"]
+        else:
+            return config["quant_bits"].get(quant_type)
+
     @staticmethod
-    def quant_backward(tensor, grad_output, quant_type):
+    def quant_backward(tensor, grad_output, scale, zero_point, qmin, qmax):
         """
         This method should be overrided by subclass to provide customized backward function,
         default implementation is Straight-Through Estimator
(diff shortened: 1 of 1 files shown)


nni/compression/pytorch/compressor.py
```diff
@@ -608,24 +626,35 @@ def quant_backward(tensor, grad_output, quant_type):
         tensor
             gradient of the input of quantization operation
         """
-        return grad_output
+        tensor_q = QuantGrad._quantize(tensor, scale, zero_point)
+        mask = (tensor_q >= qmin) * (tensor_q <= qmax)
```
Cjkkkk, line 630: mask = (tensor_q < qmin) | (tensor_q > qmax)
grad_output[mask] = 0
return grad_output

nni/compression/pytorch/compressor.py
```diff
@@ -608,24 +626,35 @@ def quant_backward(tensor, grad_output, quant_type):
...
         elif quant_type == QuantType.QUANT_OUTPUT:
-            return wrapper.quantizer.quantize_output(tensor, wrapper, **kwargs)
+            quant_type = 'output'
+            output = wrapper.quantizer.quantize_output(tensor, wrapper, **kwargs)
         else:
             raise ValueError("unrecognized QuantType.")
 
+        bits = QuantGrad.get_bits_length(wrapper.config, quant_type)
+        qmin, qmax = torch.Tensor([0], device=tensor.device), torch.Tensor([(1 << bits) - 1], device=tensor.device)
+        ctx.save_for_backward(tensor, wrapper.module.scale, wrapper.module.zero_point, qmin, qmax)
```
Cjkkkk, line 651: you only need save mask here if you want to do clip gradient, which is a bytetensor and therefore 4 times smaller than saving a floattensor.

Cjkkkk: I think the reason why we used to use a pass-through-estimator is that we always set rmin, rmax to be the min(tensor), max(tensor). But your propose is correct if rmin is larger than the min(tensor) and rmax is smaller than the max(tensor).

nni/compression/pytorch/compressor.py
```diff
@@ -608,24 +626,35 @@ def quant_backward(tensor, grad_output, quant_type):
         tensor
             gradient of the input of quantization operation
         """
-        return grad_output
+        tensor_q = QuantGrad._quantize(tensor, scale, zero_point)
```
Cjkkkk, line 629: No need to calculate` tensor_q here`, return a mask from `wrapper.quantizer.quantize_input` for example.

nni/compression/pytorch/compressor.py
eedalong: yes， for weight， rmin and rmax is always the min and max of the tensor, but think about output. Because we use ema for updating rmin and rmax, so not all item in output is in [rmin, rmax]

nni/compression/pytorch/compressor.py
eedalong: No, we need unclampped quantized data, wrapper.quantizer.quantize_input is clampped to [qmin, qmax], so the mask will always be ones everywhere

eedalong:
check pytorch's implmentation guys:

nni/compression/pytorch/compressor.py
eedalong: good point

nni/compression/pytorch/compressor.py
Cjkkkk: Yes, your propose is necessary in quantizing ouput, thanks for submitting this PR!

nni/compression/pytorch/compressor.py
Cjkkkk: I mean you can do this inside the `wrapper.quantizer.quantize_input` and return the mask which can be used in backward.

Cjkkkk: Therefore you don't do quantization twice and tensor saved will be smaller by 4 times.

eedalong: Yeah, u are right. I did write the code like that.. For doing this ,we need to modify _quantize() to return mask. But i give it up because i dont think it's a good design. Just let _quantize() to return quantized_weight is a better design i think.

Cjkkkk: True, but keeping this design will actually make many quantization algorithms with special backward function harder and less efficient to implement. The actual problem is quant backward is inside the autograd while the forward is inside the quantizer module. I think it will be much better if forward and backward logic is moved into autograd function together and quantizer only maintain some states.

Cjkkkk: What I said is actually what Pytorch implementation does.

nni/compression/pytorch/compressor.py
```diff
@@ -608,24 +626,35 @@ def quant_backward(tensor, grad_output, quant_type):
         tensor
             gradient of the input of quantization operation
         """
```
J-shang, line 628: Comment of this function need to be updated or delete.

eedalong: yeah, u r right.... . I'll update it directly.

nni/compression/pytorch/compressor.py
```diff
@@ -608,24 +626,34 @@ def quant_backward(tensor, grad_output, quant_type):
...
+        tensor_q = QuantGrad._quantize(tensor, scale, zero_point)
+        mask = (tensor_q < qmin) | (tensor_q > qmax)
+        grad_output[mask] = 0
         return grad_output
 
     @staticmethod
     def forward(ctx, tensor, quant_type, wrapper, **kwargs):
-        ctx.save_for_backward(tensor, torch.Tensor([quant_type]))
         if quant_type == QuantType.QUANT_INPUT:
```
linbinskn, line 636: Suggest changing defination of QuantType to 'input', 'weight' and 'output'.

nni/compression/pytorch/compressor.py
```diff
@@ -589,8 +589,26 @@ class QuantGrad(torch.autograd.Function):
...
+        r"""Reference function for quantizing x -- non-clamped.
+        """
+        return ((x / scale) + zp).round()
+    @classmethod
+    def get_bits_length(cls, config, quant_type):
+        """
+        get bit for quantize config
+        :param config:
+        :param quant_type:
+        :return:
```
QuanluZhang, line 603: please update docstring to be consistent with others

nni/compression/pytorch/compressor.py
```diff
@@ -589,8 +589,26 @@ class QuantGrad(torch.autograd.Function):
     """
     Base class for overriding backward function of quantization operation.
     """
+    @classmethod
+    def _quantize(cls, x, scale, zp):
```
linbinskn, line 593: Suggest changing zp to zero_point to keep consensus.

nni/compression/pytorch/compressor.py
eedalong: Um.., I dont think it is relavant with this PR. May be you need to update all definition for quant type in another pr

nni/compression/pytorch/compressor.py
eedalong: yeah, changed already

nni/compression/pytorch/compressor.py
eedalong: cool

nni/compression/pytorch/compressor.py
eedalong: Yes u r right. And this calls for a HUGE design change with current implementation.

eedalong: And since we've talked about this. I think the current design indeed need to be changed.
1. I agree a lot on what u said that move forward and backward into QuantGrad, and Quantizer only matain the status

eedalong: for now, i think i will just push the correct STE code and what u mentioned is left for next pr~

nni/compression/pytorch/compressor.py
QuanluZhang: could you help update line 583-585 to update the values from 0, 1, 2 to 'input', 'weight', 'output'? we have checked the code, it is safe to make this modification :)

nni/compression/pytorch/compressor.py
QuanluZhang: @eedalong @Cjkkkk , thanks for the discussion. The design of this pr looks reasonable. let's keep it in this pr. We will do some refactor later to put forward logic and backward logic together within the same class, which should be much easier for users to customize new quantization algorithms.

QuanluZhang (approved):

Cjkkkk (approved):

nni/compression/pytorch/compressor.py
eedalong: Ok～ 👌

J-shang (approved):

linbinskn (approved):

PR #3163, opened 2020-12-07
Fix check gpu index logic in local mode

SparkSnail:
In local training service, NNI will check if the gpu index is valid. But there is a kind of scene that gpu_metric file is not generated yet, the gpuCount number equals to 0, but check gpu index will cause experiment failed.


ts/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -436,7 +436,7 @@ class LocalTrainingService implements TrainingService {
 
     private checkSpecifiedGpuIndices(): void {
         const gpuCount: number = this.gpuScheduler.getSystemGpuCount();
-        if (this.designatedGpuIndices !== undefined) {
+        if (this.designatedGpuIndices !== undefined && gpuCount > 0) {
```
liuzhe-lz, line 439: What will happen if gpu count IS 0?

SparkSnail: Only detect gpu count when user set gpuNum>0 in config file. If gpu count is 0 in the machine, the trial job will always be waitting. We could not guarantee the detected 0 is a valid number in gpuScheduler component, then should not validate gpuIndices when count is 0.

liuzhe-lz: I suggest to return undefined or null before first detection. `getSystemGpuCount()` is only used here so I guess it won't be too much effort.

SparkSnail: agree, updated.

liuzhe-lz (approved):

J-shang (approved):

PR #3172, opened 2020-12-10
Support Custom Kubernetes Namespace in AdaptDL Mode

hao-howard-zhang:
This PR adds the support for launching trials in custom K8s namespace in AdaptDL mode. I have tested my change in our internal training cluster.

Note that other K8s based training services (Kubeflow, framework controller, etc.) still hard-code the `default` namespace. In long term, I think it is worth supporting this feature in all K8s training services because creating multiple namespaces in the K8s cluster is a common practice.

@SparkSnail @liuzhe-lz Please review when you have time. This should be the last feature we want to add to the AdaptDL integration before the 2.0 release.


hao-howard-zhang:
@SparkSnail Could you close this one and review https://github.com/microsoft/nni/pull/3176 instead? Git rebase messed up with this PR...

PR #3173, opened 2020-12-10
change SIGKILL to SIGTERM in local mode cancel trial job

J-shang:


ts/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -253,7 +253,14 @@ class LocalTrainingService implements TrainingService {
...
             return Promise.resolve();
         }
-        tkill(trialJob.pid, 'SIGKILL');
+        tkill(trialJob.pid, 'SIGTERM');
+        const pid = trialJob.pid;
+        setTimeout((pid: number) => {
+            if (isAlive(pid)) {
+                tkill(pid, 'SIGKILL');
+            }
+        }, 5 * 1000, pid);
```
QuanluZhang, line 262: with this solution, we have to wait for 5 seconds even the trail is stopped within 1 second?

J-shang: no, `setTimeout()` will not block the function, just put callback into the queue after delay.

ts/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -253,7 +253,14 @@ class LocalTrainingService implements TrainingService {
 
             return Promise.resolve();
         }
-        tkill(trialJob.pid, 'SIGKILL');
+        tkill(trialJob.pid, 'SIGTERM');
+        const pid = trialJob.pid;
+        setTimeout((pid: number) => {
+            if (isAlive(pid)) {
```
liuzhe-lz, line 259: This if check is not atomic.
Please call tkill can catch exception.

J-shang: fixed it.

ts/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -253,7 +253,16 @@ class LocalTrainingService implements TrainingService {
 
             return Promise.resolve();
         }
-        tkill(trialJob.pid, 'SIGKILL');
+        tkill(trialJob.pid, 'SIGTERM');
+        const pid = trialJob.pid;
+        setTimeout(((pid: number): void => {
+            tkill(pid, 'SIGKILL', (err) => {
+                if (err){
+                    this.log.warning(`cancel trial job {pid: ${pid}} failed: ${err.message}`);
```
liuzhe-lz, line 261: I suddenly thought of a problem.
If the pid get reused between `SIGTERM` and `SIGKILL`, this will kill a random process.
If you don't want to elegantly solve the problem in this release, I suggest to reduce the delay to minimize risk.

J-shang: Yes, and other place use `kill pid` also has this risk, maybe elegant way is let them kill themselves, need discuss. This release reduce the delay to 1 second.

liuzhe-lz (approved):

maxxx580:
is this change targeted. for v2.0 ?

J-shang:
> is this change targeted. for v2.0 ?

yes

ts/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -253,7 +253,16 @@ class LocalTrainingService implements TrainingService {
 
             return Promise.resolve();
         }
-        tkill(trialJob.pid, 'SIGKILL');
+        tkill(trialJob.pid, 'SIGTERM');
+        let waitingTime = 0;
+        while(await isAlive(trialJob.pid)) {    
+            if (waitingTime > 4999) {
+                tkill(trialJob.pid, 'SIGKILL');
```
QuanluZhang, line 260: how about add `break` here, in case pid is not terminated even with sigkill?

J-shang: `SIGKILL` can not kill process is almost impossible, if it is happened, I think there may have a serious system error. I will add `catch error` and `break` in this case.

QuanluZhang (approved):

PR #3181, opened 2020-12-11
remove 'port' when experiment stopped

J-shang:


J-shang:
> Why remove port field? Which port will be used when users `resume` or `view` a stopped experiment?

Because the original port information is out of date when the experiment is `stopped`. `resume` and `view` use `--port` to set port, and by default is `8080`.

liuzhe-lz (approved):

SparkSnail (approved):

Issue #3151, opened 2020-12-02
ModuleNotFoundError: No module named 'nni.algorithms'


JiaojiaoYe1994:
**Environment**:
- NNI version: 1.9
- NNI mode (local|remote|pai): local
- Client OS: 
- Server OS (for remote mode only):
- Python version: 3.6
- PyTorch/TensorFlow version: 1.4.0
- Is conda/virtualenv/venv used?: yes
- Is running in Docker?: no

I met this error after I update my PyTorch, no idea what happens. Is there someone has the same problem or any idea?

JiaojiaoYe1994:
> **Environment**:
> 
> * NNI version: 1.9
> * NNI mode (local|remote|pai): local
> * Client OS:
> * Server OS (for remote mode only):
> * Python version: 3.6
> * PyTorch/TensorFlow version: 1.4.0
> * Is conda/virtualenv/venv used?: yes
> * Is running in Docker?: no
> 
> I met this error after I update my PyTorch, no idea what happens. Is there someone has the same problem or any idea?

 NNI python API has changed. Algorithms module has been removed and we need to import nas diectly from NNI module, it is necessary to update the example scripts to error.

For example,

`from nni.algorithms.nas.pytorch.darts import DartsTrainer`

should be 

`from nni.nas.pytorch.darts import DartsTrainer`

PR #3191, opened 2020-12-14
fix issue #3151

mlbo:
ModuleNotFoundError: No module named 'nni.algorithms' #3151


liuzhe-lz:
You are running v2.0 example with v1.9 SDK.
Please checkout v1.9 branch or tag.

PR #3194, opened 2020-12-15
Prune archive directories

ultmaster:


QuanluZhang (approved):

liuzhe-lz:
I think we need to further investigate which files should ported back. (dockerfile doc, bash completion, etc)
But I'm okay with this PR since git retains the history.

liuzhe-lz (approved):

Tudor33:
😍😍😍 more more Its Věry good

Dne út 15. 12. 2020 6:51 uživatel Yuge Zhang <notifications@github.com>
napsal:

PR #3202, opened 2020-12-16
[Retiarii] cross-graph optimization: device placement and input deduplication

hzhua:


hzhua:
test_cgo_engine and test_dedup_input have passed the tests

nni/retiarii/execution/cgo_engine.py
```diff
@@ -79,12 +81,14 @@ def register_graph_listener(self, listener: AbstractGraphListener) -> None:
         self._listeners.append(listener)
 
     def _send_trial_callback(self, paramater: dict) -> None:
-        for listener in self._listeners:
-            listener.on_resource_used(0)  # FIXME: find the real resource id
+        if self.resources <= 0:
+            _logger.warning('There is no available resource, but trial is submitted.')
+        self.resources -= 1
```
QuanluZhang, line 86: `self.resources` is supposed to manage resource, in dedup case, one trial still consumes one gpu?

nni/retiarii/integration.py
```diff
@@ -132,7 +132,8 @@ def handle_report_metric_data(self, data):
 
     @staticmethod
     def _process_value(value) -> Any:  # hopefully a float
-        value = json_tricks.loads(value)
+        if isinstance(value, str):
```
QuanluZhang, line 135: here `value` must be `str`

nni/retiarii/execution/cgo_engine.py
hzhua: Fixed. One trial may consume multiple GPUs. It is fixed to consume how many GPUs a trial uses.

nni/retiarii/integration.py
hzhua: OK. In the test, it was a dict. I have converted it to str in the test. I also change L135 to "assert isintance(value, str)".

nni/retiarii/execution/cgo_engine.py
```diff
@@ -143,8 +148,14 @@ def trial_execute_graph(cls) -> None:
...
 class AssemblePolicy:
     @staticmethod
-    def group(logical_plan):
+    def group(logical_plan, n_model_per_graph):
+        #TODO: Packing multiple model in one GPU
+        # Currently, we only support one model per GPU
+        all_grouped_models = []
         group_model = {}
         for idx, m in enumerate(logical_plan.models):
             group_model[m] = PhysicalDevice('server', f'cuda:{idx}')
```
QuanluZhang, line 157: there is a bug here, users can submit many models, then logical plan merge them together. You enumerate them here and assign gpu with model's `idx`, it is possible that `idx` is out of range

hzhua: Fixed. I add "available_devices" in CGOExecutionEngine. I also change CGO in the environment variable to CGO_DEVICES to allow users to set available devices.

nni/retiarii/execution/cgo_engine.py
```diff
@@ -79,12 +81,15 @@ def register_graph_listener(self, listener: AbstractGraphListener) -> None:
         self._listeners.append(listener)
 
     def _send_trial_callback(self, paramater: dict) -> None:
-        for listener in self._listeners:
-            listener.on_resource_used(0)  # FIXME: find the real resource id
+        if self.resources <= 0:
+            _logger.warning('There is no available resource, but trial is submitted.')
+        print(paramater)
+        self.resources -= paramater['training_kwargs']['n_model']
```
QuanluZhang, line 87: so for now you have not considered the case that multiple model are packed on one GPU?

QuanluZhang: see your comments

nni/retiarii/execution/cgo_engine.py
```diff
@@ -14,15 +14,17 @@
 
 
 class CGOExecutionEngine(AbstractExecutionEngine):
-    def __init__(self, n_model_per_graph=4) -> None:
+    def __init__(self, available_devices = None) -> None:
```
QuanluZhang, line 17: please write docstring for classes and public member functions

nni/retiarii/execution/cgo_engine.py
```diff
@@ -143,8 +148,15 @@ def trial_execute_graph(cls) -> None:
...
         group_model = {}
+        assert(len(available_devices) > 0) # There should be at least 1 device, set in CGO_DEVICES
         for idx, m in enumerate(logical_plan.models):
-            group_model[m] = PhysicalDevice('server', f'cuda:{idx}')
-        return [group_model]
+            group_model[m] = PhysicalDevice('server', available_devices[idx % len(available_devices)])
+            if len(group_model) == len(available_devices) or idx == len(logical_plan.models)-1:
+                all_grouped_models.append(group_model)
+                group_model = {}
+        return all_grouped_models
```
QuanluZhang, line 162: seems this logic is not correct. first, it can only be applied on local mode with a single machine. second, even for a single machine, the gpu resource cannot be fully utilized, for example, there are 4 gpus, logical_plan has 3 gpu, you assign 0/1/2 for them, then a new model comes, you will assign No. 0 gpu, right?

hzhua: You are right. I assumed the models arrive and execute in batch. So the new model should arrive after the previous model finishes.

I will fix it to remove this assumption, so that two models in different batch can be executed together.

test/ut/retiarii/test_dedup_input.py
```diff
@@ -51,32 +51,26 @@ def _test_add_model(self):
...
         lp, models = self._build_logical_with_mnist(3)
         opt = DedupInputOptimizer()
         opt.convert(lp)
-        with open('dedup_logical_graph.json', 'r') as fp:
-            correct_dump = fp.readlines()
-        lp_dump = lp.logical_graph._dump()
-
-        self.assertTrue(correct_dump[0] == json.dumps(lp_dump))
-
+        # TODO: topo_sort may not be stable that leads to different dump. skip
```
QuanluZhang, line 58: could you help find out the reason?

hzhua: It was because the previous implementation of topo_sort does not have slots. The edges can be in any order.
The current implementation may not be stable either, if the graph.nodes can be stored in any order.
A safe implementation is to use a priority queue to store the nodes with a degree of 0, then always pick the one with the smallest ID/name.

lastthyme (approved):

nni/retiarii/evaluator/pytorch/cgo/evaluator.py
```diff
@@ -0,0 +1,210 @@
...
+class _ClassificationModule(MultiModelSupervisedLearningModule):
+    def __init__(self, criterion: nn.Module = nn.CrossEntropyLoss,
+                 learning_rate: float = 0.001,
+                 weight_decay: float = 0.,
+                 optimizer: optim.Optimizer = optim.Adam):
+        super().__init__(criterion, {'acc': _AccuracyWithLogits},
+                         learning_rate=learning_rate, weight_decay=weight_decay, optimizer=optimizer)
+
+
+class Classification(Lightning):
```
ultmaster, line 127: Is it possible to inherit `Classification` and `CgoMixin`, to avoid copy all the code.

nni/retiarii/experiment/pytorch.py
```diff
@@ -3,6 +3,7 @@
 
 import atexit
 import logging
+from nni.retiarii.execution.logical_optimizer.logical_plan import PhysicalDevice
```
ultmaster, line 6: Suggest add shortcut to simplify import

.gitignore
```diff
@@ -101,3 +101,4 @@ venv.bak/
 .history
 generated/
 test/ut/retiarii/_debug_graph_data.json
+data/cifar10/*
```
liuzhe-lz, line 104: Suggest use full path

nni/retiarii/evaluator/pytorch/cgo/evaluator.py
hzhua: It seems to be hard. Most lines of code of `Classification` and `Regression` are documentation. The two lines of code in `__init__` are both customized for CGO.

ultmaster (approved):
It seems that most changes are self-contained and doesn't overlap with other modules. I suggest we merge this PR (as it's large enough) and we can improve later.

hzhua:
Since this PR depends on the modification of training service #3897 , I prefer not merging this PR before #3897 is merged.

QuanluZhang (approved):

PR #3207, opened 2020-12-17
fix file lock bug

J-shang:


nni/tools/nnictl/common_utils.py
```diff
@@ -130,4 +130,4 @@ def _release(self):
         return None
 
 def get_file_lock(path: string, stale=-1):
-    return SimplePreemptiveLock(path + '.lock', stale=-1)
+    return SimplePreemptiveLock(path + '.lock', stale=stale)
```
QuanluZhang, line 133: have you checked that if you manually create a lock file, does it block new experiment?

J-shang: yes, it does not block new experiment. The previous situation was due to always `stale=-1`, means lock never expires, now fix it.

liuzhe-lz (approved):

QuanluZhang (approved):

PR #3210, opened 2020-12-18
add experiment info when use python launch

J-shang:


nni/experiment/launcher.py
```diff
@@ -64,7 +67,7 @@ def _ensure_port_idle(port: int, message: Optional[str] = None) -> None:
         raise RuntimeError(f'Port {port} is not idle {message}')
 
 
-def _start_rest_server(config: ExperimentConfig, port: int, debug: bool, experiment_id: str, pipe_path: str) -> Popen:
+def _start_rest_server(config: ExperimentConfig, port: int, debug: bool, experiment_id: str, pipe_path: str) -> [int, Popen]:
```
liuzhe-lz, line 70: The type hint should be `Tuple[int, Popen]`

liuzhe-lz (approved):

nni/experiment/launcher.py
J-shang: fix it

QuanluZhang (approved):

PR #3218, opened 2020-12-22
detached tuner process under windows

J-shang:
Fix dispatcher exit before rest main process clean up. This issue due to CTRL BREAK or CTRL C is sent to all console processes that are attached to the console in Windows.

### DIFF
--- a/nni/experiment/experiment.py
+++ b/nni/experiment/experiment.py
@@ -169,7 +169,7 @@ class Experiment:
             while True:
                 time.sleep(10)
                 status = self.get_status()
-                if status == 'STOPPED':
+                if status == 'DONE' or status == 'STOPPED':
                     return True
                 if status == 'ERROR':
                     return False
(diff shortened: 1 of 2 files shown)


nni/experiment/experiment.py
```diff
@@ -169,7 +169,7 @@ def run(self, port: int = 8080, debug: bool = False) -> bool:
             while True:
                 time.sleep(10)
                 status = self.get_status()
-                if status == 'STOPPED':
+                if status == 'DONE' or status == 'STOPPED':
```
SparkSnail, line 172: Why add `DONE`? What is the behavior of `SUCCEED` or `FAILED`?

J-shang: It is the status of experiment, not trials. And we expect this process automatic exit when the experiment is finished. You reminded me maybe we should also add `ERROR`?

SparkSnail: Seems `ERROR` already exist in line 174.

SparkSnail (approved):

nni/experiment/experiment.py
J-shang: Oh yes, I miss it.

liuzhe-lz (approved):

PR #3220, opened 2020-12-22
Support mixed-precision quantization speed up by using tensorrt

linbinskn:
This PR is used to support real mixed-precision quantization speed up in hardware by using tensorrt.


nni/compression/pytorch/speedup/quantization_speedup/integrated_tensorrt.py
```diff
@@ -0,0 +1,134 @@
...
+        model_onnx = fonnx.torch_to_onnx(self.model, input_shape=self.input_shape, model_path=self.onnx_path, input_names=self.input_names, output_names=self.output_names)
+
+        if self.using_calibrate:
+            assert self.calibrate_type is not None
+            context = self.tensorrt_build_withcalib(self.onnx_path)
+        else:
+            try:
+                raise NameError('quantized without calibrate has not been supported.')
+            except NameError:
+                raise
```
J-shang, line 108: why use `try except` how about directly `raise` or just `assert`?

nni/compression/pytorch/speedup/quantization_speedup/integrated_tensorrt.py
```diff
@@ -0,0 +1,134 @@
...
+        inputs, outputs, bindings, stream = common.allocate_buffers(self.context.engine)
+        result = []
+        for start_idx in range(0, test_data.shape[0], self.batchsize):
+            # If the number of images in the test set is not divisible by the batch size, the last batch will be smaller.
+            # This logic is used for handling that case.
+            end_idx = min(start_idx + self.batchsize, test_data.shape[0])
+            effective_batch_size = end_idx - start_idx
+
+            # Do inference for every batch.
+            inputs[0].host = test_data[start_idx:start_idx + effective_batch_size]
```
J-shang, line 127: only change `inputs[0]`? Does `len(inputs) == 1` in this case?

nni/compression/pytorch/speedup/quantization_speedup/common.py
```diff
@@ -0,0 +1,85 @@
...
+        if engine.binding_is_input(binding):
+            inputs.append(HostDeviceMem(host_mem, device_mem))
+        else:
+            outputs.append(HostDeviceMem(host_mem, device_mem))
+    return inputs, outputs, bindings, stream
+
+# This function is generalized for multiple inputs/outputs.
+# inputs and outputs are expected to be lists of HostDeviceMem objects.
+def do_inference(context, bindings, inputs, outputs, stream, batch_size=1):
+    # Transfer input data to the GPU.
```
J-shang, line 62: Under what situation would we use this function？

nni/compression/pytorch/speedup/quantization_speedup/calibrator.py
```diff
@@ -0,0 +1,51 @@
+import tensorrt as trt
+import os
+import pycuda.driver as cuda
+from urllib.request import urlretrieve
+import numpy as np
+
+
+class Calibrator(trt.IInt8Calibrator):
+    def __init__(self, training_data, cache_file, batch_size=64, algorithm=trt.CalibrationAlgoType.ENTROPY_CALIBRATION_2):
```
zheng-ningxin, line 9: Please add the doc string for this function.

linbinskn: Ok, added.

nni/compression/pytorch/speedup/quantization_speedup/common.py
linbinskn: This function is useless now and I will delete it. Under my consideration, this function will be used to do inference with given inference data and given batch size. I have using another function `do_inference_v2` to replace it.

nni/compression/pytorch/speedup/quantization_speedup/integrated_tensorrt.py
linbinskn: For the model with single input, `len(inputs) == 1` would always holds. This tool doesn't support model with multiple input right now.

nni/compression/pytorch/speedup/quantization_speedup/integrated_tensorrt.py
linbinskn: Good point, modified.

nni/compression/pytorch/speedup/quantization_speedup/calibrator.py
```diff
@@ -0,0 +1,51 @@
...
+        # Allocate enough memory for a whole batch.
+        self.device_input = cuda.mem_alloc(self.data[0].nbytes * self.batch_size)
+
+    def get_algorithm(self):
+        return self.algorithm
+
+    def get_batch_size(self):
+        return self.batch_size
+
+    def get_batch(self, names):
```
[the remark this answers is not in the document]
linbinskn: Have deleted it.

QuanluZhang:
```
class BaseModelSpeedup:
    def __init__(self, self, model, dummy_input, masks_file, backend='GPU'):
        if backend == 'tensorrt':
            BackendEngine()

    def test_latency(self):

    def inference(self, x):

    def export_compressed_module(self):

class TensorRTModuleSpeedup(BaseModuleSpeedup):
```

PR #3235, opened 2020-12-28
nnictl get port and pid from .experiment

J-shang:
1. To support `launch from python`, change from the original get `port` and `pid` from `.config` to the current get `port` and `pid` from `.experiment`.
2. Do not del `port` in `.experiment` when experiment stopped.
3. fix `EXPERIMENT_DETAIL_FORMAT` time print format

Note:
Unsupport `nnictl` commands when use `launch from python`: `view`, `resume`, `tensorboard`, `trial codegen`, `log trial`. `webui url`, `experiment save/load`, `experiment imoprt data`, `experiemnt delete`


nni/tools/nnictl/launcher.py
```diff
@@ -509,6 +509,11 @@ def launch_experiment(args, experiment_config, mode, experiment_id):
     rest_process, start_time = start_rest_server(args.port, experiment_config['trainingServicePlatform'], \
                                                  mode, experiment_id, foreground, log_dir, log_level)
     nni_config.set_config('restServerPid', rest_process.pid)
+    # save experiment information
+    nnictl_experiment_config = Experiments()
+    nnictl_experiment_config.add_experiment(experiment_id, args.port, start_time,
+                                            experiment_config['trainingServicePlatform'],
+                                            experiment_config['experimentName'], pid=rest_process.pid, logDir=log_dir)
```
QuanluZhang, line 516: why put this code snippet here? what type of error will be induced if keep it in the previous place?

nni/tools/nnictl/nnictl_utils.py
```diff
@@ -202,8 +201,9 @@ def convert_time_stamp_to_date(content):
 
 def check_rest(args):
     '''check if restful server is running'''
-    nni_config = Config(get_config_filename(args))
-    rest_port = nni_config.get_config('restServerPort')
+    experiment_config = Experiments()
+    experiment_dict = experiment_config.get_all_experiments()
+    rest_port = experiment_dict.get(get_config_filename(args), None).get('port', None)
```
QuanluZhang, line 206: you put `None` here means `None` may be returned, but None does not have attribute `get`

nni/tools/nnictl/launcher.py
J-shang: due to now we get `port` and `pid` from `.experiment`, we need set `pid` in `.experiment` when setting `pid` in `.config`.

nni/tools/nnictl/updater.py
```diff
@@ -58,8 +58,9 @@ def get_query_type(key):
 
 def update_experiment_profile(args, key, value):
     '''call restful server to update experiment profile'''
-    nni_config = Config(get_config_filename(args))
-    rest_port = nni_config.get_config('restServerPort')
+    experiment_config = Experiments()
+    experiment_dict = experiment_config.get_all_experiments()
+    rest_port = experiment_dict.get(get_config_filename(args), None).get('port', None)
```
QuanluZhang, line 63: you changed many places, just curious how to know all the related parts are changed, do we have ut to cover nnictl logic?

nni/tools/nnictl/nnictl_utils.py
J-shang: In fact `get_config_filename()` already has check if `get_config_filename(args)` is in `experiment_dict`, so if the value is `None`, there must be an unknown bug. And `{}.get('something', None)` is equal to `{}.get('something')`.

nni/tools/nnictl/updater.py
J-shang: To test this, we need ut that use `launch from python` launch a experiment, and use nnictl to manage it...

nni/tools/nnictl/nnictl_utils.py
QuanluZhang: got it. then suggest to remove `None` from the first `get`

nni/tools/nnictl/launcher.py
QuanluZhang: got it

nni/tools/nnictl/nnictl_utils.py
J-shang: ok, I will move all `None` in `{}.get()`

nni/tools/nnictl/updater.py
QuanluZhang: do we have ut to cover nnictl?

nni/tools/nnictl/nnictl_utils.py
```diff
@@ -186,7 +184,7 @@ def get_experiment_port(args):
         exit(1)
     experiment_config = Experiments()
     experiment_dict = experiment_config.get_all_experiments()
-    return experiment_dict[experiment_id]['port']
+    return experiment_dict[experiment_id].get('port', 'N/A')
```
QuanluZhang, line 187: just curious, why some `get('port'` has default value 'N/A' while some others do not have default value?

QuanluZhang (approved):
better to test tensorboard after this modification

nni/tools/nnictl/nnictl_utils.py
J-shang: it is for print `EXPERIMENT_DETAIL_FORMAT`, if `'port'` is `None`, print `N/A`. And this place actually not need default value, I will del these default values.

nni/tools/nnictl/launcher.py
```diff
@@ -608,8 +608,7 @@ def create_experiment(args):
     try:
         launch_experiment(args, experiment_config, 'new', experiment_id)
     except Exception as exception:
-        nni_config = Config(experiment_id)
-        restServerPid = nni_config.get_config('restServerPid')
+        restServerPid = Experiments().get_all_experiments().get(experiment_id).get('pid')
```
SparkSnail, line 611: If get(experiment_id) returns `None`, nnictl will crash here.

J-shang: indeed, fix it

SparkSnail (approved):

nni/tools/nnictl/launcher.py
```diff
@@ -641,8 +640,7 @@ def manage_stopped_experiment(args, mode):
     try:
         launch_experiment(args, experiment_config, mode, experiment_id)
     except Exception as exception:
-        nni_config = Config(experiment_id)
-        restServerPid = nni_config.get_config('restServerPid')
+        restServerPid = Experiments().get_all_experiments().get(experiment_id).get('pid')
```
SparkSnail, line 643: same here, avoid None response.

J-shang: oh yes, fix it

PR #3237, opened 2020-12-30
fix some issues about webui overview page

Lijiaoa:
**main issue:**
1. like this picture, error message is null because its color is white. And this icon's position needs to adjust.

image

2. These title(`Training platform`, `Best metric`) is **bold** , but don't need this style. 

image

image

**After fix:**
image
image
image

**The cause of the problem**
The color of error message is white because the class of `status` and `color` is not nested when render:

image


Lijiaoa:
+1
when experiment status is `STOPPED`, set port color as this picture(#E2E2E2)

image

Lijiaoa:
+1

TODO：
- [ ] experiment management page：latest experiment on the top line

ultmaster (approved):

colorjam (approved):

liuzhe-lz (approved):

PR #3238, opened 2020-12-30
Refactor wrap module as "blackbox_module"

ultmaster:
~~TODO: update graph gen to stop parsing when blackbox is encountered.~~

This PR updates register_module to blackbox_module, which makes the graph converter stop parsing and see the subgraph as one single node.

@blackbox_trainer looks a little weird. ~~Any better suggestions?~~ Use @register_trainer per Quanlu's suggestion.

### DIFF
--- a/nni/retiarii/experiment.py
+++ b/nni/retiarii/experiment.py
@@ -125,14 +125,14 @@ class RetiariiExperiment(Experiment):
         except Exception as e:
             _logger.error('Your base model cannot be parsed by torch.jit.script, please fix the following error:')
             raise e
-        base_model = convert_to_graph(script_module, self.base_model, self.recorded_module_args)
+        base_model_ir = convert_to_graph(script_module, self.base_model)
 
-        assert id(self.trainer) in self.recorded_module_args
-        trainer_config = self.recorded_module_args[id(self.trainer)]
-        base_model.apply_trainer(trainer_config['modulename'], trainer_config['args'])
+        recorded_module_args = get_records()
+        trainer_config = recorded_module_args[id(self.trainer)]
+        base_model_ir.apply_trainer(trainer_config['modulename'], trainer_config['args'])
 
         # handle inline mutations
-        mutators = self._process_inline_mutation(base_model)
+        mutators = self._process_inline_mutation(base_model_ir)
         if mutators is not None and self.applied_mutators:
             raise RuntimeError('Have not supported mixed usage of LayerChoice/InputChoice and mutators, \
                 do not use mutators when you use LayerChoice/InputChoice')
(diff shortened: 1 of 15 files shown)


nni/retiarii/utils.py
```diff
@@ -29,73 +31,76 @@ def add_record(key, value):
...
+    return _blackbox_cls(cls, module_name, 'args')
 
-def register_trainer():
-    def _register(cls):
-        m = _register_trainer(
-            original_class=cls)
-        return m
 
-    return _register
+def blackbox_trainer(cls):
```
QuanluZhang, line 97: I think its name could be "register_trainer", simple and clear

nni/retiarii/experiment.py
```diff
@@ -125,22 +125,22 @@ def _start_strategy(self):
...
             _logger.error('Your base model cannot be parsed by torch.jit.script, please fix the following error:')
             raise e
-        base_model = convert_to_graph(script_module, self.base_model, self.recorded_module_args)
+        base_model_ir = convert_to_graph(script_module, self.base_model)
 
-        assert id(self.trainer) in self.recorded_module_args
-        trainer_config = self.recorded_module_args[id(self.trainer)]
-        base_model.apply_trainer(trainer_config['modulename'], trainer_config['args'])
+        recorded_module_args = get_records()
+        trainer_config = recorded_module_args[id(self.trainer)]
```
QuanluZhang, line 131: better to add error message here to tell users that they may forgot to decorate their customized trainer

nni/retiarii/utils.py
```diff
@@ -29,73 +31,76 @@ def add_record(key, value):
...
-    original_class.__init__ = __init__  # Set the class' __init__ to the new one
-    return original_class
+            assert len(args) <= len(argname_list), f'Length of {args} is greater than length of {argname_list}.'
+            for argname, value in zip(argname_list, args):
+                full_args[argname] = value
 
+            # eject un-serializable arguments
+            for k in list(full_args.keys()):
+                if not isinstance(full_args[k], (int, float, str, dict, list)):
+                    full_args.pop(k)
```
QuanluZhang, line 58: do we need to output warning here to tell users that there is unserializable argument?

nni/retiarii/utils.py
```diff
@@ -29,73 +31,76 @@ def add_record(key, value):
...
-            # TODO: support both pytorch and tensorflow
-            from .nn.pytorch import Module
-            if isinstance(args[i], Module):
-                # ignore the base model object
-                continue
-            full_args[argname_list[i]] = arg
-        add_record(id(self), {'modulename': full_class_name, 'args': full_args})
 
-        orig_init(self, *args, **kws)  # Call the original __init__
+def blackbox(cls, *args, **kwargs):
```
QuanluZhang, line 81: what is this `blackbox` function used for?

QuanluZhang (approved):

nni/retiarii/utils.py
ultmaster: To create an instance inline without decorator. For example, `self.op = blackbox(MyCustomOp, hidden_units=128)`.

liuzhe-lz (approved):

PR #3240, opened 2020-12-30
Fixes for crashed NAS examples

ultmaster:
To be compatible with latest changes on PyTorch and NNI code hierarchy.

### DIFF
--- a/examples/nas/textnas/utils.py
+++ b/examples/nas/textnas/utils.py
@@ -14,7 +14,7 @@ logger = logging.getLogger("nni.textnas")
 
 def get_length(mask):
     length = torch.sum(mask, 1)
-    length = length.long()
+    length = length.long().cpu()
     return length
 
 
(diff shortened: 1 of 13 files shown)


examples/nas/textnas/utils.py
```diff
@@ -14,7 +14,7 @@
 
 def get_length(mask):
     length = torch.sum(mask, 1)
-    length = length.long()
+    length = length.long().cpu()
```
QuanluZhang, line 17: why "cpu()" is necessary, this is a bug fix?

QuanluZhang (approved):

examples/nas/textnas/utils.py
ultmaster: I think it's caused by PyTorch version upgrade. It reports some device misalignment error. I'm sure the problem was not there before.

liuzhe-lz (approved):

PR #3244, opened 2020-12-31
Add aten::detach in infer_shape to support .detach

tigert1998:


nni/compression/pytorch/speedup/infer_shape.py
```diff
@@ -273,7 +273,8 @@ def __repr__(self):
     'aten::mean': lambda module_masks, mask, shape: mean_inshape(module_masks, mask, shape),
     'Dropout': lambda module_masks, mask: dropout_inshape(module_masks, mask),
     'Dropout2d': lambda module_masks, mask: dropout_inshape(module_masks, mask),
-    'aten::dropout': lambda module_masks, mask: dropout_inshape(module_masks, mask)
+    'aten::dropout': lambda module_masks, mask: dropout_inshape(module_masks, mask),
+    'aten::detach': lambda module_masks, mask: dropout_inshape(module_masks, mask)
```
zheng-ningxin, line 277: Why detach should behave like a dropout OP?

tigert1998: Detach only affects the behavior in gradient backpropagation. And sparsity should be exactly the same between input and output tensor. I found this logic is already implemented in dropout callback so I directly put it here.

J-shang (approved):

zheng-ningxin:
Would you please show us a model that uses `aten:detach`, so that I can test this patch on it.

tigert1998:
```python
import torch
import torch.nn as nn

from nni.algorithms.compression.pytorch.pruning import L1FilterPruner
from nni.compression.pytorch import ModelSpeedup


class Net(nn.Module):
    def __init__(self):
        super().__init__()
        self.conv0 = nn.Conv2d(16, 16, 3, 1, 1)
        self.conv1 = nn.Conv2d(16, 16, 3, 1, 1)

    def forward(self, x):
        x = self.conv0(x)
        return self.conv1(x.detach())


if __name__ == "__main__":
    net = Net()
    net.eval()
    dummy_input = torch.rand((1, 16, 224, 224))

    config_list = [
        {'sparsity': 0.5, 'op_types': ['Conv2d']},
    ]

    pruner = L1FilterPruner(
        net, config_list, dependency_aware=True,
        dummy_input=dummy_input
    )
    pruner.compress()
    pruner.export_model("ckpt.pth", "mask.pth")
    pruner._unwrap_model()

    model_speedup = ModelSpeedup(net, dummy_input, "mask.pth")
    model_speedup.speedup_model()

```

zheng-ningxin (approved):

PR #3245, opened 2020-12-31
Fix RemoteConfig bug and add save log to dispatcher.log for nni.Experiment

liuzhe-lz:


nni/runtime/log.py
```diff
@@ -46,6 +46,9 @@ def init_logger_experiment() -> None:
     """
     formatter.format = _colorful_format
 
+    log_path = _prepare_log_dir(dispatcher_env_vars.NNI_LOG_DIRECTORY) / 'dispatcher.log'
+    _setup_root_logger(FileHandler(log_path), logging.DEBUG)
+
```
QuanluZhang, line 51: what is the relation between `init_logger_experiment()` and `init_logger()`?

QuanluZhang (approved):

J-shang (approved):

PR #3247, opened 2020-12-31
Fix dorefa bnn

linbinskn:
Because of the modification of compressor to solve the grad error, BNN and Dorefa can't operate normally on current code. This PR fixes this problem.


nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -148,6 +157,7 @@ def __init__(self, model, config_list, optimizer=None):
                     types of nn.module you want to apply quantization, eg. 'Conv2d'
         """
         super().__init__(model, config_list, optimizer)
+        self.quant_grad = QATGrad
```
J-shang, line 160: where we use `self.quant_grad`?

linbinskn: image
This is used to clip backward gradient between wmin and wmax.

nni/compression/pytorch/compressor.py
```diff
@@ -668,15 +670,20 @@ def forward(ctx, tensor, quant_type, wrapper, **kwargs):
         else:
             raise ValueError("unrecognized QuantType.")
 
-        bits = QuantGrad.get_bits_length(wrapper.config, quant_type)
-        qmin, qmax = torch.Tensor([0], device=tensor.device), torch.Tensor([(1 << bits) - 1], device=tensor.device)
-        ctx.save_for_backward(tensor, wrapper.module.scale, wrapper.module.zero_point, qmin, qmax)
+        bits = QuantGrad.get_bits_length(wrapper.config, QType_Dict[quant_type])
+        qmin, qmax = torch.Tensor([0]).to(tensor.device), torch.Tensor([(1 << bits) - 1]).to(tensor.device)
```
Cjkkkk, line 674: How to deal with the case of signed integer?

nni/compression/pytorch/compressor.py
```diff
@@ -668,15 +670,20 @@ def forward(ctx, tensor, quant_type, wrapper, **kwargs):
...
+        if hasattr(wrapper.module, 'scale') and hasattr(wrapper.module, 'zero_point'):
```
Cjkkkk, line 675: some quantizer only has scale to simplify the quantization calculation.

nni/compression/pytorch/compressor.py
```diff
@@ -580,10 +580,15 @@ class QuantType:
...
+    QUANT_WEIGHT = 1
+    QUANT_OUTPUT = 2
 
+QType_Dict = {
+    0: "input",
+    1: "weight",
+    2: "output"
+}
 
 class QuantGrad(torch.autograd.Function):
```
Cjkkkk, line 593: I think the best way to accomplish the quantization gradient calculation is to add `quant_weight`, `quant_input`, `quant_output` method to `quantizer` class and `QuantGrad` class only calls this methods instead.

nni/compression/pytorch/compressor.py
linbinskn: Yeah, some case may only use scale. And because it is the uniform quantization, I think using scale and zero-point is a more universal method which contain this special case.

nni/compression/pytorch/compressor.py
linbinskn: I think it is another thing which is not related to this pr. May be we will discuss it in the future release.

nni/compression/pytorch/compressor.py
linbinskn: Agree, this will take us time to discuss a new design and `quantizer` may be refactored in another pr. In this pr, we only fix dorefa and bnn unfit problem.

J-shang (approved):

Cjkkkk (approved):

QuanluZhang (approved):

PR #3248, opened 2021-01-03
Add config v2 example and refine validation error message

ultmaster:
Adds a minimum example for mnist-tfv2 written in v2 schema and reports validation error message when the config yaml has errors.


nni/tools/nnictl/launcher.py
```diff
@@ -592,16 +592,21 @@ def create_experiment(args):
...
-    except Exception as e:
-        print_error(e)
-        exit(1)
+    except Exception:
+        print_warning('Validation with V1 schema failed. Trying to convert from V2 format...')
+        try:
+            config = ExperimentConfig(**experiment_config)
+            experiment_config = convert.to_v1_yaml(config)
+        except Exception as e:
+            print_error(f'Conversion from v2 format failed: {repr(e)}')
```
QuanluZhang, line 604: better to add `exit(1)` here, as here it is already failed and output the error message. outputing the following error message would be a little misleading

QuanluZhang (approved):

nni/tools/nnictl/launcher.py
QuanluZhang: got it

liuzhe-lz (approved):

PR #3258, opened 2021-01-05
fix process early exit and handle invalid in windows

J-shang:


nni/experiment/launcher.py
```diff
@@ -88,7 +88,12 @@ def _start_rest_server(config: ExperimentConfig, port: int, debug: bool, experim
     for arg_key, arg_value in args.items():
         cmd.append('--' + arg_key)
         cmd.append(str(arg_value))
-    return int(time.time() * 1000), Popen(cmd, cwd=node_dir)
+
+    if sys.platform == 'win32':
+        proc = Popen(cmd, cwd=node_dir, creationflags=CREATE_NEW_PROCESS_GROUP)
```
SparkSnail, line 93: what's the purpose to add `CREATE_NEW_PROCESS_GROUP` here?

J-shang: due to if we send 'CTRL_BREAK' to this `proc`, it will send 'CTRL_BREAK' to all processes attached to the group that the `proc` belong to, and this will terminate the experiment at once. And `CREATE_NEW_PROCESS_GROUP` can fix this problem.

SparkSnail (approved):

liuzhe-lz (approved):

PR #3279, opened 2021-01-07
Fix bug for speedup module and enhance the Ut for speedup

zheng-ningxin:
Fix bug for speedup and enhance the UT for speedup.
(1)  The generated mask tensor should be on the same device as the original tensor.
(2) Due to `Cat` operator, the current speedup module may add a full-ones mask for some layers, the second bug will make the speedup module skip this mask(added by fix mask conflict) and lead to a shape mismatch.

### DIFF
--- a/nni/compression/pytorch/speedup/infer_shape.py
+++ b/nni/compression/pytorch/speedup/infer_shape.py
@@ -959,9 +959,7 @@ def conv2d_inshape(module_masks, mask):
         # mask conflict should be solved by fix_mask_conflict before speedup
 
         assert module_masks.input_mask == mask
-        
-        # assert module_masks.input_mask <= mask
-        # module_masks.input_mask.merge(mask)
+
     # shape changes pass through depths wise conv layers
     m = module_masks.module
     if m.in_channels == m.out_channels == m.groups:
--- a/nni/compression/pytorch/speedup/infer_shape.py
+++ b/nni/compression/pytorch/speedup/infer_shape.py
@@ -891,23 +891,18 @@ def conv2d_mask(module_masks, mask):
         sum_idx = (1, 2, 3) if dim == 0 else (0, 2, 3)
         index = torch.nonzero(weight_mask.abs().sum(
             sum_idx) != 0, as_tuple=True)[0]
-        if len(index) == weight_mask.shape[dim]:  # full mask
-            index = None
 
-        if index is None:
-            return None, None, None
-        else:
-            index = index.long().to(weight_mask.device)
-            weight_cmask = CoarseMask(num_dim=4)
-            weight_cmask.add_index_mask(dim=dim, index=index)
-            bias_cmask = None
-            if dim == 0 and 'bias' in mask and mask['bias'] is not None:
-                bias_index = torch.nonzero(mask['bias'], as_tuple=True)[0]
-                assert torch.all(torch.eq(index, bias_index)), \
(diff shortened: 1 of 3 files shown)


nni/compression/pytorch/speedup/infer_shape.py
```diff
@@ -962,8 +957,11 @@ def conv2d_inshape(module_masks, mask):
         # the same conv layer may be accessed more
         # than once, such as a concat operation.
         # mask conflict should be solved by fix_mask_conflict before speedup
-        assert module_masks.input_mask == mask
 
+        assert module_masks.input_mask == mask
+        
+        # assert module_masks.input_mask <= mask
+        # module_masks.input_mask.merge(mask)
```
QuanluZhang, line 964: what are these two comments used for?

nni/compression/pytorch/speedup/infer_shape.py
```diff
@@ -891,23 +891,18 @@ def convert_to_coarse_mask(mask, dim=0):
         sum_idx = (1, 2, 3) if dim == 0 else (0, 2, 3)
         index = torch.nonzero(weight_mask.abs().sum(
             sum_idx) != 0, as_tuple=True)[0]
-        if len(index) == weight_mask.shape[dim]:  # full mask
-            index = None
 
-        if index is None:
-            return None, None, None
```
QuanluZhang, line 898: so index cannot be None, why?

nni/compression/pytorch/speedup/infer_shape.py
```diff
@@ -891,23 +891,18 @@ def convert_to_coarse_mask(mask, dim=0):
         sum_idx = (1, 2, 3) if dim == 0 else (0, 2, 3)
         index = torch.nonzero(weight_mask.abs().sum(
             sum_idx) != 0, as_tuple=True)[0]
-        if len(index) == weight_mask.shape[dim]:  # full mask
-            index = None
```
J-shang, line 895: what issue does `index = None` cause, are there other situations that can cause `index == None`?

nni/compression/pytorch/speedup/infer_shape.py
J-shang, line 964: these two comments seem useless, removed?

test/ut/sdk/test_model_speedup.py
```diff
@@ -128,6 +128,18 @@ def generate_random_sparsity(model):
                              'sparsity': sparsity})
     return cfg_list
 
+def generate_random_sparsity_v2(model):
```
J-shang, line 131: maybe better to use ratio as a parameter, like `def generate_random_sparsity(model, layer_ratio):`, just personal opinion, the current implementation is fine to me.

nni/compression/pytorch/speedup/infer_shape.py
zheng-ningxin: removed, thanks~

nni/compression/pytorch/speedup/infer_shape.py
zheng-ningxin: if `index` is None, then we will skip this layer, it's fine when this mask is a fine-grained mask, however, this may also skip the all-ones mask added by the CatMaskPadding which will lead to a shape mis-match.

QuanluZhang (approved):

J-shang (approved):

PR #3295, opened 2021-01-13
Fix sqlite3 build on pipeline

ultmaster:
To fix sqlite3 at 5.0.0 because binary packages are missing for new versions. I don't know why such change works.

Also deleted unused pipeline files.

Another possible solution is to upgrade node-gyp version. It works as tested in #3296.


liuzhe-lz (approved):

ultmaster:
Seems that part of this PR is already done in #3288. Pending for merge.

liuzhe-lz:
> 
> 
> Seems that part of this PR is already done in #3288. Pending for merge.

I updated that PR to prevent conflict. Suggest to merge this one also.

liuzhe-lz (approved):

QuanluZhang (approved):

ts/nni_manager/package.json
```diff
@@ -21,7 +21,7 @@
     "lockfile": "^1.0.4",
     "python-shell": "^2.0.1",
     "rx": "^4.1.0",
-    "sqlite3": "^5.0.0",
+    "sqlite3": "5.0.0",
```
SparkSnail, line 24: Use a fixed version for sqlite3? then suggest to add annotation here to explain the reason.

ultmaster: Already added in description of this PR. Is there anyway to comment a JSON file?

SparkSnail (approved):

ts/nni_manager/package.json
SparkSnail: Ok, seems it hack for adding comments in this file.

Issue #2964, opened 2020-10-16
[webui enhancement] about the tooltip content for default metric chart


Lijiaoa:
**What would you like to be added**:
show more infomation on webui default metric chart, like chart, add trial id message. And on overview page, the default metric chart show less message when hover to the point, It's not friendly.
![image](https://user-images.githubusercontent.com/61399850/96207033-c9ce7e00-0f9c-11eb-877e-0c13278182fb.png)
![image](https://user-images.githubusercontent.com/61399850/96207195-35b0e680-0f9d-11eb-8e76-e2de98889b3c.png)

**Why is this needed**: 信息一致，更加友好

**Without this feature, how does current nni work**：

**Components that may involve changes**:

**Brief description of your proposal if any**:

kvartet:
closed as it has been merged.

PR #3316, opened 2021-01-18
[webui] Update default metric tooltip

Lijiaoa:
#2964


ts/webui/src/components/trial-detail/DefaultMetricPoint.tsx
```diff
@@ -25,6 +25,8 @@ const EmptyGraph = {
 interface DefaultPointProps {
     trialIds: string[];
     visible: boolean;
+    chartHeight: number;
+    isHasbestCurve: boolean;
```
ultmaster, line 29: Suggest hasBestCurve

Lijiaoa: ok

ts/webui/src/App.tsx
```diff
@@ -6,8 +6,13 @@ import { isManagerExperimentPage } from './static/function';
 import NavCon from './components/NavCon';
 import MessageInfo from './components/modals/MessageInfo';
 import { SlideNavBtns } from './components/slideNav/SlideNavBtns';
+const echarts = require('echarts/lib/echarts');
+echarts.registerTheme('my_theme', {
```
liuzhe-lz, line 10: Maybe "nni_theme" or "nniTheme" is a better name?

ts/webui/src/components/Overview.tsx
```diff
@@ -167,7 +165,12 @@ class Overview extends React.Component<{}, OverviewState> {
                                         </div>
                                     </Stack>
                                     <div className='overviewChart'>
-                                        <Accuracy accuracyData={accuracyGraphData} accNodata={noDataMessage} />
+                                        <DefaultPoint
+                                            trialIds={bestTrials.map(trial => trial.info.trialJobId)}
+                                            visible={true}
```
liuzhe-lz, line 170: I think "visible" can be handled in this level. No need to pass a property into `DefaultPoint`.

ts/webui/src/App.tsx
Lijiaoa: yes, will change it to nni_theme

ts/webui/src/components/Overview.tsx
Lijiaoa: yes, here no need this `visible`. And also no need `whichGraph`. `whichGraph` is appeared in v0.8 for `ant design`. It should be deleted when using `fabric/fluent` because `current tab` will be uninstall when you choose other `tab`. So webui doesn't need to consider `current tab` when page refresh. Thanks for your comment.

Lijiaoa:
image

Test also find that: `hyper-parameter | Duration | Intermediate` result graph will render twice when page refreshing. And drawing `Default metric` graph function is in `render()` component. It doesn't correct. These status maybe will fix in next release.

liuzhe-lz (approved):

ultmaster (approved):

PR #3334, opened 2021-01-25
migrate nnicli

J-shang:
need do in next pr:
1. start experiment by config file
2. update search space by filepath
3. support stop experiment when connect experiment by port
4. update doc & test


liuzhe-lz (approved):

test/config/integration_tests.yml
```diff
@@ -142,17 +142,6 @@ testCases:
     kwargs:
       import_data_file_path: config/nnictl_experiment/test_import.json
 
-- name: nnicli
```
QuanluZhang, line 145: similar test should be added back to test the refactored code?

nni/experiment/experiment.py
```diff
@@ -193,8 +207,369 @@ def run(self, port: int = 8080, debug: bool = False) -> bool:
             self.stop()
 
 
-    def get_status(self) -> str:
+    def connect_experiment(self, port: int):
```
QuanluZhang, line 210: it would be better to have a BaseExperiment class which has basic functionality of an experiment, then write different Experiment classes inheriting BaseExperiment to provide view/launch experiment and to manage experiments

QuanluZhang:
this pr is about the design of experiment apis, let's fully discuss it

nni/experiment/experiment.py
```diff
@@ -76,23 +77,35 @@ def __init__(self, tuner: Tuner, training_service: Union[str, List[str]]) -> Non
...
+                config, training_service = None, config
 
-        if config is None:
-            self.config = ExperimentConfig(training_service)
-        else:
-            self.config = config
+            if config is None:
+                self.config = ExperimentConfig(training_service)
+            else:
+                self.config = config
```
ultmaster, line 108: Suggest print a message when tuner is None.

nni/experiment/experiment.py
```diff
@@ -193,8 +207,369 @@ def run(self, port: int = 8080, debug: bool = False) -> bool:
...
+        Parameters
+        ----------
+        port
+            The port of web UI.
+        """
+        self.port = port
+        self.get_status()
+
+
+    def _experiment_rest_get(self, port: int, api: str) -> Any:
```
ultmaster, line 223: double-line between member functions looks weird to me. Suggest reformatting this file with tools like autopep8.

nni/experiment/experiment.py
```diff
@@ -193,8 +207,369 @@ def run(self, port: int = 8080, debug: bool = False) -> bool:
...
+        self.trialJobId = None
+        for key in json_obj.keys():
+            setattr(self, key, json_obj[key])
+        self.value = json.loads(self.value)
+
+    def __repr__(self):
+        return "TrialResult(parameter: {} value: {} trialJobId: {})".format(self.parameter, self.value, self.trialJobId)
+
+
+class TrialMetricData:
```
ultmaster, line 452: Maybe we should rewrite these classes with @dataclass?

test/nni_test/nnitest/validators.py
```diff
@@ -93,7 +93,7 @@ class NnicliValidator(ITValidator):
     def __call__(self, rest_endpoint, experiment_dir, nni_source_dir, **kwargs):
         print(rest_endpoint)
         exp = Experiment()
-        exp.connect_experiment(rest_endpoint)
+        exp.connect_experiment(int(rest_endpoint.split(':')[-1]))
```
ultmaster, line 96: Is this the correct way to get the port part? What if the endpoint ends with `/`?

test/config/integration_tests.yml
J-shang: we don't support launch from config file and stop a connected experiment right now, will add back in next pr.

nni/experiment/experiment.py
J-shang: good point, we need to discuss this, in the previous discussion, we use parameters to distinguish different launch ways.

nni/experiment/experiment.py
J-shang: I will confirm with @liuzhe-lz

nni/experiment/experiment.py
J-shang: yes, I will add a message.

nni/experiment/experiment.py
J-shang: good suggestion, I will rewrite them

test/nni_test/nnitest/validators.py
J-shang: it's a workaround, and will refactor in next pr, this will never be called right now.

QuanluZhang (approved):

ultmaster (approved):

PR #3365, opened 2021-02-04
[retiarii] refactor of pytorch operators

QuanluZhang:
- [x] support more unit tests
- [x] support shared module
- [x] refactor of pytorch operators
- [ ] write more unittest: pytorch models
- [x] write more unittest: basic operations (partial)
- [ ] support control flow (v2.2)


tczhangzhi:
I like this refactor, it seems everything is on the right way again. In order to find a good design pattern, I browsed some `tvm` repositories. Extracting and retranslating computational graphs is widely discussed, and they usually exist as part of the library. In fact, I guess a more perfect solution is not to create a calculation graph intermediate representation for the entire module, and just replace the parts related to `nni`, like `nn.LayerChoice`.

nni/retiarii/codegen/pytorch.py
```diff
@@ -33,8 +33,25 @@ def _sorted_incoming_edges(node: Node) -> List[Edge]:
 
 
 def _format_inputs(node: Node) -> List[str]:
```
liuzhe-lz, line 35: `-> Tuple[List[str], List[str]]:`

nni/retiarii/operation_def/torch_op_def.py
```diff
@@ -1,46 +1,331 @@
...
+    'torch.half',         # 5
+    'torch.float',        # 6
+    'torch.double',       # 7
+    'torch.complex32',    # 8
+    'torch.complex64',    # 9
+    'torch.complex128',   # 10
+    'torch.bool',         # 11
+]
 
+class NoopIdentity(PyTorchOperation):
```
liuzhe-lz, line 32: `NoOp`

nni/retiarii/operation_def/torch_op_def.py
```diff
@@ -1,46 +1,331 @@
...
+    def emit_args(args):
+        # filter out the `out` argument here
+        return [(arg.name, str(arg.type), str(arg.default_value)) for arg in args] #  if arg.name != 'out'
+
+    op_names = []
+    op_args = {}
+    # discover methods
+    for elem in dir(torch.Tensor):
+        if not hidden(elem):
+            schemas = torch._C._jit_get_schemas_for_operator("aten::" + elem)
```
liuzhe-lz, line 266: This API looks dangerous to me...

QuanluZhang: yes, kind of. this one is copied from this file: https://github.com/pytorch/pytorch/blob/master/torch/jit/supported_ops.py

QuanluZhang:
> I like this refactor, it seems everything is on the right way again. In order to find a good design pattern, I browsed some `tvm` repositories. Extracting and retranslating computational graphs is widely discussed, and they usually exist as part of the library. In fact, I guess a more perfect solution is not to create a calculation graph intermediate representation for the entire module, but just replace the parts related to `nni`, like `nn.LayerChoice`.

thanks for your suggestion! it would be great if you can also provide the links of the discussion here :). I agree with your suggestion. we also thought about this design. the problem is that we have to convert graph back to code. if we only parse a submodule while keeping the top-level module unchanged, it is not easy to stitch the unchanged part and the generated part together. we will keep thinking about this. If you have good suggestions, feel free to tell us.

nni/retiarii/codegen/pytorch.py
QuanluZhang: updated, thanks

nni/retiarii/converter/graph_gen.py
```diff
@@ -189,8 +209,19 @@ def _generate_expr(tensor):
                     left = _generate_expr(tensor.node().inputsAt(0))
                     right = _generate_expr(tensor.node().inputsAt(1))
                     return f'({left} == {right})'
+                elif tensor.node().kind() == 'aten::le':
+                    left = _generate_expr(tensor.node().inputsAt(0))
+                    right = _generate_expr(tensor.node().inputsAt(1))
+                    return f'({left} <= {right})'
```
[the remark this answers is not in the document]
QuanluZhang: added

nni/retiarii/operation.py
```diff
@@ -106,59 +131,27 @@ def to_init_code(self, field: str) -> str:
...
+        inputs_value : List[Any]
+            some variables are actually constant, their real values are recorded in ```inputs_value```.
+            if not constant, we simply put None at the corresponding index
+
+        Returns
+        -------
+        str
+            generated code line
+        """
+        if self.type == 'aten::slice':
```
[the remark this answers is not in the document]
QuanluZhang: will refactor this part in the next pr

nni/retiarii/operation_def/torch_op_def.py
QuanluZhang: updated

test/ut/retiarii/inject_nn.py
```diff
@@ -0,0 +1,280 @@
...
+        wrapped_class.__init__ = wrapped_class.bak_init_for_inject
+        delattr(wrapped_class, 'bak_init_for_inject')
+    if hasattr(wrapped_class, 'bak_del_for_inject'):
+        if wrapped_class.bak_del_for_inject is not None:
+            wrapped_class.__del__ = wrapped_class.bak_del_for_inject
+        delattr(wrapped_class, 'bak_del_for_inject')
+    return None
+
+def remove_inject_pytorch_nn():
+    Identity = unwrap_module(nn.Identity)
```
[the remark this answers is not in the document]
QuanluZhang: let's leave it for now

ultmaster (approved):

nni/retiarii/converter/graph_gen.py
```diff
@@ -217,8 +248,128 @@ def handle_if_node(node):
...
+                    shared_type_operation = Operation.new('shared', {'reference': submodule_full_name})
+                    subcell = ir_graph.add_node(shared_node_name, shared_type_operation)
+                else:
+                    # this module is processed for the first time, build cell for it
+                    if subgraph is None:
+                        # if we do not parse this module's graph, we create Node for this module
+                        subcell = ir_graph.add_node(submodule_full_name, submodule_type_str, sub_m_attrs)
+                        if isinstance(submodule_obj, Placeholder):
+                            subcell.update_label(submodule_obj.label)
+                        elif isinstance(submodule_obj, (LayerChoice, InputChoice)):
```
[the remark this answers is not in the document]
QuanluZhang: because only our provided APIs can be specified a label. If ValueChoice is a graph node, we should deal with it here, otherwise no need to deal with it here.

liuzhe-lz (approved):

Issue #3385, opened 2021-02-18
AutoCompression Class


MohamedAliRashad:
## What would you like to be added ?
A unified class for Automatic Compression (Pruning & Quantization).

## Why is this needed ?
Ease of Usage (and all the required blocks are already there).

## Without this feature, how does current nni work ?
The closest solution for something like this is this [script](https://github.com/microsoft/nni/blob/master/examples/model_compress/pruning/auto_pruners_torch.py), but it's still only for pruning and doesn't even use all purners.

## Brief description of my prposal
Simple Class called `AutoCompresser` that can be used like this
```
from nni.compression.pytorch import AutoCompressor

comp = AutoCompressor(pruners=["LevelPruner", "AGPPruner"], quantizers="All", training_dataset=training_dataloader, testing_dataset=testing_dataloader, evaluation_metrics=MSELoss())
model = comp(model, progress_bar=True) 
```

J-shang:
Thanks for your issue. It is a great idea to make using pruner and quantizer more convenient. We will discuss the specific implementation plan and more ways to combine pruner and quantizer in later versions. We will continue to update the improvement plan under this issue and welcome to discuss together.

MohamedAliRashad:
I will start a PR to work on this feature, feel free to join anytime @J-shang

kvartet:
@J-shang, is AutoCompression Framework supported in NNI v2.3 satisfy the user's need? if it is, I will close this issue~

J-shang:
> @J-shang, is AutoCompression Framework supported in NNI v2.3 satisfy the user's need? if it is, I will close this issue~

Yes, AutoCompression Framework is partial what this issue needs, and we need to improve sparsity recommendations in the future.

PR #3390, opened 2021-02-19
Abstract class for automatic compression

MohamedAliRashad:
PR to track the addition of auto compression interface requested here #3385


MohamedAliRashad:
We will need to discuss the attributes of the class and how it will work

QuanluZhang:
> We will need to discuss the attributes of the class and how it will work

thanks for your contribution, feel free to post more detailed design. we will first discuss internally and provide more related information and design suggestions soon.

J-shang:
Hi @MohamedAliRashad , we want to implement this feature in the next version, and I commit our basic idea about this. Include `MultiCompressor` which can specify the pruner and quantizer for each layer, and base on it `AutoCompressor` can auto change the pruner and quantizer to find their effect.

J-shang:
The latest implementation is in #3627 and #3631, and this pr will be closed.

PR #3400, opened 2021-02-24
[webui] fix some bugs on v2.1 bug bash

Lijiaoa:
- use localstorage to store displayed columns status
- fix bug about the point look like this picture is not clickable in the overview page

     image


ts/webui/src/components/trial-detail/DefaultMetricPoint.tsx
```diff
@@ -36,6 +36,8 @@ interface DefaultPointState {
 }
 
 class DefaultPoint extends React.Component<DefaultPointProps, DefaultPointState> {
+    private isDetailPage = window.location.pathname === '/oview' ? false : true;
```
liuzhe-lz, line 39: Suggest to use property.

liuzhe-lz (approved):

ts/webui/src/components/trial-detail/DefaultMetricPoint.tsx
Lijiaoa: ok

ultmaster (approved):

PR #3419, opened 2021-03-04
Align `nni.experiment` tuner behavior with nnictl

liuzhe-lz:
This PR also refactors NNI manager to support quit from REST request.


nni/experiment/config/common.py
```diff
@@ -106,7 +106,10 @@ def _validation_rules(self):
     'trial_code_directory': util.canonical_path,
     'max_experiment_duration': lambda value: f'{util.parse_time(value)}s' if value is not None else None,
     'experiment_working_directory': util.canonical_path,
-    'tuner_gpu_indices': lambda value: [int(idx) for idx in value.split(',')] if isinstance(value, str) else value
+    'tuner_gpu_indices': lambda value: [int(idx) for idx in value.split(',')] if isinstance(value, str) else value,
+    'tuner': lambda config: None if config.name == '_none_' else config,
+    'assessor': lambda config: None if config.name == '_none_' else config,
+    'advisor': lambda config: None if config.name == '_none_' else config,
```
QuanluZhang, line 112: where '_none_' is assigned to `name` and in which case?

QuanluZhang: found...

liuzhe-lz: In `Experiment.__init__()`. The purpose is to let users omit calling `AlgorithmConfig()` explicitly.
`AlgorithmConfig` with name "_none_" is designed to be equivalent to `None`, just like `Path('/home')` is equivalent to `str('/home/')`. This means everywhere `tuner` can be `None`, it can also be `AlgorithmConfig(name='_none_')`.

nni/experiment/launcher.py
```diff
@@ -97,7 +86,8 @@ def _start_rest_server(config: ExperimentConfig, port: int, debug: bool, experim
         from subprocess import CREATE_NEW_PROCESS_GROUP
         proc = Popen(cmd, cwd=node_dir, creationflags=CREATE_NEW_PROCESS_GROUP)
     else:
-        proc = Popen(cmd, cwd=node_dir)
+        import os
+        proc = Popen(cmd, cwd=node_dir, preexec_fn=os.setpgrp)
```
QuanluZhang, line 89: why `preexec_fn` is not set in the previous version?

ts/nni_manager/main.ts
```diff
@@ -174,25 +174,8 @@ mkDirP(getLogDir())
...
-    let hasError: boolean = false;
-    try {
-        const nniManager: Manager = component.get(Manager);
-        await nniManager.stopExperiment();
-        const experimentManager: ExperimentManager = component.get(ExperimentManager);
-        await experimentManager.stop();
-        const ds: DataStore = component.get(DataStore);
-        await ds.close();
-        const restServer: NNIRestServer = component.get(NNIRestServer);
-        await restServer.stop();
```
QuanluZhang, line 188: it is a little strange that these components are created in this file, but stopped in nnimanager.ts

nni/experiment/config/convert.py
```diff
@@ -14,7 +14,7 @@
 
 
 def to_v1_yaml(config: ExperimentConfig, skip_nnictl: bool = False) -> Dict[str, Any]:
-    config.validate(skip_nnictl)
+    config.validate(False)
```
QuanluZhang, line 17: `skip_nnictl` is not used in this function. what is the meaning of `skip_nnictl`?

ts/nni_manager/rest_server/restHandler.ts
```diff
@@ -317,6 +318,15 @@ class NNIRestHandler {
         });
     }
 
+    private stop(router: Router): void {
+        router.delete('/experiment', (req: Request, res: Response) => {
+            this.nniManager.stopExperimentTopHalf().then(() => {
+                res.send();
+                this.nniManager.stopExperimentBottomHalf();
```
J-shang, line 325: Maybe we can wait for the whole experiment to stop, then `res.send()`? To avoid something like the dispatcher has an independent process in the future. Or we need to put this kind of independent process `cleanup()` in `stopExperimentTopHalf()`? Anyway, it is fine in this version.

J-shang (approved):

nni/experiment/launcher.py
liuzhe-lz: Because it's not needed.
This is about terminating subprocess.

ts/nni_manager/main.ts
liuzhe-lz: Maybe. But I don't want to make huge refactor at this point.

ts/nni_manager/rest_server/restHandler.ts
liuzhe-lz: Because killing trials is too time consuming. I'm afraid users (like Quanlu) will kill the process with second ctrl-c if we let them wait that long.

nni/experiment/config/convert.py
liuzhe-lz: This function is likely to be removed soon. I'm too lazy to make it elegant.

ts/nni_manager/core/nnimanager.ts
```diff
@@ -296,10 +297,70 @@ class NNIManager implements Manager {
     }
 
     public async stopExperiment(): Promise<void> {
+        await this.stopExperimentTopHalf();
+        await this.stopExperimentBottomHalf();
+    }
+
+    public async stopExperimentTopHalf(): Promise<void> {
```
ultmaster, line 304: The API sounds very internal. Can we think of a formal name?

QuanluZhang (approved):

ts/nni_manager/core/nnimanager.ts
liuzhe-lz: The name comes from Linux interrupt handler. Top-half means it should be completed before handling control to caller.

PR #3424, opened 2021-03-05
Retiarii exp launch

QuanluZhang:


ts/nni_manager/core/nnimanager.ts
```diff
@@ -341,7 +341,6 @@ class NNIManager implements Manager {
                 }
             }
         }
-
```
ultmaster, line 344: This is a unrelated change?

nni/retiarii/experiment/pytorch.py
```diff
@@ -131,7 +151,37 @@ def start(self, port: int = 8080, debug: bool = False) -> None:
         debug
             Whether to start in debug mode.
         """
-        super().start(port, debug)
+        atexit.register(self.stop)
+
+        self.id = management.generate_experiment_id()
+
+        if self.config.experiment_working_directory is not None:
```
ultmaster, line 158: Is it a good idea to put these logics into Retiarii? Might as well create a new class in `nni.experiment`, called, for example, `_TunerCenteredExperiment`

ultmaster (approved):

nni/retiarii/experiment/pytorch.py
liuzhe-lz: I'll consider this a temporary implementation and should be refactored in upcoming release.

liuzhe-lz (approved):

nni/retiarii/experiment/pytorch.py
QuanluZhang: agree

ts/nni_manager/core/nnimanager.ts
QuanluZhang: sorry, not related...

PR #3426, opened 2021-03-05
Fix sharedstorage bug

J-shang:


ts/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
```diff
@@ -182,7 +182,7 @@ export class OpenPaiEnvironmentService extends EnvironmentService {
         let environmentRoot: string;
         if (environment.useSharedStorage) {
             environmentRoot = component.get<SharedStorageService>(SharedStorageService).remoteWorkingRoot;
-            environment.command = `${component.get<SharedStorageService>(SharedStorageService).remoteMountCommand} && cd ${environmentRoot} && ${environment.command}`;
+            environment.command = `${component.get<SharedStorageService>(SharedStorageService).remoteMountCommand.replace(/echo -e /g, `echo `).replace(/echo /g, `echo -e `)} && cd ${environmentRoot} && ${environment.command}`;
```
liuzhe-lz, line 185: What is this doing?

J-shang: Change all `echo` to `echo -e`. Because in pai we need `-e` to escape character, but in other training services we don't need. Seems if we run the command by `python` or `ts`, we don't need `-e`, but in bash, we need `-e`. I am not sure, but it runs well under this.

docs/en_US/Tutorial/ExperimentConfig.rst
```diff
@@ -838,14 +838,14 @@ localMountPoint
...
-The absolute path that the storage has been or will be mounted in local.
+The absolute path that the storage has been or will be mounted in local. i.e. ``/tmp/nni-shared-storage``.
 
 remoteMountPoint
 ^^^^^^^^^^^^^^^^
 
 Required. String.
 
-The absolute path that the storage will be mounted in remote.
+The absolute path that the storage will be mounted in remote. i.e. ``./nni-shared-storage``. Note that the directory must be empty if using AzureBlob.
```
SparkSnail, line 848: If the path does not exist in remote, will it be created automatically?

J-shang: Yes, it will be created automatically.

SparkSnail (approved):

ts/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
liuzhe-lz: So why it's replaced twice?

liuzhe-lz (approved):

ts/nni_manager/training_service/reusable/environments/openPaiEnvironmentService.ts
J-shang: To avoid the situation of `echo` and `echo -e` exist at the same time.

Issue #3276, opened 2021-01-06
ModuleNotFoundError: No module named 'ruamel'


m4rkl1u:
**Environment**:
- NNI version: 1.9.0
- NNI mode (local|remote|pai): Local
- Client OS: Ubuntu
- Server OS (for remote mode only): Local
- Python version: 3.6.0
- PyTorch/TensorFlow version: tensorflow==2.3.0
- Is conda/virtualenv/venv used?: conda
- Is running in Docker?:Yes

**Log message**:

```
   /opt/conda/lib/python3.6/site-packages/nni/package_utils.py in <module>
          8 from pathlib import Path
          9 import importlib
    ---> 10 import ruamel.yaml as yaml
         11 
         12 from .constants import BuiltinAlgorithms
    
    ModuleNotFoundError: No module named 'ruamel'
 
```

**What issue meet, what's expected?**:
It seems the conda install ruamel is ruamel_yaml, I expect that if the nni running in docker+conda environment could identify the ruamel path

**How to reproduce it?**: 

**Additional information**:

liuanhua110:
change the 'import ruamel.yaml as yaml' to 'import ruamel_yaml as yaml' it works for me

kvartet:
> change the 'import ruamel.yaml as yaml' to 'import ruamel_yaml as yaml' it works for me

Thanks for your solution, maybe @m4rkl1u can try his method.

kvartet:
met this error in Windows :(

potoo0:
- windows 10
- python 3.8
 could install `ruamel.yaml` to solve this:
```
pip install --ignore-installed ruamel.yaml
```

kvartet:
To avoid this issue, NNI substitutes ``ruamel.yaml`` with PyYAML in pr #3702, which will be released in v2.3.

PR #3430, opened 2021-03-07
fix: Fix ruamel import

rushtehrani:
This PR fixes the following error when the `ruamel.yaml` version is greater than 0.11.4:

```bash
ModuleNotFoundError: No module named 'ruamel'
```

It also makes the code consistent with other code that uses this module.

### DIFF
--- a/nni/experiment/config/base.py
+++ b/nni/experiment/config/base.py
@@ -6,7 +6,7 @@ import dataclasses
 from pathlib import Path
 from typing import Any, Dict, Optional, Type, TypeVar
 
-from ruamel.yaml import yaml
+import ruamel.yaml as yaml
 
 from . import util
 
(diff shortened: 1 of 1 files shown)


QuanluZhang (approved):

J-shang:
@rushtehrani Hi, thanks for your pr, but we don't meet this issue with `ruamel.yaml == 0.16.12`, could you show your `ruamel.yaml` version? Let's check if the version range of `ruamel.yaml` should be restricted.

And in some scenarios we will meet error using `import ruamel.yaml as yaml`, FYI.
https://github.com/microsoft/nni/issues/3276

rushtehrani:
Hi @J-shang 

> @rushtehrani Hi, thanks for your pr, but we don't meet this issue with `ruamel.yaml == 0.16.12`, could you show your `ruamel.yaml` version? Let's check if the version range of `ruamel.yaml` should be restricted.

The version I was having issues with is `0.15.80`, I think at some point they changed the name of the PyPi package from `ruamel_yaml` to `ruamel.yaml`.

> And in some scenarios we will meet error using `import ruamel.yaml as yaml`, FYI.
> #3276

Yes, I believe this is due to the name change in the most recent versions. 

Note that `import ruamel.yaml as yaml` format is [used in a few other places](https://github.com/microsoft/nni/search?q=import+ruamel.yaml+as+yaml) as well so I'm sure if the issues can crop up there too.

J-shang:
@rushtehrani Hi, you are right, I find the doc https://yaml.readthedocs.io/en/latest/api.html , and this pr will be merged. Thanks for your contribution and if there is any other issue, welcome to communicate with us.

J-shang (approved):

PR #3435, opened 2021-03-08
fix update search space

J-shang:


nni/experiment/experiment.py
```diff
@@ -377,6 +378,7 @@ def update_search_space(self, value: dict):
         value: dict
             New search_space.
         """
+        value = json.dumps(value)
```
liuzhe-lz, line 381: I think this should be handled inside dispatcher.

liuzhe-lz (approved):

nni/experiment/experiment.py
liuzhe-lz, line 381: I think it should use json_tricks here. It might contains numpy value or infinity/nan.

J-shang: fix it

SparkSnail (approved):

PR #3439, opened 2021-03-09
Fix relative path bug in config v2

liuzhe-lz:
Fix a bug found by J-Shang: when reading v2 config file nnictl will use pwd to resolve relative paths, while the correct behavior is to use config file's directory.

### DIFF
--- a/nni/experiment/config/base.py
+++ b/nni/experiment/config/base.py
@@ -47,6 +47,8 @@ class ConfigBase:
         They will be converted to snake_case automatically.
         If a field is missing and don't have default value, it will be set to `dataclasses.MISSING`.
         """
+        if 'basepath' in kwargs:
+            _base_path = kwargs.pop('basepath')
         kwargs = {util.case_insensitive(key): value for key, value in kwargs.items()}
         if _base_path is None:
             _base_path = Path()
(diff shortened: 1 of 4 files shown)


nni/experiment/config/base.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, *, _base_path: Optional[Path] = None, **kwargs):
         They will be converted to snake_case automatically.
         If a field is missing and don't have default value, it will be set to `dataclasses.MISSING`.
         """
+        if 'basepath' in kwargs:
+            _base_path = kwargs.pop('basepath')
```
QuanluZhang, line 51: what is basepath used for?

QuanluZhang: what is the difference between basepath and _base_path and base_path. looks messy. could you briefly explain this pr?

liuzhe-lz: Base path is used to resolve relative paths.
The name `_base_path` is used in interface to indicate internal API; and `base_path` is used for local variables.
And there is a problem: to support snake_case (used in Python code) and camelCase (used in YAML) configuration at same time, the `Config` base class removes all underscore when loading a dict, which may incorrectly rename `_base_path` to `basepath`. Line 50 and 51 patch this problem.

J-shang (approved):

ultmaster (approved):

PR #3356, opened 2021-02-01
Support mixed-precision quantization speed up by using tensorrt[original pr #3220 has been closed]

linbinskn:
This PR is used to support real mixed-precision quantization speed up in hardware by using tensorrt.  Original PR #3220 has been deleted for its old base version which will cause much modification of merging new version.

PR #3458, opened 2021-03-19
Add model export for QAT

linbinskn:
Add `export_model` function for quantization algorithm QAT. Users can save quantized weight and calibration parameters to specific path. What' s more, this function will be a prerequisite for #3356 (Support mixed-precision quantization speed up by using tensorrt).

### DIFF
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -110,6 +110,27 @@ def get_bits_length(config, quant_type):
     else:
         return config["quant_bits"].get(quant_type)
 
+def del_simulated_attr(module):
+    if hasattr(module, 'old_weight'):
+        delattr(module, 'old_weight')
+    if hasattr(module, 'ema_decay'):
+        delattr(module, 'ema_decay')
+    if hasattr(module, 'tracked_min_biased'):
+        delattr(module, 'tracked_min_biased')
+    if hasattr(module, 'tracked_max_biased'):
+        delattr(module, 'tracked_max_biased')
+    if hasattr(module, 'tracked_min'):
+        delattr(module, 'tracked_min')
+    if hasattr(module, 'tracked_max'):
+        delattr(module, 'tracked_max')
+    if hasattr(module, 'scale'):
+        delattr(module, 'scale')
+    if hasattr(module, 'zero_point'):
+        delattr(module, 'zero_point')
+    if hasattr(module, 'weight_bit'):
+        delattr(module, 'weight_bit')
+    if hasattr(module, 'activation_bit'):
+        delattr(module, 'activation_bit')
 
 class QATGrad(QuantGrad):
     @staticmethod
@@ -153,7 +174,10 @@ class QAT_Quantizer(Quantizer):
         for layer, config in modules_to_compress:
             layer.module.register_buffer("zero_point", torch.Tensor([0.0]))
             layer.module.register_buffer("scale", torch.Tensor([1.0]))
+            if "weight" in config.get("quant_types", []):
+                layer.module.register_buffer('weight_bit', torch.zeros(1))
             if "output" in config.get("quant_types", []):
+                layer.module.register_buffer('activation_bit', torch.zeros(1))
                 layer.module.register_buffer('ema_decay', torch.Tensor([0.99]))
                 layer.module.register_buffer('tracked_min_biased', torch.zeros(1))
                 layer.module.register_buffer('tracked_min', torch.zeros(1))
@@ -256,6 +280,7 @@ class QAT_Quantizer(Quantizer):
         module.scale, module.zero_point = update_quantization_param(weight_bits, rmin, rmax)
         weight = self._quantize(weight_bits, module, weight)
         weight = self._dequantize(module, weight)
+        module.weight_bit = torch.Tensor([weight_bits])
         wrapper.module.weight = weight
         return weight
 
@@ -263,6 +288,7 @@ class QAT_Quantizer(Quantizer):
         config = wrapper.config
         module = wrapper.module
         output_bits = get_bits_length(config, 'output')
+        module.activation_bit = torch.Tensor([output_bits])
         quant_start_step = config.get('quant_start_step', 0)
         assert output_bits >= 1, "quant bits length should be at least 1"
 
@@ -281,6 +307,45 @@ class QAT_Quantizer(Quantizer):
         out = self._quantize(output_bits, module, output)
         out = self._dequantize(module, out)
         return out
+
+    def export_model(self, model_path, calibration_path=None):
+        """
+        Export pruned model weights, masks and onnx model(optional)
+
+        Parameters
+        ----------
+        model_path : str
+            path to save pruned model state_dict
+        calibration_path : str
+            (optional) path to save quantize parameters after calibration
+        """
+        assert model_path is not None, 'model_path must be specified'
+        self._unwrap_model()
+        calibration_config = {}
+        support_op = [torch.nn.Conv2d, torch.nn.Linear, torch.nn.ReLU6]
+
+        for name, module in self.bound_model.named_modules():
+            if type(module) not in support_op:
+                del_simulated_attr(module)
+                continue
+            if hasattr(module, 'weight_bit') or hasattr(module, 'activation_bit'):
+                calibration_config[name] = {}
+            if hasattr(module, 'weight_bit'):
+                calibration_config[name]['weight_bit'] = int(module.weight_bit)
+            if hasattr(module, 'activation_bit'):
+                calibration_config[name]['activation_bit'] = int(module.activation_bit)
+                calibration_config[name]['tracked_min'] = float(module.tracked_min_biased)
+                calibration_config[name]['tracked_max'] = float(module.tracked_max_biased)
+            del_simulated_attr(module)
+
+        torch.save(self.bound_model.state_dict(), model_path)
+        logger.info('Model state_dict saved to %s', model_path)
+        if calibration_path is not None:
+            torch.save(calibration_config, calibration_path)
+            logger.info('Mask dict saved to %s', calibration_path)
+
+        return calibration_config
+
 
     def fold_bn(self, config, **kwargs):
         # TODO simulate folded weight
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -162,30 +162,15 @@ class QAT_Quantizer(Quantizer):
                 layer.module.register_buffer('tracked_max_biased', torch.zeros(1))
                 layer.module.register_buffer('tracked_max', torch.zeros(1))
 
-    def del_simulated_attr(self, module):
+    def _del_simulated_attr(self, module):
         """
         delete redundant parameters in quantize module
         """
-        if hasattr(module, 'old_weight'):
-            delattr(module, 'old_weight')
-        if hasattr(module, 'ema_decay'):
-            delattr(module, 'ema_decay')
-        if hasattr(module, 'tracked_min_biased'):
-            delattr(module, 'tracked_min_biased')
-        if hasattr(module, 'tracked_max_biased'):
-            delattr(module, 'tracked_max_biased')
-        if hasattr(module, 'tracked_min'):
-            delattr(module, 'tracked_min')
-        if hasattr(module, 'tracked_max'):
-            delattr(module, 'tracked_max')
-        if hasattr(module, 'scale'):
-            delattr(module, 'scale')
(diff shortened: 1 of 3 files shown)


nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -110,6 +110,30 @@ def get_bits_length(config, quant_type):
...
+    if hasattr(module, 'tracked_max'):
+        delattr(module, 'tracked_max')
+    if hasattr(module, 'scale'):
+        delattr(module, 'scale')
+    if hasattr(module, 'zero_point'):
+        delattr(module, 'zero_point')
+    if hasattr(module, 'weight_bit'):
+        delattr(module, 'weight_bit')
+    if hasattr(module, 'activation_bit'):
+        delattr(module, 'activation_bit')
```
QuanluZhang, line 136: how do you choose these attributes? different quantization algorithms have different subset of these attributes? is it possible that a new quantization algorithm has more attributes?

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -282,6 +311,45 @@ def quantize_output(self, output, wrapper, **kwargs):
...
+        ----------
+        model_path : str
+            path to save quantized model weight
+        calibration_path : str
+            (optional) path to save quantize parameters after calibration
+        """
+        assert model_path is not None, 'model_path must be specified'
+        self._unwrap_model()
+        calibration_config = {}
+        support_op = [torch.nn.Conv2d, torch.nn.Linear, torch.nn.ReLU6]
```
QuanluZhang, line 328: why `ReLU6`?

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: This function is only for QAT now, other quantization algorithms might have other attributes and we need to define another function to handle it.

QuanluZhang: then it is better to put it in `QAT_Quantizer`

QuanluZhang: BTW, why only support export for QAT?

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Because activation layers also need to be set bit width in inference framework like **tensorrt**, we have to choose activation layers which we can support during calibration. I have tested `ReLU6` in some examples and it can be fully supported by **tensorrt**. But I am not sure whether other activation ops can be supported.

QuanluZhang: what types of ops does QAT support in our current implementation?

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Because QAT is enough for **linear quantization** simulation. **Binarized quantization** like BNN needs no calibration and users can save it weights as normal.

QuanluZhang: could you refer to model export implemented in pruners, and try to align the export feature of pruner and quantizer?

linbinskn: Have aligned `export_model()` function in pruners and quantizers.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Removed `support_op` constrain in `export_model()`.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -153,13 +152,41 @@ def __init__(self, model, config_list, optimizer=None):
...
+                layer.module.register_buffer('weight_bit', torch.zeros(1))
             if "output" in config.get("quant_types", []):
+                layer.module.register_buffer('activation_bit', torch.zeros(1))
                 layer.module.register_buffer('ema_decay', torch.Tensor([0.99]))
                 layer.module.register_buffer('tracked_min_biased', torch.zeros(1))
                 layer.module.register_buffer('tracked_min', torch.zeros(1))
                 layer.module.register_buffer('tracked_max_biased', torch.zeros(1))
                 layer.module.register_buffer('tracked_max', torch.zeros(1))
 
+    def del_simulated_attr(self, module):
```
QuanluZhang, line 165: -> _del_simulated_attr

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -153,13 +152,41 @@ def __init__(self, model, config_list, optimizer=None):
...
+        if hasattr(module, 'tracked_max'):
+            delattr(module, 'tracked_max')
+        if hasattr(module, 'scale'):
+            delattr(module, 'scale')
+        if hasattr(module, 'zero_point'):
+            delattr(module, 'zero_point')
+        if hasattr(module, 'weight_bit'):
+            delattr(module, 'weight_bit')
+        if hasattr(module, 'activation_bit'):
+            delattr(module, 'activation_bit')
```
QuanluZhang, line 188: suggest to use `for`.
to_del = ['old_weight', 'ema_decay', ...]
for each in to_del:
    if hasattr():
        delattr()

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -282,6 +311,55 @@ def quantize_output(self, output, wrapper, **kwargs):
...
+            torch.save(calibration_config, calibration_path)
+            logger.info('Mask dict saved to %s', calibration_path)
+        if onnx_path is not None:
+            assert input_shape is not None, 'input_shape must be specified to export onnx model'
+            # input info needed
+            if device is None:
+                device = torch.device('cpu')
+            input_data = torch.Tensor(*input_shape)
+            torch.onnx.export(self.bound_model, input_data.to(device), onnx_path)
+            logger.info('Model in onnx with input shape %s saved to %s', input_data.shape, onnx_path)
```
QuanluZhang, line 359: have you tested export onnx? and better to write test for this feature

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -301,6 +379,19 @@ class DoReFaQuantizer(Quantizer):
 
     def __init__(self, model, config_list, optimizer=None):
         super().__init__(model, config_list, optimizer)
+        modules_to_compress = self.get_modules_to_compress()
+        for layer, config in modules_to_compress:
+            if "weight" in config.get("quant_types", []):
+                layer.module.register_buffer('weight_bit', torch.zeros(1))
+
+    def del_simulated_attr(self, module):
```
QuanluZhang, line 387: -> _del_simulated_attr

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -391,3 +527,37 @@ def quantize_output(self, output, wrapper, **kwargs):
...
+            device of the model, used to place the dummy input tensor for exporting onnx file.
+            the tensor is placed on cpu if ```device``` is None
+        """
+        assert model_path is not None, 'model_path must be specified'
+        self._unwrap_model()
+
+        torch.save(self.bound_model.state_dict(), model_path)
+        logger.info('Model state_dict saved to %s', model_path)
+        if calibration_path is not None:
+            logger.info('No calibration config will be saved because no calibration data in BNN quantizer')
```
QuanluZhang, line 555: i think we should export bit number even they are all 1 bit. Because the speedup module should know this information to use 1 bit. the speedup module does not know you are using BNNQuantizer

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -338,6 +430,50 @@ def quantize(self, input_ri, q_bits):
...
+            the tensor is placed on cpu if ```device``` is None
+        """
+        assert model_path is not None, 'model_path must be specified'
+        self._unwrap_model()
+        calibration_config = {}
+
+        for name, module in self.bound_model.named_modules():
+            if hasattr(module, 'weight_bit'):
+                calibration_config[name] = {}
+                calibration_config[name]['weight_bit'] = int(module.weight_bit)
```
QuanluZhang, line 458: so this quantizer does not calibrate activation?

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -391,3 +527,37 @@ def quantize_output(self, output, wrapper, **kwargs):
...
+        if calibration_path is not None:
+            logger.info('No calibration config will be saved because no calibration data in BNN quantizer')
+        if onnx_path is not None:
+            assert input_shape is not None, 'input_shape must be specified to export onnx model'
+            # input info needed
+            if device is None:
+                device = torch.device('cpu')
+            input_data = torch.Tensor(*input_shape)
+            torch.onnx.export(self.bound_model, input_data.to(device), onnx_path)
+            logger.info('Model in onnx with input shape %s saved to %s', input_data.shape, onnx_path)
```
J-shang, line 563: For other implementations return a dict, add `return {}` is better?

And all `export model()` implementations seem to have mostly the same logic, can we use the implementation in `QAT_Quantizer` for all `Quantizer`? Or just specify how to construct `calibration_config` in different `Quantizer`?

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Has tested `export_model` function including pytorch state_dict() and onnx in three algorithms.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: In our current implementation of **Dorefa**, it does not quantize activation, so we don't need to calibrate activation.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Modified.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Good point! Modified.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Modified.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Make sense. Have added.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Useful suggestions! Have modified.

J-shang (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
QuanluZhang: could you double check the paper, does the paper mention how to calibrate activation?

linbinskn: After discussion, we reach an agreement that the refactor of `Dorefa` should starts after survey and it will be done in another PR. What' s more, ut related to `export_model()` has been added into code.

test/ut/sdk/test_compressor_torch.py
```diff
@@ -274,6 +274,55 @@ def test_torch_QAT_quantizer(self):
...
+            'quant_types': ['output'],
+            'quant_bits': 1,
+            'op_types': ['ReLU']
+        }]
+        config_set = [config_list_qat, config_list_dorefa, config_list_bnn]
+        quantize_algorithm_set = [torch_quantizer.QAT_Quantizer, torch_quantizer.DoReFaQuantizer, torch_quantizer.BNNQuantizer]
+
+        for config, quantize_algorithm in zip(config_set, quantize_algorithm_set):
+            model = TorchModel()
+            model.relu = torch.nn.ReLU()
```
QuanluZhang, line 309: what is this line used for?

QuanluZhang:
@linbinskn , looks good! please update doc accordingly

test/ut/sdk/test_compressor_torch.py
linbinskn: Follow the code of `test_torch_QAT_quantizer()` in ut. Change the op relu type to **ReLU** and it will match the type in config_list.

QuanluZhang (approved):

PR #3462, opened 2021-03-20
Model Speedup Refactor

zheng-ningxin:
Initial commit of Model Speedup Refactor.


nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,414 @@
...
+    indexes: torch.Tensor
+        The indexes of the sparsity that can be structurally removed.
+    remained_indexes: torch.Tensor
+        The indexes of values that need to be remained.
+    """
+    assert isinstance(t_mask, torch.Tensor)
+    shape = list(t_mask.size())
+    n_dims = len(shape)
+    dim_list = list(range(n_dims))
+    # try to reduce the mask from the 0-th dimension
```
J-shang, line 52: try to reduce the mask from the dim-th dimension

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,414 @@
...
+    in_mask = auto_infer.in_masks[0]
+    # only need the first batch of the constant
+    in_constant = auto_infer.in_constants[0][:1]
+    output_mask = auto_infer.output_mask
+    weight_mask = auto_infer.weight_mask['weight']
+    # N C K
+    pruned_in, remained_in = convert_to_coarse_mask(in_mask, 1)
+    pruned_out, remained_out = convert_to_coarse_mask(output_mask, 1)
+    n_remained_in = weight_mask.size(1) - pruned_in.size(0)
+    n_remained_out = weight_mask.size(0) - pruned_out.size(0)
```
J-shang, line 101: If i got it right, `n_remained_in == remained_in.size(0)`?

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,414 @@
...
+    ----------
+    norm : torch.nn.BatchNorm1d
+        The batchnorm module to be replace
+    auto_infer : AutoMaskInference
+        The auto mask inference object that contains the input,
+        parameter and output masks.
+
+    Returns
+    -------
+    torch.nn.BatchNorm2d
```
J-shang, line 153: torch.nn.BatchNorm1d

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,414 @@
...
+    output_mask = auto_infer.output_mask
+    # print('BN CONSTANT')
+    # print(auto_infer.in_constants[0])
+    # N, C, H, W
+    _, remained_in = convert_to_coarse_mask(in_mask, 1)
+    _, remained_out = convert_to_coarse_mask(output_mask, 1)
+    assert remained_in.size(0) == remained_out.size(0)
+
+    num_features = remained_in.size(0)
+    _logger.info("replace batchnorm2d with num_features: %d", num_features)
```
J-shang, line 167: batchnorm2d -> batchnorm1d

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: Yes, indeed.

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: fixed.

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: fixed.

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: fixed

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,381 @@
...
+    def __init__(self, module, bias):
+        super(BiasModule, self).__init__()
+        self.ori_module = module
+        self.register_buffer('speedup_bias', bias)
+
+    def forward(self, x):
+        return self.ori_module(x)[:] + self.speedup_bias
+
+
+def replace_conv2d(conv, auto_infer):
```
QuanluZhang, line 231: could you brief explain in high-level why the logic of replacing module increases a lot?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,347 @@
...
-            The dummy input for ```jit.trace```, users should put it on right device before pass in
+        dummy_input : pytorch tensor, tuple of tensor, list of tensor
+            Note: The first dimension of the dummy_input should be the batchsize.
+            The dummy input for ```jit.trace```, users should put it on the right
+            device.
         masks_file : str
             The path of user provided mask file
         map_location : str
             the device on which masks are placed, same to map_location in ```torch.load```
+            confidence: the confidence coefficient of the sparsity inference. This value is
```
QuanluZhang, line 41: wrong indent, and missed `batch_dim`

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: Basically, I enhance the replace function of the conv2d, in the original version, it can remove the whole group of filter only when this conv is depth-wise. Currently, we can remove the whole group of filters once the filters in the same groups are masked out.

nni/common/graph_utils.py
```diff
@@ -388,8 +389,14 @@ def _expand_module_node(self, node, node_name, unique_name, op_type, nodes,
...
                         else:
                             outputs.append(output_name)
+                            # if(len(outputs)>1):
+                            #     import pdb; pdb.set_trace()
+                            break
                 else:
                     outputs.append(output_name)
+                    # if(len(outputs)>1):
+                    #     import pdb; pdb.set_trace()
+                # print(_output)
```
QuanluZhang, line 399: why keep these comments?

zheng-ningxin: Removed, thanks~

nni/common/graph_utils.py
```diff
@@ -388,6 +389,7 @@ def _expand_module_node(self, node, node_name, unique_name, op_type, nodes,
                                 visited.add(successor_node)
                         else:
                             outputs.append(output_name)
+                            break
```
QuanluZhang, line 392: why add `break` here, there may be other `outputs`/`successor_node` in `nodes`?

nni/common/graph_utils.py
```diff
@@ -724,6 +726,8 @@ def _build_graph(self):
                     unique_name = module_name
                     if use_count > 0:
                         unique_name = module_name + '.%d' % use_count
+                        self.reused_module.add(unique_name)
+                        self.reused_module.add(module_name)
```
QuanluZhang, line 730: though `use_count > 0`, `node` is not necessarily reused module. you can double check when `use_count` is increased by 1

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,381 @@
...
-def replace_linear(linear, mask):
+def replace_linear(linear, auto_infer):
     """
     Parameters
     ----------
     linear : torch.nn.Linear
         The linear module to be replace
-    mask : ModuleMasks
-        The masks of this module
+    auto_infer : AutoMaskInference
```
QuanluZhang, line 79: `AutoMaskInference` not only contains masks, but also has auto inference logic? if so, i suggest to make this input argument more concrete, that is, only pass the masks it needs

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,381 @@
...
+def no_replace(module, auto_infer):
     """
     No need to replace
     """
     _logger.debug("no need to replace")
     return module
 
 
-def replace_linear(linear, mask):
+def replace_linear(linear, auto_infer):
```
QuanluZhang, line 73: it would be better to briefly describe what types of sparsity this function supports. in high-level, coarse-grained sparsity

nni/compression/pytorch/speedup/compress_modules.py
QuanluZhang: more specific, the input argument `1` in `convert_to_coarse_mask` determines the sparsity type that this function can handle

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,381 @@
...
+        tmp_weight_data = torch.index_select(
+            linear.weight.data, 0, remained_out)
+        new_linear.weight.data = torch.index_select(
+            tmp_weight_data, 1, remained_in)
+
+        if linear.bias is not None:
+            new_linear.bias.data = torch.index_select(
+                linear.bias.data, 0, remained_out)
+
+    if NEED_FOLD_BIAS and torch.sum(in_constant) > 0:
```
QuanluZhang, line 124: let's discuss more about bias folding. can this type of folding be easily generalized to other operators?

nni/compression/pytorch/speedup/compress_modules.py
QuanluZhang: maybe you can add a member function for `AutoMaskInference` to return all the masks as a tuple or dict

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: added

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: Nice catch, I agree with you. However, the replace function also need the `in_constants` in the `auto_infer` (bias folding is integrated with the replace function), so I guess we need figure out how should we deal with the `Fold_bias` feature first, if we split it from the replace functions, then we can modify the interface as you said. We can have a quick call when you are available.

zheng-ningxin: updated.

nni/common/graph_utils.py
```diff
@@ -71,7 +71,11 @@ def __init__(self, model=None, dummy_input=None, traced_model=None):
     def _trace(self, model, dummy_input):
         training = model.training
         model.eval()
-        self.trace = torch.jit.trace(model, dummy_input)
+        kw_args = {}
+        if torch.__version__ >= '1.6.0':
+            # only pytorch with version greater than 1.6.0 has the strict option
+            kw_args['strict'] = False
+        self.trace = torch.jit.trace(model, dummy_input, **kw_args)
```
QuanluZhang, line 78: does `strict=False` mean the same behavior as the jit.trace in the versions before 1.6.0?

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,364 @@
...
     """
     _logger.debug("no need to replace")
     return module
 
 
-def replace_linear(linear, mask):
+def replace_linear(linear, masks):
     """
+    This function will replace the original linear according to
+    the infered masks. This function support the fine-grained and
```
QuanluZhang, line 94: does this function support fine-grained sparsity?

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,364 @@
...
-                                 bias=linear.bias is not None)
-    new_linear.to(linear.weight.device)
-    new_linear.weight.data = torch.index_select(
-        linear.weight.data, -1, index.to(linear.weight.device))
+    in_masks, output_mask, weight_mask = masks
+    assert isinstance(linear, nn.Linear)
+    assert len(in_masks) == 1
+    assert isinstance(output_mask, torch.Tensor)
+    in_mask = in_masks[0]
+    # only need the first batch of the constant
```
QuanluZhang, line 117: this can be removed? or move to the right place

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: Yes, I update the `convert_to_coarse_mask` function, it will remove the whole column/row that happened be covered by the fine-grained sparsity.

nni/common/graph_utils.py
zheng-ningxin: strict (bool, optional) – run the tracer in a strict mode or not (default: True). Only turn this off when you want the tracer to record your mutable container types (currently list/dict) and you are sure that the container you are using in your problem is a constant structure and does not get used as control flow (if, for) conditions.

The `strict` parameter enables the list of tensors and dict of tensors which is commonly used in detection model, in the version older than 1.6.0, it cannot support the list/dict as the input.

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: fixed, thanks.

nni/common/graph_utils.py
zheng-ningxin: Actually, this is to remove duplicated output tensors which caused by a legacy problem(use list to save the output name rather than set, and we should use the list indeed because it preserve the order of the input parameters).
I have updated the code and use a better way to remove the duplication in the outputs, thanks.

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: removed, thanks.

nni/common/graph_utils.py
zheng-ningxin: `use_count>1` happens only when there are more than one separated subgraphs that have the same `module_name`. Note that this module is a leaf module of the model, therefore, this leaf module is used more than once in the model graph(reused).

nni/compression/pytorch/speedup/compress_modules.py
QuanluZhang: got it

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
 
 class ModelSpeedup:
     """
-    This class is to speedup the model with provided weight mask
+    This class is to speedup the model with provided weight mask.
     """
 
-    def __init__(self, model, dummy_input, masks_file, map_location=None):
+    def __init__(self, model, dummy_input, masks_file, map_location=None,
+                batch_dim=0, confidence=8, fold_bias=False):
```
QuanluZhang, line 27: remove `fold_bias`

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+        self.fold_bias = fold_bias
 
-        For a module:
-            Infer its input and output shape from its weight mask
-            Infer its output shape from its input shape
-            Infer its input shape from its output shape
 
-        If its input shape is changed, continue infering its predecessors
-        If its output shape is changed, continue infering its successors
+    def _random_model_input(self, dummy_input, confidence, batch_dim):
```
QuanluZhang, line 77: suggest to return `dummy_input` and `device` then assign them to `self.dummy_input` and `self.device`. it is better than implicitly updating these two member variables within this function

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,352 @@
...
+            The number of examples used to infer the mask.
+        fix_conflict_by_padding: bool
+            If this flag is enabled, we will modify the network architecture to resolve
+            the sparsity conflict. Compared to modifying the mask to resolve the conflict,
+            the size of the speedup model in this method will be smaller, however,
+            the inference speed may be slower than the first method. Currently, we donnot
+            recommand this feature as default.
         """
+        assert confidence > 1
         from nni.common.graph_utils import build_module_graph
```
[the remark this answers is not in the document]
QuanluZhang: yes, can move it to the top

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+                input_shape[batch_dim] = confidence
+                self.dummy_input[in_name] = rand_like_with_shape(
+                    input_shape, t_input)
+        else:
+            raise TypeError(input_errmsg)
 
+    def _prepare_dummy_input(self, node):
+        """
+        Prepare the dummy_input for the auto mask inference.
         Parameters
```
QuanluZhang, line 119: add blank line here

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
-        in_shape : ModuleMasks
-            Input shape of this node
-        out_shape : ModuleMasks
-            Output shape of this node
-        """
-        input_cmask = output_cmask = None
-        if module_name in self.inferred_masks:
-            module_masks = self.inferred_masks[module_name]
+        node: NodePyGroup
+        Returns
```
QuanluZhang, line 122: add blank line here

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+        -------
+        dummy_input: list
+            List of tensors that will be used as input for the target node.
```
QuanluZhang, line 125: there are two returns, but you only wrote one here

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+                # node, such as `conv`.
+                # This is caused by the `meage_module_node` function in the
+                # _graph_utils.py, because it doesn't merge the prim::GetAttr
+                # node into the key node. In current version of _graph_utils.py,
+                # we will only merge the nodes that have same scope name, however,
+                # the scope name of the correponding prim::GetAttr node of `weight` tensor
+                # is None.
+                continue
+            # TODO why detach??
+            # TODO what if a list/tuple of tensor
```
QuanluZhang, line 155: do you want to resolve these two "TODO"s?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+
+        # this name is consistent with the name returned by named_modules()
+        module_name = node.name
+        _logger.info('Update mask for %s', module_name)
+        unique_name = node.unique_name
+        dummy_input, input_debugname = self._prepare_dummy_input(node)
+        # get the input mask from self.masks
+        # Note: the input mask of the successor nodes are
+        # already created by the predecessor node
+        in_masks = [self.masks[debugname] for debugname in input_debugname]
```
QuanluZhang, line 175: do you deal with weight here?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+        # update the mask tensor and the internal output of the submodules
+        # after manually unpack the tuple/list of tensors, the number of the outputs
+        # of each node should always be one(Except for the TupleUnpack node at the end
+        # of the whole model)
+        assert len(node.outputs) == 1, 'The number of the output should be one after the Tuple unpacked manually'
+
+        out_debugname = node.outputs[0]
+        # update the output mask into self.masks
+        self.masks[out_debugname] = _auto_infer.output_mask
+        self.constant[out_debugname] = _auto_infer.out_constant
```
QuanluZhang, line 215: what is the meaning of `out_constant`?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+        assert isinstance(c_node, torch._C.Value), errmsg
+        if isinstance(c_node.type(), torch._C.TensorType):
+            shape = tuple(c_node.type().sizes())
+            dtype = c_node.type().scalarType()
+            # TODO should use a more general way to get the input
+            # TODO ugly code here
+            if dtype.startswith('Float') or dtype.startswith('Double'):
+                return torch.rand(shape).to(self.device)
+            else:
+                return torch.randint(0, 10, shape, device=self.device)
```
QuanluZhang, line 263: why limit it to 0-10?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,336 @@
...
+        self.masks[module_name] = _auto_infer.weight_mask
+
+    def update_indirect_sparsity(self, node):
+        """
+        update the indirect sparisty for the target node.
+        """
+        module_name = node.name
+        _logger.info('Update indirect sparsity for %s', module_name)
+        unique_name = node.unique_name
+        if unique_name in self.auto_inferences and self.auto_inferences[unique_name] is not None:
```
QuanluZhang, line 230: in what cases that this condition is not met?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -148,40 +341,128 @@ def replace_compressed_modules(self):
...
-        _, conv_prune_dim = fix_mask_conflict(self.masks, self.bound_model, self.dummy_input)
-        set_conv_prune_dim(conv_prune_dim)
+        _logger.info("start to speed up the model")
+        self.initialize_speedup()
+        training = self.bound_model.training
+        # set to the evaluation mode
+        self.bound_model.train(False)
+        # TODO suppose to fix the conflict after the sparsity propagation
+        # which is more elegent
+        fix_mask_conflict(self.masks, self.bound_model, self.dummy_input)
```
QuanluZhang, line 456: do we still need this function?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -148,40 +341,128 @@ def replace_compressed_modules(self):
...
+        assert unique_name in self.auto_inferences
+        g_node = self.torch_graph.name_to_node[unique_name]
+        _logger.debug("replace %s, in %s type, with op_type %s",
+                      unique_name, g_node.type, g_node.op_type)
+        auto_infer = self.auto_inferences[unique_name]
+        if g_node.type == 'module':
+            if g_node.unique_name in self.torch_graph.reused_module:
+                if reindex_dim is not None:
+                    _logger.warning(
+                        'Cannot replace a reused module with padding operator!!')
```
QuanluZhang, line 388: what is the meaning of "padding operator"?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -148,40 +341,128 @@ def replace_compressed_modules(self):
...
+                raise RuntimeError(
+                    "Has not supported replacing the module: `{}`".format(m_type))
+            _logger.info("replace module (name: %s, op_type: %s)",
+                         g_node.name, m_type)
+            compressed_module = replace_module[m_type](leaf_module, auto_infer.get_masks())
+            new_submodule = compressed_module
+            if reindex_dim is None:
+                setattr(super_module, g_node.name.split(
+                    '.')[-1], compressed_module)
+            elif reindex_dim is not None and reindex is not None:
```
QuanluZhang, line 403: do not quite understand this branch...

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+import torch
+import torch.nn as nn
+from ..utils import randomize_tensor, torch_float_dtype, torch_integer_dtype
+_logger = logging.getLogger(__name__)
+_logger.setLevel(logging.INFO)
+
+STD_DELTA = 1e-6
+
+
+class AutoMaskInference:
```
QuanluZhang, line 14: add docstring for this class

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+        self.__apply_weight_mask()
+
+    def __apply_input_mask(self):
+        """
+        Apply the mask of the input tensor.
+        """
+        with torch.no_grad():
+            # apply the input mask
+            for tid, in_tensor in enumerate(self.dummy_input):
+                if isinstance(in_tensor, torch.Tensor) and self.in_masks[tid] is not None:
```
QuanluZhang, line 132: how about `else`?

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+                    'Only support the OP whose output is tensor/tuple of tensor/list of tensor')
+
+        # We also need random the parameters of the module, because if the weight of the model has
+        # a unmasked 0, then our out sparsity inference may be wrong
+        # However, after radomizing the weight/parameters, the constant in the output tensors may
+        # be different from the constants that calculated from its original stata_dict. However,
+        # so to get the right constant to eliminate the bias between model before and after sparsity
+        # inference, we need to reload its state_dict and recalculate the constant
+
+        # TODO can be optimized here, move the randomization at the begining
```
QuanluZhang, line 237: remove this comment?

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+        if len(self.weights) > 0 and self.state_dict is not None:
```
QuanluZhang, line 238: let's discuss this part

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+            elif isinstance(out, (list, tuple)):
+                constant = []
+                for i, tout in enumerate(out):
+                    _tmp = torch.zeros_like(tout)
+                    sparsity_pos = out_mask[i] == 0
+                    _tmp[sparsity_pos] = tout[sparsity_pos]
+                    constant.append(_tmp)
+
+        return out_mask, constant
+        # return out_mask
```
QuanluZhang, line 257: remove

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+        Find those hidden sparsity through gradient.
+        """
+        # Each node only update the output mask when we backwards
+        # update the output mask
+        if isinstance(self.output, torch.Tensor) and self.output.grad is not None:
+            # if output have gradient which means this node has successor
+            # nodes and the successor nodes have already update their indirect
+            # sparsity
+            # we can mask the values whose gradient is always zeros
+            gradient_sum = torch.sum(torch.abs(self.output.grad.data), dim=0)
```
QuanluZhang, line 270: you tried to deal with different `dim` values, but `dim` is fixed to 0 here. I suggest to only support dim=0 for now. BTW, are there many cases that `dim` is not 0?

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+            return
+        # Note: output maybe tensor or list/tuple of tensors
+        if isinstance(output, torch.Tensor):
+            output.backward(self.output_mask)
+        elif isinstance(output, list) or isinstance(output, tuple):
+            for tid, t_out in enumerate(output):
+                t_out.backward(self.output_mask[tid])
+
+        # update the sparsity of the paramters
+        for para_name in self.weights:
```
QuanluZhang, line 309: do we need to update the mask of its input, besides its weight?

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+            grad_zero = self.weights[para_name].grad.data == 0
+            self.weight_mask[para_name][grad_zero] = 0
+
+    def update_direct_sparsity(self):
+
+
```
QuanluZhang, line 315: remove blank lines

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,331 @@
...
+        elif isinstance(output, list) or isinstance(output, tuple):
+            for tid, t_out in enumerate(output):
+                t_out.backward(self.output_mask[tid])
+
+        # update the sparsity of the paramters
+        for para_name in self.weights:
+            grad_zero = self.weights[para_name].grad.data == 0
+            self.weight_mask[para_name][grad_zero] = 0
+
+    def update_direct_sparsity(self):
```
QuanluZhang, line 313: why split updating direct sparsity into two functions?

nni/compression/pytorch/speedup/jit_translate.py
```diff
@@ -0,0 +1,535 @@
...
+    dim_2 = inputs[2].toIValue()
+    new_transpose = partial(torch.transpose, dim0=dim_1, dim1=dim_2)
+    return new_transpose
+
+
+def toint_python(node, speedup):
+    class ToIntModule(torch.nn.Module):
+        def forward(self, x):
+            return x.to(torch.int)
+    return ToIntModule()
```
QuanluZhang, line 242: you wrap some funcs to module but keep some others to funcs. what is the reason or guideline to do this?

nni/compression/pytorch/utils/mask_conflict.py
```diff
@@ -4,10 +4,10 @@
 import logging
 import torch
 import numpy as np
-from .shape_dependency import ChannelDependency, GroupDependency, CatPaddingDependency, InputChannelDependency
+from .shape_dependency import ChannelDependency, GroupDependency, InputChannelDependency
```
QuanluZhang, line 7: let's discuss this dependency and conflict

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: fixed.

nni/compression/pytorch/speedup/compressor.py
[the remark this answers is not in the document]
zheng-ningxin: updated~ thanks.

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: updated~

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: `out_constant` is the constant values of the corresponding output mask which is used as the input constants for the following layers.

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: It may fail on "ReLU6" when this range is too large.

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: Yes. Basically, we have two ways to resolve the mask conflict:(1) `fix_mask_conflict` (2) `padding/reindex the module`(currently removed from this pr). With the second way(in the future), we may achieve an automatic conflict fixing without analyzing the model architecture.

nni/compression/pytorch/speedup/infer_mask.py
zheng-ningxin: added, thanks.

nni/compression/pytorch/speedup/infer_mask.py
zheng-ningxin: if one of the inputs is not in `self.in_masks`, it means it may be a constant values, such as, `prim::GetAttr` or `prim::Constant`.

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,333 @@
...
+            dummy_input.append(self.internal_result[_input].detach())
+            debugnames.append(_input)
+
+        return dummy_input, debugnames
+
+    def update_direct_sparsity(self, node):
+        """
+        Update the mask for the target node.
+        """
+        AutoMaskInferenceClass = AutoMaskInference
```
J-shang, line 164: why we trans the name

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -3,196 +3,363 @@
...
+    'AvgPool2d': lambda module, masks: no_replace(module, masks),
+    'AdaptiveAvgPool2d': lambda module, masks: no_replace(module, masks),
+    'ReLU': lambda module, masks: no_replace(module, masks),
+    'ReLU6': lambda module, masks: no_replace(module, masks),
+    'LeakyReLU': lambda module, masks: no_replace(module, masks),
+    'ELU': lambda module, masks: no_replace(module, masks),
+    'Hardtanh': lambda module, masks: no_replace(module, masks),
+    'Hardsigmoid': lambda module, masks: no_replace(module, masks),
+    'LogSigmoid': lambda module, masks: no_replace(module, masks),
+    'PReLU': lambda module, masks: no_replace(module, masks),
```
J-shang, line 25: need replace

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -1,143 +1,333 @@
...
-                self.infer_module_mask(_module_name, module_name, out_shape=input_cmask)
-        if output_cmask:
-            successors = self.torch_graph.find_successors(module_name)
-            for _module_name in successors:
-                self.infer_module_mask(_module_name, module_name, in_shape=output_cmask)
+        self.masks[module_name] = _auto_infer.weight_mask
+
+    def update_indirect_sparsity(self, node):
+        """
+        update the indirect sparisty for the target node.
```
QuanluZhang, line 224: improve docstring

nni/compression/pytorch/speedup/compressor.py
QuanluZhang: maybe add warning in `else` branch

nni/compression/pytorch/speedup/compressor.py
QuanluZhang: add comments to explain why 0-10

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -148,40 +338,130 @@ def replace_compressed_modules(self):
...
+        Replace the submodule according to the inferred sparsity.
+        unique_name: str
+            The unique_name of the submodule to replace.
+        reindex_dim: int
+            The dimension of the re-index operation.
+        reindex: Reindex
+            The index tensor. Normally this variable is None. If we want to reindex the
+            output of this submodule, we can pass the index by this parameter.
+        """
+        class ReindexModule(nn.Module):
```
QuanluZhang, line 356: remove

nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -0,0 +1,365 @@
...
+            # calculate the std of the output among batch dimension
+            std = torch.std(tout, dim=0)
+            # calculate the mean value of the output among the batch dimension
+            mean = torch.mean(tout, dim=0)
+            mask_pos = std < STD_DELTA
+            out_mask[:, mask_pos] = 0
+            constant[:, mask_pos] = mean[mask_pos]
+        return out_mask, constant
+
+    def clac_out_sparsity(self):
```
QuanluZhang, line 235: calc

nni/compression/pytorch/speedup/infer_mask.py
QuanluZhang: separate the pass for mask inference and constant inference

nni/compression/pytorch/speedup/infer_mask.py
QuanluZhang: please add comments here to explain where is input_mask

nni/compression/pytorch/speedup/jit_translate.py
QuanluZhang: reorganize this file to two classes: func and module

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: fixed.

nni/compression/pytorch/speedup/compress_modules.py
zheng-ningxin: fixed.

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: updated.

nni/compression/pytorch/speedup/jit_translate.py
zheng-ningxin: fixed.

nni/compression/pytorch/speedup/infer_mask.py
zheng-ningxin: updated

docs/zh_CN/Compression/CompressionReference.rst
```diff
@@ -140,9 +140,6 @@ Quantizers
 ..  autoclass:: nni.compression.pytorch.utils.shape_dependency.GroupDependency
     :members:
 
-..  autoclass:: nni.compression.pytorch.utils.mask_conflict.CatMaskPadding
-    :members:
-
```
J-shang, line 145: I don't know if we can directly modify the CN doc in this place, please check this. And others look great, I will approve this pr.

J-shang (approved):

docs/zh_CN/Compression/CompressionReference.rst
zheng-ningxin: We cannot directly modify the CN docs, have reverted the modification.

QuanluZhang (approved):
looks great

PR #3464, opened 2021-03-22
[Retiarii] Export topk models

ultmaster:


nni/retiarii/experiment/pytorch.py
```diff
@@ -257,16 +259,30 @@ def stop(self) -> None:
         self._dispatcher_thread = None
         _logger.info('Experiment stopped')
 
-    def export_top_models(self, top_n: int = 1):
+    def export_top_models(self, top_k: int = 1, optimize_mode: str = 'maximize', formatter: str = 'code') -> Any:
```
QuanluZhang, line 262: it is better to obtain `optimize_mode` from strategy

nni/retiarii/execution/api.py
```diff
@@ -43,6 +44,12 @@ def submit_models(*models: Model) -> None:
     engine.submit_models(*models)
 
 
+def list_models(*models: Model) -> Iterable[Model]:
```
liuzhe-lz, line 47: I think it's good to make return type concrete, unless it's likely to change in future.

liuzhe-lz (approved):

nni/retiarii/execution/api.py
ultmaster: I think iterable is fine. Some execution engine might favor to return a generator since there are a lot of models.

nni/retiarii/experiment/pytorch.py
ultmaster: This requires re-designing of strategy as some strategy does not need to know what mode it is by design (e.g., random). I'll add a comment that `optimize_mode` is likely to be removed and defined in strategy in future.

QuanluZhang (approved):

PR #3465, opened 2021-03-22
init dngo tuner generally

98may:


98may:
todo:
   explore more search space, so that values can be other than number but strings etc

PR #3472, opened 2021-03-23
Fix env folder

SparkSnail:
1. Move trials folder under env/{envId} folder to map trial jobs
2. Remove remote mode dup env folder


nni/tools/trial_tool/trial.py
```diff
@@ -47,7 +47,7 @@ def run(self):
 
         nni_log(LogType.Info, "%s: start to run trial" % self.name)
 
-        trial_working_dir = os.path.realpath(os.path.join(os.curdir, "..", "..", "trials", self.id))
+        trial_working_dir = os.path.realpath(os.path.join(os.curdir, "trials", self.id))
```
J-shang, line 50: There is a concern that if users want to find their outputs, they need know the trail's runner. And do we have a place to save these mapping?

SparkSnail: Yes, already add trials' envId and platform information in `nnictl trial ls` and webui.

J-shang: yes, I find `envId`, but it seems there no place to store the `runnerId`, that I need it to splice path like `expId/envs/runnerId/trials/trialId`. And I am worried if the checkpoint or output is easy to find under this way. Before, I know it is under `expId/trials/trialId`, but now I need a way to get `runnerId`.

J-shang: Oh, I find we saved `envName`, but no `envId`, I made a mistake upstairs.

J-shang (approved):

liuzhe-lz (approved):

PR #3474, opened 2021-03-25
[x] fix issue “The path argument must be of type string“

Lijiaoa:
when PR https://github.com/microsoft/nni/pull/3449 merged, build webui find that `yarn start` is successful, but the webportal show that issue:

image

It is resolved by **this proposal**: (related link: https://stackoverflow.com/questions/60234640/typeerror-err-invalid-arg-type-the-path-argument-must-be-of-type-string-re/62287792#62287792)
image


Lijiaoa:
image

It cause another issue, webpack 热更新失效。It is very unfriendly for development, so I want to set `react-dev-utils` to `^9.1.0` back.

Lijiaoa:
in latest commit, i had set `react-dev-utils` version to `^9.0.0` from `^11.0.4`

Lijiaoa:
close for another way to roback PR #3449

PR #3476, opened 2021-03-25
[Retiarii] support debug mode for easy debuggability

QuanluZhang:
- [x] support local debug mode. This mode only runs one trial locally, specifically, it applies mutators (default to choose the first candidate for the choices) to generate a new model, then run this model locally.
- [x] make strategy the main thread for easily debugging user customized strategies
- [x] update doc accordingly

### DIFF
--- a/nni/retiarii/strategy/local_debug_strategy.py
+++ b/nni/retiarii/strategy/local_debug_strategy.py
@@ -5,7 +5,6 @@ import logging
 import os
 import random
 import string
-import time
 
 from .. import Sampler, codegen, utils
 from ..execution.base import BaseGraphData
--- a/nni/retiarii/strategy/local_debug_strategy.py
+++ b/nni/retiarii/strategy/local_debug_strategy.py
@@ -1,0 +1,42 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import logging
+import os
+import random
+import string
+import time
+
+from .. import Sampler, codegen, utils
+from ..execution.base import BaseGraphData
+from .base import BaseStrategy
+
+_logger = logging.getLogger(__name__)
+
+class ChooseFirstSampler(Sampler):
+    def choice(self, candidates, mutator, model, index):
+        return candidates[0]
+
+class LocalDebugStrategy(BaseStrategy):
+
+    def run_one_model(self, model):
+        graph_data = BaseGraphData(codegen.model_to_pytorch_script(model), model.evaluator)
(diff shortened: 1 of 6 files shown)


nni/retiarii/strategy/local_debug_strategy.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def run_one_model(self, model):
+        graph_data = BaseGraphData(codegen.model_to_pytorch_script(model), model.evaluator)
+        random_str = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(6))
+        file_name = f'_generated_model/{random_str}.py'
+        os.makedirs(os.path.dirname(file_name), exist_ok=True)
+        with open(file_name, 'w') as f:
+            f.write(graph_data.model_script)
+        model_cls = utils.import_(f'_generated_model.{random_str}._model')
+        graph_data.evaluator._execute(model_cls)
```
ultmaster, line 29: I don't recommend expose internal logics (e.g., `BaseGraphData`) to a strategy. But it's fine to keep it now. I'll rethink about it when I implement PurePythonExecEngine later.

QuanluZhang: thanks!

liuzhe-lz:
My major concern is that the feature is split into two modules, experiment and strategy, which increases the complexity and in my opinion, is unnecessary.
`LocalDebugStrategy` IS a strategy, so it seems intuitive to create a normal experiment with that strategy. But that's not the expected way to use it. Similarly, user may set a strategy before invoking `local_debug_run()`, and the strategy will be silently ignored.
Since current `RetiariiExperiment` API hides `strategy.run()` from end user, my suggestion is to define a "local debug experiment" and put all logic into that class. And in case we changed experiment API to let user invoke `strategy.run()` directly, I would suggest to make it a static method.

QuanluZhang:
> My major concern is that the feature is split into two modules, experiment and strategy, which increases the complexity and in my opinion, is unnecessary.
> `LocalDebugStrategy` IS a strategy, so it seems intuitive to create a normal experiment with that strategy. But that's not the expected way to use it. Similarly, user may set a strategy before invoking `experiment.local_debug_run()`, and the strategy will be silently ignored.
> Since current `RetiariiExperiment` API hides `strategy.run()` from end user, my suggestion is to define a "local debug experiment" and put all logic into that class or function. And in case we changed experiment API to let user invoke `strategy.run()` directly, I would suggest to make debug mode a static method.

good points. To me, the current interface is the most convenient as it only requires changing one line of code.

liuzhe-lz (approved):

ultmaster (approved):

PR #3487, opened 2021-03-29
[Model Compression / TensorFlow] Support exporting pruned model

liuzhe-lz:


nni/compression/tensorflow/compressor.py
```diff
@@ -167,6 +211,43 @@ def compress(self):
...
+            Because Keras cannot save tensors without a ``Model``,
+            this will create a model, set all masks as its weights, and then save that model.
+            Masks in saved model will be named by corresponding layer name in compressed model.
+
+        Returns
+        -------
+        None
+        """
+        _logger.info('Saving model to %s', model_path)
+        input_shape = self.compressed_model._build_input_shape  # cannot find a public API
```
QuanluZhang, line 234: where is `_build_input_shape` assigned?

nni/compression/tensorflow/compressor.py
```diff
@@ -167,6 +211,43 @@ def compress(self):
...
+        Parameters
+        ----------
+        model_path : path-like
+            The path passed to ``Model.save()``.
+            You can use ".h5" extension name to export HDF5 format.
+        mask_path : path-like or None
+            Export masks to the path when set.
+            Because Keras cannot save tensors without a ``Model``,
+            this will create a model, set all masks as its weights, and then save that model.
+            Masks in saved model will be named by corresponding layer name in compressed model.
```
QuanluZhang, line 227: so even the mask file has different format between pytorch and tensorflow?...

QuanluZhang (approved):

nni/compression/tensorflow/compressor.py
```diff
@@ -233,6 +312,10 @@ def __init__(self, layer, config, pruner):
         _logger.info('Layer detected to compress: %s', self.layer.name)
 
     def call(self, *inputs):
+        self._update_weights()
+        return self.layer(*inputs)
+
+    def _update_weights(self):
```
J-shang, line 318: why we split this function?

J-shang (approved):

nni/compression/tensorflow/compressor.py
liuzhe-lz: This is tensorflow's private API.

nni/compression/tensorflow/compressor.py
liuzhe-lz: PyTorch masks are exported as PyTorch tensors. So there's no way to make TF masks compatible.

nni/compression/tensorflow/compressor.py
liuzhe-lz: Because forward should not be the only allowed place to update weights.

PR #3503, opened 2021-03-31
Add LSQ quantizer

chenbohua3:
This PR contains an implementation of LSQ quantizer (Learned Step Size Quantization, ICLR 2020, see [here](https://arxiv.org/pdf/1902.08153.pdf)). It uses gradients to update quantization scales and can achieve sound results in our production environment, especially for lower bits.

In the mnist experiment, it can get about 99.20% top1 acc. And the results on imagenet-1k is on going.


nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, model, config_list, optimizer=None):
                     types of nn.module you want to apply quantization, eg. 'Conv2d'
         """
         super().__init__(model, config_list, optimizer)
-        self.quant_grad = QATGrad
+        self.quant_grad = QATGrad.apply
```
linbinskn, line 149: Why we have to move `apply` here instead of using it directly?

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -545,4 +545,81 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
 
         self.export_model_save(self.bound_model, model_path, calibration_config, calibration_path, onnx_path, input_shape, device)
 
-        return calibration_config
\ No newline at end of file
+        return calibration_config
+
+
+class LsqQuantizer(Quantizer):
```
linbinskn, line 551: Please add docstring as the other Quantizers.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -545,4 +545,81 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+    def __init__(self, model, config_list, optimizer=None):
+        super().__init__(model, config_list, optimizer)
+        self.quant_grad = QuantForward()
+        modules_to_compress = self.get_modules_to_compress()
+        for layer, config in modules_to_compress:
+            layer.module.register_parameter("zero_point", torch.nn.Parameter(torch.Tensor([0.0])))
+            layer.module.register_parameter("scale", torch.nn.Parameter(torch.Tensor([1.0])))
+            if "weight" in config.get("quant_types", []):
+                # todo: support per-channel quantization for weight since TensorRT it for conv weight
+                q_bit = get_bits_length(config, "weight")
```
linbinskn, line 566: In current implementation, we only support single bit quantization in LsqQuantizer? Can we support mixed precision right now?

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -545,4 +545,81 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+                qmax = 2 ** (q_bit - 1) - 1
+                qmin = -2 ** (q_bit - 1)
+                init_weight_scale = layer.module.weight.data.detach().abs().mean() * 2 / (qmax ** 0.5)
+                layer.module.scale = torch.nn.Parameter(init_weight_scale)
+                layer.module.weight_qmax = qmax
+                layer.module.weight_qmin = qmin
+
+            # todo: in the origin paper, the initial value of activation is calculated from first input batch
+            if "output" in config.get("quant_types", []):
+                q_bit = get_bits_length(config, "")
```
linbinskn, line 576: Same question with single bit weight.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -545,4 +545,81 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+        """
+            A simple way to execute `round` operation with grad set to 1
+        """
+        y = x.round()
+        y_grad = x
+        return (y - y_grad).detach() + y_grad
+
+    def quantize(self, x, scale, zero_point, qmin, qmax):
+        grad_scale_factor = 1.0 / ((qmax * x.numel()) ** 0.5)
+        scale = self.grad_scale(scale, grad_scale_factor)
```
linbinskn, line 605: A little confused about the name of value and function. Can we polish naming here or in `grad_scale` function? For instance, change the second parameter name 'scale' to 'scale_factor'.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -545,4 +545,81 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+        # bias
+        old_weight = module.old_weight
+        weight = self.quantize(old_weight, module.scale, module.zero_point, module.weight_qmin, module.weight_qmax)
+        module.weight = weight
+        return weight
+
+    def quantize_output(self, output, wrapper, **kwargs):
+        module = wrapper.module
+        output = self.quantize(output, module.scale, module.zero_point, module.activation_qmin, module.activation_qmax)
+        return output
```
linbinskn, line 625: Can this quantization algorithm support exporting model and related quantization parameters? If yes, maybe we can consider adding function `export_model()` based on what parameters should export to inference framework like TensorRT.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: So it is for avoiding STE in LSQ quantizer.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -545,4 +545,81 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+
+class LsqQuantizer(Quantizer):
+    """Quantizer defined in:
+       Learned Step Size Quantization (ICLR 2020)
+       https://arxiv.org/pdf/1902.08153.pdf
+       """
+
+    def __init__(self, model, config_list, optimizer=None):
+        super().__init__(model, config_list, optimizer)
+        self.quant_grad = QuantForward()
```
linbinskn, line 559: If we keep the original forward and backward structure, the Lsq can forward as usual and backward by **STE**. In this way, will it be anything wrong? May be have something to do with the update of scale and zeropoint.

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: Yes, it is aimed at unifying the framework of quantizers with customized gradient and quantizers with auto-grad gradient. Also, use`.apply` is the way recommended by PyTorch (see [here](https://pytorch.org/tutorials/beginner/examples_autograd/two_layer_net_custom_function.html))

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: I will check it out

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: The names of functions and variables are the same as those defined in the paper. However, I think 'scale_factor' may be a better one:)

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: It seems that mixed precision of quantization is supported in this implementation since each layer has its own `q_bit`. We can achieve mixed quantization through some specific settings in `config_list` like：
```
configure_list = [{
        'quant_types': ['weight'],
        'quant_bits': 8,
        'op_types': ['Conv2d'],
        'op_names': ['features.3']
    }, {
        'quant_types': ['weight'],
        'quant_bits': 7,
        'op_types': ['Conv2d'],
        'op_names': ['features.6']
    }]
```

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: same as weight

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: There will not be anything wrong if the gradients are handled carefully. However, there exists one major limitation for the origin framework, that is, we must customize all gradients for all learnable parameters. If the gradient-based algorithms become complex, it will be troubling and  error-prone to do the customization. In this situation, I think using the auto-grad system to determine the gradient is more convenient for users.

linbinskn:
Many good points in this PR! Please test exported model on TensorRT and modify initialization of activation scale.

chenbohua3:
I have added codes about using the first batch data to initialize activation scale. After that, we can get about 99.20% top1 acc with the provided example.

Also, I have tested exporting the model to TensorRT. The transformed TensorRT model get almost the same acc with the PyTorch model.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -559,4 +559,170 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+        return weight
+
+    def quantize_output(self, output, wrapper, **kwargs):
+        module = wrapper.module
+
+        # initialize the scale
+        if self.bound_model.steps == 1:
+            qmax = module.activation_qmax
+            init_oup_scale = output.data.detach().abs().mean() * 2 / (qmax ** 0.5)
+            module.scale.data = init_oup_scale
```
linbinskn, line 663: It seems that weight and activation use the same scale in single module which means weight and activation have the same rescale parameter, and the value of scale will update by the gradient of weight and activation simultaneously. What consequence would be caused if we quantized both weight and activation of the same layer? Would it cause something wrong?

chenbohua3: Yes you are right:) Now each layer will construct input_scale/weight_scale/output_sclae according to the config setting.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -692,18 +722,18 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
-                abs_max_weight = float(module.scale * module.weight_qmax)
-                calibration_config[name]['tracked_min_input'] = -abs_max_weight
-                calibration_config[name]['tracked_max_input'] = abs_max_weight
-            if hasattr(module, 'activation_bit'):
-                calibration_config[name]['activation_bit'] = int(module.activation_bit)
-                abs_max_activation = float(module.scale * module.activation_qmax)
-                calibration_config[name]['tracked_min_activation'] = -abs_max_activation
-                calibration_config[name]['tracked_max_activation'] = abs_max_activation
+            if hasattr(module, 'input_bit'):
+                calibration_config[name]['weight_bit'] = int(module.input_bit)
```
linbinskn, line 728: Why assigning `calibration_config[name]['weight_bit']` with `module.input_bit` instead of `module.weight_bit`. If 'weight_bit' is not equal to `input_bit` when setting the config, the export result will be incorrect.

chenbohua3: According to [here](https://github.com/microsoft/nni/blob/dddf0b9efa310443a52467854dea053854f6a134/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py#L176), `weight_bit ` is used to determine whether set input tensor's dynamic ranges or not, which I think may be not appropriate. Assigning `input_bit` to `weight_bit ` here is just to be consistent with it.

linbinskn: Currently, we choose to record range of input tensor during the process of quantizing weight in the algorithm QAT. The reason why we handle it in this way is the requirement of integration with TensorRT which needs input tensor's dynamic range when setting layer precision to 8bit. So we record input dynamic range as [here](https://github.com/microsoft/nni/blob/dddf0b9efa310443a52467854dea053854f6a134/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py#L176).
And if we want to export LSQ model to TensorRT, `input dynamic range` should also be set in most situations and `input_bit` should be the same as `weight_bit`. 
However, it is still strange not to set `calibration_config[name]['weight_bit]` with `weight_bit` since we already have the value of `weight_bit` ==.

linbinskn (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: Got it. How about changing the codes like:
```
 if hasattr(module, 'weight_bit'):
     calibration_config[name]['weight_bit'] = int(module.weight_bit)
     abs_max_input = float(module.input_scale * module.input_qmax)
     calibration_config[name]['tracked_min_input'] = -abs_max_input
     calibration_config[name]['tracked_max_input'] = abs_max_input
```

linbinskn: Looks good. Completing related doc is necessary. Please refer to [overview](https://github.com/microsoft/nni/blob/master/docs/en_US/Compression/Overview.rst), [quantization](https://github.com/microsoft/nni/blob/master/docs/en_US/Compression/quantization.rst) and [Quantizer](https://github.com/microsoft/nni/blob/master/docs/en_US/Compression/Quantizer.rst). Feel free to ask me if have any questions.

linbinskn:
Looks good. Completing related doc is necessary. Please refer to overview, quantization and Quantizer. Feel free to ask me if have any questions.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -559,4 +559,200 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+        module = wrapper.module
+        # initialize the scale
+        if self.bound_model.steps == 1:
+            qmax = module.input_qmax
+            init_oup_scale = inputs[0].data.detach().abs().mean() * 2 / (qmax ** 0.5)
+            module.input_scale.data = init_oup_scale
+
+        new_input = self.quantize(inputs[0], module.input_scale, module.input_qmin, module.input_qmax)
+        list_inp = list(inputs)
+        list_inp[0] = new_input
```
J-shang, line 695: why we only quantize the first input

chenbohua3:
Docs have been added:)

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -559,4 +559,200 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
-        return calibration_config
\ No newline at end of file
+        return calibration_config
+
+
+class LsqQuantizer(Quantizer):
+    """Quantizer defined in:
+       Learned Step Size Quantization (ICLR 2020)
+       https://arxiv.org/pdf/1902.08153.pdf
+       """
```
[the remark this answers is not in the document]
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: It seem that currently the quantization framework only supports layers with single input (see [here](https://github.com/microsoft/nni/blob/7add1c694a1ab8849168523a506f2254e750b090/nni/compression/pytorch/compressor.py#L486), so is trt backend, see [here](https://github.com/microsoft/nni/blob/dddf0b9efa310443a52467854dea053854f6a134/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py#L184) ). So current implementation does not support layers with multi inputs.  It may be a better choice to modify the lsq quantizer to support layers with multi inputs after the framework supports it.

J-shang: got it, it is reasonable

J-shang (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -559,4 +559,200 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+        ----------
+        model : torch.nn.Module
+            the model to be quantized
+        config_list : list of dict
+            list of configurations for quantization
+            supported keys for dict:
+                - quant_types : list of string
+                    type of quantization you want to apply, currently support 'weight', 'input', 'output'
+                - quant_bits : int or dict of {str : int}
+                    bits length of quantization, key is the quantization type, value is the length, eg. {'weight', 8},
```
linbinskn, line 583: {'weight', 8} -> {'weight': 8}

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -559,4 +559,200 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+                layer.module.register_buffer('output_bit', torch.Tensor([q_bit]))
+                qmax = 2 ** (q_bit - 1) - 1
+                qmin = -2 ** (q_bit - 1)
+                layer.module.output_qmax = qmax
+                layer.module.output_qmin = qmin
+
+                self.optimizer.add_param_group({"params": layer.module.output_scale})
+
+            if "input" in config.get("quant_types", []):
+                # scale of activation will be initialized using the first batch data
```
linbinskn, line 623: activation -> input

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -559,4 +559,200 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
+                qmin = -2 ** (q_bit - 1)
+                layer.module.input_qmax = qmax
+                layer.module.input_qmin = qmin
+
+                self.optimizer.add_param_group({"params": layer.module.input_scale})
+
+    @staticmethod
+    def grad_scale(x, scale):
+        """
+            Used to scale the gradient
```
linbinskn, line 637: Recommend explaining this function in detail since both of reviewers were confused during reviewing this part. Whatever, I think this function is also part of key implementation of LSQ which can helps others understand the insight of this algorithm.

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: done

J-shang (approved):

linbinskn (approved):

docs/en_US/Compression/Quantizer.rst
```diff
@@ -86,6 +87,61 @@ note
...
+----
+
+LSQ Quantizer
+-------------
+
+In `LEARNED STEP SIZE QUANTIZATION <https://arxiv.org/pdf/1902.08153.pdf>`__\ , authors Steven K. Esser and Jeffrey L. McKinstry provide an algorithm to train the scales with gradients.
+
+..
+
+   We introduce a novel means to estimate and scale the task loss gradient at each weight and activation layer’s quantizer step size, such that it can be learned in conjunction with other network parameters.
```
QuanluZhang, line 99: We -> The authors

docs/en_US/Compression/Quantizer.rst
```diff
@@ -86,6 +87,61 @@ note
...
+        }, {
+            'quant_types': ['output'],
+            'quant_bits': {'output': 8,},
+            'op_names': ['relu1']
+    }]
+
+    quantizer = LsqQuantizer(model, configure_list, optimizer)
+    quantizer.compress()
+
+You can view example for more information
```
QuanluZhang, line 135: better to add a hyperlink to the example

examples/model_compress/quantization/LSQ_torch_quantizer.py
```diff
@@ -0,0 +1,145 @@
...
+        datasets.MNIST('data', train=True, download=True, transform=trans),
+        batch_size=64, shuffle=True)
+    test_loader = torch.utils.data.DataLoader(
+        datasets.MNIST('data', train=False, transform=trans),
+        batch_size=1000, shuffle=True)
+
+    model = Mnist()
+    '''you can change this to DoReFaQuantizer to implement it
+    DoReFaQuantizer(configure_list).compress(model)
+    '''
```
QuanluZhang, line 94: this comment can be removed

chenbohua3: done

docs/en_US/Compression/Quantizer.rst
chenbohua3: done

docs/en_US/Compression/Quantizer.rst
chenbohua3: done

QuanluZhang (approved):

QuanluZhang:
@chenbohua3 looks great, thanks for your contribution!

PR #3507, opened 2021-04-02
[Model Compression] Update api of iterative pruners

colorjam:
This pr reconstructs the category of pruners and improves the examples for model compression.
* Separate iterative pruners from one-shot pruners. Add necessary parameters (trainer, training_epochs, criterion) to iterative pruners. 
* Modify StructuredFilterPruner -> DependencyAwarePruner.
* Unify the models used in pruning and quantization, organize them into a folder.
* Add one example (combined_compression.py) for end-to-end compression pipeline in 7 steps. 
* Update related docs.


nni/compression/pytorch/compressor.py
```diff
@@ -433,6 +433,17 @@ def load_model_state_dict(self, model_state):
         else:
             self.bound_model.load_state_dict(model_state)
 
+    def get_pruned_weights(self, dim=0):
+        for wrapper_idx, wrapper in enumerate(self.get_modules_wrapper()):
+            weight_mask = wrapper.weight_mask
+            if len(weight_mask.size()) == 4:
+                sum_idx = (1, 2, 3) if dim == 0 else (0, 2, 3)
```
J-shang, line 440:
```
sum_idx = [0, 1, 2, 3]
sum_idx.remove(dim)
```
or if we only handle `dim=0` and `dim=1`, we could use a `boolean` to mark it.

nni/algorithms/compression/pytorch/pruning/one_shot.py
```diff
@@ -124,8 +125,11 @@ class SlimPruner(OneshotPruner):
             Optimizer used to train model
     """
 
-    def __init__(self, model, config_list, optimizer=None):
-        super().__init__(model, config_list, pruning_algorithm='slim', optimizer=optimizer)
+    def __init__(self, model, config_list, trainer, training_epochs=5, scale=0.0001):
```
J-shang, line 128: Why we never need an optimizer in this pruner?

nni/compression/pytorch/compressor.py
colorjam: Thanks, fix it.

docs/en_US/Compression/CustomizeCompressor.rst
```diff
@@ -28,7 +28,7 @@ An implementation of ``weight masker`` may look like this:
            # mask = ...
            return {'weight_mask': mask}
 
-You can reference nni provided :githublink:`weight masker <nni/algorithms/compression/pytorch/pruning/structured_pruning.py>` implementations to implement your own weight masker.
+You can reference nni provided :githublink:`weight masker <nni/algorithms/compression/pytorch/pruning/structured_pruning_maskerpy>` implementations to implement your own weight masker.
```
J-shang, line 31: structured_pruning_maskerpy -> structured_pruning_masker.py ?

docs/en_US/Compression/Pruner.rst
```diff
@@ -1,14 +1,13 @@
...
+We provide several pruning algorithms that support fine-grained weight pruning and structural filter pruning. **Fine-grained Pruning** generally results in  unstructured models, which need specialized hardware or software to speed up the sparse network. **Filter Pruning** achieves acceleration by removing the entire filter. Some pruning algorithms use one-shot method that prune weights at once based on an importance metric (It is necessary to finetune the model to compensate for the loss of accuracy). Other pruning algorithms **iteratively** prune weights during optimization, which control the pruning schedule, including some automatic pruning algorithms.
```
J-shang, line 4: Filter Pruning -> One-shot Pruning

docs/zh_CN/Compression/CompressionReference.rst
```diff
@@ -34,7 +34,7 @@ module 的包装
 ..  autoclass:: nni.algorithms.compression.pytorch.pruning.weight_masker.WeightMasker
     :members:
 
-..  autoclass:: nni.algorithms.compression.pytorch.pruning.structured_pruning.StructuredWeightMasker
+..  autoclass:: nni.algorithms.compression.pytorch.pruning.structured_pruning_maskerStructuredWeightMasker
```
J-shang, line 37: no need to modify Chinese doc?

docs/zh_CN/Compression/CustomizeCompressor.rst
```diff
@@ -28,7 +28,7 @@
            # mask = ...
            return {'weight_mask': mask}
 
-参考 NNI 提供的 :githublink:`权重掩码 <nni/algorithms/compression/pytorch/pruning/structured_pruning.py>` 来实现自己的权重掩码。
+参考 NNI 提供的 :githublink:`权重掩码 <nni/algorithms/compression/pytorch/pruning/structured_pruning_maskerpy>` 来实现自己的权重掩码。
```
J-shang, line 31: Chinese doc the same

colorjam: Thanks for point out that. Fix it

docs/en_US/Compression/Pruner.rst
colorjam: *Filter Pruning* and *Fine-grained Pruning* represent the alternative methods from the perspective of pruning granularity.

examples/model_compress/pruning/basic_pruners_torch.py
```diff
@@ -256,22 +206,82 @@ def main(args):
...
-        model = nn.DataParallel(model)
+    pruner.export_model(model_path=model_path, mask_path=mask_path)
 
     if args.test_only:
         test(args, model, device, criterion, test_loader)
 
+    # Unwrap all modules to normal state
+    pruner._unwrap_model() 
+    m_speedup = ModelSpeedup(model, dummy_input, mask_path, device)
+    m_speedup.speedup_model()
```
J-shang, line 280: seems `args.speed_up` can be remove from `__main__`

examples/model_compress/quantization/QAT_torch_quantizer.py
```diff
@@ -87,7 +68,7 @@ def main():
     quantizer.compress()
 
     model.to(device)
-    for epoch in range(40):
+    for epoch in range(1):
```
J-shang, line 71: why `1`?

nni/algorithms/compression/pytorch/pruning/__init__.py
```diff
@@ -1,14 +1,15 @@
...
+from .finegrained_pruning_masker import *
+from .structured_pruning_masker import *
+from .one_shot_pruner import *
+from .iterative_pruner import *
+# from .agp_pruner import *
 from .lottery_ticket import LotteryTicketPruner
 from .simulated_annealing_pruner import SimulatedAnnealingPruner
 from .net_adapt_pruner import NetAdaptPruner
-from .admm_pruner import ADMMPruner
+# from .admm_pruner import ADMMPruner
```
J-shang, line 12: remove

nni/algorithms/compression/pytorch/pruning/__init__.py
```diff
@@ -1,14 +1,15 @@
...
 
-from .finegrained_pruning import *
-from .structured_pruning import *
-from .one_shot import *
-from .agp import *
+from .finegrained_pruning_masker import *
+from .structured_pruning_masker import *
+from .one_shot_pruner import *
+from .iterative_pruner import *
+# from .agp_pruner import *
```
J-shang, line 8: remove

nni/algorithms/compression/pytorch/pruning/admm_pruner.py
```diff
@@ -159,7 +161,7 @@ def callback():
             # step 1: optimize W with AdamOptimizer
             for epoch in range(self._training_epochs):
                 self._trainer(self.bound_model, optimizer=optimizer,
-                              criterion=criterion, epoch=epoch, callback=callback)
+                              criterion=self._criterion, epoch=epoch, callback=callback)
```
J-shang, line 164: can we handle the criterion in trainer?

nni/algorithms/compression/pytorch/pruning/constants.py
```diff
@@ -16,3 +16,5 @@
     'apoz': ActivationAPoZRankFilterPrunerMasker,
     'mean_activation': ActivationMeanRankFilterPrunerMasker
 }
+
+MAX_EPOCHS = 9999
```
J-shang, line 20: where we use this?

nni/compression/pytorch/compressor.py
```diff
@@ -7,7 +7,7 @@
 from . import default_layers
 
 _logger = logging.getLogger(__name__)
-
+MAX_EPOCHS = 9999
```
J-shang, line 10: the same in `constants.py`?

docs/en_US/Compression/Pruner.rst
J-shang: got it, so we have two categories, (Filter Pruning and Fine-grained Pruning), (Iteratively Pruning and One-shot Pruning)

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
```diff
@@ -677,13 +683,7 @@ class SlimPrunerMasker(WeightMasker):
...
     def __init__(self, model, pruner, **kwargs):
         super().__init__(model, pruner)
-        weight_list = []
-        for (layer, _) in pruner.get_modules_to_compress():
-            weight_list.append(layer.module.weight.data.abs().clone())
-        all_bn_weights = torch.cat(weight_list)
-        k = int(all_bn_weights.shape[0] * pruner.config_list[0]['sparsity'])
-        self.global_threshold = torch.topk(
-            all_bn_weights.view(-1), k, largest=False)[0].max()
+        self.global_threshold = None
```
J-shang, line 686: why we take this part in pruner? I think masker means how to calculate mask and calculate threshold is a part of the mask calculation? Pruner means when and where to calculate mask?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        pass
+
+    def _callback(self):
+        """
+        Callback function to do additonal optimization
+        """
+        pass
+
+    def calc_mask(self, wrapper, wrapper_idx=None):
```
J-shang, line 86: seems the implementation is the same in `DependencyAwarePruner`

nni/compression/pytorch/compressor.py
J-shang, line 10: `MAX_EPOCHS` is defined in `nni/algorithms/compression/pytorch/pruning/constants.py`

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,584 @@
...
+        super().__init__(model, config_list, optimizer, pruning_algorithm, dependency_aware, dummy_input, **algo_kwargs)
+
+        self.training_epochs = training_epochs
+        self._trainer = trainer
+        self._criterion = criterion
+
+    def compress(self):
+        training = self.bound_model.training
+        self.bound_model.train()
+        for epoch in range(MAX_EPOCHS):
```
J-shang, line 60: why we do not use just `for epoch in range(self.training_epochs):`

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,584 @@
...
+        dependency between the conv layers.
+    dummy_input : torch.Tensor
+        The dummy input to analyze the topology constraints. Note that, the dummy_input
+        should on the same device with the model.
+
+    """
+
+    def __init__(self, model, config_list, optimizer, trainer, criterion, training_epochs=1,
+                 dependency_aware=False, dummy_input=None):
+        super().__init__(model, config_list, optimizer=optimizer,         pruning_algorithm='taylorfo', trainer=trainer, criterion=criterion, training_epochs=training_epochs, dependency_aware=dependency_aware, dummy_input=dummy_input)
```
J-shang, line 493: remove space and add `Parameters` comments, the following pruner comments also need `Parameters` comments.

nni/algorithms/compression/pytorch/pruning/one_shot_pruner.py
```diff
@@ -0,0 +1,211 @@
...
+        dict
+            dictionary for storing masks, keys of the dict:
+            'weight_mask':  weight mask tensor
+            'bias_mask': bias mask tensor (optional)
+        """
+        if wrapper.if_calculated:
+            return None
+        print("-> one shot pruner calc mask")
+        sparsity = wrapper.config['sparsity']
+        if not wrapper.if_calculated:
```
J-shang, line 83: we have already judged whether it is `wrapper.if_calculated` at the first line in this function

examples/model_compress/pruning/basic_pruners_torch.py
colorjam: Thanks for point out that.

examples/model_compress/quantization/QAT_torch_quantizer.py
colorjam: This is modified for fast validation. Changed back to 40.

nni/algorithms/compression/pytorch/pruning/__init__.py
colorjam: Removed

nni/algorithms/compression/pytorch/pruning/admm_pruner.py
colorjam: Criterion should be defined by the user, so it has to be passed as a parameter.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
colorjam: Thanks for your suggestion, move the `_get_global_threshold` to masker. Please review the latest version.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
colorjam: Thanks for point out that. Remove the duplicate functions.

nni/compression/pytorch/compressor.py
colorjam: This constant is useless. Remove it.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
colorjam: Change to `for epoch in range(self.training_epochs):`

docs/en_US/Compression/Pruner.rst
QuanluZhang: what is the difference between "filter pruning" and "channel pruning"? "channel pruning" is also a type of "filter pruning"?

docs/en_US/Compression/Pruner.rst
```diff
@@ -1,14 +1,13 @@
...
 
 
 **Fine-grained Pruning**
-
 * `Level Pruner <#level-pruner>`__
```
QuanluZhang, line 8: is this one a one-shot pruner? if it is, it's better to put it into that category. Maybe we can add a note to it, for example, "Level Pruner (fine-grained pruning)"

examples/model_compress/combined_compression.py
```diff
@@ -0,0 +1,309 @@
+# Copyright (c) Microsoft Corporation.
```
QuanluZhang, line 1: looks good, maybe it is better to write this example using Jupyter Notebook, let's leave this task to another one, maybe @linbinskn :)

QuanluZhang: better to rename this file to "end2end_compression.py"

docs/en_US/Compression/CompressionReference.rst
```diff
@@ -43,40 +43,40 @@ Pruners
...
 
-..  autoclass:: nni.algorithms.compression.pytorch.pruning.one_shot.ActivationAPoZRankFilterPruner
+..  autoclass:: nni.algorithms.compression.pytorch.pruning.iterative_pruner.TaylorFOWeightFilterPruner
     :members:
 
-..  autoclass:: nni.algorithms.compression.pytorch.pruning.one_shot.ActivationMeanRankFilterPruner
+..  autoclass:: nni.algorithms.compression.pytorch.pruning.iterative_pruner.ActivationAPoZRankFilterPruner
     :members:
 
-..  autoclass:: nni.algorithms.compression.pytorch.pruning.lottery_ticket.LotteryTicketPruner
```
zheng-ningxin, line 73: We doesn't provide the `LotteryTicketPruner` any more?

docs/en_US/Compression/CompressionReference.rst
```diff
@@ -43,40 +43,40 @@ Pruners
...
 
-..  autoclass:: nni.algorithms.compression.pytorch.pruning.agp.AGPPruner
+..  autoclass:: nni.algorithms.compression.pytorch.pruning.iterative_pruner.AGPPruner
     :members:
 
-..  autoclass:: nni.algorithms.compression.pytorch.pruning.admm_pruner.ADMMPruner
+..  autoclass:: nni.algorithms.compression.pytorch.pruning.iterative_pruner.ADMMPruner
     :members:
 
 ..  autoclass:: nni.algorithms.compression.pytorch.pruning.auto_compress_pruner.AutoCompressPruner
```
zheng-ningxin, line 82: Why the `AutoCompressPruner` is not classified into the iterative_pruner?

docs/en_US/Compression/CompressionReference.rst
```diff
@@ -88,6 +88,9 @@ Pruners
 ..  autoclass:: nni.algorithms.compression.pytorch.pruning.simulated_annealing_pruner.SimulatedAnnealingPruner
```
zheng-ningxin, line 88: Why are some pruners neither in the one-shot pruner nor in the iterative category?  I think it's a little bit confusing.

docs/en_US/Compression/Overview.rst
```diff
@@ -26,7 +35,7 @@ The algorithms include pruning algorithms and quantization algorithms.
 Pruning Algorithms
 ^^^^^^^^^^^^^^^^^^
 
-Pruning algorithms compress the original network by removing redundant weights or channels of layers, which can reduce model complexity and address the over-ﬁtting issue. 
+Pruning algorithms compress the original network by removing redundant weights or channels of layers, which can reduce model complexity and address the over-ﬁtting issue.
```
zheng-ningxin, line 38: Address is too strong, how about `mitigate`?

docs/en_US/Compression/Pruner.rst
zheng-ningxin: Their selection rules are different, one is based on the weight of the filter(out-channel), the other is based on the weight of the input-channel. Filter pruning prunes the output channel of the conv layers and the channel pruning prunes the input channel of the conv layers. The have the same effect after the propagation( pruning an input channel is equivalent to pruning a filter of the previous conv) .

zheng-ningxin: > what is the difference between "filter pruning" and "channel pruning"? "channel pruning" is also a type of "filter pruning"?

Their selection rules are different, one is based on the weight of the filter(out-channel), the other is based on the weight of the input-channel. Filter pruning prunes the output channel of the conv layers and the channel pruning prunes the input channel of the conv layers. The have the same effect after the propagation( pruning an input channel is equivalent to pruning a filter of the previous conv) .

examples/model_compress/combined_compression.py
zheng-ningxin: I think our sample is a bit unfriendly to users（with too many input parameters). Shall we add a README to this example like AMC(https://github.com/microsoft/nni/blob/25dea8f5528be7f3f74ca2c2fa42f528f6cdd4eb/examples/model_compress/pruning/amc/README.md)?

docs/en_US/Compression/Pruner.rst
colorjam: Thanks for your suggestion. Please review the latest verison.

docs/en_US/Compression/CompressionReference.rst
colorjam: Actually, some pruners (e.g., `AutoCompressPruner` and `SimulatedAnnealingPruner`) belong to the iterative category. While the length of the code of these pruners is relatively long. So put such pruners in the independent files.

docs/en_US/Compression/Overview.rst
colorjam: Thanks for your suggestion. Have changed to `mitigate`, please review the latest version.

examples/model_compress/pruning/basic_pruners_torch.py
```diff
@@ -256,22 +206,82 @@ def main(args):
...
+        if args.pruner == 'slim':
+            config_list = [{
+            'sparsity': args.sparsity,
+            'op_types': ['BatchNorm2d'],
+        }]
+        else:
+            config_list = [{
+            'sparsity': args.sparsity,
+            'op_types': ['Conv2d'],
+            'op_names': ['feature.0', 'feature.24', 'feature.27', 'feature.30', 'feature.34', 'feature.37']
```
zheng-ningxin, line 254: Why so specific?

examples/model_compress/pruning/basic_pruners_torch.py
```diff
@@ -256,22 +206,82 @@ def main(args):
...
+    # Unwrap all modules to normal state
+    pruner._unwrap_model() 
+    m_speedup = ModelSpeedup(model, dummy_input, mask_path, device)
+    m_speedup.speedup_model()
+
+    print('start finetuning...')
     best_top1 = 0
+    save_path = os.path.join(args.experiment_data_dir, f'finetuned.pth')
     for epoch in range(args.fine_tune_epochs):
         pruner.update_epoch(epoch)
```
zheng-ningxin, line 286: what's this interface for?

nni/algorithms/compression/pytorch/pruning/admm_pruner.py
```diff
@@ -59,12 +59,14 @@ def trainer(model, criterion, optimizer, epoch, callback):
```
zheng-ningxin, line 59: Update the docstring accordingly

nni/algorithms/compression/pytorch/pruning/admm_pruner.py
```diff
@@ -59,12 +59,14 @@ def trainer(model, criterion, optimizer, epoch, callback):
 
     """
 
-    def __init__(self, model, config_list, trainer, num_iterations=30, training_epochs=5, row=1e-4, base_algo='l1'):
+    def __init__(self, model, config_list, trainer, criterion, optimizer=None, num_iterations=30, training_epochs=5, row=1e-4, base_algo='l1'):
```
zheng-ningxin, line 62: Why increase two arguments?

zheng-ningxin: How about giving `criterion` a default value (torch.nn.CrossEntropyLoss)

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,572 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class IterativePruner(DependencyAwarePruner):
+    """
+    Prune model during the training process. 
+    """
+
+    def __init__(self, model, config_list, optimizer=None, pruning_algorithm='level', trainer=None, criterion=None, training_epochs=1, dependency_aware=False, dummy_input=None, **algo_kwargs):
```
zheng-ningxin, line 24: training_epochs=1 ?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,572 @@
...
+        pruning_algorithm: str
+            algorithms being used to prune model
+        trainer: function
+            Function used to train the model.
+            Users should write this function as a normal function to train the Pytorch model
+            and include `model, optimizer, criterion, epoch, callback` as function arguments.
+        criterion: function
+            Function used to calculate the loss between the target and the output.
+        training_epochs : int
+            Totoal number of training epochs for iterative pruning.
```
zheng-ningxin, line 43: confusing.
You mean the training epochs between the pruning iteration?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,572 @@
...
+        self.patch_optimizer(self.update_mask)
+        
+        return self.bound_model
+
+
+    def _callback(self):
+        """
+        Callback function to do additonal optimization
+        """
+        pass
```
zheng-ningxin, line 75:
```
raise NotImplementedError
```

zheng-ningxin:
Suggest add an integration test/unit test(set the training epochs to 0) to cover all the pruners? Otherwise too many modification are difficult to ensure correctness.

nni/algorithms/compression/pytorch/pruning/auto_compress_pruner.py
```diff
@@ -219,7 +221,8 @@ def compress(self):
                 self._experiment_data_dir, 'model_admm_masked.pth')))
 
             masks_file = os.path.join(self._experiment_data_dir, 'mask.pth')
-            device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+            # device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
zheng-ningxin, line 224: remove the useless comments

examples/model_compress/pruning/basic_pruners_torch.py
colorjam: It aims to reproduce the result in the paper 'PRUNING FILTERS FOR EFFICIENT CONVNETS', in which prune `Conv_1, Conv_8, Conv_9, Conv_10, Conv_11, Conv_12`.

examples/model_compress/pruning/basic_pruners_torch.py
colorjam: This interface is useless after refactor. Thanks for point out that. This line has been deleted, please review the latest version.

nni/algorithms/compression/pytorch/pruning/admm_pruner.py
colorjam: Thanks! AdmmPruner is moved into `iterative_pruner.py`. Two arguments are aligned with the parameters of `IterativePruner`. The default values are added for parameters, please review the latest version.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
colorjam: An additional optimization is not necessary, which is only required in some pruners that have to do something on weights after `loss.backward()`, such as `ADMMPruner` and `SlimPruner`.

colorjam:
> Suggest to add an integration test/unit test(set the training epochs to 0) to cover all the pruners? Otherwise too many modification are difficult to ensure correctness.

@zheng-ningxin Thanks for your valuable suggestions, let's leave that in follow-up PRs.

nni/algorithms/compression/pytorch/pruning/admm_pruner.py
QuanluZhang: update docstring accordingly

nni/algorithms/compression/pytorch/pruning/dependency_aware_pruner.py
```diff
@@ -0,0 +1,159 @@
...
+from nni.compression.pytorch.compressor import Pruner
+from .constants import MASKER_DICT
+
+__all__ = ['DependencyAwarePruner']
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class DependencyAwarePruner(Pruner):
```
QuanluZhang, line 19: @zheng-ningxin is this pruner written by you? where is the previous version?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,572 @@
...
+from .constants import MASKER_DICT
+from .dependency_aware_pruner import DependencyAwarePruner
+
+__all__ = ['AGPPruner', 'ADMMPruner', 'SlimPruner', 'TaylorFOWeightFilterPruner', 'ActivationAPoZRankFilterPruner', 'ActivationMeanRankFilterPruner', ]
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class IterativePruner(DependencyAwarePruner):
```
QuanluZhang, line 19: all our iterative pruners support dependency_aware?

examples/model_compress/end2end_compression.py
```diff
@@ -0,0 +1,309 @@
...
+from nni.compression.pytorch import ModelSpeedup
+
+from nni.algorithms.compression.pytorch.pruning import L1FilterPruner
+from nni.algorithms.compression.pytorch.quantization import QAT_Quantizer
+
+from models.mnist.naive import NaiveModel
+from nni.compression.pytorch.quantization_speedup import ModelSpeedupTensorRT
+
+
+def get_model_time_cost(model, dummy_input):
```
linbinskn, line 31: A little strange that this function is used to test 100 times batch inference time rather than one batch inference time(eg. final result / n_times)?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
QuanluZhang: maybe it is not total number of training epochs during the whole pruning...

examples/model_compress/end2end_compression.py
```diff
@@ -0,0 +1,309 @@
...
+    model.eval()
+    n_times = 100
+    time_list = []
+    for _ in range(n_times):
+        torch.cuda.synchronize()
+        tic = time.time()
+        _ = model(dummy_input)
+        torch.cuda.synchronize()
+        time_list.append(time.time()-tic)
+    time_list = time_list[1:]
```
linbinskn, line 41: Why record inference time without batch num 0, if it is for cold start-up, one batch maybe not enough.

linbinskn (commented):
Part of quantization looks good.

nni/algorithms/compression/pytorch/pruning/dependency_aware_pruner.py
zheng-ningxin: Yes, it's at https://github.com/microsoft/nni/blob/9444e27508ab9d6e4184436af2661cd4efd387bd/nni/algorithms/compression/pytorch/pruning/structured_pruning.py#L70

zheng-ningxin:
> > Suggest to add an integration test/unit test(set the training epochs to 0) to cover all the pruners? Otherwise too many modification are difficult to ensure correctness.
> 
> @zheng-ningxin Thanks for your valuable suggestion, let's leave that in follow-up PRs.

I think we must add these tests before the version is released.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
zheng-ningxin: Yes, I mean we can replace the `pass`  with `raise NotImplementedError`. If a function is not implemented, then we can just raise a not implemented error.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
colorjam: not all pruners support the dependency_aware mode

colorjam:
> > > Suggest to add an integration test/unit test(set the training epochs to 0) to cover all the pruners? Otherwise too many modification are difficult to ensure correctness.
> > 
> > 
> > @zheng-ningxin Thanks for your valuable suggestion, let's leave that in follow-up PRs.
> 
> I think we must add these tests before the version 2.3 is released, if you are too busy to handle the tests, please let me know and maybe I can add several test for the pruners.

@zheng-ningxin It would be great if you could help do this.

examples/model_compress/end2end_compression.py
colorjam: Thanks for point out that. I changed the first 10 batches and averaged the inference time. Please review the latest version.

J-shang (approved):

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
QuanluZhang: maybe it means the number of pruning iterations? @colorjam

docs/en_US/Compression/Overview.rst
```diff
@@ -14,10 +14,19 @@ NNI provides a model compression toolkit to help user compress and speed up thei
...
 
+
+Compression Pipeline
+--------------------
+
+.. image:: ../../img/compression_flow.jpg
+   :target: ../../img/compression_flow.jpg
+   :alt: 
+
+The overall compression pipeline in NNI. For compress a pretrained model, pruning and quantization can be used alone or in combination.
```
linbinskn, line 25: compress -> compressing?

examples/model_compress/end2end_compression.py
```diff
@@ -0,0 +1,309 @@
...
+    for batch_idx, (data, target) in enumerate(train_loader):
+        data, target = data.to(device), target.to(device)
+        optimizer.zero_grad()
+        output = model(data)
+        loss = criterion(output, target)
+        loss.backward()
+
+        # callback should be inserted between loss.backward() and optimizer.step()
+        if callback:
+            callback()
```
QuanluZhang, line 56: @colorjam why not put the logic in `callback` into optimizer as a hook/patch?

QuanluZhang: image

nni/algorithms/compression/pytorch/pruning/auto_compress_pruner.py
```diff
@@ -100,7 +101,7 @@ def evaluator(model):
         PATH to store temporary experiment data.
     """
 
-    def __init__(self, model, config_list, trainer, evaluator, dummy_input,
+    def __init__(self, model, config_list, trainer, criterion, evaluator, dummy_input,
```
QuanluZhang, line 104: some criterion has default value but some don't have...
update docstring accordingly

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,574 @@
...
+
+    def compress(self):
+        training = self.bound_model.training
+        self.bound_model.train()
+        for epoch in range(self.training_epochs):
+            self._trainer(self.bound_model, optimizer=self.optimizer,
+            criterion=self._criterion, epoch=epoch, callback=self._callback)
+        self.bound_model.train(training)
+
+        self.update_mask()
```
QuanluZhang, line 66: is this `compress` function iterative? seems `update_mask` is put outside of the for loop

QuanluZhang: is our SlimPruner not precisely implemented?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,574 @@
...
+            current training epoch
+        """
+
+        if epoch > 0:
+            self.now_epoch = epoch
+            for wrapper in self.get_modules_wrapper():
+                wrapper.if_calculated = False
+
+    def compress(self):
+        for epoch in range(self.training_epochs):
```
QuanluZhang, line 213: does this `self.training_epochs` means `num_iterations`?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,574 @@
...
+        harvest the speed benefit from the pruned model. Note that, if this flag is set True
+        , the dummy_input cannot be None, because the pruner needs a dummy input to trace the
+        dependency between the conv layers.
+    dummy_input : torch.Tensor
+        The dummy input to analyze the topology constraints. Note that, the dummy_input
+        should on the same device with the model.
+
+    """
+
+    def __init__(self, model, config_list, optimizer=None, trainer=None, criterion=None, activation='relu',
```
QuanluZhang, line 520: could `optimizer` `trainer` `criterion` be None?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,574 @@
...
+        The dummy input to analyze the topology constraints. Note that, the dummy_input
+        should on the same device with the model.
+
+    """
+
+    def __init__(self, model, config_list, optimizer, trainer, criterion, training_epochs=1, dependency_aware=False, dummy_input=None):
+        super().__init__(model, config_list, optimizer=optimizer, pruning_algorithm='taylorfo', trainer=trainer, criterion=criterion, training_epochs=training_epochs, dependency_aware=dependency_aware, dummy_input=dummy_input)
+
+
+class ActivationAPoZRankFilterPruner(IterativePruner):
```
QuanluZhang, line 486: why this pruner should inherit `InterativePruner`?

nni/algorithms/compression/pytorch/pruning/one_shot_pruner.py
```diff
@@ -0,0 +1,174 @@
...
+        that have dependencies to prune the same channels, so the speedup module can better
+        harvest the speed benefit from the pruned model. Note that, if this flag is set True
+        , the dummy_input cannot be None, because the pruner needs a dummy input to trace the
+        dependency between the conv layers.
+    dummy_input : torch.Tensor
+        The dummy input to analyze the topology constraints. Note that, the dummy_input
+        should on the same device with the model.
+    """
+
+    def __init__(self, model, config_list, optimizer=None, trainer=None, criterion=None, dependency_aware=False, dummy_input=None):
```
QuanluZhang, line 107: why oneshot pruner also need `trainer`?

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
```diff
@@ -677,16 +678,24 @@ class SlimPrunerMasker(WeightMasker):
...
-        k = int(all_bn_weights.shape[0] * pruner.config_list[0]['sparsity'])
+        k = int(all_bn_weights.shape[0] * self.pruner.config_list[0]['sparsity'])
         self.global_threshold = torch.topk(
             all_bn_weights.view(-1), k, largest=False)[0].max()
+        print(f'set global threshold to {self.global_threshold}')
 
     def calc_mask(self, sparsity, wrapper, wrapper_idx=None):
         assert wrapper.type == 'BatchNorm2d', 'SlimPruner only supports 2d batch normalization layer pruning'
+
+        if self.global_threshold is None:
```
QuanluZhang, line 696: remove this `if` and only keep the if body

nni/compression/pytorch/compressor.py
```diff
@@ -319,8 +318,8 @@ class Pruner(Compressor):
 
     def __init__(self, model, config_list, optimizer=None):
         super().__init__(model, config_list, optimizer)
-        if optimizer is not None:
-            self.patch_optimizer(self.update_mask)
+        # if optimizer is not None:
+        #     self.patch_optimizer(self.update_mask)
```
QuanluZhang, line 322: where is `self.update_mask` placed in this version?

nni/compression/pytorch/compressor.py
```diff
@@ -433,6 +432,21 @@ def load_model_state_dict(self, model_state):
         else:
             self.bound_model.load_state_dict(model_state)
 
+    def get_pruned_weights(self, dim=0):
```
QuanluZhang, line 435: please add docstring for this function

nni/compression/pytorch/compressor.py
```diff
@@ -433,6 +432,21 @@ def load_model_state_dict(self, model_state):
...
+        for _, wrapper in enumerate(self.get_modules_wrapper()):
+            weight_mask = wrapper.weight_mask
+            mask_size = weight_mask.size()
+            if len(mask_size) == 4:
+                sum_idx = [0, 1, 2, 3]
+                sum_idx.remove(dim)
+                index = torch.nonzero(weight_mask.abs().sum(sum_idx) != 0).tolist()
+            elif len(mask_size) == 2:
+                index = torch.nonzero(weight_mask.abs().sum(1) != 0).tolist()
+            elif len(mask_size) == 1:
```
QuanluZhang, line 445: why specifically deal with 4, 2, 1. what about other values? how to deal with them?

nni/compression/pytorch/compressor.py
QuanluZhang: there is logic changes if this patch is put at the send of `compress` function

docs/en_US/Compression/CompressionReference.rst
QuanluZhang: we can put `AutoCompressPruner` into iterative in future

examples/model_compress/end2end_compression.py
QuanluZhang: if using this approach, the optimizer is modified by us, we should let user know about this.

nni/algorithms/compression/pytorch/pruning/one_shot_pruner.py
zheng-ningxin: L1FilterPruner does not need the `optimizer`, `trainer` and `criterion`.  Current interface is too complicated.

nni/algorithms/compression/pytorch/pruning/one_shot_pruner.py
```diff
@@ -0,0 +1,174 @@
...
+    Parameters
+    ----------
+    model : torch.nn.Module
+        Model to be pruned
+    config_list : list
+        Supported keys:
+            - sparsity : This is to specify the sparsity operations to be compressed to.
+            - op_types : Only Conv2d is supported in L1FilterPruner.
+    optimizer: torch.optim.Optimizer
+            Optimizer used to train model
```
zheng-ningxin, line 93: The docstring is mismatched with the interface. Please simplify the interface and update the docstring accordingly.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -0,0 +1,574 @@
...
+        if epoch > 0:
+            self.now_epoch = epoch
+            for wrapper in self.get_modules_wrapper():
+                wrapper.if_calculated = False
+
+    def compress(self):
+        for epoch in range(self.training_epochs):
+            self.update_epoch(epoch)
+            self._trainer(self.bound_model, optimizer=self.optimizer, criterion=self._criterion, epoch=epoch)
+            self.update_mask()
```
linbinskn, line 216: Is current position of function `update_mask` appropriate? I think put it before trainer is more reasonable since training is the process of finetuning after masking weight. If it was put after training, we will skip the last iteration finetuning and the results of training may not align with testing.

nni/algorithms/compression/pytorch/pruning/one_shot_pruner.py
```diff
@@ -0,0 +1,174 @@
...
+        that have dependencies to prune the same channels, so the speedup module can better
+        harvest the speed benefit from the pruned model. Note that, if this flag is set True
+        , the dummy_input cannot be None, because the pruner needs a dummy input to trace the
+        dependency between the conv layers.
+    dummy_input : torch.Tensor
+        The dummy input to analyze the topology constraints. Note that, the dummy_input
+        should on the same device with the model.
+    """
+
+    def __init__(self, model, config_list, optimizer=None, trainer=None, criterion=None, dependency_aware=False, dummy_input=None):
```
zheng-ningxin, line 137: Same with the `L1FilterPruner`, please update the interface and the docstring~

nni/algorithms/compression/pytorch/pruning/one_shot_pruner.py
```diff
@@ -0,0 +1,174 @@
...
+from schema import And, Optional, SchemaError
+from .dependency_aware_pruner import DependencyAwarePruner
+
+__all__ = ['LevelPruner', 'L1FilterPruner', 'L2FilterPruner', 'FPGMPruner']
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class OneshotPruner(DependencyAwarePruner):
```
zheng-ningxin, line 21: Why does the `OneshotPruner` inherit from the `DependencyAwarePruner`? Shouldn't the `DependencyAwarePruner` be inherited from OneShotPruner?  Not all one shot pruners can be dependency-aware(level-pruner).

J-shang: Is it possible to achieve dependency-aware for level-pruner even though it is hard to implement?

This may because both iterative pruners (slim, Taylor...) and one-shot pruner(l1, l2..) can support dependency-aware.
My plan is to add not implementation error or warning for those un-supported pruners.

zheng-ningxin: Got it.

zheng-ningxin: Level-pruner is basically impossible to achieve dependency-aware.

J-shang: Got it, and seems we only have level pruner as a special pruner that hard to achieve dependency-aware? maybe we can detach level pruner from inheriting `DependencyAwarePruner` in follow-up pr.

zheng-ningxin: It sounds good to me after your kindly explanation. No need to separate the level-pruner specifically. Please go ahead.

test/ut/sdk/test_compressor_torch.py
```diff
@@ -152,7 +152,7 @@ def test_torch_slim_pruner(self):
         config_list = [{'sparsity': 0.2, 'op_types': ['BatchNorm2d']}]
         model.bn1.weight.data = torch.tensor(w).float()
         model.bn2.weight.data = torch.tensor(-w).float()
-        pruner = torch_pruner.SlimPruner(model, config_list)
+        pruner = torch_pruner.SlimPruner(model, config_list, optimizer=None, trainer=None, criterion=None)
```
zheng-ningxin, line 155: `optimizer` `trainer` `criterion` is necessary for SlimPruner and user needs to give three more `None` every time?

test/ut/sdk/test_compressor_torch.py
```diff
@@ -202,7 +202,7 @@ def test_torch_taylorFOweight_pruner(self):
 
         model = TorchModel()
         optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.5)
-        pruner = torch_pruner.TaylorFOWeightFilterPruner(model, config_list, optimizer, statistics_batch_num=1)
+        pruner = torch_pruner.TaylorFOWeightFilterPruner(model, config_list, optimizer, trainer=None, criterion=None, training_epochs=1)
```
zheng-ningxin, line 205: Same problem, this is very unfriendly to the users.

test/ut/sdk/test_compressor_torch.py
```diff
@@ -165,7 +165,7 @@ def test_torch_slim_pruner(self):
         config_list = [{'sparsity': 0.6, 'op_types': ['BatchNorm2d']}]
         model.bn1.weight.data = torch.tensor(w).float()
         model.bn2.weight.data = torch.tensor(w).float()
-        pruner = torch_pruner.SlimPruner(model, config_list)
+        pruner = torch_pruner.SlimPruner(model, config_list, optimizer=None, trainer=None, criterion=None)
```
zheng-ningxin, line 168: Too complicated, please use the default value to simplify the usage. Thanks~

ultmaster (approved):

QuanluZhang (approved):

linbinskn (approved):

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
J-shang: This is mainly because some pruners need to get some information during training, like `TaylorFOWeightFilterPruner`, so we `update_mask` after `train`.

And maybe we need to discuss if users should finetune the pruned model by themselves.

nni/compression/pytorch/compressor.py
J-shang: This mainly affects pruners which handle optimizer, i.e. `TaylorFOWeightFilterPruner`.
In the previous version, `update_mask` happened at the beginning of the finetune stage (outside `compress()`) after the first `optimizer.step()`.
Now, `update_mask` happened after a sparsifying training iteration (inside `compress()`).

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
J-shang: already renaming `self.training_epochs` -> `self.num_iterations` and `self.epochs_per_iteration`

J-shang:
Need follow-up:
* check compression doc and update expired content.
* `TaylorFOWeightFilterPruner`, `ActivationAPoZRankFilterPruner`, `ActivationMeanRankFilterPruner` check these pruners if need multi-epoch training.
* `SlimPruner` check if we need support multi-iteration.
* `AutoCompressPruner` `ADMMPruner` `SimulatedAnnealingPruner` check the relationship between these pruner.
* test `patch_optimizer_before()` (callback)
* `AGP` and `ADMM` may need to refactor `compress()`.

PR #3508, opened 2021-04-02
[Retiarii] Nest `ValueChoice` in `LayerChoice` and dict/list in `ValueChoice`

ultmaster:


nni/retiarii/nn/pytorch/api.py
```diff
@@ -279,11 +281,29 @@ def forward(self):
...
+            v = value
+            for a in self._accessor:
+                v = v[a]
+        except KeyError:
+            raise KeyError(''.join([f'[{a}]' for a in self._accessor]) + f' does not work on {value}')
+        return v
+
+    def __getitem__(self, item):
+        access = copy.deepcopy(self)
+        access._accessor.append(item)
```
QuanluZhang, line 302: why use `append`?

nni/retiarii/converter/graph_gen.py
```diff
@@ -590,7 +581,13 @@ def convert_module(self, script_module, module, module_name, ir_model):
         if original_type_name in MODULE_EXCEPT_LIST:
             pass  # do nothing
         elif original_type_name == OpTypeName.LayerChoice:
-            m_attrs = self._handle_layerchoice(module)
+            graph = Graph(ir_model, -100, module_name, _internal=True)  # graph_id is not used now
+            candidate_name_list = [f'layerchoice_{module.label}_{cand_name}' for cand_name in module.names]
+            for cand_name, cand in zip(candidate_name_list, module):
+                cand_type = '__torch__.' + get_importable_name(cand.__class__)
+                graph.add_node(cand_name, cand_type, get_init_parameters_or_fail(cand))
```
QuanluZhang, line 588: this subgraph is a little strange because it has no connection, which would be not friendly to graph visualization

nni/retiarii/nn/pytorch/mutator.py
```diff
@@ -14,13 +14,20 @@ def __init__(self, nodes: List[Node]):
...
+        chosen = self.choice(candidates)
         for node in self.nodes:
-            target = model.get_node_by_name(node.name)
-            chosen_cand = node.operation.parameters['candidates'][chosen_index]
-            target.update_operation(chosen_cand['type'], chosen_cand['parameters'])
+            target = model.graphs[node.operation.cell_name]
+            chosen_node = target.get_node_by_name(chosen)
+            assert chosen_node is not None
+            target.add_edge((target.input_node, 0), (chosen_node, None))
+            target.add_edge((chosen_node, None), (target.output_node, None))
```
QuanluZhang, line 24: yeah, you add connections here.

nni/retiarii/nn/pytorch/mutator.py
```diff
@@ -14,13 +14,20 @@ def __init__(self, nodes: List[Node]):
...
-        n_candidates = len(self.nodes[0].operation.parameters['candidates'])
-        indices = list(range(n_candidates))
-        chosen_index = self.choice(indices)
+        candidates = self.nodes[0].operation.parameters['candidates']
+        chosen = self.choice(candidates)
         for node in self.nodes:
-            target = model.get_node_by_name(node.name)
-            chosen_cand = node.operation.parameters['candidates'][chosen_index]
-            target.update_operation(chosen_cand['type'], chosen_cand['parameters'])
+            target = model.graphs[node.operation.cell_name]
```
QuanluZhang, line 20: I suggest to add more comments to describe that this node is a cell.

nni/retiarii/nn/pytorch/mutator.py
```diff
@@ -61,20 +68,14 @@ def __init__(self, nodes: List[Tuple[Node, str]], candidates: List[Any]):
     def mutate(self, model):
         chosen = self.choice(self.candidates)
         for node, argname in self.nodes:
+            chosen_value = node.operation.parameters[argname].access(chosen)
             target = model.get_node_by_name(node.name)
-            target.update_operation(target.operation.type, {**target.operation.parameters, argname: chosen})
+            target.update_operation(target.operation.type, {**target.operation.parameters, argname: chosen_value})
```
QuanluZhang, line 73: seems you deal with indices of ValueChoice when ValueChoice is used as initialization parameters. what if the indices are used in `forward` function?

nni/retiarii/converter/graph_gen.py
```diff
@@ -557,7 +547,8 @@ def _handle_inputchoice(self, module):
     def _handle_valuechoice(self, module):
         return {
             'candidates': module.candidates,
-            'label': module.label
+            'label': module.label,
+            'accessor': module._accessor
```
liuzhe-lz, line 551: Could you explain the idea of accessor?

nni/retiarii/converter/graph_gen.py
ultmaster: I don't think base graph should be visualized.

nni/retiarii/nn/pytorch/api.py
ultmaster: The underlying implementation is to clone the current instance, and append item to "accessor", which records all the history getitem calls. For example, when accessor is ``[a, b, c]``, the value choice will return `vc[a][b][c]` where `vc` is the original value choice.

nni/retiarii/nn/pytorch/mutator.py
ultmaster: You can find this case in the tests.

nni/retiarii/converter/graph_gen.py
ultmaster: Please refer to newly-added comments and response to Quanlu.

nni/retiarii/nn/pytorch/api.py
QuanluZhang: wow, very fancy, supporting multi-level indices. but does valuechoice support multi-level choice?

nni/retiarii/nn/pytorch/mutator.py
QuanluZhang: yes, in test you uses valuechoice as a function's parameter. What if it is used naively? for example, `param = vc[0] + 0.1`? does it work?

nni/retiarii/converter/graph_gen.py
QuanluZhang: ok

nni/retiarii/nn/pytorch/api.py
ultmaster: I think I checked that. Will throw an exception in case the multi-level choice on any of the candidates does not work.

nni/retiarii/nn/pytorch/mutator.py
ultmaster: Test added. It works.

QuanluZhang: haha, lihaile!

QuanluZhang (approved):

liuzhe-lz (approved):

PR #3522, opened 2021-04-11
Cast quniform and qloguniform values to be float

g1910:
Hyperopt returns values in np.float64 for quniform and qloguniform which causes deduplication at line 280 "if total_params in self.total_data.values():" to fail. They need to be cast to native float.

### DIFF
--- a/nni/algorithms/hpo/hyperopt_tuner.py
+++ b/nni/algorithms/hpo/hyperopt_tuner.py
@@ -80,7 +80,7 @@ def json2parameter(in_x, parameter, name=NodeType.ROOT):
                 }
             else:
                 if _type in ['quniform', 'qloguniform']:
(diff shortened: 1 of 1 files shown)


kvartet:
hello @g1910, sorry for my late reply. As you said, hyperopt returns values in ``np.float64`` for quniform and qloguniform, but the values for quniform and qloguniform in ``self.total_data`` are also  ``np.float64`` as they were added in [#L283](https://github.com/microsoft/nni/blob/master/nni/algorithms/hpo/hyperopt_tuner.py#L283). I tried some data and the mechanism of deduplication did not fail. 

The reason for the duplicate parameters in ``self.total_data`` is to avoid falling into an infinite loop, NNI does not judge whether the data randomly searched due to repetition is repeated.

QuanluZhang:
@g1910 feel free to reopen if you have any comments

PR #3526, opened 2021-04-13
Delete redundant code

maxsuren:
Delete redundant code in:
ts/nni_manager/training_service/kubernetes/frameworkcontroller/frameworkcontrollerTrainingService.ts
ts/nni_manager/training_service/kubernetes/kubeflow/kubeflowTrainingService.ts

These lines will never be reached

### DIFF
--- a/ts/nni_manager/training_service/kubernetes/frameworkcontroller/frameworkcontrollerTrainingService.ts
+++ b/ts/nni_manager/training_service/kubernetes/frameworkcontroller/frameworkcontrollerTrainingService.ts
@@ -64,7 +64,6 @@ class FrameworkControllerTrainingService extends KubernetesTrainingService imple
             await delay(3000);
             await this.fcJobInfoCollector.retrieveTrialStatus(this.kubernetesCRDClient);
             if (this.kubernetesJobRestServer.getErrorMessage !== undefined) {
-                this.stopping = true;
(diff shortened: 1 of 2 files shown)


ts/nni_manager/training_service/kubernetes/frameworkcontroller/frameworkcontrollerTrainingService.ts
```diff
@@ -65,7 +65,6 @@ class FrameworkControllerTrainingService extends KubernetesTrainingService imple
             await this.fcJobInfoCollector.retrieveTrialStatus(this.kubernetesCRDClient);
             if (this.kubernetesJobRestServer.getErrorMessage !== undefined) {
                 throw new Error(this.kubernetesJobRestServer.getErrorMessage);
-                this.stopping = true;
```
J-shang, line 68: Seems better to move this in front of the `throw new Error`.

maxsuren: Hello J-shang, I thought about this at the beginning, but than I realized `this.stopping` never been assigned to `true` in the old logic, so if I move `this.stopping=true` to the front of `throw new Error`, `this.stopping` will be `true` when throwing Error and the old logic will be broken.

SparkSnail (approved):

J-shang (approved):

PR #3528, opened 2021-04-14
cherry pick from PR3455

J-shang:


SparkSnail (approved):

ts/nni_manager/training_service/remote_machine/shellExecutor.ts
```diff
@@ -169,6 +169,16 @@ class ShellExecutor {
         return this.tempPath;
     }
 
+    public async getHomePath(): Promise<string> {
```
liuzhe-lz, line 172: The name looks like it returns `/home/zzz`, but it actually returns cwd?

J-shang: Yes, `cwd` is we need, I will change it to `getCurrentPath`.

liuzhe-lz (approved):

PR #3547, opened 2021-04-19
[Retiarii] fix experiment early exit

QuanluZhang:
- [x] fix the bug that retiarii experiment early exits
- [x] provide a new api `budget_exhausted` to strategies


nni/retiarii/experiment/pytorch.py
```diff
@@ -240,7 +242,12 @@ def _check_exp_status(self) -> bool:
         try:
             while True:
                 time.sleep(10)
-                status = self.get_status()
+                # this if is to deal with the situation that
```
ultmaster, line 245: Better to put this logic into a member function?

And I'm not sure whether this is correct? @liuzhe-lz

nni/retiarii/execution/api.py
```diff
@@ -67,3 +69,8 @@ def query_available_resources() -> int:
 
 def is_stopped_exec(model: Model) -> bool:
     return model.status in (ModelStatus.Trained, ModelStatus.Failed)
+
+
+def budget_exhausted() -> bool:
+    advisor = get_advisor()
```
ultmaster, line 75: Suggest access to execution engine rather than directly to advisor.

Suggest rename budget to `resource` for consistency

QuanluZhang: `resource` is a little misleading, because in our current implementation, resource means concurrency

ultmaster (approved):

liuzhe-lz (approved):

nni/retiarii/experiment/pytorch.py
QuanluZhang: leave the refactor to the next release

PR #3607, opened 2021-05-06
fix retiarii config experiment_working_directory

J-shang:


nni/retiarii/experiment/pytorch.py
```diff
@@ -52,7 +52,7 @@ class RetiariiExeConfig(ConfigBase):
     nni_manager_ip: Optional[str] = None
     debug: bool = False
     log_level: Optional[str] = None
-    experiment_working_directory: Optional[PathLike] = None
+    experiment_working_directory: Optional[PathLike] = '~/nni-experiments'
```
ultmaster, line 55: I'm not sure. What will happen if it's none?

J-shang: if we don't have a default value and the user don't set it, we will get
image

ultmaster: I had realized the problem. But I'm not sure whether this is a correct fix.

J-shang: Yes, is `PathLike`.

We have similar ExperimentConfig, FYI
https://github.com/microsoft/nni/blob/02eab99b9bb280e385a7af71b4c6c4a73bae1f04/nni/experiment/config/common.py#L69

ultmaster: But I just tested. It only works when I changed `~/nni-experiments` to `/home/xxx/nni-...`.

Have you checked the two blue buttons to view logs?

J-shang: Yes, I can view the log, and `expanduser` is here:

https://github.com/microsoft/nni/blob/02eab99b9bb280e385a7af71b4c6c4a73bae1f04/nni/experiment/config/base.py#L60

ultmaster: Could it be an issue with `Optional`? I'll retest it when this PR gets merged.

ultmaster (approved):

liuzhe-lz (approved):

PR #3614, opened 2021-05-07
Add TensorFlow slim pruner

liuzhe-lz:


nni/algorithms/compression/tensorflow/pruning/one_shot.py
```diff
@@ -57,15 +62,54 @@ def calc_masks(self, sparsity, wrapper, wrapper_idx=None):
...
+class SlimPrunerMasker(WeightMasker):
+    def __init__(self, model, pruner, **kwargs):
+        super().__init__(model, pruner)
+        weight_list = []
+        for wrapper in pruner.wrappers:
+            weights = [w for w in wrapper.layer.weights if '/gamma:' in w.name]
+            assert len(weights) == 1, f'Bad weights: {[w.name for w in wrapper.layer.weights]}'
+            weight_list.append(tf.math.abs(weights[0].read_value()))
+        all_bn_weights = tf.concat(weight_list, 0)
+        k = int(all_bn_weights.shape[0] * (1 - pruner.wrappers[0].config['sparsity']))
```
linbinskn, line 82: Why use `shape[0]` here which represents the first dimension of `all_bn_weights`? If we want to pick (1-sparsity) max values from weight, shouldn't we use count of all elements in `all_bn_weights` instead of `shape[0]`?

linbinskn: The value k is correct if it represents preserved channel. But then the calculation of `top_k` and `self.global_threshold` are wrong.

linbinskn: Current implementation is correct since this pruner only prune BatchNormalization layer....

nni/algorithms/compression/tensorflow/pruning/one_shot.py
```diff
@@ -57,15 +62,54 @@ def calc_masks(self, sparsity, wrapper, wrapper_idx=None):
...
+        for variable in wrapper.layer.weights:
+            if '/gamma:' in variable.name:
+                weight = variable.read_value()
+                weight_name = variable.name
+            elif '/beta:' in variable.name:
+                bias_name = variable.name
+
+        assert weight is not None
+        if wrapper.masks.get(weight_name) is not None:
+            weight *= wrapper.masks[weight_name]
```
linbinskn, line 103: I can't fully understand why multiply mask here. Maybe it is used to prune layer step by step.

nni/algorithms/compression/tensorflow/pruning/one_shot.py
```diff
@@ -3,15 +3,15 @@
 from nni.compression.tensorflow import Pruner
 
 __all__ = [
-    'OneshotPruner',
```
J-shang, line 6: Maybe we can keep this for user who wants to write their own pruner?

nni/algorithms/compression/tensorflow/pruning/one_shot.py
J-shang, line 82: This place is a little different from pytorch implementation and may lead to different thresholds with the same weight and sparsity.

https://github.com/microsoft/nni/blob/761732ab5f8e2606d6b5a5c6e0bf025d3a085b60/nni/algorithms/compression/pytorch/pruning/structured_pruning.py#L684

nni/algorithms/compression/tensorflow/pruning/one_shot.py
J-shang: yes, slim is iterative pruning

nni/algorithms/compression/tensorflow/pruning/one_shot.py
liuzhe-lz: I agree it looks strange, but that's how PyTorch version works.
What's your suggestion?

liuzhe-lz: The topk API is different between PyTorch and TensorFlow. If we don't use `1 - ` the logic will be much more complicated and hard to read.
Maybe we can alter the rounding behavior to make them more similar, but I guess the original implementation does not care it at all...

nni/algorithms/compression/tensorflow/pruning/one_shot.py
liuzhe-lz: If we export it than it's a public API and we need to maintain backward compatibility.
I think it's an internal helper and don't want to make it a formal API.
If a user really want it, they should copy the code.

nni/algorithms/compression/tensorflow/pruning/one_shot.py
J-shang: I think this is because SlimPruner is a channel level pruner, so by default, we use `shape[0]`. Maybe can refactor if we support more kind of channel scaling factor, but for now, it is good enough.

J-shang: Yes, maybe alter the rounding behavior is better. Or each layer may gap at most two channels be pruned between pytorch and tensorflow, then I'm afraid the final effect may have a bigger gap.

linbinskn: I see, SlimPruner is for pruning BatchNorm operator whose number of weight parameter is equal to channel number. In this case, all weight parameters are in one dimension and pruning each of them equal to pruning convolution channel. So current implementation is actually enough.

linbinskn: Maybe we can also use `-tf.nn.top_k(-weights,k,sorted=False).values.numpy()` to get the minimum k values.

nni/algorithms/compression/tensorflow/pruning/one_shot.py
```diff
@@ -57,15 +62,54 @@ def calc_masks(self, sparsity, wrapper, wrapper_idx=None):
...
+            elif '/beta:' in variable.name:
+                bias_name = variable.name
+
+        assert weight is not None
+        if wrapper.masks.get(weight_name) is not None:
+            weight *= wrapper.masks[weight_name]
+
+        filters = weight.shape[0]
+        num_prune = int(filters * sparsity)
+        if filters >= 2 and num_prune >= 1:
```
linbinskn, line 107: Does it contain some corner cases? Why not casting weight by `self.global_threshold` directly instead of doing such branch identification.

J-shang (approved):

linbinskn:
Please resolve conflicts.

linbinskn (approved):

ultmaster (approved):

PR #3643, opened 2021-05-14
Suport prefix url for nnimanager

acured:
Suport prefix url for nnimanager

### DIFF
--- a/nni/tools/nnictl/launcher.py
+++ b/nni/tools/nnictl/launcher.py
@@ -16,7 +16,7 @@ from nni.tools.package_utils import get_builtin_module_class_name
 import nni_node  # pylint: disable=import-error
 from .launcher_utils import validate_all_content
 from .rest_utils import rest_put, rest_post, check_rest_server, check_response
-from .url_utils import cluster_metadata_url, experiment_url, get_local_urls
+from .url_utils import cluster_metadata_url, experiment_url, get_local_urls, set_prefix_url
 from .config_utils import Config, Experiments
 from .common_utils import get_yml_content, get_json_content, print_error, print_normal, print_warning, \
                           detect_port, get_user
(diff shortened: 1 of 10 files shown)


ts/nni_manager/rest_server/nniRestServer.ts
```diff
@@ -19,7 +20,7 @@ import { createRestHandler } from './restHandler';
  */
 @component.Singleton
 export class NNIRestServer extends RestServer {
-    private readonly API_ROOT_URL: string = '/api/v1/nni';
+    private readonly API_ROOT_URL: string = `/api/v1/nni/${getUrlPrefix()}`;
```
liuzhe-lz, line 23: I think prefix URL should better override `/api/v1/nni`.

nni/tools/nnictl/launcher.py
```diff
@@ -81,6 +81,10 @@ def start_rest_server(port, platform, mode, experiment_id, foreground=False, log
         cmds += ['--log_level', log_level]
     if foreground:
         cmds += ['--foreground', 'true']
+    if url_prefix:
+        path_validation(url_prefix)
```
J-shang, line 85: maybe validate in `create_experiment` is better?
and if we `resume` or `view` an experiment, should we support reconfigure `url_prefix`?

ts/nni_manager/rest_server/nniRestServer.ts
J-shang: The same opinion with @liuzhe-lz , in https://github.com/microsoft/nni/blob/5a71ce916b9298d7d3cda77b3fee24d4474f05e8/nni/tools/nnictl/url_utils.py#L10

so that we don't need to modify too many internal interfaces.

acured: Changed.

nni/tools/nnictl/launcher.py
acured: 1. Changed.
2. I'm ok for that, since there is no way to change "url_prefix". Any concerns @liuzhe-lz ?

liuzhe-lz (approved):

J-shang (approved):

PR #3648, opened 2021-05-15
Support hybrid and V2 config pipieline

SparkSnail:


test/config/examples/mnist-annotation-v2.yml
```diff
@@ -0,0 +1,16 @@
+experimentName: default_test
+searchSpaceFile: ../../../examples/trials/mnist-keras/search_space.json
+trialCommand: python3 mnist.py --batch_num 10
+trialCodeDirectory: ../../../examples/trials/mnist-annotation
```
J-shang, line 4: do we support annotation in config v2?

J-shang (approved):

test/config/examples/mnist-annotation-v2.yml
liuzhe-lz: Yes, though there is a bug to fix.

liuzhe-lz (approved):

PR #3650, opened 2021-05-17
[Retiarii] Policy-based RL Strategy

ultmaster:
This PR supports a family of RL strategy based on tianshou. The default built-in algorithm is PPO.

TODOs:

- [x] logging
- [x] tests

### DIFF
--- a/nni/retiarii/strategy/rl.py
+++ b/nni/retiarii/strategy/rl.py
@@ -10,7 +10,7 @@ try:
     import torch
     from tianshou.data import AsyncCollector, Collector, VectorReplayBuffer
     from tianshou.env import SubprocVectorEnv
-    from tianshou.policy import BasePolicy, PPOPolicy
+    from tianshou.policy import BasePolicy, PPOPolicy  # pylint: disable=unused-import
     from ._rl_impl import ModelEvaluationEnv, Preprocessor, Actor, Critic
 except ImportError:
     has_tianshou = False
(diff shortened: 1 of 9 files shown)


liuzhe-lz (approved):

nni/retiarii/strategy/rl.py
```diff
@@ -0,0 +1,90 @@
...
+    max_collect : int
+        How many times collector runs to collect trials for RL. Default 100.
+    trial_per_collect : int
+        How many trials (trajectories) each time collector collects.
+        After each collect, trainer will sample batch from replay buffer and do the update. Default: 20.
+    policy_fn : function
+        Takes ``ModelEvaluationEnv`` as input and return a policy. See ``_default_policy_fn`` for an example.
+    asynchronous : bool
+        If true, in each step, collector won't wait for all the envs to complete.
+        This should generally not affect the result, but might affect the efficiency. Note that a slightly more trials
```
QuanluZhang, line 41: don't understand, why asynchronous does not affect the result?

ultmaster: Synchronous doesn't mean single-process sampling. Both synchronous and asynchronous has parallelism. "Asynchronous" induces a mechanism to give up on some environment when it's not finished.

Refer to https://tianshou.readthedocs.io/en/master/tutorials/cheatsheet.html#parallel-sampling if you feel interested. It's a bit complicated and I don't think I can make it clear here in a few words.

QuanluZhang (approved):

PR #3652, opened 2021-05-19
Support nested ModuleList and fix an issue in list append

QuanluZhang:
- [x] support nested ModuleList
- [x] fix an issue of in-place python list append: incorrect control flow


test/ut/retiarii/test_convert_models.py
```diff
@@ -0,0 +1,72 @@
+import os
```
ultmaster, line 1: Why another file?

QuanluZhang: this file is committed before, it is supposed to test whole models. so I put this test into this file

ultmaster (approved):

liuzhe-lz (approved):

PR #3672, opened 2021-05-26
fix url and api conflict

acured:


ts/webui/src/static/const.ts
```diff
@@ -6,7 +6,7 @@ const METRIC_GROUP_UPDATE_SIZE = 20;
 
 const prefix = getPrefix();
 
-const MANAGER_IP = prefix === undefined ? '/api/v1/nni' : `${prefix}`;
+const MANAGER_IP = prefix === undefined ? '/api/v1/nni' : `api/v1/nni${prefix}`;
```
ultmaster, line 9: prefix should be before `/api`?

Lijiaoa: But when rest api is `prefix/api/v1/nni`, it still is conflict with ui router. because they are in the same directory `prefix/xxx` .

Lijiaoa: image
test with ui router `ip:port/api/oview` and rest api `ip:port/api/v1/nni/trial-jobs`, webportal is normal!

acured: Hi @liuzhe-lz , do you have any concern?

ultmaster (approved):

liuzhe-lz (approved):

ts/webui/src/static/const.ts
liuzhe-lz: Let's make it work first.

PR #3674, opened 2021-05-26
update prefix rest api format and fix getPrefix function bug

Lijiaoa:


ultmaster (approved):

liuzhe-lz (approved):

liuzhe-lz:
I don't know if this PR is correct. But current master does not work at all. So it can't be worse...

PR #3677, opened 2021-05-26
Set reuse as default mode for local trainingService

SparkSnail:


ultmaster:
Need to test on local pipeline.

liuzhe-lz (approved):

PR #3688, opened 2021-05-27
[#3507 follow up] update doc & ut

J-shang:


docs/en_US/Compression/Framework.rst
```diff
@@ -103,7 +102,8 @@ Users can also remove this collector like this:
 Pruner
 ------
 
-A pruner receives ``model``\ , ``config_list`` and ``optimizer`` as arguments. It prunes the model per the ``config_list`` during training loop by adding a hook on ``optimizer.step()``.
+A pruner receives ``model``\ , ``config_list`` as arguments.
```
QuanluZhang, line 105: remove \

J-shang: fixed

nni/algorithms/compression/pytorch/pruning/auto_compress_pruner.py
```diff
@@ -34,6 +34,8 @@ class AutoCompressPruner(Pruner):
         Function used for the first subproblem of ADMM Pruner.
         Users should write this function as a normal function to train the Pytorch model
         and include `model, optimizer, criterion, epoch` as function arguments.
+    criterion: function
+        Function used to calculate the loss between the target and the output. By default, we use CrossEntropyLoss.
```
QuanluZhang, line 38: so it should be a function not an instance? so users should provide `loss` for criterion, where `loss = nn.CrossEntropyLoss()`? then it is better to give this concrete example

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -554,6 +559,8 @@ class ActivationMeanRankFilterPruner(IterativePruner):
         The activation type.
     sparsity_training_epochs: int
         The number of batches to statistic the activation.
+    statistics_batch_num: int
```
QuanluZhang, line 562: why there is no such parameter before?

nni/algorithms/compression/pytorch/pruning/auto_compress_pruner.py
J-shang: add doc str

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
J-shang: workaround by using `sparsifying_training_batches`

ultmaster (approved):

QuanluZhang (approved):

ultmaster (approved):

PR #3703, opened 2021-06-01
fix wrong lr scheduler for cream

chenbohua3:
It seems that cream uses linear scheduler to train the super network. The initial multiplicative factor should be 1.0 instead of the initial learning rate (Although the initial learning rate of training the supernet is exactly 1.0 ).

### DIFF
--- a/examples/nas/legacy/cream/lib/utils/util.py
+++ b/examples/nas/legacy/cream/lib/utils/util.py
@@ -176,5 +176,5 @@ def create_supernet_scheduler(cfg, optimizer):
     ITERS = cfg.EPOCHS * \
         (1280000 / (cfg.NUM_GPU * cfg.DATASET.BATCH_SIZE))
     lr_scheduler = torch.optim.lr_scheduler.LambdaLR(optimizer, lambda step: (
-        cfg.LR - step / ITERS) if step <= ITERS else 0, last_epoch=-1)
+        1.0 - step / ITERS) if step <= ITERS else 0, last_epoch=-1)
     return lr_scheduler, cfg.EPOCHS


ultmaster (approved):

Z7zuqer:
Hi,

Thanks for your interest in our project!

I'm afraid that such modification would make the training of the supernet inflexible. As in [config file](https://github.com/microsoft/nni/blob/6aaca5f7ad2f0f169cc9439407a2137c44a25de9/examples/nas/legacy/cream/configs/train.yaml#L34), you could set different learning rate.

Best,

Hao.

penghouwen:
Our setting is correct. PR Blocked.

chenbohua3:
@Z7zuqer The initial learning rate for each parameter is set [here](https://github.com/microsoft/nni/blob/b7c91e733062699f8bdb13dbd8269c2ce31d2c7a/examples/nas/legacy/cream/lib/utils/util.py#L91), which uses the LR defined in [config file](https://github.com/microsoft/nni/blob/6aaca5f7ad2f0f169cc9439407a2137c44a25de9/examples/nas/legacy/cream/configs/train.yaml#L34). However, the codes I changed are about the learning rate scheduler. According to [hear](https://pytorch.org/docs/1.2.0/optim.html#torch.optim.lr_scheduler.LambdaLR), the `lr_lambda` is used to calculate the `multiplicative factor` for the learning rate.  If `cfg.LR < 1.0`, than `cfg.LR - step / ITERS` would produce negative results.

QuanluZhang:
@chenbohua3  closed. feel free to reopen if there is other questions

PR #3705, opened 2021-06-01
Fix a logging related bug

liuzhe-lz:
There is a bug in AML caused by trying to json stringify a circular object for logging.
Now the logging system automatically uses `util.inspect()` to print objects. So it's better to avoid `JSON.stringify()` for logging in case a circular object crashes the whole process and eats the message.
This PR also adds tag (class name) to most logging info.


ultmaster (approved):

ts/nni_manager/core/nniDataStore.ts
```diff
@@ -71,7 +71,7 @@ class NNIDataStore implements DataStore {
 
     public storeTrialJobEvent(
         event: TrialJobEvent, trialJobId: string, hyperParameter?: string, jobDetail?: TrialJobDetail): Promise<void> {
-        //this.log.debug(`storeTrialJobEvent: event: ${event}, data: ${hyperParameter}, jobDetail: ${JSON.stringify(jobDetail)}`);
+        //this.log.debug(`storeTrialJobEvent: event: ${event}, data: ${hyperParameter}, jobDetail:`, jobDetail);
```
SparkSnail, line 74: If this log info is not used, could delete this line.

PR #3707, opened 2021-06-01
Fix DNGO tuner class name

ultmaster:
Fixed an issue that DNGO tuner name is not consistent in python file and yaml.

### DIFF
--- a/nni/algorithms/hpo/dngo_tuner.py
+++ b/nni/algorithms/hpo/dngo_tuner.py
@@ -40,7 +40,7 @@ def _random_config(search_space, random_state):
     return chosen_config
 
 
-class DngoTuner(Tuner):
+class DNGOTuner(Tuner):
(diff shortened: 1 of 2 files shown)


J-shang (approved):

QuanluZhang (approved):

nni/algorithms/hpo/dngo_tuner.py
```diff
@@ -40,7 +40,7 @@ def _random_config(search_space, random_state):
     return chosen_config
 
 
-class DngoTuner(Tuner):
+class DNGOTuner(Tuner):
```
QuanluZhang, line 43: BTW, do we update doc accordingly?

ultmaster: The docs were correct in the first place.

PR #3713, opened 2021-06-02
Fix a few bugs in Retiarii and upgrade Dockerfile

ultmaster:
1. Upgrade dependencies in Dockerfile and include pytorch-lightning.
2. Support export in pure-python execution engine, update docs and examples.
3. Fix an issue that `nni.retiarii.nn` cannot be imported in generated code.
4. Change the RL implementation to multithreading because multiprocessing doesn't work with current execution engine implementation. As a result, we don't support AsyncCollector for now.

This PR should unblock #3700.

### DIFF
--- a/nni/retiarii/experiment/pytorch.py
+++ b/nni/retiarii/experiment/pytorch.py
@@ -318,7 +318,7 @@ class RetiariiExperiment(Experiment):
         """
         Export several top performing models.
 
-        For one-shot algorithms, only top-1 is supported. For others, ``optimize_mode`` asnd ``formatter`` is
+        For one-shot algorithms, only top-1 is supported. For others, ``optimize_mode`` and ``formatter`` are
         available for customization.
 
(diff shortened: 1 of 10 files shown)


nni/retiarii/experiment/pytorch.py
```diff
@@ -317,7 +318,7 @@ def export_top_models(self, top_k: int = 1, optimize_mode: str = 'maximize', for
         """
         Export several top performing models.
 
-        For one-shot algorithms, only top-1 is supported. For others, ``optimize_mode`` asnd ``formater`` is
+        For one-shot algorithms, only top-1 is supported. For others, ``optimize_mode`` asnd ``formatter`` is
```
QuanluZhang, line 321: asnd -> and, is -> are

QuanluZhang (approved):

liuzhe-lz (approved):

PR #3721, opened 2021-06-03
Port trial examples' config file to v2

liuzhe-lz:
Notes:
- config_pai and other non-local training services in most examples are removed. Instead added a comment asking users to check mnist-pytorch for training service examples.
- Shared storage example is not ported. I don't have environment to test.
- system_auto_tuning is not ported. I don't know how to test.
- mnist-tfv1 and -keras are not ported. They are deprecated.
- Non-reusable k8s training services recommends v1 config for now.
- Added remote example for mnist-pytorch.
- Added "config_detailed.yml" for mnist-pytorch and -tfv2.
- mnist-distributed is renamed to mnist-distributed-tfv1, to match mnist-distributed-pytorch.

Fixed bugs:
- When GPU indices contains only one index, the YAML field becomes a int instead of string. This causes problems.
- Custom tuner's field codeDirectory has different name in code and doc.
- Expanding annotation only uses v1 trial code directory.


ultmaster (approved):

SparkSnail:
Add example for config_hybrid.yml?

J-shang (approved):

liuzhe-lz:
> Add example for config_hybrid.yml and config_windows_v2.yml?

I added a comment line in every "basic" example to inform windows users. I think a separate Windows example is bad because it makes me think NNI behaves differently on Windows and Linux. Let's discuss it in the meeting.
A hybrid example should make sense. I'm testing it now.

examples/trials/efficientnet/config.yml
```diff
@@ -0,0 +1,15 @@
+searchSpaceFile: search_net.json
+trialCodeDirectory: EfficientNet-PyTorch
+trialCommand: python main.py /data/imagenet -j 12 -a efficientnet --batch-size 48 --lr 0.048 --wd 1e-5 --epochs 5 --request-from-nni
```
SparkSnail, line 3: python3?

SparkSnail (approved):

examples/trials/cifar10_pytorch/config.yml
```diff
@@ -1,23 +1,14 @@
-authorName: default
-experimentName: example_pytorch_cifar10
+searchSpaceFile: search_space.json
```
QuanluZhang, line 1: suggest at least for this simple example, we add "experimentName", because if users really use nni to run experiments, they want to give each experiment an easy-to-remember name (not experiment ID). If this field is not added in the example, users have to check config references, which is not friendly

examples/trials/mnist-pytorch/config.yml
```diff
@@ -1,21 +1,14 @@
-authorName: default
-experimentName: example_mnist_pytorch
+# This is the minimal config file for an NNI experiment.
+# Use "nnictl create --config config.yml" to launch this experiment.
+# Afterwards, you can check "config_detailed.yml" for more explaination.
```
QuanluZhang, line 3: explaination -> explanation

examples/trials/mnist-pytorch/config_detailed.yml
```diff
@@ -0,0 +1,42 @@
+# This example shows more configurable fields comparing to the minimal "config.yml"
+# You can use "nnictl create --config config_detailed.yml" to launch this experiment.
+# If you see an error message saying "port 8080 is used", use "nnictl stop --all" to stop previous experiment.
```
QuanluZhang, line 3: -> use "nnictl stop --port 8080" to stop that experiment, or use "nnictl stop --all" to stop all the previous experiments.

examples/trials/mnist-pytorch/config_detailed.yml
```diff
@@ -0,0 +1,42 @@
...
+  lr:
+    _type: choice
+    _value: [0.0001, 0.001, 0.01, 0.1]
+  momentum:
+    _type: uniform
+    _value: [0, 1]
+
+trialCommand: python3 mnist.py  # The command to launch a trial. NOTE: change "python3" to "python" if you are using Windows.
+trialCodeDirectory: .           # The path of trial code. By default it's ".", which means the same directory of this config file.
+trialGpuNumber: 1               # How many GPUs should each trial use. CUDA is required when it's greator than zero.
```
QuanluZhang, line 25: greator -> greater

examples/trials/mnist-pytorch/config_detailed.yml
```diff
@@ -0,0 +1,42 @@
...
+  hidden_size:
+    _type: choice
+    _value: [128, 256, 512, 1024]
+  lr:
+    _type: choice
+    _value: [0.0001, 0.001, 0.01, 0.1]
+  momentum:
+    _type: uniform
+    _value: [0, 1]
+
```
QuanluZhang, line 22: yeah, "experimentName" should be put here, we can tell users that they can omit it if they don't want to write it.

QuanluZhang:
looks great!

examples/trials/mnist-pytorch/config_detailed.yml
liuzhe-lz: I prefer not to provide that much details in example. At this point the user does not need to know how to manage multiple experiments.
I'm afraid that "--port 8080" might threaten newbie users.

scarlett2018 (approved):

PR #3725, opened 2021-06-04
Fix a few issues in Retiarii

ultmaster:
1. Fix a data race issue on Retiarii RL that induces unexpected result for models within the same iteration.
2. Add docstring to TPE strategy.
3. Fix the key name in mutators created in BaseExecutionEngine.
4. Fix an issue that causes budget to be negative. Thanks @hzhua for pointing out.

### DIFF
--- a/nni/retiarii/strategy/_rl_impl.py
+++ b/nni/retiarii/strategy/_rl_impl.py
@@ -1,6 +1,7 @@
 # This file might cause import error for those who didn't install RL-related dependencies
 
 import logging
+import threading
 from multiprocessing.pool import ThreadPool
 
 import gym
@@ -18,6 +19,7 @@ from ..execution import submit_models, wait_models
 
 
 _logger = logging.getLogger(__name__)
+_thread_lock = threading.Lock()
 
 
 class MultiThreadEnvWorker(EnvWorker):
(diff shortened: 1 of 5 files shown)


nni/retiarii/strategy/_rl_impl.py
```diff
@@ -100,7 +102,8 @@ def step(self, action):
                 if self.cur_step < self.num_steps else self.action_dim
         }
         if self.cur_step == self.num_steps:
-            model = get_targeted_model(self.base_model, self.mutators, self.sample)
+            with _thread_lock:
+                model = get_targeted_model(self.base_model, self.mutators, self.sample)
```
QuanluZhang, line 106: could explain more, why there is data race here?

ultmaster: Why mutator is mutating, it will rely on some of its member attributes. This can cause trouble if you are calling multiple `mutator.mutate` at the same time.

QuanluZhang: got it

QuanluZhang (approved):

liuzhe-lz (approved):

PR #3726, opened 2021-06-04
Fix 3rd-party training service bug

liuzhe-lz:
Environment service with web command channel needs to know the port. Otherwise it only supports 8080.
To fix this, pass whole ExperimentStartupInfo to environment service.

The template / example training service: https://github.com/liuzhe-lz/nni-training-service-template

### DIFF
--- a/ts/nni_manager/common/utils.ts
+++ b/ts/nni_manager/common/utils.ts
@@ -25,7 +25,7 @@ import { ExperimentManager } from './experimentManager';
 import { HyperParameters, TrainingService, TrialJobStatus } from './trainingService';
 
 function getExperimentRootDir(): string {
-    return getExperimentStartupInfo().getLogDir();
+    return getExperimentStartupInfo().logDir;
 }
 
 function getLogDir(): string {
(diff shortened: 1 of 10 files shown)


ts/nni_manager/common/utils.ts
```diff
@@ -213,7 +212,7 @@ function cleanupUnitTest(): void {
     Container.restore(TrainingService);
     Container.restore(DataStore);
     Container.restore(Database);
-    Container.restore(ExperimentStartupInfo);
+    setExperimentStartupInfo(true, 'unittest', 8080, 'unittest', undefined, logLevel);
```
SparkSnail, line 215: Why use 'unittest' as experiment id here?

liuzhe-lz: The original logic was snapshot ExperimentStartupInfo in `prepareUnitTest` and restore it here with IOC's utility.
To share ExperimentStartupInfo with 3rd-party training service, it is no longer an IOC container and has to be manually restored.
This line is copied from `prepareUnitTest` to restore the same state.

SparkSnail (approved):

ultmaster (approved):

PR #3729, opened 2021-06-07
Catch clean up error to prevent crash

liuzhe-lz:
Named pipe can sometimes fall into extremely sophisticated states and there is very little benefit to handle them. So when strange things happen, just wait the OS to recycle resources.
The only potential problem is, if the user stops and resumes an experiment in one process, then resuming might fail. I believe this use case is too rare to worth a fix.

### DIFF
--- a/nni/experiment/pipe.py
+++ b/nni/experiment/pipe.py
@@ -1,6 +1,9 @@
 from io import BufferedIOBase
+import logging
 import os
 import sys
+
+_logger = logging.getLogger(__name__)
 
 if sys.platform == 'win32':
     import _winapi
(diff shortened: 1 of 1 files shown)


ultmaster:
This PR doesn't seem to fix a problem, but to ignore a problem.

Does it mean that the exception is very rare? In that case, why don't we just raise an exception?

liuzhe-lz:
> This PR doesn't seem to fix a problem, but to ignore a problem.
> 
> Does it mean that the exception is very rare? In that case, why don't we just raise an exception?

Yes. This PR ignores a problem which only happens on experiment exiting. Since the experiment is already done, it's not harmful to have some invalid internal state.
The root cause is that NNI manager and Python script share the pipe and therefore its state depends on NNI manager's exit timing. To fix this, it needs to check the pipe's status before closing, and it must deal with atomic problem to be "correct".

ultmaster (approved):

QuanluZhang (approved):

PR #3808, opened 2021-06-10
fix foreground log

J-shang:


ultmaster (approved):
Please test with and without foreground and make sure everything is working as expected.

SparkSnail (approved):

PR #3833, opened 2021-06-15
Expose rstate in update_search_space to fix seed

g1910:
Expose rstate argument in update_search_space function to control/fix random seed of FMinIter in hyperopt


liuzhe-lz:
`update_search_space` is a virtual method / callback. The added parameter can never be used in standard NNI workflow.
To fix random seed, I suggest add a parameter to the constructor and it can be specified with `classArgs` field in config file.

QuanluZhang:
closed due to no further update

PR #3837, opened 2021-06-16
flops counter formatting fix

xiaowu0162:
Fix a minor formatting issue. Previously the counter prints out the FLOPs as a tensor directly.
before:
before
after:
after

### DIFF
--- a/nni/compression/pytorch/utils/counter.py
+++ b/nni/compression/pytorch/utils/counter.py
@@ -313,12 +313,13 @@ class ModelProfiler:
 
         table.field_names = headers
         for i, result in enumerate(self.results):
+            flops_count = int(result['flops'].item()) if isinstance(result['flops'], torch.Tensor) else int(result['flops'])
             row_values = [
                 i,
                 result['name'],
                 result['module_type'],
                 str(result['weight_shape']),
-                result['flops'],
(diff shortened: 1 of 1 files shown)


xiaowu0162:
@ultmaster @J-shang

J-shang (approved):

nni/compression/pytorch/utils/counter.py
```diff
@@ -313,12 +313,13 @@ def format_results(self):
 
         table.field_names = headers
         for i, result in enumerate(self.results):
+            flops_count = int(result['flops'].item()) if isinstance(result['flops'], torch.Tensor) else int(result['flops'])
```
ultmaster, line 316: `torch.is_tensor(...)`

xiaowu0162: It seems that line 366 in the same file also uses isinstance for checking. Also, do we need to consider the recommendation from the official website? [https://pytorch.org/docs/stable/generated/torch.is_tensor.html](url)

ultmaster: In the official recommendation is isinstance, I'm okay with that.

ultmaster (approved):

PR #3844, opened 2021-06-17
update pruning example

J-shang:
`optimizer` used in pruner might be patched, so need a new `optimizer` in finetuning stage.

### DIFF
--- a/examples/model_compress/pruning/basic_pruners_torch.py
+++ b/examples/model_compress/pruning/basic_pruners_torch.py
@@ -143,7 +143,7 @@ def get_model_optimizer_scheduler(args, device, train_loader, test_loader, crite
         model.load_state_dict(torch.load(args.pretrained_model_dir))
         best_acc = test(args, model, device, criterion, test_loader)
 
-    # setup new opotimizer for fine-tuning
+    # setup new opotimizer for pruning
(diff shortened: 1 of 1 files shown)


examples/model_compress/pruning/basic_pruners_torch.py
```diff
@@ -278,6 +278,8 @@ def trainer(model, optimizer, criterion, epoch):
         m_speedup.speedup_model()
 
     print('start finetuning...')
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.9, weight_decay=5e-4)
+    scheduler = MultiStepLR(optimizer, milestones=[int(args.pretrain_epochs * 0.5), int(args.pretrain_epochs * 0.75)], gamma=0.1)
```
hippogr, line 282: I found the same code at line 147 and 148...

J-shang: yes, I copy from there...
This because `optimizer` might be modified in pruner, ref line 230, 258, then we need a new `optimizer` in finetuning stage.

J-shang: add a comment for these lines

hippogr (approved):
:shipit:

linbinskn (approved):

QuanluZhang (approved):

ultmaster (approved):

PR #3856, opened 2021-06-21
[quantization speedup]fix data device type bug

linbinskn:
This PR is for fixing data's device type which has to be "cpu". Otherwise, it can not convert to numpy directly and will raise errors.

### DIFF
--- a/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py
+++ b/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py
@@ -290,6 +290,9 @@ class ModelSpeedupTensorRT(BaseModelSpeedup):
                 calib_data_set.append(data)
             calib_data = np.concatenate(calib_data_set)
         elif type(self.calib_data_loader) == torch.Tensor:
+            # trt need numpy as calibration data, only cpu data can convert to numpy directly
+            if self.calib_data_loader.device != torch.device("cpu"):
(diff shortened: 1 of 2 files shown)


nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py
```diff
@@ -326,6 +329,8 @@ def inference(self, test_data):
             Model input tensor
         """
         # convert pytorch tensor to numpy darray
+        if test_data.device != torch.device("cpu"):
+            test_data = test_data.to("cpu")
```
QuanluZhang, line 333: it is a little strange that inference can only be executed on cpu... why?

linbinskn: image
Actually the inference executed on GPU after the memory copy from host memory to device memory which is done in our tool and the process of host2device memory copy is unavoidable. We use pycuda' s `memcpy_htod_async` api to do such copy. It is just a way of unifying that we choose to handle all host data and we will do memory copy in our engine before calibration and inference. We can also take another to handle it such as finding other way or API which can use torch cuda data directly. However, it will be more complicated and no extra advantage.

QuanluZhang: ok, got it

QuanluZhang (approved):

J-shang (approved):

PR #3861, opened 2021-06-22
Update Quickstart and remove autoCompletion in the doc

kvartet:
some improvements:
1. update the quickstart doc: change the v1 config to the v2 config and make it more user-friendly
2. remove the `nnictl` auto-completion part in the doc, which is unsupported now
3. fix typos in docs and codes: `alogrithm`, `achine`, `complted`
4. fix the wrong header level


docs/en_US/Tutorial/QuickStart.rst
```diff
@@ -4,7 +4,7 @@ QuickStart
 Installation
 ------------
 
-We currently support Linux, macOS, and Windows. Ubuntu 16.04 or higher, macOS 10.14.1, and Windows 10.1809 are tested and supported. Simply run the following ``pip install`` in an environment that has ``python >= 3.6``.
+Currently, NNI supports running on Linux, MacOS and Windows. Ubuntu 16.04 or higher, macOS 10.14.1, and Windows 10.1809 are tested and supported. Simply run the following ``pip install`` in an environment that has ``python >= 3.6``.
```
liuzhe-lz, line 7: It's `macOS`.

docs/en_US/Tutorial/QuickStart.rst
```diff
@@ -128,27 +138,29 @@ Three steps to start an experiment
...
-     codeDir: .
-     gpuNum: 0
+Config the experiment in the YAML file which declares the ``path`` to the search space and trial files. It also gives other information such as the tuning algorithm, max trial number, and max duration arguments.
 
+.. code-block:: yaml
+
+   experimentName: MNIST               # An optional name to distinguish the experiments
+   searchSpaceFile: search_space.json  # The path to the search space file
+   trialCommand: python3 mnist.py      # NOTE: change "python3" to "python" if you are using Windows
+   trialCodeDirectory: .               # The path to the trial code
```
liuzhe-lz, line 152: As a template, I think it's save to remove this line and ask the user to put config file in trial code's top directory.

docs/en_US/Tutorial/QuickStart.rst
```diff
@@ -158,19 +170,23 @@ Three steps to start an experiment
...
 .. code-block:: bash
 
    nnictl create --config nni/examples/trials/mnist-pytorch/config.yml
 
 Windows
-^^^^^^^
+*******
 
-Run the **config_windows.yml** file from your command line to start an MNIST experiment.
+Run the **config_windows.yml** file from your command line to start the experiment.
```
liuzhe-lz, line 189: Ooops, I removed config_windows.yml. I thought it's suggesting NNI behaves differently on Windows and Linux, which is not the case.
If it makes doc or quick start hard or complex, please add it back. Just copy config.yml and change `python3` to `python`.

docs/en_US/Tutorial/QuickStart.rst
```diff
@@ -80,12 +76,20 @@ NNI is born to help the user do tuning jobs; the NNI working process is presente
...
+   In the this part, we will focus on the first approach. For the second approach, please refer to `this tutorial <HowToLaunchFromPython.rst>`__\ .
+
 
-Three steps to start an experiment
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Step 1: Define the Seach Space
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**Step 1**: Write a ``Search Space`` file in JSON, including the ``name`` and the ``distribution`` (discrete-valued or continuous-valued) of all the hyperparameters you need to search.
+Define a ``Search Space`` in the JSON file, including the ``name`` and the ``distribution`` (discrete-valued or continuous-valued) of all the hyperparameters you want to search.
```
liuzhe-lz, line 92: Maybe we can suggest embedded search space. It's more convenient in most cases.

ultmaster: agree

docs/en_US/CommunitySharings/AutoCompletion.rst
```diff
@@ -1,57 +0,0 @@
-Auto Completion for nnictl Commands
-===================================
-
```
ultmaster, line 3: Do we have a future plan to support this feature? If yes, we can add a ``.. warning`` here that states the feature is broken in current releases and will be fixed in feature.

docs/en_US/Tutorial/QuickStart.rst
kvartet: Thanks for pointing it out, I will update it.

docs/en_US/CommunitySharings/AutoCompletion.rst
kvartet: It seems that @liuzhe-lz is the owner of this feature. - would you mind tell us the future plan?

docs/en_US/Tutorial/QuickStart.rst
kvartet: okay, will update it

docs/en_US/Tutorial/QuickStart.rst
kvartet: hahaha, a user asked in the WeChat group why there is no `config_windows.yml` example for v2.3, so I add it back?

docs/en_US/Tutorial/QuickStart.rst
kvartet: ok, so do we need to remove the JSON part? or suggest two ways to define the search space in the quickstart?

liuzhe-lz: I think in quick start we can remove JSON part.

docs/en_US/Tutorial/QuickStart.rst
liuzhe-lz: Another choice is to tell the users to update the config. I don't know which is better.

docs/en_US/CommunitySharings/AutoCompletion.rst
liuzhe-lz: It is planned, with very priority.

docs/en_US/Tutorial/QuickStart.rst
kvartet: I modified all the config files in quickstart to `config_detailed.yml`, which including the inline search space and swap the search space part and the trial code part so that users can config the experiment directly in the YAML file after writing the search space.

docs/en_US/CommunitySharings/AutoCompletion.rst
kvartet: This feature has been deferred since v2.0, and I don’t know when it will be implemented, so I temporarily removed it...

ultmaster (approved):

scarlett2018 (approved):

docs/en_US/Tutorial/QuickStart.rst
```diff
@@ -128,55 +118,83 @@ Three steps to start an experiment
...
+         _value: [16, 32, 64, 128]
+      hidden_size:
+         _type: choice
+         _value: [128, 256, 512, 1024]
+      lr:
+         _type: choice
+         _value: [0.0001, 0.001, 0.01, 0.1]
+      momentum:
+         _type: uniform
+         _value: [0, 1]
```
QuanluZhang, line 141: it is a little strange that you are supposed to demonstrate "1. Write a config file and start the experiment from the command line.", but the format of this search space is for "launch from python"...

QuanluZhang: and if you are demonstrate the first approach, suggest to use "config.yml" rather than "config_detailed.yml"

kvartet: This is inline(aka. embedded) search space, which can be defined in the YAML file, instead of the JSON file. [They](https://github.com/microsoft/nni/pull/3861#discussion_r658465990) think this approach is more convenient, for example, users only need to write this additional file, and can comment out each line at will. Only "config_detailed.yml" adopts the inline search space, in consistency, I suggest using this file.

QuanluZhang: got it, thanks

QuanluZhang (approved):

khituras:
Thanks for the response @QuanluZhang.
I gladly would, however it has already been done per this issue. The current file at https://github.com/microsoft/nni/blob/master/docs/en_US/Tutorial/QuickStart.rst does ot mention the auto completion script.
However, the online documentation seems to be old. It needs to be updated from sources, I guess.

khituras:
I see. I just followed the link from https://github.com/microsoft/nni. Perhaps this link should point to `latest` then?

Issue #3857, opened 2021-06-22
`max_workspace_size` is not compatible with tensorrt8.x


un-knight:
The following codes will invoke `AttributeError: 'tensorrt.tensorrt.Builder' object has no attribute 'max_workspace_size'` in the TensorRT 8.0.0.3. So it seems that `max_workspace_size` attribute has been removed in TensorRT8

https://github.com/microsoft/nni/blob/ef15fc81967fa13c74da6147931fde236464bf04/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py#L126

reference issues: https://github.com/NVIDIA-AI-IOT/torch2trt/issues/557

mallamsripooja:
Yes, same issue here. Had to downgrade to older version of Tensorrt.

linbinskn:
Exactly, this problem exists due to api updating after TensorRT8.0. Have supported the latest version in PR #3866 , please wait for it merged.

sriram-dsl:
builder = trt.Builder(TRT_LOGGER)
network = builder.create_network()

builder_config = builder.create_builder_config()

builder_config.max_workspace_size = 1 << 30

PR #3866, opened 2021-06-24
[Quantization speedup]support TensorRT8.0.0

linbinskn:
This PR aims to support the latest TensorRT version. Current quantization speedup tool is implemented based on TensorRT7.0. However, the latest tensorrt python api has changed which separates network definition and configuration after version 8.0. All configuration of low precision have been moved to IBuilderConfig and our current implementation can't work on it (the problem raised in issue #3857 ). This PR supports the new TensorRT version and the new api.

### DIFF
--- a/docs/en_US/Compression/QuantizationSpeedup.rst
+++ b/docs/en_US/Compression/QuantizationSpeedup.rst
@@ -48,7 +48,11 @@ Prerequisite
 ------------
 CUDA version >= 11.0
 
-TensorRT version >= 7.2
+TensorRT version >= 8.0
+
+Note
+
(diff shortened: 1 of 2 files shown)


J-shang:
this means we upgrade TensorRT dependency to >= 8.0? Is these upgrade available for most users?
And pls update the quantization speed up doc for this change.

linbinskn:
> 

> this means we upgrade TensorRT dependency to >= 8.0? Is these upgrade available for most users?
> And pls update the quantization speed up doc for this change.

@J-shang Good point! Have updated quantization speedup doc. I think the departure of network definition and configuration in latest TensorRT version is rational, and support the latest version is necessary for us. I believe most of people will use the latest version especially these people who want to try mixed precision in TensorRT.

J-shang (approved):

QuanluZhang (approved):

PR #3881, opened 2021-06-28
Use python 3.9 on pipeline

ultmaster:
This PR upgrade Python to 3.9 on pipeline unittests.

Lint is not updated due to smac issue (see follows):

- SMAC/BOHB test case is disabled because ConfigSpace used something called `tp_print` in its cython implementation, and the API is removed in Python 3.9. We can either contribute to upstream or wait until python 3.8 is EOL.
- Vision patch is no longer needed for latest version of PyTorch.
- Fixed an issue that makes DNGO tuner not work on later version of PyTorch. Not sure why. I think it has something to do with its model implementation, which is beyond the scope of this repo.
- Dropped the test cases on PyTorch 1.5.1, as I think it's a version that is too legacy to worth a maintenance.

Need to run IT to verify whether it works.


test/ut/sdk/test_builtin_tuners.py
```diff
@@ -22,8 +22,10 @@
 from nni.algorithms.hpo.regularized_evolution_tuner import RegularizedEvolutionTuner
 from nni.runtime.msg_dispatcher import _pack_parameter, MsgDispatcher
 
-if sys.platform != 'win32':
+smac_imported = False
+if sys.platform != 'win32' and sys.version_info < (3, 9):
```
J-shang, line 26: SMAC only works under 3.9?

dependencies/recommended.txt
```diff
@@ -1,12 +1,13 @@
...
-torch == 1.6.0 ; sys_platform == "darwin"
-torchvision == 0.7.0+cpu ; sys_platform != "darwin"
-torchvision == 0.7.0 ; sys_platform == "darwin"
+tensorflow == 2.5.0
+keras == 2.4.3
+tensorboard == 2.5.0
+torch == 1.9.0+cpu ; sys_platform != "darwin"
+torch == 1.9.0 ; sys_platform == "darwin"
+torchvision == 0.10.0+cpu ; sys_platform != "darwin"
+torchvision == 0.10.0 ; sys_platform == "darwin"
```
J-shang, line 10: do we need update the dockerfile?

ultmaster: I suggest we keep the Dockerfile unchanged for a while.

J-shang (approved):

QuanluZhang (approved):

PR #3884, opened 2021-06-29
Transformer Head Pruner

xiaowu0162:
This pr adds a pruner for pruning attention heads in transformers. 
To-do:
- [x] basic: module matching, name-based weight grouping, group-aware maskers using weight norm as criteria
- [x] graph-based weight grouping 
- [x] maskers relying on activation 
- [x] maskers relying on gradient
- [x] global sort in maskers
- [x] support iterative pruning from scratch ~/ integration with pruning scheduler for iterative features~
- [x] example 
- [x] docs


xiaowu0162:
Warning: this pr depends on the pr #3462 with `strict=False` parameter given to `torch.jit.trace`

xiaowu0162:
Unit test not added; will add in a later commit

docs/en_US/Compression/Pruner.rst
```diff
@@ -722,3 +723,125 @@ User configuration for Sensitivity Pruner
...
+
+Currently, the following head sorting criteria are supported:
+
+    * "l1_weight": rank heads by the L1-norm of weights of the query, key, and value projection matrices.
+    * "l2_weight": rank heads by the L2-norm of weights of the query, key, and value projection matrices.
+    * "l1_activation": rank heads by the L1-norm of their attention computation output.
+    * "l2_activation": rank heads by the L2-norm of their attention computation output.
+    * "taylorfo": rank heads by l1 norm of the output of attention computation * gradient for this output. Check more details in `this paper <https://arxiv.org/abs/1905.10650>`__ and `this one <https://arxiv.org/abs/1611.06440>`__.
+
+We support local sorting (i.e., sorting heads within a layer) and global sorting (sorting all heads together), and you can control by setting the ``global_sort`` parameter. Note that if ``global_sort=True`` is passed, all weights must have the same sparsity in the config list.
```
zheng-ningxin, line 752: why all weights have the same sparsity in the global_sort mode?

xiaowu0162: I think the user should specify the same sparsity for all modules when using global sort. But that value will be interpreted as a global sparsity, and each layer is likely to have different sparsity after pruning by global sort. I will clarify that in the docs.

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
```diff
@@ -0,0 +1,306 @@
...
+            if sparsity is None:
+                sparsity = group[0].config['sparsity']
+            if self.global_sort:
+                assert sparsity == group[0].config['sparsity'], \
+                    errmsg + ': for global_sort=True, the sparsity for all modules must be the same'
+            t = group[0].module.weight.size(0) / self.head_hidden_dim
+            assert t % 1 == 0, errmsg + ': head_hidden_dim must be a divisor of the output dimension of the ' \
+                                        'projection weights'
+
+    def remove_ungrouped_modules(self):
```
xiaowu0162, line 190: Should unwrap and wrap the model before and after calling this function

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
```diff
@@ -0,0 +1,306 @@
...
+                               'Exception message: ' + str(e))
+
+    def validate_weight_groups(self):
+        """
+        Sanity checks:
+            - Q, K, V projection weights in each groups must have the same shape
+            - output projection weight shape must match total hidden dimension (inferred from Q, K, V projection)
+            - Four weights in a group must have the same sparsity in their config
+            - If global_sort is specified, all weights must have the same sparsity
+            - head_hidden_dim must be a divisor of the output dimension of the projection weights
```
xiaowu0162, line 167: i.e., head number must be an int

nni/compression/pytorch/utils/shape_dependency.py
```diff
@@ -501,3 +502,133 @@ def export(self, filepath):
...
+                continue
+            if curnode.op_type == 'LayerNorm':
+                continue
+            children = self.graph.find_successors(curnode.unique_name)
+            children = [self.graph.name_to_node[name] for name in children]
+            for child in children:
+                queue.append(child)
+        return children_layers
+
+    def build_dependency(self):
```
xiaowu0162, line 587: Copy the algorithm to the docstring

examples/model_compress/pruning/transformers/run.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# Usage: ./run.sh gpu_id task
+
+export CUDA_VISIBLE_DEVICES=$1
+TASK_NAME=$2
+PRETRAINED_MODEL='bert-base-uncased'          # example: 'distilbert-base-uncased'  'roberta-base'   'bert-base-cased'   'bert-base-uncased'
+
+# parameters for pruning
+USAGE=2                       # change to different numbers to run examples with different configs
```
xiaowu0162, line 10: explain

examples/model_compress/pruning/transformers/transformer_pruning.py
```diff
@@ -0,0 +1,596 @@
...
+                        help="A csv or a json file containing the training data.")
+    parser.add_argument("--validation_file", type=str, default=None,
+                        help="A csv or a json file containing the validation data.")
+    parser.add_argument("--max_length", type=int, default=128,
+                        help=("The maximum total input sequence length after tokenization. Sequences longer than this "
+                              "will be truncated, sequences shorter will be padded if `--pad_to_max_lengh` is passed."))
+    parser.add_argument("--pad_to_max_length", action="store_true",
+                        help="If passed, pad all samples to `max_length`. Otherwise, dynamic padding is used.")
+    parser.add_argument("--use_slow_tokenizer", action="store_true",
+                        help="If passed, will use a slow tokenizer (not backed by the 🤗 Tokenizers library).",)
```
J-shang, line 92: 🤗?

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
```diff
@@ -0,0 +1,306 @@
...
+                training = self.bound_model.training
+                self.bound_model.eval()
+                self._trainer(self.bound_model, optimizer=self._optimizer, criterion=self._criterion, epoch=0)
+                self.update_mask()
+                self.bound_model.train(training)
+            else:
+                self.update_mask()
+
+            # for iterative pruning, finetune before next iteration
+            if self.num_iterations > 1:
```
J-shang, line 226: maybe `if current_iteration < self.num_iterations:` is better?

nni/algorithms/compression/pytorch/pruning/transformer_pruning_head_masker.py
```diff
@@ -0,0 +1,434 @@
...
+    """
+    def get_head_importance_scores(self, weight_group):
+        q_proj, k_proj, v_proj, _ = weight_group
+
+        n_heads = q_proj.module.weight.size()[0] // self.head_hidden_dim
+        query_proj_weights = q_proj.module.weight.data.view([n_heads, -1])
+        key_proj_weights = k_proj.module.weight.data.view([n_heads, -1])
+        value_proj_weights = v_proj.module.weight.data.view([n_heads, -1])
+
+        query_norm_avg = torch.sum(query_proj_weights ** 2, -1)
```
J-shang, line 252: why we don't use `torch.norm()`? and missing `torch.sqrt()`?

xiaowu0162: `torch.norm()` is cleaner. I will fix this.

examples/model_compress/pruning/transformers/transformer_pruning.py
xiaowu0162: This is from the original code. It seems that this emoji is called the "huggingface"

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
xiaowu0162: Then I think for one-shot pruning we also enter this condition because `0 < 1`?

xiaowu0162: As we discussed yesterday, I think I will merge this condition with the condition below so that finetuning inside pruner will only be performed between iterations of iterative pruning, not after the last iteration, nor after one-shot pruning

examples/model_compress/pruning/transformers/transformer_pruning.py
QuanluZhang: we cannot copy code with Apache license though it is for user-friendly, because our repo is MIT license.

QuanluZhang: we can try to provide instructions about how to get this transformer model and import it in our example code

docs/en_US/Compression/Pruner.rst
```diff
@@ -722,3 +723,125 @@ User configuration for Sensitivity Pruner
 **PyTorch**
 
 ..  autoclass:: nni.algorithms.compression.pytorch.pruning.SensitivityPruner
+
+Transformer Head Pruner
+-----------------------
+
+Transformer Head Pruner is a tool designed for pruning attention heads from the models belonging to the `Transformer family <https://proceedings.neurips.cc/paper/2017/file/3f5ee243547dee91fbd053c1c4a845aa-Paper.pdf>`__.
+Typically, each attention layer in the Transformer models consists of four weights: three projection matrices for query, key, value, and an output projection matrix. The outputs of the former three matrices contains the projected results for all heads. Normally, the results are then reshaped so that each head performs that attention computation independently. The final results are concatenated back before fed into the output projection. Please refer to the original paper for more details.
```
zheng-ningxin, line 731: Suggest to add a diagram to show the structure.

examples/model_compress/pruning/transformers/transformer_pruning.py
xiaowu0162: Fixed. Re-wrote the example from scratch.

docs/en_US/Compression/Pruner.rst
xiaowu0162: Added

J-shang:
the huggingface example has 400 lines, do we have a simple example?

xiaowu0162:
> the huggingface example has 400 lines, do we have a simple example?

I think the entire logic is not that long, but writting config list in 3 ways takes 100 lines, and comments take another 50 lines. I will try to make the example more concise.

J-shang (approved):
I think we can discuss the config_list format for structure some kind like head.
{'op_names': ['...'], 'step_on_dim': 64, 'dim': 0}
{'op_names': ['...'], 'group_on_dim': 12, 'dim': 0}

docs/en_US/Compression/Pruner.rst
```diff
@@ -722,3 +723,128 @@ User configuration for Sensitivity Pruner
...
+             'num_iterations': 1,
+             'epochs_per_iteration': 1,        # this is ignored when num_iterations = 1
+             'head_hidden_dim': 64,
+             'dummy_input': dummy_input,
+             'trainer': trainer,
+             'optimizer': optimizer
+             }
+    config_list = [{
+        'sparsity': 0.5,
+        'op_types': ["Linear"]
```
zheng-ningxin, line 782: Why Linear here? feel a little bit confused

xiaowu0162: Currently we can only handle attention layers with their projection weights written as separate Linear modules. I will further clarify this in the docs and the example.

xiaowu0162: improved docs

zheng-ningxin: Yes, I mean if this config list is passed to TransformerHeadPruner(which is only used to prune the head only), then why should we need the `op_type` attribute, not to mention `Linear` is not the type of attention head.

examples/model_compress/pruning/transformers/transformer_pruning.py
```diff
@@ -0,0 +1,411 @@
+import argparse
```
zheng-ningxin, line 1: # Copyright (c) Microsoft Corporation.

examples/model_compress/pruning/transformers/transformer_pruning.py
```diff
@@ -0,0 +1,411 @@
...
+def dry_run_or_finetune(args, model, train_dataloader, optimizer, device, epoch_num=None):
+    """
+    This function is used for to create a "trainer" that is passed to the pruner. 
+    If epoch_num is 0, the function just runs forward and backward without updating the parameters. 
+    This allows the pruner to collect data without parameter update (for activation or gradient based 
+    pruning methods). 
+    Otherwise, finetune the model for 1 epoch. This is called by the pruner during pruning iterations.
+    """
+    if epoch_num == 0:
+        print("Running forward and backward on the entire dataset without updating parameters...")
```
zheng-ningxin, line 186: logger.info?

examples/model_compress/pruning/transformers/transformer_pruning.py
```diff
@@ -0,0 +1,411 @@
...
+        for group_idx, group in enumerate(pruner.attention_name_groups):
+            # get the layer index
+            layer_idx = None
+            for part in group[0].split("."):
+                try:
+                    layer_idx = int(part)
+                    break
+                except:
+                    continue
+            if layer_idx is not None:
```
zheng-ningxin, line 380: Is there a better way to get the index of the attention head? the first integer may be not strong.

docs/en_US/Compression/Pruner.rst
xiaowu0162: I think `Compressor`'s config validation requires that either `op_types` or `op_names` is specified. But more importantly, if the user does not give the `attention_name_groups` parameter, then the user is also unlikely to give `op_names`. As a result, we will obtain graph and infer the name groups, and this process basically requires the `op_type` being `Linear`. So I think by letting the user write out explicitly we can remind them that we can only handle the scenario where different projection weights are represented as separate `Linear` layers.

examples/model_compress/pruning/transformers/transformer_pruning.py
xiaowu0162: This `layer_idx` is the layer index of the BERT encoder. Here I include these lines of code only to show the user how they may take advantage of the `pruned_heads` dict inside pruner to get the pruned heads for each group, and then match each group to the original layer, and finally call the built-in transformers `_prune_heads()` function to do model speedup. This is meant to be a temporary workaround before we can properly handle speedup for transformers.

xiaowu0162: If our speedup code after refactor can handle transformer, then I will replace these lines with our speedup methods (maybe in a separate pr)

xiaowu0162: Since the users are aware of the naming of their own model to prune, I think they can also use their own rules to match layers to groups

examples/model_compress/pruning/transformers/transformer_pruning.py
xiaowu0162: fixed

examples/model_compress/pruning/transformers/transformer_pruning.py
xiaowu0162: fixed

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
```diff
@@ -0,0 +1,316 @@
...
+        If num_iterations is 1, then no finetuning is performed by the pruner after pruning.
+    optimizer: torch.optim.Optimizer
+        Optimizer used to train model
+    trainer: function
+        Function used to train the model.
+        Users should write this function as a normal function to train the Pytorch model
+        and include `model, optimizer, criterion, epoch` as function arguments.
+    criterion: function
+        Function used to calculate the loss between the target and the output.
+        For example, you can use ``torch.nn.CrossEntropyLoss()`` as input.
```
zheng-ningxin, line 74: Feel like that the `TransformerHeadPruner` is too heavy. I prefer to locate this pruner as a one-shot pruner, which means we do not need handle with the `num_iteration`, `optimizer`, 'criterion' things. That's much clearer. All those finetuning related things we can offload to the outer search algorithms. We can discuss with Quanlu @QuanluZhang .

xiaowu0162: Yes, we can further discuss on that. One challenge is that this does not fit well in our current compression V1 framework (since the current iterative and dependency aware pruner are limited to convolutions), and compression V2 is not ready yet. My initial thought was to first integrate all these logic in one pruner (because of empirically good performance compared to one-shot pruning), and then factor out when compression V2 is ready.

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
```diff
@@ -0,0 +1,321 @@
...
+            assert group[0].config['sparsity'] == group[1].config['sparsity'] == \
+                   group[2].config['sparsity'] == group[3].config['sparsity'], \
+                errmsg + ': the sparsity of matrices in the same layer must be the same'
+            if sparsity is None:
+                sparsity = group[0].config['sparsity']
+            if self.global_sort:
+                assert sparsity == group[0].config['sparsity'], \
+                    errmsg + ': for global_sort=True, the sparsity for all modules must be the same'
+            t = group[0].module.weight.size(0) / self.head_hidden_dim
+            assert t % 1 == 0, errmsg + ': head_hidden_dim must be a divisor of the output dimension of the ' \
```
zheng-ningxin, line 193: ?

xiaowu0162: That was meant to check that `t` is an integer. Not sure if there are better ways to implement that

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
```diff
@@ -0,0 +1,321 @@
...
+        }], model, logger)
+
+        schema.validate(config_list)
+
+    def compress(self):
+        for pruning_iter in range(self.num_iterations):
+            if self.ranking_criterion in ['l1_activation', 'l2_activation', 'taylorfo']:
+                training = self.bound_model.training
+                self.bound_model.eval()
+                self._trainer(self.bound_model, optimizer=self._optimizer, criterion=self._criterion, epoch=0)
```
zheng-ningxin, line 234: why activation based criterion need training?

xiaowu0162: We need to run a `forward` once to collect activations and (optionally) `backward` once to collect gradients (for `taylorfo`). This might be problematic if the trainer provided by the user also do parameter updates (i.e., `optimizer.step`). In the docs and example I mention that the user can write the trainer so that when `epoch=0` no parameter update is performed. The pruner also guarantees that when it needs to collect data, it always passes 0 to the parameter `epoch`. When it uses the trainer for finetuning, it never passes 0 to `epoch`.

xiaowu0162: Or maybe we can pass -1 instead of 0.

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
zheng-ningxin: assert group[0].module.weight.size(0) % self.head_hidden_dim == 0

xiaowu0162: fixed

docs/en_US/Compression/Pruner.rst
```diff
@@ -722,3 +723,134 @@ User configuration for Sensitivity Pruner
...
+             }
+    config_list = [{
+        'sparsity': 0.5,
+        'op_types': ["Linear"],
+        'op_names': [x for layer in attention_name_groups[:6] for x in layer]      # first six layers
+    },
+    {
+        'sparsity': 0.25,
+        'op_types': ["Linear"],
+        'op_names': [x for layer in attention_name_groups[:6] for x in layer]      # last six layers
```
QuanluZhang, line 844: this is for the last six layers? it is the same as line 839

docs/en_US/Compression/Pruner.rst
QuanluZhang: does usage 1 reliable?

QuanluZhang: i suggest we only show usage 3 in doc

docs/en_US/Compression/Pruner.rst
xiaowu0162: fixed

docs/en_US/Compression/Pruner.rst
xiaowu0162: > does usage 1 reliable?

It is reliable for Huggingface transformers, but may not work for other implementations. In that case, the user would have to try usage 2.

xiaowu0162: > i suggest we only show usage 3 in doc

I will first promote usage 3 to the front, and delete usage 2

xiaowu0162: fixed docs

nni/algorithms/compression/pytorch/pruning/transformer_pruner.py
xiaowu0162: fixed; now the pruner passes `None` to the trainer's `epoch` argument when it is using this trainer to collect data.

QuanluZhang (approved):

PR #3891, opened 2021-06-30
Add reconfig API to Pruner

J-shang:
it is not a public API for users, just for pruning scheduler now.

did not support in 'AutoCompressPruner' 'LotteryTicketPruner' 'AMCPruner', because they will refactor to PruningScheduler or too complex.


nni/compression/pytorch/compressor.py
```diff
@@ -38,12 +39,36 @@ def __init__(self, model, config_list, optimizer=None):
...
+            if self.optimizer_cls.__name__ == 'SGD':
+                self.optimizer = self.optimizer_cls(model.parameters(), lr=0.001)
+            else:
+                self.optimizer = self.optimizer_cls(model.parameters())
+            groups = self.optimizer.param_groups
+            assert len(groups) == len(self.optimizer_param_groups)
+            for group, temp_group in zip(groups, self.optimizer_param_groups):
+                assert len(group) - 1 == len(temp_group)
+                for k, v in temp_group.items():
+                    group[k] = deepcopy(v)
```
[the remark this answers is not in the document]
J-shang: Because this step is using the original optimizer param groups to reset the optimizer new by us. We want the optimizer in each `reconfig` to be the same as the optimizer transfer by the user.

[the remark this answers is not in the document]
linbinskn: But in line45 `self.optimizer_param_groups = optimizer.state_dict()['param_groups']`, `self.optimizer_param_groups` get the reference of optimizer param_groups. It means that the following `reconfig` process will give new optimizer the final changed param_groups in the first optimizer.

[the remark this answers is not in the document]
J-shang: yes, and actually we never use the incoming optimizer from the user, we just get its state_dict, and all optimizers we used are initialized in line62-65. In this case, if the user has some changes on the incoming optimizer before initializing the compressor, we will get the change in the state_dict, but I think it is intuitive. And another problem is we can't get the initial param from an optimizer instance.

nni/compression/pytorch/compressor.py
```diff
@@ -38,12 +39,36 @@ def __init__(self, model, config_list, optimizer=None):
...
+        """
         assert isinstance(model, torch.nn.Module)
         self.validate_config(model, config_list)
 
         self.bound_model = model
         self.config_list = config_list
-        self.optimizer = optimizer
+
+        if self.optimizer is not None:
+            if self.optimizer_cls.__name__ == 'SGD':
```
ultmaster, line 62: Why SGD is so special?

nni/algorithms/compression/pytorch/pruning/amc/channel_pruning_env.py
```diff
@@ -227,7 +227,7 @@ def step(self, action):
 
     def reset(self):
         # restore env by loading the checkpoint
-        self.pruner.reset(self.checkpoint)
+        self.pruner.reset_bound_model(self.checkpoint)
```
ultmaster, line 230: Suggest testing this PR on IT.

nni/compression/pytorch/compressor.py
J-shang: because in SGD, lr is required param.

ultmaster (approved):

nni/algorithms/compression/pytorch/pruning/amc/channel_pruning_env.py
QuanluZhang: @J-shang please manually trigger IT for this PR

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -120,14 +119,16 @@ class AGPPruner(IterativePruner):
...
                          num_iterations=num_iterations, epochs_per_iteration=epochs_per_iteration)
         assert isinstance(optimizer, torch.optim.Optimizer), "AGP pruner is an iterative pruner, please pass optimizer of the model to it"
-        self.masker = MASKER_DICT[pruning_algorithm](model, self)
+
+    def reconfig(self, model, config_list):
+        super().reconfig(model, config_list)
         self.now_epoch = 0
-        self.freq = epochs_per_iteration
-        self.end_epoch = epochs_per_iteration * num_iterations
-        self.set_wrappers_attribute("if_calculated", False)
```
QuanluZhang, line 130: where is this line moved to?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -120,14 +119,16 @@ class AGPPruner(IterativePruner):
 
     def __init__(self, model, config_list, optimizer, trainer, criterion,
                  num_iterations=10, epochs_per_iteration=1, pruning_algorithm='level'):
-        super().__init__(model, config_list, optimizer=optimizer, trainer=trainer, criterion=criterion,
+        self.freq = epochs_per_iteration
+        self.end_epoch = epochs_per_iteration * num_iterations
```
QuanluZhang, line 123: why move these two lines to the top?

nni/compression/pytorch/compressor.py
```diff
@@ -63,7 +88,7 @@ def validate_config(self, model, config_list):
         """
         pass
 
-    def reset(self, checkpoint=None):
+    def reset_bound_model(self, checkpoint=None):
```
QuanluZhang, line 91: what is this function used for?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
J-shang: move all constant assignments before init, because we  call `reconfig()` in `__init__`, and these constant may be used in `reconfig()`.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
J-shang: we have set this in `DependencyAwarePruner.__init__()`.

nni/compression/pytorch/compressor.py
J-shang: this function is used for re-wrap the model, used in `AMCPruner` and `reconfig()`

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
QuanluZhang: got it

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -120,14 +119,16 @@ class AGPPruner(IterativePruner):
...
+
+        super().__init__(model, config_list, optimizer=optimizer, pruning_algorithm=pruning_algorithm, trainer=trainer, criterion=criterion,
                          num_iterations=num_iterations, epochs_per_iteration=epochs_per_iteration)
         assert isinstance(optimizer, torch.optim.Optimizer), "AGP pruner is an iterative pruner, please pass optimizer of the model to it"
```
QuanluZhang, line 127: suggest to put this line above `super().__init__` to be consistent with other pruners

nni/algorithms/compression/pytorch/pruning/net_adapt_pruner.py
```diff
@@ -84,31 +84,34 @@ def evaluator(model):
...
         if not os.path.exists(self._experiment_data_dir):
             os.makedirs(self._experiment_data_dir)
 
         self._tmp_model_path = os.path.join(self._experiment_data_dir, 'tmp_model.pth')
 
+        super().__init__(model, config_list)
+
+    def reconfig(self, model, config_list):
+        # models used for iterative pruning and evaluation
+        self._model_to_prune = copy.deepcopy(model)
```
QuanluZhang, line 105: if the model is pruned from the original state just with a different sparsity ratio, what is the benefit? in other words, what type of information is inherited between two consecutive iteration?

nni/compression/pytorch/compressor.py
```diff
@@ -38,12 +39,36 @@ def __init__(self, model, config_list, optimizer=None):
...
+        self.optimizer = optimizer
+        if optimizer is not None:
+            self.optimizer_cls = optimizer.__class__
+            self.optimizer_param_groups = optimizer.state_dict()['param_groups']
+            # don't save 'params' in a 'param_group' in 'param_groups'
+            [group.pop('params') for group in self.optimizer_param_groups]
+        if model is not None and config_list is not None:
+            self.reconfig(model, config_list)
+
+    def reconfig(self, model, config_list):
```
QuanluZhang, line 51: maybe we can split this function to two functions. one is `reset_status`, the other is `reset_config`. There could be another function `reset` which contains the above two functions

J-shang: `reset_config` rewrap the `new_model` with `new_config`, and `reset_status` used to reset the median params or all other params?

QuanluZhang: `reset_with_new_config` is for resetting the pruner, it invokes `reset_status` and `set_config`.

J-shang: split it.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
J-shang: move this line above `super().__init__`

nni/algorithms/compression/pytorch/pruning/net_adapt_pruner.py
J-shang: It is a good question and in `net_adapt`, it will find just a layer that has the best performance for the current iteration sparsity. Then update config_list.
i.e. iteratoin one, `sparsity=0.1` `conv1 performance: 0.8; conv2 performance: 0.6`, then `[] <- {'op_names': 'conv1', 'sparsity': 0.1}`.
iteratoin two, `sparsity=0.2`  `conv1 performance: 0.4; conv2 performance: 0.7`, then `[{'op_names': 'conv1', 'sparsity': 0.1}] <- {'op_names': 'conv2', 'sparsity': 0.2}`.
...

nni/compression/pytorch/compressor.py
```diff
@@ -38,35 +39,52 @@ def __init__(self, model, config_list, optimizer=None):
...
-        self.bound_model = model
-        self.config_list = config_list
         self.optimizer = optimizer
+        if optimizer is not None:
+            self.optimizer_cls = optimizer.__class__
+            self.optimizer_param_groups = optimizer.state_dict()['param_groups']
+            # don't save 'params' in a 'param_group' in 'param_groups'
+            [group.pop('params') for group in self.optimizer_param_groups]
+        if model is not None and config_list is not None:
+            self.reset_with_new_config(model, config_list)
```
QuanluZhang, line 49: does this refactor affect quantizers?

QuanluZhang:
closed. the modifications will be incorporated in model compression 2.0

PR #3895, opened 2021-07-01
[webui] Support dict intermediate metrics(only for number type)

Lijiaoa:
1. trial intermediate result:

image

2. compare model:

image

### DIFF
--- a/ts/webui/src/components/modals/Compare.tsx
+++ b/ts/webui/src/components/modals/Compare.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { renderToString } from 'react-dom/server';
-import { Stack, Modal, IconButton, IDragOptions, ContextualMenu } from '@fluentui/react';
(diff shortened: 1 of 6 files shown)


ts/webui/src/components/modals/Compare.tsx
```diff
@@ -218,9 +240,23 @@ class Compare extends React.Component<CompareProps, {}> {
             duration: convertDuration(trial.duration),
             parameters: flatten(trial.parameters(inferredSearchSpace)),
             metrics: flatten(trial.metrics(TRIALS.inferredMetricSpace())),
-            intermediates: _parseIntermediates(trial)
+            intermediates: _parseIntermediates(trial, intermediateKey)
         }));
 
+        if (trials[0].intermediates !== undefined && trials[0].intermediates[0]) {
```
ultmaster, line 246: What if `trials[0]` has empty intermediates?

Lijiaoa: I think `empty intermediates` is same as `x.intermediates is undefined`...

ultmaster (approved):

acured (approved):

PR #3896, opened 2021-07-02
[Model Compression] Add global sort for taylor pruner

linbinskn:
This PR supports global sort for taylor pruner.
Constraint set to 1 to prevent all channels pruned so that at least one channel will be kept.


nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
```diff
@@ -33,11 +33,12 @@ class StructuredWeightMasker(WeightMasker):
 
     """
 
-    def __init__(self, model, pruner, preserve_round=1, dependency_aware=False):
+    def __init__(self, model, pruner, preserve_round=1, dependency_aware=False, global_sort=False):
```
J-shang, line 36: I think it may be better to implement it for Taylor alone without changing the underlying interface, because we don't have another masker use these interfaces, and `slim` is natural using global sort.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
```diff
@@ -477,6 +488,29 @@ def __init__(self, model, pruner, statistics_batch_num=1):
...
+            all_channel_contributions.view(-1), k, largest=False)[0].max()
+        print(f'set global threshold to {self.global_threshold}')
+    
+    def _get_global_num_prune(self, wrapper, wrapper_idx):
+        if self.global_threshold is None:
+            self._get_global_threshold()
+        weight = wrapper.module.weight.data
+        filters = weight.size(0)
+        channel_contribution = self.get_channel_sum(wrapper, wrapper_idx)
+        num_prune = channel_contribution[channel_contribution < self.global_threshold].size()[0]
```
xiaowu0162, line 510: Just a dumb question, do we want < or <= here?

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
linbinskn: But based our discussion before, we agree that we should modify the condition in 

> https://github.com/microsoft/nni/blob/master/nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py#L63

May be we can add `assert` to tell user that we only support Taylor currently?

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
linbinskn: It is a key point and I thought it before, we can choose using '<=' which may cause `num_prune` larger than specific sparsity or `<` which may cause `num_prune` smaller than specific sparsity. And I finally choose to use `<` since little sparsity is much more safe and iterative pruning process would help it do further pruning.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
J-shang: the current implementation is good for me, just a little opinion, don't need to modify. And we need update doc for Taylor.

linbinskn: Have updated doc.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
```diff
@@ -477,6 +488,29 @@ def __init__(self, model, pruner, statistics_batch_num=1):
...
         self.pruner.patch_optimizer(self.calc_contributions)
+        self.global_threshold = None
+
+    def _get_global_threshold(self):
+        channel_contribution_list = []
+        for wrapper_idx, wrapper in enumerate(self.pruner.get_modules_wrapper()):
+            channel_contribution = self.get_channel_sum(wrapper, wrapper_idx)
+            channel_contribution_list.append(channel_contribution)
+        all_channel_contributions = torch.cat(channel_contribution_list)
+        k = int(all_channel_contributions.shape[0] * self.pruner.config_list[0]['sparsity'])
```
J-shang, line 499: what if the filters' sizes are different?

xiaowu0162 (approved):

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
linbinskn: That's truly a key problem, I will fix it.

linbinskn: Have fixed.

J-shang: seems we need `view(-1)` to the contribution, or if they can cat together with different size?

linbinskn: That's a problem. Have fixed.

J-shang (approved):

nni/algorithms/compression/pytorch/pruning/dependency_aware_pruner.py
```diff
@@ -27,7 +27,7 @@ class DependencyAwarePruner(Pruner):
     """
 
     def __init__(self, model, config_list, optimizer=None, pruning_algorithm='level', dependency_aware=False,
-                 dummy_input=None, **algo_kwargs):
+                 dummy_input=None, global_sort=False, **algo_kwargs):
```
zheng-ningxin, line 30: Do all the dependency pruners support `global_sort` mode? If not, I'm a little concerned.

zheng-ningxin:
Please add some unit tests for the global mode.

nni/algorithms/compression/pytorch/pruning/dependency_aware_pruner.py
QuanluZhang: agree, how many pruners support `global_sort`?

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
```diff
@@ -51,10 +51,13 @@ def __init__(self, model, config_list, optimizer=None, pruning_algorithm='slim',
         dummy_input: torch.Tensor
             The dummy input to analyze the topology constraints. Note that,
             the dummy_input should on the same device with the model.
+        global_sort: bool
+            If prune the model in a global-sort way.
+            Only support TaylorFOWeightFilterPruner currently.
```
QuanluZhang, line 56: if only `TaylorFOWeightFilterPruner` supports `global_sort`, suggest to put this argument in `TaylorFOWeightFilterPruner`.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
```diff
@@ -477,6 +488,32 @@ def __init__(self, model, pruner, statistics_batch_num=1):
...
+            channel_contribution = self.get_channel_sum(wrapper, wrapper_idx)
+            wrapper_size = wrapper.module.weight.size().numel()
+            channel_size = wrapper.module.weight.size(0)
+            contribution_expand = channel_contribution.expand(int(wrapper_size / channel_size), channel_size).reshape(-1)
+            channel_contribution_list.append(contribution_expand)
+        all_channel_contributions = torch.cat(channel_contribution_list)
+        k = int(all_channel_contributions.shape[0] * self.pruner.config_list[0]['sparsity'])
+        self.global_threshold = torch.topk(
+            all_channel_contributions.view(-1), k, largest=False)[0].max()
+        print(f'set global threshold to {self.global_threshold}')
```
QuanluZhang, line 505: better to remove this `print`

nni/algorithms/compression/pytorch/pruning/dependency_aware_pruner.py
linbinskn: Not all of them support `global_sort`, have deleted parameters in them.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
linbinskn: Have removed.

nni/algorithms/compression/pytorch/pruning/iterative_pruner.py
linbinskn: Done.

linbinskn:
> Please add some unit tests for the global mode.

Will add it later.

nni/algorithms/compression/pytorch/pruning/structured_pruning_masker.py
QuanluZhang: so you put global sort in `StructuredWeightMasker` but not in `TaylorFOWeightFilterPrunerMasker`. what is the reason for it? is it because this would make `TaylorFOWeightFilterPrunerMasker` have different initial arguments with other structured weight masker?

linbinskn: Because in [here](https://github.com/microsoft/nni/pull/3896/files#diff-3c3764e5ee06fcf6e53f6bf6a96fe6d9c493906106f6714b040590cb2ab35255R64), we modify the conditional judgement statement which discriminate different situations including global-sort and dependency-aware. The reason why we do conditional judgement [here](https://github.com/microsoft/nni/pull/3896/files#diff-3c3764e5ee06fcf6e53f6bf6a96fe6d9c493906106f6714b040590cb2ab35255R64) is because we think global-sort should be the same level with dependency-aware. This conditional judgement is done in `StructuredWeightMasker ` so we put `global_sort` here.

QuanluZhang (approved):

PR #3898, opened 2021-07-02
Escape environment variable PATH for Windows scripts

liuzhe-lz:
I once tried to fix all similar problems together. But it looks too dangerous.


J-shang:
seems we need test this in pipeline

acured (approved):

liuzhe-lz:
> seems we need test this in pipeline

I found UT is completely broken. Testing exit experiment stops the process and most test cases are not run…
I need to fix it before adding testcase.

J-shang (approved):

SparkSnail (approved):

PR #3899, opened 2021-07-02
fix multi-thread event handling issue

QuanluZhang:
- [x] fix the bug that sometimes retiarii experiment cannot exit successfully after finishing all the trials, because nnimanager requests more trials from dispatcher/strategy
- [x] fix incorrect order of items for cleaning up retiarii experiment
- [x] fully test the modification


liuzhe-lz (approved):

ultmaster (approved):
I'm not confident of this approve, because I have no idea why it works now and why it didn't work before.

SparkSnail (approved):

PR #3910, opened 2021-07-06
[Retiarii]: Add info required by nn-meter to graph ir

kaleid-liner:
Tested on #3876 ShuffleNetV2


nni/retiarii/graph.py
```diff
@@ -309,7 +309,7 @@ def add_node(self, name: str, operation: Operation) -> 'Node': ...
     @overload
     def add_node(self, name: str, type_name: str, parameters: Dict[str, Any] = {}) -> 'Node': ...
 
-    def add_node(self, name, operation_or_type, parameters={}):
+    def add_node(self, name, operation_or_type, parameters=None):
```
ultmaster, line 312: Why change the behavior of this API?

nni/retiarii/nn/pytorch/api.py
```diff
@@ -78,16 +78,21 @@ def __init__(self, candidates: Union[Dict[str, nn.Module], List[nn.Module]], lab
...
                     "Please don't use a reserved name '{}' for your module.".format(name)
                 self.add_module(name, module)
                 self.names.append(name)
+                if not self.chosen:
+                    self.chosen = module
         elif isinstance(candidates, list):
             for i, module in enumerate(candidates):
                 self.add_module(str(i), module)
                 self.names.append(str(i))
+                if not self.chosen:
```
ultmaster, line 94: Why do we need a chosen? This seems a `fix` mode and should be done in `__new__`.

nni/retiarii/graph.py
mydmdm: From my knowledge, it is not a good practice to use `{}` as default values, here is a reference 
https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

Not sure whether there is any other special reason, but I believe it worth a refactor to change all default `{}` and `[]` to `None`

kaleid-liner: > Why change the behavior of this API?

And this will cause parameters of the nodes to share the same dict object (the initial `{}`). When one of the parameters is changed, all of the others will also be changed at the same time. 

This results in some strange bugs.

nni/retiarii/nn/pytorch/api.py
kaleid-liner: I need to call the first candidate. But as `candidates` is a list/dict, instead of `ModuleList`, so it can't be directly accessed in `forward` (eg., `candidates[0]`). I don't know if there are better ways to achieve that.

nni/retiarii/graph.py
ultmaster: I'm okay with this change if you insist. But you should put a `if parameters is None: parameters = {}` in the body of this function.

ultmaster: If parameters is immutable in the function body, actually putting `parameters={}` in the arguments works, except that IDE will complain.

nni/retiarii/nn/pytorch/api.py
ultmaster: In that case, you can write: `self._modules[self.names[0]](x)`

nni/retiarii/graph.py
kaleid-liner: > I'm okay with this change if you insist. But you should put a if parameters is None: parameters = {} in the body of this function.

Yes. I have added that to `Opeartion::__init__()` and `Operation::__new__()`.

> If parameters is immutable in the function body, actually putting parameters={} in the arguments works, except that IDE will complain.

But if you assign it to the attribute of the object, when next time you modify that attribute, the other objects will also be affected. (Actually this is why it incurs some strange bugs and it costs me a night to debug out 😂

nni/retiarii/converter/graph_gen.py
```diff
@@ -597,10 +579,15 @@ def convert_module(self, script_module, module, module_name, ir_model):
...
+            candidate_name_list = []
+            for cand_name in module.names:
+                cand = module[cand_name]
+                script_cand = script_module._modules[cand_name]
+                cand_name = build_cand_name(cand_name, module.label)
+                candidate_name_list.append(cand_name)
                 cand_type = '__torch__.' + get_importable_name(cand.__class__)
-                graph.add_node(cand_name, cand_type, get_init_parameters_or_fail(cand))
+                graph.add_node(cand_name, cand_type, get_init_parameters_or_fail(cand, silently=True))
+                self._convert_module(script_cand, cand, cand_name, ir_model)
```
ultmaster, line 590: I think choices in LayerChoice should stop parsing. Please test with examples in https://github.com/microsoft/nni/tree/master/test/retiarii_test/darts to see if that works.

nni/retiarii/converter/graph_gen.py
```diff
@@ -654,6 +641,137 @@ def convert_module(self, script_module, module, module_name, ir_model):
...
+        dict
+            the input arguments of this module
+        """
+
+        self._convert_module(script_module, module, module_name, ir_model)
+
+
+class HardwareAwareGraphConverter(GraphConverter):
+
+    def convert_module(self, script_module, module, module_name, ir_model, example_inputs):
```
ultmaster, line 672: Suggest adding docstring and unittests for this module.

nni/retiarii/nn/pytorch/api.py
kaleid-liner: > In that case, you can write: self._modules[self.names[0]](x)
Thanks, I got it.

nni/retiarii/converter/graph_gen.py
kaleid-liner: But nn-meter needs the subgraph of layerchoice's candidates.

If there are really some errors, users just need to wrap the candidate with `serialize` to stop parsing.

kaleid-liner: Actually `serialize` is required under the former implementation. There is no difference if `serialize` is provided. What I implement is that when the candidate is not wrapped with `serialize`, I will parse it recursively.

ultmaster: Make sense to me. If you have already tested the example, I'm okay with this change.

nni/retiarii/nn/pytorch/api.py
kaleid-liner: > In that case, you can write: `self._modules[self.names[0]](x)`

It seems I can't access `self._modules` in `forward`.

kaleid-liner:
I rebase and forced push to include multi-trial SPOS example (#3876) as later development is based on it.

nni/retiarii/strategy/filter.py
```diff
@@ -0,0 +1,30 @@
+from nn_meter import get_default_config, load_latency_predictors  # pylint: disable=import-error
+
+
+class LatencyFilter:
```
ultmaster, line 4: Suggest putting this filter into examples.

nni/retiarii/strategy/bruteforce.py
```diff
@@ -86,15 +86,28 @@ class Random(BaseStrategy):
         Do not try the same configuration twice. When variational is true, deduplication is not supported. Default: true.
     """
 
-    def __init__(self, variational=False, dedup=True):
+    def __init__(self, variational=False, dedup=True, model_filter=None):
```
ultmaster, line 89: Please update the docstring correspondingly.

nni/retiarii/experiment/pytorch.py
```diff
@@ -106,15 +107,20 @@ def _validation_rules(self):
...
+                     parse_shape=False, example_inputs=None):
     # TODO: this logic might need to be refactored into execution engine
     if full_ir:
         try:
             script_module = torch.jit.script(base_model)
         except Exception as e:
             _logger.error('Your base model cannot be parsed by torch.jit.script, please fix the following error:')
             raise e
-        base_model_ir = convert_to_graph(script_module, base_model)
+        if parse_shape:
```
ultmaster, line 119: This looks like another execution engine.
I suppose you can merge `parse_shape` and `example_inputs` with `full_ir`, and rename `full_ir` to something like `ir_format`.

nni/retiarii/experiment/pytorch.py
```diff
@@ -171,7 +180,8 @@ def __init__(self, base_model: nn.Module, trainer: Union[Evaluator, BaseOneShotT
 
     def _start_strategy(self):
         base_model_ir, self.applied_mutators = preprocess_model(
-            self.base_model, self.trainer, self.applied_mutators, full_ir=self.config.execution_engine != 'py')
+            self.base_model, self.trainer, self.applied_mutators, full_ir=self.config.execution_engine != 'py',
```
ultmaster, line 183: Please add another value to `execution_engine`.

nni/retiarii/nn/pytorch/api.py
ultmaster: In that case, you can write `self._first_module = self._modules[self.names[0]](x)`. Might be more clear that `self.chosen`.
My previous concern is that some one-shot algorithms use layer choice directly and might treat `self.chosen` as another module due to wrong implementation. But I found that actually `self.names` is used in `__iter__`. So never mind...

examples/nas/oneshot/spos/multi_trial.py
```diff
@@ -142,9 +143,12 @@ def _main(port):
                                 val_dataloaders=pl.DataLoader(test_dataset, batch_size=64),
                                 max_epochs=2, gpus=1)
 
-    simple_strategy = strategy.Random()
+    simple_strategy = strategy.Random(model_filter=LatencyFilter(100))
 
-    exp = RetiariiExperiment(base_model, trainer, [], simple_strategy)
+    example_inputs = torch.randn(1, 3, 32, 32)
+
+    base_model.eval()
```
QuanluZhang, line 150: it is a little strange to config dummy input and "eval" here, let's discuss in the meeting

ultmaster: Agree.
I think this input should be obtained from dataloader.

nni/retiarii/converter/graph_gen.py
```diff
@@ -654,8 +641,212 @@ def convert_module(self, script_module, module, module_name, ir_model):
...
+            the whole graph ir
 
-def convert_to_graph(script_module, module):
+        Returns
+        -------
+        Graph
+            the built graph ir from module, ```None``` means do not further parse the module
+        dict
+            the input arguments of this module
+        """
```
QuanluZhang, line 665: why wrap `_convert_module` with exactly the same input arguments? and seems `convert_module` does not have returns, but there are returns in the docstring

nni/retiarii/converter/graph_gen.py
QuanluZhang: suggest to add ut for non-serialize case.

kaleid-liner:
The tests failed because I move `LatencyFilter` to `utils.py`. The import `from nn_meter import ` of this file incurs `ImportError`, as `nn_meter` hasn't been released,

nni/retiarii/strategy/utils.py
```diff
@@ -1,11 +1,16 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+
 import collections
+import logging
 from typing import Dict, Any, List
+from nn_meter import get_default_config, load_latency_predictors  # pylint: disable=import-error
```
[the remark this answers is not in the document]
kaleid-liner: ok

nni/retiarii/experiment/pytorch.py
```diff
@@ -154,7 +160,8 @@ def debug_mutated_model(base_model, trainer, applied_mutators):
 
 class RetiariiExperiment(Experiment):
     def __init__(self, base_model: nn.Module, trainer: Union[Evaluator, BaseOneShotTrainer],
-                 applied_mutators: List[Mutator] = None, strategy: BaseStrategy = None):
+                 applied_mutators: List[Mutator] = None, strategy: BaseStrategy = None,
+                 parse_shape: bool = False, example_inputs = None):
```
ultmaster, line 164: Any chance we can put them into config?

PR #3913, opened 2021-07-07
add nfs shared storage integration test

acured:


ultmaster:
Have you tested it on IT?

test/config/integration_tests.yml
```diff
@@ -48,19 +48,18 @@ testCases:
...
-#      exportedDirectory: /home/nni/mnt/
-#  validator:
-#    class: FileExistValidator
-#    kwargs:
-#      rootpath: /tmp/nnimount/testlocalrootpath
+- name: shared-storage-remote-nfs
+  configFile: test/config/sharedstorage_test/config_sharedstorage_remote_nfs.yml
+  config:
+    sharedStorage:
+      localMountPoint: /tmp/nnimount/testlocalrootpath
```
SparkSnail, line 55: Suggest to set mount point value to pipeline variables, use value in $(LOCAL_MOUNT_POINT) way in case the value will be changed when we reset new path in the future.

acured:
> Have you tested it on IT?

IT passed, https://msrasrg.visualstudio.com/NNIOpenSource/_build/results?buildId=24493&view=results

test/config/integration_tests.yml
acured: Changed.

SparkSnail (approved):

ultmaster (approved):

PR #3915, opened 2021-07-08
Add post training observer_quantizer

chenbohua3:
This PR adds a post training quantizer `ObserverQuantizer`. It uses PyTorch official `HistogramObserver` to calculate quantization information of activation and uses `MinMaxObserver` for weight. This quantizer has two advantages:

1. Users can directly get quantization information of a pre-trained model with a little calibration data. Training is not needed.
2. The quantization information generated by this quantizer can be used to initialize the scale and zero point of QAT quantizer, which will lead to a faster convergence and a better accuracy.

Also, users can easily customize their own observer, as long as the usage of the observer is consistent with that of PyTorch. I will upload our implementation of a classic post-training quantizer (e.g. Kullback-Leibler observer, used by TensorRT) in the next pr.

### DIFF
--- a/examples/model_compress/quantization/observer_quantizer.py
+++ b/examples/model_compress/quantization/observer_quantizer.py
@@ -41,8 +41,8 @@ def test(model, device, test_loader):
 def calibration(model, device, test_loader):
     model.eval()
     with torch.no_grad():
-        for data, target in test_loader:
-            data, target = data.to(device), target.to(device)
+        for data, _ in test_loader:
+            data = data.to(device)
             model(data)
 
 
--- a/examples/model_compress/quantization/observer_quantizer.py
+++ b/examples/model_compress/quantization/observer_quantizer.py
@@ -1,0 +1,155 @@
+import torch
+import torch.nn.functional as F
+from torchvision import datasets, transforms
+from nni.algorithms.compression.pytorch.quantization import ObserverQuantizer
+
+
+class Mnist(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv1 = torch.nn.Conv2d(1, 20, 5, 1)
+        self.conv2 = torch.nn.Conv2d(20, 50, 5, 1)
+        self.fc1 = torch.nn.Linear(4 * 4 * 50, 500)
+        self.fc2 = torch.nn.Linear(500, 10)
+        self.relu1 = torch.nn.ReLU6()
+        self.relu2 = torch.nn.ReLU6()
+        self.relu3 = torch.nn.ReLU6()
+        self.max_pool1 = torch.nn.MaxPool2d(2, 2)
+        self.max_pool2 = torch.nn.MaxPool2d(2, 2)
+
+    def forward(self, x):
+        x = self.relu1(self.conv1(x))
+        x = self.max_pool1(x)
+        x = self.relu2(self.conv2(x))
+        x = self.max_pool2(x)
+        x = x.view(-1, 4 * 4 * 50)
+        x = self.relu3(self.fc1(x))
+        x = self.fc2(x)
+        return F.log_softmax(x, dim=1)
+
+
+def train(model, device, train_loader, optimizer):
+    model.to(device)
+    model.train()
+    for batch_idx, (data, target) in enumerate(train_loader):
+        data, target = data.to(device), target.to(device)
+        optimizer.zero_grad()
+        output = model(data)
+        loss = F.nll_loss(output, target)
+        loss.backward()
+        optimizer.step()
+        if batch_idx % 100 == 0:
+            print('{:2.0f}%  Loss {}'.format(100 * batch_idx / len(train_loader), loss.item()))
+
+
+def test(model, device, test_loader):
+    model.eval()
+    test_loss = 0
+    correct = 0
+    with torch.no_grad():
+        for data, target in test_loader:
+            data, target = data.to(device), target.to(device)
+            output = model(data)
+            test_loss += F.nll_loss(output, target, reduction='sum').item()
+            pred = output.argmax(dim=1, keepdim=True)
+            correct += pred.eq(target.view_as(pred)).sum().item()
+    test_loss /= len(test_loader.dataset)
+
+    print('Loss: {}  Accuracy: {}%)\n'.format(
+        test_loss, 100 * correct / len(test_loader.dataset)))
+
+
+def calibration(model, device, test_loader):
+    model.eval()
+    with torch.no_grad():
+        for data, target in test_loader:
+            data, target = data.to(device), target.to(device)
+            model(data)
+
+
+def test_trt(engine, test_loader):
+    test_loss = 0
+    correct = 0
+    time_elasped = 0
+    for data, target in test_loader:
+        output, time = engine.inference(data)
+        test_loss += F.nll_loss(output, target, reduction='sum').item()
+        pred = output.argmax(dim=1, keepdim=True)
+        correct += pred.eq(target.view_as(pred)).sum().item()
+        time_elasped += time
+    test_loss /= len(test_loader.dataset)
+
+    print('Loss: {}  Accuracy: {}%'.format(
+        test_loss, 100 * correct / len(test_loader.dataset)))
+    print("Inference elapsed_time (whole dataset): {}s".format(time_elasped))
+
+
+def main():
+    torch.manual_seed(0)
+    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+    trans = transforms.Compose([transforms.ToTensor(), transforms.Normalize((0.1307,), (0.3081,))])
+    train_loader = torch.utils.data.DataLoader(
+        datasets.MNIST('data', train=True, download=True, transform=trans),
+        batch_size=64, shuffle=True)
+    test_loader = torch.utils.data.DataLoader(
+        datasets.MNIST('data', train=False, transform=trans),
+        batch_size=1000, shuffle=True)
+
+    model = Mnist()
+    configure_list = [{
+            'quant_types': ['weight', 'input'],
+            'quant_bits': {'weight': 8, 'input': 8},
+            'op_names': ['conv1'],
+        }, {
+            'quant_types': ['output'],
+            'quant_bits': {'output': 8, },
+            'op_names': ['relu1'],
+        }, {
+            'quant_types': ['weight', 'input'],
+            'quant_bits': {'weight': 8, 'input': 8},
+            'op_names': ['conv2'],
+        }, {
+            'quant_types': ['output'],
+            'quant_bits': {'output': 8},
+            'op_names': ['relu2'],
+        }, {
+            'quant_types': ['output'],
+            'quant_bits': {'output': 8},
+            'op_names': ['max_pool2'],
+        }
+    ]
+    optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.5)
+
+    # Train the model to get a baseline performance
+    for epoch in range(5):
+        print('# Epoch {} #'.format(epoch))
+        train(model, device, train_loader, optimizer)
+
+    test(model, device, test_loader)
+
+    # Construct the ObserverQuantizer. Note that currently ObserverQuantizer only works
+    # in evaluation mode.
+    quantizer = ObserverQuantizer(model.eval(), configure_list, optimizer)
+    # Use the test data set to do calibration, this will not change the model parameters
+    calibration(model, device, test_loader)
+    # obtain the quantization information and switch the model to "accuracy verification" mode
+    quantizer.compress()
+
+    # measure the accuracy of the quantized model.
+    test(model, device, test_loader)
+
+    model_path = "mnist_model.pth"
+    calibration_path = "mnist_calibration.pth"
+    calibration_config = quantizer.export_model(model_path, calibration_path)
+    print("calibration_config: ", calibration_config)
+
+    # For now the quantization settings of ObserverQuantizer does not match the TensorRT,
+    # so TensorRT conversion are not supported
+    # current settings:
+    # weight      : per_tensor_symmetric, qint8
+    # activation  : per_tensor_affine, quint8, reduce_range=True
+
+
+if __name__ == '__main__':
+    main()
--- a/examples/model_compress/quantization/observer_quantizer.py
+++ b/examples/model_compress/quantization/observer_quantizer.py
@@ -2,30 +2,9 @@ import torch
 import torch.nn.functional as F
 from torchvision import datasets, transforms
 from nni.algorithms.compression.pytorch.quantization import ObserverQuantizer
-
-
-class Mnist(torch.nn.Module):
-    def __init__(self):
-        super().__init__()
-        self.conv1 = torch.nn.Conv2d(1, 20, 5, 1)
-        self.conv2 = torch.nn.Conv2d(20, 50, 5, 1)
-        self.fc1 = torch.nn.Linear(4 * 4 * 50, 500)
-        self.fc2 = torch.nn.Linear(500, 10)
-        self.relu1 = torch.nn.ReLU6()
-        self.relu2 = torch.nn.ReLU6()
-        self.relu3 = torch.nn.ReLU6()
-        self.max_pool1 = torch.nn.MaxPool2d(2, 2)
-        self.max_pool2 = torch.nn.MaxPool2d(2, 2)
-
-    def forward(self, x):
-        x = self.relu1(self.conv1(x))
-        x = self.max_pool1(x)
-        x = self.relu2(self.conv2(x))
-        x = self.max_pool2(x)
-        x = x.view(-1, 4 * 4 * 50)
-        x = self.relu3(self.fc1(x))
-        x = self.fc2(x)
-        return F.log_softmax(x, dim=1)
+import sys
+sys.path.append('../models')
+from mnist.naive import NaiveModel
 
 
 def train(model, device, train_loader, optimizer):
@@ -67,23 +46,6 @@ def calibration(model, device, test_loader):
             model(data)
 
 
-def test_trt(engine, test_loader):
-    test_loss = 0
-    correct = 0
-    time_elasped = 0
-    for data, target in test_loader:
-        output, time = engine.inference(data)
-        test_loss += F.nll_loss(output, target, reduction='sum').item()
-        pred = output.argmax(dim=1, keepdim=True)
-        correct += pred.eq(target.view_as(pred)).sum().item()
-        time_elasped += time
-    test_loss /= len(test_loader.dataset)
-
-    print('Loss: {}  Accuracy: {}%'.format(
-        test_loss, 100 * correct / len(test_loader.dataset)))
-    print("Inference elapsed_time (whole dataset): {}s".format(time_elasped))
-
-
 def main():
     torch.manual_seed(0)
     device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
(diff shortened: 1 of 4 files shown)


examples/model_compress/quantization/observer_quantizer.py
```diff
@@ -0,0 +1,155 @@
+import torch
+import torch.nn.functional as F
+from torchvision import datasets, transforms
+from nni.algorithms.compression.pytorch.quantization import ObserverQuantizer
+
+
+class Mnist(torch.nn.Module):
```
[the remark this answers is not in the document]
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -119,6 +122,217 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
         return grad_output
 
 
+class ObserverQuantizer(Quantizer):
+    """
```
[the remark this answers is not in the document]
chenbohua3: done

examples/model_compress/quantization/observer_quantizer.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # measure the accuracy of the quantized model.
+    test(model, device, test_loader)
+
+    model_path = "mnist_model.pth"
+    calibration_path = "mnist_calibration.pth"
+    calibration_config = quantizer.export_model(model_path, calibration_path)
+    print("calibration_config: ", calibration_config)
+
+    # For now the quantization settings of ObserverQuantizer does not match the TensorRT,
```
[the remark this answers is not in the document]
chenbohua3: Because the `dtype` and `qscheme` of PyTorch **default** observer are different with that in TensorRT. For example, TensorRT uses per_tensor_symmetric with uint8 for activation, PyTorch uses per_tensor_affine with quint8 for activation.
When customization of `dtype` and `qscheme` is ready, we can support TensorRT.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -119,6 +122,217 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+                all_observers[layer_name]["output"] = default_histogram_observer()
+                setattr(module, "output_qmax", activation_q_max)
+                setattr(module, "output_qmin", activation_q_min)
+        self.all_observers = all_observers
+        self.bound_model.to(self.device)
+
+    def validate_config(self, model, config_list):
+        schema = CompressorSchema([{
+            Optional('quant_types'): Schema([lambda x: x in ['weight', 'output', 'input']]),
+            Optional('quant_bits'): Or(And(int, lambda n: n == 8), Schema({
```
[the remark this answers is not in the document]
chenbohua3: I think that there would not be major obstacles. The reason why we only support int8 right now is that PyTorch quantization observers only support 8 bit quantization (see [here](https://github.com/pytorch/pytorch/blob/8a2c7d902fb159bb57a0cf46f3a5ba74095cc34a/torch/quantization/observer.py#L152) ). To support them, we should extend/customize the observers to support all bit type.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -119,6 +122,217 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+            observer(*tensor)
+        else:
+            observer(tensor.cpu())
+
+    def calculate_qparams(self, name, quant_type):
+        observer = self.all_observers[name][quant_type]
+        scale, zero_point = observer.calculate_qparams()
+        return scale, zero_point
+
+    def quantize(self, x, scale, zero_point, qmin, qmax):
```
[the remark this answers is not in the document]
chenbohua3: done

examples/model_compress/quantization/observer_quantizer.py
```diff
@@ -0,0 +1,155 @@
...
+
+def calibration(model, device, test_loader):
+    model.eval()
+    with torch.no_grad():
+        for data, target in test_loader:
+            data, target = data.to(device), target.to(device)
+            model(data)
+
+
+def test_trt(engine, test_loader):
```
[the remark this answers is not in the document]
chenbohua3: Have deleted it

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -120,6 +123,222 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+    Note that the observer type, tensor dtype and quantization qscheme are hard coded for now. Their customization
+    are under development and will be ready soon.
+    """
+
+    def __init__(self, model, config_list, optimizer=None):
+        super().__init__(model, config_list, optimizer)
+        # NOTE: this quantizer is experimental for now. The dtype and qscheme of quantization
+        # is hard-coded.
+        # TODO:
+        # 1. support dtype and qscheme customization through config_list. Current settings:
```
QuanluZhang, line 141: if dtype and qscheme can be applied on each layer separately, then it is better to support them in config_list. otherwise, it is better to support them as quantizer's initialization argument

linbinskn: Agree, and I think it should be supported in `config_list` since weight and activation may use different settings and even different layers can use different settings.

linbinskn:
Post-training quantization is a critical feature in model compression and I think this pr is a good start for nni to support it. I have some thought related to this pr and current NNI quantization design:

- Both bit type choice and observers should be supported in `config_list` in layer level. The reason is:

  1. For post-training quantization, in addition to 8bit, the other bit types like 4-bit and 1bit are also supported in hardware such as NVIDIA GPU(> turing architecture).  And more bit type would be supported in future hardwares which enable more quantization data types.
  2. For deployment, different backend frameworks may use different observers. We have to offer general  support for them by setting `config_list`. For research, researchers may want to support different observers intro layer(activation/weight) and inter layer to get best performance.
  
- Current design of quantization in NNI is not good enough and it is important for us to define post-training quantization and quantization aware training(not only QAT) clearly in model compression and design in NNI. For instance, I think one potential abstraction is post-training is a one-shot quantization algorithm, and quantization aware training algorithms(not only QAT) are different finetuning methods above it. Should we separate them as different quantizers? Or they are different stages in the whole quantization pipeline(post-training quantization is necessary and quantization aware training is optional) and they can be applied sequentially.

I think we don't need to modify a lot in this pr. We can discuss them and make an appropriate design and customize corresponded API gradually. I can also contribute to this part to make it better.

QuanluZhang:
> Post-training quantization is a critical feature in model compression and I think this pr is a good start for nni to support it. I have some thoughts related to this pr and current NNI quantization design:
> 
> * Both bit type choice and observers should be supported in `config_list` in layer level. The reason is:
>   
>   1. For post-training quantization, in addition to 8bit, the other bit types like 4-bit and 1bit are also supported in hardware such as NVIDIA GPU(>= turing architecture).  And more bit type would be supported in future hardwares which enable more quantization data types.
>   2. For deployment, different backend frameworks may use different observers. We have to offer general  support for them by setting `config_list`. For research, researchers may want to support different observers intro layer(activation/weight) and inter layer to get best performance.
> * Current design of quantization in NNI is not good enough and it is important for us to define post-training quantization and quantization aware training(not only QAT) clearly in model compression and design in NNI. For instance, I think one potential abstraction is post-training is a one-shot quantization algorithm, and quantization aware training algorithms(not only QAT) are different finetuning methods above it. Should we separate them as different quantizers? Or they are different stages in the whole quantization pipeline(post-training quantization is necessary and quantization aware training is optional) and they can be applied sequentially.
> 
> I think we don't need to modify a lot in this pr. We can discuss them and make an appropriate design and customize corresponded API gradually. I am very glad to attend contributions of this part to make it better.

thanks for the discussion. in the current stage, it is not good to put bit types in `config_list`, because users can easily specify different bit types for different layers. If the supported quantization algorithm does not support different bit types on different layers, it would be error-prone, not user friendly.

about the refactor of quantization in NNI. we can think about how to adapt quantizer into similar modularized framework as NNI pruners. observer in quantization is very similar to "metric calculator" in our refined NNI pruning framework. on the other hand, we should survey all the quantization aware training to figure out whether it can be seen as a type of fine-tuning.

J-shang:
> > Post-training quantization is a critical feature in model compression and I think this pr is a good start for nni to support it. I have some thoughts related to this pr and current NNI quantization design:
> > 
> > * Both bit type choice and observers should be supported in `config_list` in layer level. The reason is:
> >   
> >   1. For post-training quantization, in addition to 8bit, the other bit types like 4-bit and 1bit are also supported in hardware such as NVIDIA GPU(>= turing architecture).  And more bit type would be supported in future hardwares which enable more quantization data types.
> >   2. For deployment, different backend frameworks may use different observers. We have to offer general  support for them by setting `config_list`. For research, researchers may want to support different observers intro layer(activation/weight) and inter layer to get best performance.
> > * Current design of quantization in NNI is not good enough and it is important for us to define post-training quantization and quantization aware training(not only QAT) clearly in model compression and design in NNI. For instance, I think one potential abstraction is post-training is a one-shot quantization algorithm, and quantization aware training algorithms(not only QAT) are different finetuning methods above it. Should we separate them as different quantizers? Or they are different stages in the whole quantization pipeline(post-training quantization is necessary and quantization aware training is optional) and they can be applied sequentially.
> > 
> > I think we don't need to modify a lot in this pr. We can discuss them and make an appropriate design and customize corresponded API gradually. I am very glad to attend contributions of this part to make it better.
> 
> thanks for the discussion. in the current stage, it is not good to put bit types in `config_list`, because users can easily specify different bit types for different layers. If the supported quantization algorithm does not support different bit types on different layers, it would be error-prone, not user friendly.
> 
> about the refactor of quantization in NNI. we can think about how to adapt quantizer into similar modularized framework as NNI pruners. observer in quantization is very similar to "metric calculator" in our refined NNI pruning framework. on the other hand, we should survey all the quantization aware training to figure out whether it can be seen as a type of fine-tuning.

agree, maybe we need a meeting to discuss these, and rethinking what should in `compressor` and what in `config_list`. Are all post-training quantization algorithms based on observers?

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -120,6 +123,222 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
         return grad_output
 
 
+class ObserverQuantizer(Quantizer):
```
J-shang, line 126: please add a ut for this quantizer

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -120,6 +123,222 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+        process will be simulated, which is used to test the accuracy of the quantization.
+        """
+        modules_to_compress = self.get_modules_to_compress()
+        for layer, config in modules_to_compress:
+            module = layer.module
+            if "weight" in config.get("quant_types", []):
+                scale, zero_point = self.calculate_qparams(layer.name, 'weight')
+                module.register_buffer('weight_scale', scale.to(self.device))
+                module.register_buffer('weight_zero_point', zero_point.to(self.device))
+                # todo: recover old_weight to weight, because the compressed
```
QuanluZhang, line 264: could you explain more about the case that should recover old_weight to weight?

QuanluZhang (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: Because after being wrapped by `QuantizerModuleWrapper`, two things may happen to the weight of each layer:
1. The type may changed from `torch.nn.Parameter` to `torch.Tensor`
2. The weight of BN may have been folded.

Basically, we need to ensure that the structure/parameter types of the model are consistent before and after PTQ. In theory, we can use the original model to perform downstream tasks (like qat or deployment), since the current ptq will not change the weight of the model.  But users may also use the model exported by the ptq for the downstream tasks, so I think it is better to recover them.

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: `dtype` and `qscheme` can be applied on each layer separately. Also we can customize `validate_config` function to control the rules for each quantizer

examples/model_compress/quantization/observer_quantizer.py
```diff
@@ -0,0 +1,117 @@
...
+    test_loss /= len(test_loader.dataset)
+
+    print('Loss: {}  Accuracy: {}%)\n'.format(
+        test_loss, 100 * correct / len(test_loader.dataset)))
+
+
+def calibration(model, device, test_loader):
+    model.eval()
+    with torch.no_grad():
+        for data, target in test_loader:
```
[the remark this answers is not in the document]
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: I have added some code logics in `export_model` for recovering weight

chenbohua3:
Is it necessary to add documentation？I think it’s better to leave this until the dtype/quant scheme customization is ready. Otherwise there will be many NOTEs or TODOs in the doc:)

J-shang (approved):

linbinskn (approved):

QuanluZhang:
> Is it necessary to add documentation？I think it’s better to leave this until the dtype/quant scheme customization is ready. Otherwise there will be many NOTEs or TODOs in the doc:)

agree

nni/algorithms/compression/pytorch/quantization/quantizers.py
QuanluZhang: 1) suppose quantizer's export should export the model after quantization. it is not proper to export original model.

2) I notice the inconsistency among the supported quantizers. some quantizers deal with fold bn, some others not. some quantizers recover weight, some others not. could you explain the current status a little bit more?

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -120,6 +123,233 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+        if self.compressed:
+            new_weight = self._quantize(old_weight,
+                                       module.weight_scale,
+                                       module.weight_zero_point,
+                                       module.weight_qmin,
+                                       module.weight_qmax)
+        else:
+            self.record(wrapper, 'weight', old_weight)
+            new_weight = old_weight
+        return new_weight
```
J-shang, line 236: seems we never use this `new_weight`? because there isn't something like `module.weight = new_weight`

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -120,6 +123,233 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+        calibration_config = {}
+
+        for name, module in self.bound_model.named_modules():
+            if hasattr(module, 'weight_scale') or hasattr(module, 'input_scale') or hasattr(module, 'output_scale'):
+                calibration_config[name] = {}
+            if hasattr(module, 'weight_scale'):
+                calibration_config[name]['weight_bit'] = 8
+                val = float(module.weight_scale * module.weight_qmax)
+                calibration_config[name]['tracked_min_weight'] = val
+                calibration_config[name]['tracked_max_weight'] = -val
```
J-shang, line 308: is `calibration_config[name]['tracked_min_weight'] = -val` and `calibration_config[name]['tracked_max_weight'] = val`?

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: Yes, `new_weight` will not be used. I just return it like what QAT quantizer does. It also returns unused new_weight. Do I need to delete it?

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: Yes, have corrected it

nni/algorithms/compression/pytorch/quantization/quantizers.py
J-shang: I think for the quantization simulation after compress(), we need to add `module.weight = new_weight`.
```
        if self.compressed:
            new_weight = self._quantize(old_weight,
                                       module.weight_scale,
                                       module.weight_zero_point,
                                       module.weight_qmin,
                                       module.weight_qmax)
            module.weight = new_weight
```
Will we get the correct inference result in this way?

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: As discussed offline, I replace the origin weight with the pseudo-quantized weight.

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: You are right, have corrected it

chenbohua3:
I have removed the codes for replacing weight with quantized one in `compress` function. @QuanluZhang @J-shang

test/ut/sdk/test_compressor_torch.py
```diff
@@ -263,6 +264,39 @@ def test_torch_taylorFOweight_pruner_global_sort(self):
...
+        input = torch.randn(1, 1, 28, 28)
+        model(input)
+        quantizer.compress()
+        model_path = "test_model.pth"
+        calibration_path = "test_calibration.pth"
+        calibration_config = quantizer.export_model(model_path, calibration_path)
+        new_parameters = dict(model.named_parameters())
+        self.assertTrue(all(torch.equal(v, new_parameters[k]) for k, v in origin_parameters.items()))
+        self.assertTrue(calibration_config is not None)
+        self.assertTrue(len(calibration_config) == 4)
```
QuanluZhang, line 298: please update test accordingly

chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -120,6 +123,241 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
+        #  activation observer : per_tensor_affine, quint8, reduce_range=True
+        # 2. add more kinds of observers, such as Kullback-Leibler divergence.
+        # 3. add batch normalization folding
+        assert not model.training, "Currently the observer quantizer only works in evaluation mode."
+        self.quant_grad = QuantForward()
+        self.device = next(model.parameters()).device
+        modules_to_compress = self.get_modules_to_compress()
+        all_observers = defaultdict(dict)
+        weight_q_min, weight_q_max = -127, 127
+        activation_q_min, activation_q_max = 0, 127  # reduce_range is set to True
```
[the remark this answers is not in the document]
chenbohua3: By default, activation observer's `reduce_range` is set to True. This means that the range of the quantized data type is reduced by 1 bit. This is sometimes required to avoid instruction overflow. 
However, there does exist a mismatch between here and that in `export_model`, I have corrected it.

PR #3917, opened 2021-07-08
Compression v2 Stage 1

J-shang:
Pruner:
- [x] level
- [x] l1filter
- [x] l2filter
- [x] FPGM
- [x] APoZ
- [x] MeanRank
- [x] Taylor
- [x] Slim

SparsityAllocator:
- [x] Normal
- [x] GroupBaseGlobalSort
- [x] DependencyAware

others:
- [ ] doc
- [x] export model
- [x] validate config

PruningScheduler will update in next pr.


nni/algorithms/compression_v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,282 @@
+# Copyright (c) Microsoft Corporation.
```
ultmaster, line 1: I suggest `nni.algorithms.compression.v2`.

J-shang: good suggestion, I will move them.

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+class LayerInfo:
+    def __init__(self, name: str, module: Module):
+        self.module = module
+        self.name = name
+        self.type = type(module).__name__
+
+
+def _setattr(model: Module, name: str, module: Module):
```
zheng-ningxin, line 26: looks like duplicated with `get_module_by_name` , please use this function instead which takes some corner case into account.

https://github.com/microsoft/nni/blob/d8127e02841f7f2afb923c22c91b254155544db8/nni/compression/pytorch/utils/utils.py#L9

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,282 @@
+# Copyright (c) Microsoft Corporation.
```
zheng-ningxin, line 1: Why do have a father class named `base`? what is `base` for?

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,282 @@
...
+            self.modules_wrapper[layer.name] = wrapper
+
+        self._wrap_model()
+
+    def _detect_modules_to_compress(self) -> List[Tuple[LayerInfo, Dict]]:
+        """
+        Detect all modules should be compressed, and save the result in `self._modules_to_compress`.
+        The model will be instrumented and user should never edit it after calling this method.
+        """
+        if self._modules_to_compress is None:
```
zheng-ningxin, line 90: suggest put `self._modules_to_compress=None` into ```__init__```, esle there will be a potential calling order dependency between `_detect_modules_to_compress` and `reset` which may lead to error.

nni/algorithms/compression/v2/pytorch/base/compressor.py
J-shang: The `base` is for some base classes like `Compressor` `Pruner` `Quantizer` and some base classes used by `Pruner` like `DataCollector`...

nni/algorithms/compression/v2/pytorch/base/compressor.py
J-shang: thanks, it is better, I have modified it.

nni/algorithms/compression/v2/pytorch/base/compressor.py
J-shang: I think `self._modules_to_compress=None` in `reset` is enough, because `reset` is a part of logic in `__init__`. And only after set `self.bound_model`, call `_detect_modules_to_compress` is meaningful?

zheng-ningxin: Got it.

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,285 @@
...
+        value
+            Value of the variable.
+        """
+        for wrapper in self.get_modules_wrapper():
+            if isinstance(value, torch.Tensor):
+                wrapper.register_buffer(name, value.clone())
+            else:
+                setattr(wrapper, name, value)
+
+    def generate_graph(self, dummy_input: Tensor) -> TorchModuleGraph:
```
xiaowu0162, line 187: Is it possible that the dummy_input is not a `Tensor`? Like we discussed this afternoon, the input to transformer encoders is often a two-element tuple.

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,285 @@
...
+        -------
+        TorchModuleGraph
+            A `TorchModuleGraph` instance.
+        """
+        self._unwrap_model()
+        graph = TorchModuleGraph(model=self.bound_model, dummy_input=dummy_input)
+        self._wrap_model()
+        return graph
+
+    def generate_module_groups(self) -> Dict[int, List[str]]:
```
xiaowu0162, line 206: Is this function only used by pruner, or also might be used by other successor classes like quantizer?

nni/algorithms/compression/v2/pytorch/base/pruner_tools.py
```diff
@@ -0,0 +1,360 @@
...
+                def trainer(model: Module, optimizer: Optimizer, criterion: Callable[[Tensor, Tensor], Tensor]):
+                    model.train()
+                    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+                    for batch_idx, (data, target) in enumerate(train_loader):
+                        data, target = data.to(device), target.to(device)
+                        optimizer.zero_grad()
+                        output = model(data)
+                        loss = criterion(output, target)
+                        loss.backward()
+                        optimizer.step()
```
xiaowu0162, line 100: Should we recommend the user to write the trainer so that parameter update is performed during data collection?

nni/algorithms/compression/v2/pytorch/common/data_collector.py
```diff
@@ -0,0 +1,58 @@
...
+from torch import Tensor
+
+from nni.algorithms.compression.v2.pytorch.base.pruner_tools import DataCollector, TrainerBasedDataCollector
+
+_logger = logging.getLogger(__name__)
+
+__all__ = ['WeightDataCollector', 'WeightTrainerBasedDataCollector', 'SingleHookTrainerBasedDataCollector']
+
+
+class WeightDataCollector(DataCollector):
```
xiaowu0162, line 16: Maybe `WeightDataCollector` and `TrainerBasedDataCollector` can be put into the same file since they belong to the same level?

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -0,0 +1,506 @@
...
+        schema.validate(config_list)
+
+    def _collector(self, buffer: List, weight_tensor: Tensor) -> Callable[[Module, Tensor, Tensor], None]:
+        def collect_taylor(grad: Tensor):
+            if len(buffer) < self.training_batches:
+                buffer.append(self._calculate_taylor_expansion(weight_tensor, grad))
+        return collect_taylor
+
+    def _calculate_taylor_expansion(self, weight_tensor: Tensor, grad: Tensor) -> Tensor:
+        return (weight_tensor * grad).data.pow(2)
```
xiaowu0162, line 493: Do we need `detach()` here?

nni/algorithms/compression/v2/pytorch/base/compressor.py
J-shang: yes, in fact, the input type is determined by the model, change it to Any.

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,282 @@
...
+        OrderedDict[str, Module]
+            An ordered dict, key is the name of the module, value is the wrapper of the module.
+        """
+        return self.modules_wrapper
+
+    def _wrap_model(self):
+        """
+        Wrap all modules that needed to be compressed.
+        """
+        for _, wrapper in reversed(self.get_modules_wrapper().items()):
```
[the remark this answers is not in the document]
J-shang: It's not required but the code is clearer after adding judgment, added it.

nni/algorithms/compression/v2/pytorch/base/compressor.py
J-shang: I'm not sure, I place this function here because it needs to traverse the entire model.

nni/algorithms/compression/v2/pytorch/base/pruner_tools.py
J-shang: It is a good suggestion, add it

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: what will happen if we don't `detach()` here

nni/algorithms/compression/v2/pytorch/base/pruner_tools.py
xiaowu0162: I mean do we want the user to always call optimizer.step() even when this trainer is used for calculating scores?

J-shang: got it, add a comment for this, this trainer is always used for data collection.

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
xiaowu0162: I think if some operations temper with the data the weight might be affected, but it seems that this chance is very small. In our previous implementation, we also used `data()` e.g., in the `get_channel_sum` function in the masker to get the data for score calculation. However, the masker immediately `detach()` the resulting mask, and does not expose the intermediate data. Here since we modularize each step and pass the data around to other components, it might be safer to `detach()`, but I'm also not sure.

J-shang: yes, as we discussed, detach() is safer, add it

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -0,0 +1,197 @@
...
+        super().reset(model=model, config_list=config_list)
+        self.data_collector: Optional[DataCollector] = None
+        self.metrics_calculator: Optional[MetricsCalculator] = None
+        self.sparsity_allocator: Optional[SparsityAllocator] = None
+        self._reset_tools()
+
+    def _reset_tools(self):
+        """
+        This function is used to reset `self.data_collector`, `self.metrics_calculator` and `self.sparsity_allocator`.
+        The subclass need implement this function to complete the pruning process.
```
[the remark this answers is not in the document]
J-shang: thanks, fix it

nni/algorithms/compression/v2/pytorch/base/pruner_tools.py
```diff
@@ -0,0 +1,390 @@
...
+        Dict[str, Tensor]
+            The key is the layer_name, value is the metric.
+            Note that the metric has the same size with the data size on `dim`.
+        """
+        raise NotImplementedError()
+
+
+class SparsityAllocator:
+    """
+    An abstract class for allocate mask based on metrics.
```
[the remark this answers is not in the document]
J-shang: fix it

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,286 @@
...
+        ret = None
+        for config in self.config_list:
+            config = config.copy()
+            # expand config if key `default` is in config['op_types']
+            if 'op_types' in config and 'default' in config['op_types']:
+                expanded_op_types = []
+                for op_type in config['op_types']:
+                    if op_type == 'default':
+                        expanded_op_types.extend(weighted_modules)
+                    else:
```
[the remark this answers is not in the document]
J-shang: `default` is for `level` pruner, and other `module` which has `weight` also can be pruned by it.

ultmaster (approved):

linbinskn:
Please add example and Unit Test for this pr.

nni/algorithms/compression/v2/pytorch/common/data_collector.py
```diff
@@ -0,0 +1,58 @@
...
+        for _ in range(self.training_epochs):
+            self.trainer(self.compressor.bound_model, self.optimizer, self.criterion)
+
+        data = {}
+        for _, wrapper in self.compressor.get_modules_wrapper().items():
+            data[wrapper.name] = wrapper.module.weight.data.clone().detach()
+        return data
+
+
+class SingleHookTrainerBasedDataCollector(TrainerBasedDataCollector):
```
linbinskn, line 46: Single can only has one hook to collect data. So it can not satisfy the pruner which calculates metric based on both activation and gradient? Would we support that?

nni/algorithms/compression/v2/pytorch/common/metrics_calculator.py
```diff
@@ -0,0 +1,179 @@
...
+from torch import Tensor
+
+from nni.algorithms.compression.v2.pytorch.base.pruner_tools import MetricsCalculator
+
+__all__ = ['NormMetricsCalculator', 'MultiDataNormMetricsCalculator', 'DistMetricsCalculator',
+           'APoZRankMetricsCalculator', 'MeanRankMetricsCalculator']
+
+
+class NormMetricsCalculator(MetricsCalculator):
+    """
```
linbinskn, line 16: Recommend adding annotation about pruner types this metrics_calculator serves.

linbinskn:
Good code style. Have learnt a lot from it.  ^_^

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -0,0 +1,197 @@
...
+from .compressor import Compressor, LayerInfo
+from .pruner_tools import DataCollector, MetricsCalculator, SparsityAllocator
+
+_logger = logging.getLogger(__name__)
+
+__all__ = ['Pruner']
+
+
+class PrunerModuleWrapper(Module):
+    def __init__(self, module: Module, module_name: str, module_type: str, config: Dict, pruner: Compressor):
```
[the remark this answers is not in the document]
J-shang: yes, it is a little strange... but `pruner: Pruner` will cause a circular reference.

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -0,0 +1,197 @@
...
+            Device of the model, used to place the dummy input tensor for exporting onnx file.
+            The tensor is placed on cpu if ```device``` is None.
+        """
+        assert model_path is not None, 'model_path must be specified'
+        mask_dict = {}
+        self._unwrap_model()  # used for generating correct state_dict name without wrapper state
+
+        for name, wrapper in self.get_modules_wrapper().items():
+            weight_mask = wrapper.weight_mask
+            bias_mask = wrapper.bias_mask
```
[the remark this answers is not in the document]
J-shang: `wrapper.bias_mask == None` if module has no bias.

nni/algorithms/compression/v2/pytorch/common/data_collector.py
J-shang: Yes, we can support this by adding more logic in `collect` function. due to we don't have this scenario right now, I am a little lazy to implement it...

nni/algorithms/compression/v2/pytorch/base/pruner_tools.py
```diff
@@ -0,0 +1,390 @@
...
+from torch.optim import Optimizer
+
+from .compressor import Compressor, LayerInfo
+
+_logger = logging.getLogger(__name__)
+
+__all__ = ['DataCollector', 'TrainerBasedDataCollector', 'HookCollectorInfo', 'MetricsCalculator', 'SparsityAllocator']
+
+
+class DataCollector:
```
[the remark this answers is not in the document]
J-shang: This is because the pruner base class uses these three tools. And puts the base class in `common` is also reasonable, we can discuss this in the next meeting.

nni/algorithms/compression/v2/pytorch/base/compressor.py
```diff
@@ -0,0 +1,286 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import collections
+import logging
+from typing import List, Dict, Optional, OrderedDict, Tuple, Any
+
+import torch
+from torch.nn import Module
+from torch.tensor import Tensor
```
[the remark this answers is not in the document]
J-shang: thanks, remove it.

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -0,0 +1,197 @@
...
+        module_type
+            The type of the module to compress.
+        pruner
+            The pruner used to calculate mask.
+        """
+        super().__init__()
+        # origin layer information
+        self.module = module
+        self.name = module_name
+        self.type = module_type
```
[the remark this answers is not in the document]
J-shang: yes, seems we can use `type(wrapper.module).__name__` , remove it.

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -0,0 +1,197 @@
...
+            The configuration for generating the mask.
+        """
+        _logger.debug("Module detected to compress : %s.", layer.name)
+        wrapper = PrunerModuleWrapper(layer.module, layer.name, layer.type, config, self)
+        assert hasattr(layer.module, 'weight'), "module %s does not have 'weight' attribute" % layer.name
+        # move newly registered buffers to the same device of weight
+        wrapper.to(layer.module.weight.device)
+        return wrapper
+
+    def load_masks(self, masks: Dict[str, Dict[str, Tensor]]):
```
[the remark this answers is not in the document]
J-shang: good suggestion, add it.

nni/algorithms/compression/v2/pytorch/common/metrics_calculator.py
J-shang: add it.

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -0,0 +1,197 @@
...
+        # origin layer information
+        self.module = module
+        self.name = module_name
+        self.type = module_type
+        # config and pruner
+        self.config = config
+        self.pruner = pruner
+
+        # register buffer for mask
+        self.register_buffer("weight_mask", torch.ones(self.module.weight.shape))
```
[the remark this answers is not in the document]
J-shang: In this scenario, what will we do to the matmul module?

nni/algorithms/compression/v2/pytorch/base/pruner.py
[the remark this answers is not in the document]
linbinskn: And during it is instantiated, in fact there is still a circular reference...  So only type declaration will raise error here while nothing happened during runtime?

nni/algorithms/compression/v2/pytorch/base/pruner.py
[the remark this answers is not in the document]
linbinskn: We have to mask its output like below picture.
image

nni/algorithms/compression/v2/pytorch/base/pruner.py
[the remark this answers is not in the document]
J-shang: Where is the circular reference if `pruner: Compressor`, I can't find it...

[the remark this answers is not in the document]
linbinskn: During runtime, this argument will be still be `Pruner`?

J-shang:
> Please add example and Unit Test for this pr.

add a naive example, ut will update in other pr.

QuanluZhang (approved):

linbinskn (approved):

zheng-ningxin (approved):

PR #3919, opened 2021-07-08
Support kubeflow reuse mode

SparkSnail:


nni/experiment/config/kubeflow.py
```diff
@@ -8,21 +8,21 @@
...
     azure_account: Optional[str] = None
     azure_share: Optional[str] = None
-    key_vault: Optional[str] = None
-    key_vault_secret: Optional[str] = None
+    key_vault_name: Optional[str] = None
+    key_vault_key: Optional[str] = None
 
 @dataclass(init=False)
-class KubeflowNfsConfig(_KubeflowStorageConfig):
+class KubeflowNfsConfig(KubeflowStorageConfig):
```
liuzhe-lz, line 25: Shouldn't reusable use shared storage service?

nni/experiment/config/kubeflow.py
```diff
@@ -32,37 +32,41 @@ class KubeflowAzureStorageConfig(ConfigBase):
...
+    master: Optional[KubeflowRoleConfig] = None
+    reuse_mode: Optional[bool] = True #set reuse mode as true for v2 config
 
     def __init__(self, **kwargs):
         kwargs = util.case_insensitive(kwargs)
-        kwargs['storage'] = util.load_config(_KubeflowStorageConfig, kwargs.get('storage'))
+        kwargs['storage'] = util.load_config(KubeflowStorageConfig, kwargs.get('storage'))
         kwargs['worker'] = util.load_config(KubeflowRoleConfig, kwargs.get('worker'))
-        kwargs['parameterserver'] = util.load_config(KubeflowRoleConfig, kwargs.get('parameterserver'))
+        kwargs['ps'] = util.load_config(KubeflowRoleConfig, kwargs.get('ps'))
```
J-shang, line 65: `ps` is `parameterserver`? why we don't have `master` before?

ultmaster (approved):

nni/experiment/config/kubeflow.py
liuzhe-lz: I once merged ps and master. They are both parameter server. No difference at all, just different names for pytorch and tensorflow.

J-shang (approved):

nni/experiment/config/kubeflow.py
SparkSnail: Only openPAI environment service support storage service.

liuzhe-lz (approved):

PR #3921, opened 2021-07-09
[Quantization Speedup] Add integration-test-trt.yml for TensorRT integration test

linbinskn:
This PR is a complement to PR #3862 .


J-shang (approved):
Can remove `Install customized tuner` if we don't use these tuners.

ultmaster (approved):

linbinskn:
> Can remove `Install customized tuner` if we don't use these tuners.

@J-shang Have updated.

PR #3933, opened 2021-07-13
"Fix" NNI manager UT

liuzhe-lz:
It moved all test cases to one folder to fix the bug that test codes are packaged into wheel.

Disabled test cases are marked with `FIXME`.


ultmaster (approved):

J-shang (approved):

ts/nni_manager/test/core/nnimanager.test.ts
```diff
@@ -134,7 +134,8 @@ describe('Unit test for nnimanager', function () {
     })
 
     after(async () => {
-        await setTimeout(() => {nniManager.stopExperiment()},15000);
+        // FIXME
```
SparkSnail, line 137: If do not stop experiment, will it occupy the port?

SparkSnail (approved):

ultmaster (approved):

PR #3934, opened 2021-07-13
Fix a bug in IP detection

liuzhe-lz:
The connection might not ready when it tries to fetch IP. Make the function async to wait the OS initializing connection.

Reported by Zhenhua.

### DIFF
--- a/ts/nni_manager/common/utils.ts
+++ b/ts/nni_manager/common/utils.ts
@@ -223,7 +223,7 @@ let cachedIpv4Address: string | null = null;
 /**
  * Get IPv4 address of current machine.
  */
-function getIPV4Address(): string {
+async function getIPV4Address(): Promise<string> {
     if (cachedIpv4Address !== null) {
         return cachedIpv4Address;
     }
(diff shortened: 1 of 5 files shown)


ts/nni_manager/common/utils.ts
```diff
@@ -232,12 +232,20 @@ function getIPV4Address(): string {
     // since udp is connectionless, this does not send actual packets.
     const socket = dgram.createSocket('udp4');
     socket.connect(1, '192.0.2.0');
-    cachedIpv4Address = socket.address().address;
+    for (let i = 0; i < 10; i++) {  // wait the system to initialize "connection"
+        await yield_();
```
J-shang, line 236: a little confused about why we `yield_()` ten times, what happens when initializing "connection"

cruiseliu: `socket.connect()` returns instantly, but `socket.address()` is not available until OS finish underlying proceeding.
There is no guarantee on how much time it takes, though all experiments shows 1 yield is enough.

J-shang: got it

J-shang (approved):

SparkSnail (approved):

PR #3952, opened 2021-07-19
Fix bug in IP detection

liuzhe-lz:
This is a follow up of #3934

The yield method was not strong enough. In certain cases it will not trigger the OS scheduler. So change to setTimeout(1ms) instead.


J-shang (approved):

hzhua:
I have tested it on darts on remote training service. It works well without EBADF error.

hzhua (approved):

PR #3968, opened 2021-07-21
[Model Compression] Expand export_model arguments: dummy input and onnx opset_version

### LINKED ISSUE #3964: Use higher opset_version when exporting ONNX models

1. A `dummy_input` argument for the `export_model` in Compressor. 
2. Allow using `opset_version=10` or higher when calling `torch.onnx.export` inside the `export_model` in Compressor. 

**Why is this needed**:

1. Some models need a tuple instead of a tensor as the model input. For example, transformer models require both input and input mask as the model input. In these cases, it seems that the original `input_shape` argument is not enough to specify the dummy input. 
2. Some models that use dynamic indexing (e.g., transformers) does not work with the default opset_version (9) of `torch.onnx.export`. The reported error looks like `RuntimeError: Unsupported: ONNX export of Slice with dynamic inputs. DynamicSlice is a deprecated experimental op.` The error disappears when `opset_version=10` is used.

It does not work for exporting transformers to ONNX. 

1. `export_model` function in `Compressor`
2. Model Speedup may be affected 

**Brief description of your proposal if any**:

Allow a `dummy_input` argument and possibly an `opset_version` argument.

xiaowu0162:
Fix #3964.

### DIFF
--- a/nni/compression/pytorch/compressor.py
+++ b/nni/compression/pytorch/compressor.py
@@ -432,7 +432,7 @@ class Pruner(Compressor):
                 'input_shape or dummy_input must be specified to export onnx model'
             # create dummy_input using input_shape if input_shape is not passed
             if dummy_input is None:
-                _logger.warning("""The argument input_shape and device will be removed in the next release. 
+                _logger.warning("""The argument input_shape and device will be removed in the next release.
                                    Please create a dummy input and pass it to dummy_input instead.""")
                 if device is None:
                     device = torch.device('cpu')
--- a/nni/compression/pytorch/compressor.py
+++ b/nni/compression/pytorch/compressor.py
@@ -374,7 +374,8 @@ class Pruner(Compressor):
         wrapper.to(layer.module.weight.device)
         return wrapper
 
-    def export_model(self, model_path, mask_path=None, onnx_path=None, input_shape=None, device=None):
+    def export_model(self, model_path, mask_path=None, onnx_path=None, input_shape=None, dummy_input=None,
+                     opset_version=None, device=None):
         """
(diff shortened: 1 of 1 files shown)


xiaowu0162:
Ready for review @J-shang @QuanluZhang

nni/compression/pytorch/compressor.py
```diff
@@ -411,17 +418,29 @@ def export_model(self, model_path, mask_path=None, onnx_path=None, input_shape=N
...
-            _logger.info('Model in onnx with input shape %s saved to %s', input_data.shape, onnx_path)
+            assert input_shape is not None or dummy_input is not None,\
+                'input_shape or dummy_input must be specified to export onnx model'
+            # create dummy_input using input_shape if input_shape is not passed
+            if dummy_input is None:
+                if device is None:
+                    device = torch.device('cpu')
+                input_data = torch.Tensor(*input_shape).to(device)
+            else:
+                input_data = dummy_input
```
J-shang, line 435: what if user both set `dummy_input` and `device`

xiaowu0162: I think device should be ignored in that case. Have updated the docstring.

J-shang: recommand we also `input_data = dummy_input.to(device)`, or this may confuse user, if user also set `device` but we ignore it.

xiaowu0162: But I think this operation may fail when e.g., `dummy_input` is a tuple

J-shang (approved):

ultmaster (approved):

PR #3972, opened 2021-07-22
Enable `fixed_arch` on Retiarii

### LINKED ISSUE #3969: How to print and save the best model when using one-shot nas
**Environment**:
- NNI version: 2.3
- PyTorch/TensorFlow version: PyTorch

I use DartsTrainer and build a model as follows:
``` python
class Net(nn.Module):

    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.LayerChoice([
            nn.Conv2d(3, 6, 3, padding=1),
            nn.Conv2d(3, 6, 5, padding=2),
            nn.Conv2d(3, 6, 7, padding=3),
        ], label='conv1')
        self.pool = nn.MaxPool2d(2, 2)
        self.conv2 = nn.LayerChoice([
            nn.Conv2d(6, 16, 3, padding=1),
            nn.Conv2d(6, 16, 5, padding=2),
            nn.Conv2d(6, 16, 7, padding=3),
        ], label='conv2')
        self.conv3 = nn.Conv2d(16, 16, 1)

        self.skipconnect1 = nn.InputChoice(n_candidates=2, label='skipconnection')
        self.bn = nn.BatchNorm2d(16)

        self.gap = nn.AdaptiveAvgPool2d(4)
        self.fc1 = nn.Linear(16*4*4, nn.ValueChoice([64, 128, 256], label='fc1'))
        self.fc2 = nn.Linear(
            nn.ValueChoice([64, 128, 256], label='fc1'), 
            nn.ValueChoice([32, 64, 128], label='fc2')
        )
        self.fc3 = nn.Linear(nn.ValueChoice([32, 64, 128], label='fc2'), 10)

    def forward(self, x):
        bs = x.size(0)
        x = self.pool(F.relu(self.conv1(x)))
        x0 = F.relu(self.conv2(x))
        x1 = F.relu(self.conv3(x0))

        x1 = self.skipconnect1([x1, x0+x1])
        x = self.pool(self.bn(x1))

        x = self.gap(x).view(bs, -1)
        x = F.relu(self.fc1(x))
        x = F.relu(self.fc2(x))
        x = self.fc3(x)

        return x
```

The first problem is that I can't get all search parameters using ```trainer.export()```. It prints ```{'conv1': 1, 'conv2': 2, 'skipconnection': [1]}```, but I expect it prints like ```{'conv1': 1, 'conv2': 2, 'skipconnection': [1], 'fc1': 64, 'fc2': 32}```. I check the source code of [DartsTrainer](https://github.com/microsoft/nni/blob/master/nni/retiarii/oneshot/pytorch/darts.py). In line 123 and 124, it only replace layer choice and input choice, doesn't replace value choice. I think that is why I can't print 'fc1' and 'fc2'. Is it a bug? And how can I print the best value when using ```nn.ValueChoice```?

The second problem is that how can I save the best architecture. In NNI 2.0, trainer.export() has parameters 'file' to save architecture as json, and load using ```apply_fixed_architecture```. How can I do this in NNI 2.3?

Thinks.

ultmaster:
Fix #3969.

Added an API called `fixed_arch` to enable users to quickly use the exported architecture.

I haven't tested on other trainers because their retraining code looks over complex. I'll do that as a follow-up task.

### DIFF
--- a/nni/retiarii/oneshot/pytorch/darts.py
+++ b/nni/retiarii/oneshot/pytorch/darts.py
@@ -19,7 +19,7 @@ _logger = logging.getLogger(__name__)
 class DartsLayerChoice(nn.Module):
     def __init__(self, layer_choice):
         super(DartsLayerChoice, self).__init__()
-        self.name = layer_choice.key
+        self.name = layer_choice.label
         self.op_choices = nn.ModuleDict(OrderedDict([(name, layer_choice[name]) for name in layer_choice.names]))
         self.alpha = nn.Parameter(torch.randn(len(self.op_choices)) * 1e-3)
 
@@ -45,7 +45,7 @@ class DartsLayerChoice(nn.Module):
 class DartsInputChoice(nn.Module):
     def __init__(self, input_choice):
         super(DartsInputChoice, self).__init__()
-        self.name = input_choice.key
+        self.name = input_choice.label
         self.alpha = nn.Parameter(torch.randn(input_choice.n_candidates) * 1e-3)
         self.n_chosen = input_choice.n_chosen or 1
 
--- a/nni/retiarii/oneshot/pytorch/darts.py
+++ b/nni/retiarii/oneshot/pytorch/darts.py
@@ -3,6 +3,7 @@
 
 import copy
 import logging
+from collections import OrderedDict
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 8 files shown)


skeletondyh (approved):

nni/retiarii/oneshot/pytorch/darts.py
```diff
@@ -19,7 +20,7 @@ class DartsLayerChoice(nn.Module):
     def __init__(self, layer_choice):
         super(DartsLayerChoice, self).__init__()
         self.name = layer_choice.key
```
skeletondyh, line 22: should be layer_choice.label, otherwise there would be warning. DartsInputChoice should also be modified.

skeletondyh (changes requested):

johsnows (approved):

jonsnows (approved):

skeletondyh (approved):

QuanluZhang (approved):

J-shang (approved):

PR #3976, opened 2021-07-23
Bump Python dependencies

liuzhe-lz:


Dockerfile
```diff
@@ -11,84 +11,53 @@ ENV DEBIAN_FRONTEND=noninteractive
...
+RUN python3 -m pip --no-cache-dir install azureml azureml-sdk
 
 #
 # Install NNI
 #
 COPY dist/nni-${NNI_RELEASE}-py3-none-manylinux1_x86_64.whl .
 RUN python3 -m pip install nni-${NNI_RELEASE}-py3-none-manylinux1_x86_64.whl
 
-# 
+#
```
ultmaster, line 55: I think you no longer need this if you have upgraded pytorch and torchvision to the latest version.
But never mind. We can do this in future...

ultmaster (approved):

Dockerfile
liuzhe-lz: If we upgrade torch to latest version, then the CUDA version is too high for tensorflow.

PR #3990, opened 2021-07-29
[Model Compression Quantization] Unify variable name

linbinskn:
This pr targets unifying two variable names in quantization code. They are too messy for developer and code reader. The first is `output`.`activation` -> `output`. The second is `bits`, `bit` -> `bits`. This pr also fixes a bug in quantization speedup conveniently ( [bug](https://github.com/microsoft/nni/blob/master/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py#L227) ) and adds warning for not setting optimizer in quantizer.

- [x] unify two variable names
- [x] quantization speedup fug fix [bug](https://github.com/microsoft/nni/blob/master/nni/compression/pytorch/quantization_speedup/integrated_tensorrt.py#L227) 
- [x] add warning for not setting optimizer in quantizer

### DIFF
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -354,7 +354,7 @@ class QAT_Quantizer(Quantizer):
     http://openaccess.thecvf.com/content_cvpr_2018/papers/Jacob_Quantization_and_Training_CVPR_2018_paper.pdf
     """
 
-    def __init__(self, model, config_list, optimizer=None, dummy_input=None):
+    def __init__(self, model, config_list, optimizer, dummy_input=None):
         """
         Parameters
         ----------
@@ -379,6 +379,7 @@ class QAT_Quantizer(Quantizer):
                     given, the batch normalization folding would be disabled.
         """
 
+        assert isinstance(optimizer, torch.optim.Optimizer), "unrecognized optimizer type"
         super().__init__(model, config_list, optimizer, dummy_input)
         self.quant_grad = QATGrad.apply
         modules_to_compress = self.get_modules_to_compress()
@@ -642,7 +643,8 @@ class DoReFaQuantizer(Quantizer):
     (https://arxiv.org/abs/1606.06160)
     """
 
-    def __init__(self, model, config_list, optimizer=None):
+    def __init__(self, model, config_list, optimizer):
+        assert isinstance(optimizer, torch.optim.Optimizer), "unrecognized optimizer type"
         super().__init__(model, config_list, optimizer)
         device = next(model.parameters()).device
         modules_to_compress = self.get_modules_to_compress()
@@ -749,7 +751,8 @@ class BNNQuantizer(Quantizer):
     (https://arxiv.org/abs/1602.02830)
     """
 
-    def __init__(self, model, config_list, optimizer=None):
+    def __init__(self, model, config_list, optimizer):
+        assert isinstance(optimizer, torch.optim.Optimizer), "unrecognized optimizer type"
         super().__init__(model, config_list, optimizer)
         device = next(model.parameters()).device
         self.quant_grad = ClipGrad.apply
@@ -848,7 +851,7 @@ class LsqQuantizer(Quantizer):
        https://arxiv.org/pdf/1902.08153.pdf
     """
 
-    def __init__(self, model, config_list, optimizer=None):
+    def __init__(self, model, config_list, optimizer):
         """
         Parameters
         ----------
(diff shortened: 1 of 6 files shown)


linbinskn:
@J-shang @chenbohua3  @xiaowu0162 
Please help review this pr.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -161,12 +161,12 @@ def __init__(self, model, config_list, optimizer=None):
                 setattr(module, "weight_qmin", weight_q_min)
             if "input" in config.get("quant_types", []):
                 all_observers[layer_name]["input"] = default_histogram_observer()
-                setattr(module, "input_qmax", activation_q_max)
-                setattr(module, "input_qmin", activation_q_min)
+                setattr(module, "input_qmax", output_q_max)
```
chenbohua3, line 164: I think that we can unify `qmax` and `q_max`

chenbohua3:
Nit: I am curious why you use `bits` instead of `bit`. `bit` seems to be a more common choice.

xiaowu0162:
Saw people using `bit_length` or `bit_width` elsewhere, which is less ambiguous but much longer than `bits`

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Good point, have unified them to `qmax`.

linbinskn:
> Nit: I am curious why you use `bits` instead of `bit`. `bit` seems to be a more common choice.

In config list, we use `quant_bits`. Here to be aligned with it.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -306,29 +306,29 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
             if hasattr(module, 'weight_scale') or hasattr(module, 'input_scale') or hasattr(module, 'output_scale'):
                 calibration_config[name] = {}
             if hasattr(module, 'weight_scale'):
-                calibration_config[name]['weight_bit'] = 8
+                calibration_config[name]['weight_bits'] = 8
                 val = float(module.weight_scale * module.weight_qmax)
                 calibration_config[name]['tracked_max_weight'] = val
                 calibration_config[name]['tracked_min_weight'] = -val
                 calibration_config[name]['tracked_weight_qmin'] = -127
                 calibration_config[name]['tracked_weight_qmax'] = 127
```
chenbohua3, line 314: it's better to use `tracked_qmin_weight`, to be consistent with `tracked_qmin_output`

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -306,29 +306,29 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
...
             # refactor these magic numbers when customizations of dtype and qscheme are ready.
             if hasattr(module, 'input_scale'):
-                calibration_config[name]['input_bit'] = 8
+                calibration_config[name]['input_bits'] = 8
                 max_input = float(module.input_scale * (module.input_qmax - module.input_zero_point))
                 min_input = float(module.input_scale * (module.input_qmin - module.input_zero_point))
                 calibration_config[name]['tracked_min_input'] = min_input
                 calibration_config[name]['tracked_max_input'] = max_input
                 calibration_config[name]['tracked_input_qmin'] = 0
```
chenbohua3, line 322: `tracked_qmin_input`

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Ok, done.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Have unified.

chenbohua3:
LGTM

nni/compression/pytorch/compressor.py
```diff
@@ -615,6 +615,8 @@ def __init__(self, model, config_list, optimizer=None, dummy_input=None):
                 # we still need to deal with the old_bias when it occurs
                 if hasattr(wrapper.module, "old_bias"):
                     self.optimizer.add_param_group({"params": getattr(wrapper.module, "old_bias")})
+        else:
+            _logger.warning("Quantization step may not happen since optimizer is not set. Suggest passing an optimizer to Quantizer.")
```
J-shang, line 619: Not all quantizer needs an optimizer, suggest removing the default value `None` of `optimizer` in `QAT` and others, and move the warning in these quantizers with judging `isinstance(optimizer, Optimizer)`.

chenbohua3: Good point. Observer quantizer currently does not need optimizer

linbinskn: Done.

acured (approved):

PR #3991, opened 2021-07-29
Fix kubeflow & frameworkcontroller pipeline

SparkSnail:


nni/tools/nnictl/launcher_utils.py
```diff
@@ -128,5 +128,3 @@ def validate_all_content(experiment_config, config_path):
 
     if 'maxExecDuration' in experiment_config:
         experiment_config['maxExecDuration'] = parse_time(experiment_config['maxExecDuration'])
-    if 'maxTrialDuration' in experiment_config:
-        experiment_config['maxTrialDuration'] = parse_time(experiment_config['maxTrialDuration'])
```
QuanluZhang, line 132: why remove "maxTrialDuration" here?

SparkSnail: It use parse_time() to convert time format to seconds here, but it convert the number in nniManager again. Only need to keep one convert.

SparkSnail: https://github.com/microsoft/nni/blob/master/ts/nni_manager/core/nnimanager.ts#L435

QuanluZhang: seems "maxExecDuration" is also converted in nnimanager. line 425

SparkSnail: It is "maxExperimentDuration", not "maxExecDuration" here. The two values seems duplicated.

ultmaster (approved):

ts/nni_manager/core/nnimanager.ts
```diff
@@ -457,6 +457,9 @@ class NNIManager implements Manager {
         } else if (platform === 'local') {
             const module_ = await import('../training_service/local/localTrainingService');
             return new module_.LocalTrainingService(config);
+        } else if (platform === 'kubeflow') {
+            const module_ = await import('../training_service/kubernetes/kubeflow/kubeflowTrainingService');
+            return new module_.KubeflowTrainingService();
```
QuanluZhang, line 462: this means kubeflow does not reuse mode by default?

SparkSnail: This can't be removed, since kubeflow non-reuse mode only support v1 config for now. It can set to reuse mode as default in v2 config nnictl side.

ts/nni_manager/rest_server/restValidationSchemas.ts
```diff
@@ -226,6 +226,7 @@ export namespace ValidationSchemas {
             trainingServicePlatform: joi.string(),
             searchSpace: joi.string().required(),
             maxExecDuration: joi.number().min(0).required(),
+            maxTrialDuration: joi.number().min(0).required(),
```
QuanluZhang, line 229: where is "STARTEXPERIMENT" used?

SparkSnail: It is validation for V1 config to start experiment, validation not enabled now, but we had added the field, and should add it here.

QuanluZhang (approved):

PR #3992, opened 2021-07-29
restore weight by default for qat quantizer

chenbohua3:
By default, the type of `weight` will change from `torch.nn.Parameter` to `torch.Tensor(buffer)`. This will make its weight unable  to be accessed by functions like `named_parameters()`. So I think it is better to restore it by default.

### DIFF
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -570,14 +570,13 @@ class QAT_Quantizer(Quantizer):
                 calibration_config[name]['tracked_max_input'] = float(module.tracked_max_input)
 
                 # Recover weight/bias for batch normalization folding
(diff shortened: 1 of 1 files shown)


QuanluZhang:
> By default, the type of `weight` will change from `torch.nn.Parameter` to `torch.Tensor(buffer)`. This will make its weight unable to be accessed by functions like `named_parameters()`. So I think it is better to restore it by default.

`old_weight` is the weight after quantization aware training, it is not the original weight. i suggest to output the model with quantized weight (neither the original weight nor the weight after quantization aware training). that is, it is better to apply `quantize` on `old_weight` before saving weights.

linbinskn (commented):
Recommend adding quantize weight logic in both export model and the first inference.

QuanluZhang (approved):

J-shang (approved):

PR #3993, opened 2021-07-29
fix sharedstorage it

J-shang:
1. mount point in local may late to umount when stop experiment in `shared-storage-remote-azureblob`, so keep two shared storage tests away from each other.
2. hide tensorboard not installed error in `nnictl_stderr`

### DIFF
--- a/ts/nni_manager/core/nniTensorboardManager.ts
+++ b/ts/nni_manager/core/nniTensorboardManager.ts
@@ -114,9 +114,9 @@ class NNITensorboardManager implements TensorboardManager {
     }
 
     private setTensorboardVersion(): void {
-        let command = `python3 -c 'import tensorboard ; print(tensorboard.__version__)'`;
+        let command = `python3 -c 'import tensorboard ; print(tensorboard.__version__)' 2>&1`;
         if (process.platform === 'win32') {
-            command = `python -c "import tensorboard ; print(tensorboard.__version__)"`;
+            command = `python -c "import tensorboard ; print(tensorboard.__version__)" 2>&1`;
         }
         try {
             const tensorboardVersion = cp.execSync(command).toString();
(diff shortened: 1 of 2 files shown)


ts/nni_manager/core/nniTensorboardManager.ts
```diff
@@ -114,9 +114,9 @@ class NNITensorboardManager implements TensorboardManager {
     }
 
     private setTensorboardVersion(): void {
-        let command = `python3 -c 'import tensorboard ; print(tensorboard.__version__)'`;
+        let command = `python3 -c 'import tensorboard ; print(tensorboard.__version__)' 2>&1`;
         if (process.platform === 'win32') {
-            command = `python -c "import tensorboard ; print(tensorboard.__version__)"`;
+            command = `python -c "import tensorboard ; print(tensorboard.__version__)" 2>&1`;
```
QuanluZhang, line 119: 1 means stdout, you redirect stderr message to stdout? nnictl also collect stdout right?

QuanluZhang: maybe " >/dev/null 2>&1", not sure whether it works in windows...

J-shang: no, stdout will be collected by nnimanager

QuanluZhang: ok

QuanluZhang (approved):

ts/nni_manager/core/nniTensorboardManager.ts
J-shang: tensorboard manager needs the stdout of this command to check the version, so `>/dev/null` will not work?

ultmaster (approved):

PR #4005, opened 2021-08-02
fix trial status not correct when trial is early stoped

acured:
Caused: the trial status was update when the trial process is exit completely(e.g. from RUNING to EARLY_STOPPED), but process exiting will cost some times which could cause status not match. 

Resue mode do not have this issue.

### DIFF
--- a/ts/nni_manager/core/nnimanager.ts
+++ b/ts/nni_manager/core/nnimanager.ts
@@ -434,7 +434,7 @@ class NNIManager implements Manager {
 
     private get maxTrialDuration(): number {
         const value = this.experimentProfile.params.maxTrialDuration;
-        let duration = (value === undefined ? Infinity : toSeconds(value));
+        const duration = (value === undefined ? Infinity : toSeconds(value)) * 1000;
         // Fix timeout warning : Infinity does not fit into a 32-bit signed integer(2147483647).
         return duration > 2147483647 ? 2147483647 : duration;
     }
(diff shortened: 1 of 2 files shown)


ts/nni_manager/training_service/local/localTrainingService.ts
```diff
@@ -236,20 +236,11 @@ class LocalTrainingService implements TrainingService {
...
-                });
-                break;
+        tkill(trialJob.pid, 'SIGKILL', (err) => {
+            if (err) {
+                this.log.error(`kill trial job error: ${err}`);
             }
-            await delay(500);
-        }
-
+        });
```
J-shang, line 243: `SIGTERM` and wait for a few time because maybe the trial has some clean logic, why this will cause the wrong job status?

acured: If user canceled(early stop) a trial, the status is not changed immediately, which will keep "RUNNING" for a while, From "nnimanager" main loop, there is a logic that set "RUNNING" to "FAILED" when status is "RUNNING" but the PID is not alive.

That will cause the status is "FAILED" which should be "EARLY_STOP" sometimes.

J-shang: got it, but change `SIGTERM` to `SIGKILL` may cause the early stop trial to skip its clean-up, is this a serious change? @QuanluZhang

QuanluZhang: let's discuss it tomorrow

acured: updates: set early_stop before isalive check.

ts/nni_manager/core/nnimanager.ts
```diff
@@ -681,7 +683,7 @@ class NNIManager implements Manager {
                     this.currSubmittedTrialNum++;
                     this.log.info('submitTrialJob: form:', form);
                     const trialJobDetail: TrialJobDetail = await this.trainingService.submitTrialJob(form);
-                    setTimeout(async () => this.stopTrialJobIfOverMaxDurationTimer(trialJobDetail.id), 1000 * this.maxTrialDuration);
+                    setTimeout(async () => this.stopTrialJobIfOverMaxDurationTimer(trialJobDetail.id), this.maxTrialDuration);
```
liuzhe-lz, line 686: I think it should only add timer when duration is not infinity.

acured: Yes, changed.

J-shang (approved):

QuanluZhang (approved):

PR #4007, opened 2021-08-02
remove vgg16 from speedup test

J-shang:
because vgg16 is too big and usually breaks fast test...

### DIFF
--- a/test/ut/sdk/test_model_speedup.py
+++ b/test/ut/sdk/test_model_speedup.py
@@ -349,7 +349,7 @@ class SpeedupTestCase(TestCase):
         self.speedup_integration(model_list)
 
     def test_speedup_integration_big(self):
-        model_list = ['vgg11', 'vgg16', 'resnet34', 'squeezenet1_1',
+        model_list = ['vgg11', 'resnet34', 'squeezenet1_1',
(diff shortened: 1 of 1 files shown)


test/ut/sdk/test_model_speedup.py
```diff
@@ -349,7 +349,7 @@ def test_speedup_integration_small(self):
         self.speedup_integration(model_list)
 
     def test_speedup_integration_big(self):
-        model_list = ['vgg11', 'vgg16', 'resnet34', 'squeezenet1_1',
+        model_list = ['vgg11', 'resnet34', 'squeezenet1_1',
```
zheng-ningxin, line 352: How about skip the `speedup_integration_test` on macos and windows rather than remove the vgg model?

J-shang: it also breaks the latest ubuntu, so I think maybe remove it is a better choice, does vgg16 have some special structure that we need to test?

zheng-ningxin: No big deal, we can remove it as long as we still keep vgg11.

zheng-ningxin (approved):

QuanluZhang (approved):

PR #4025, opened 2021-08-05
Disable GPU in unit test

liuzhe-lz:
Currently the UT will not exit if the machine has GPU. This is because local training service test case will open GPU metric file and will never close it. We could add closing logic to fix it, but I think it's not the right way.
Our default UT environment is CPU-only, and by design an experiment with GPU number > 1 should fail on this type of machine. The test case was passing by accident because the error of GPU scheduler is silently ignored. And since it failed to generate GPU metric file, there is no closing issue.
This does not harm real experiments because they are stopped by `process.exit()`, which will not be blocked by file handlers.


QuanluZhang (approved):

ultmaster:
I triggered an IT to test this PR. Please ensure it passes before merge.

https://msrasrg.visualstudio.com/NNIOpenSource/_build/results?buildId=25918

ultmaster (approved):

PR #4031, opened 2021-08-06
update compression exclude example

J-shang:


dependencies/required.txt
```diff
@@ -21,3 +21,5 @@ scipy < 1.6 ; python_version < "3.7"
 scipy ; python_version >= "3.7"
 pandas < 1.2 ; python_version < "3.7"
 matplotlib < 3.4 ; python_version < "3.7"
+tornado
```
ultmaster, line 24: Why do we need tornado as required dep?

dependencies/required.txt
```diff
@@ -21,3 +21,5 @@ scipy < 1.6 ; python_version < "3.7"
...
+jupyter_core
```
ultmaster, line 25: Why is jupyter required?

dependencies/required.txt
J-shang: because when I pip install nni in a new env, these two packages are required.

J-shang: image

J-shang: reset the requirement

QuanluZhang (approved):

ultmaster (approved):

PR #4034, opened 2021-08-06
Catch exception from ONNX conversion

ultmaster:


nni/retiarii/evaluator/pytorch/lightning.py
```diff
@@ -178,7 +178,10 @@ def validation_step(self, batch, batch_idx):
         y_hat = self(x)
 
         if not self._already_exported:
-            self.to_onnx(self.export_onnx, x, export_params=True)
+            try:
+                self.to_onnx(self.export_onnx, x, export_params=True)
+            except RuntimeError as e:
+                warnings.warn(f'ONNX conversion failed. As a result, you might not be able to use visualization. Error message: {e}')
```
chenbohua3, line 184: when fail to export onnx, `self._already_exported` should be `False`

ultmaster: It shouldn't try to export again.

QuanluZhang (approved):

J-shang (approved):

PR #4036, opened 2021-08-06
Fix 2 bugs found in bug bash

liuzhe-lz:
1. nnictl view —experiment path/to/exp-id/ (with backslash) does not work
2. log files have content type “json”


ultmaster (approved):
Looks like I've made a silly mistake. ;(

SparkSnail (approved):

PR #4038, opened 2021-08-06
fix wrong quantization target in weight quantization

chenbohua3:
This pr contains two things:
1. When bn fold is enable, the folded weight is assign to `module.weight`, QAT quantizer should quantize `module.weight` instead of `module.old_weight`
2. pass the weight that should be quantized to each quantizer directly to make the training graph more clearly.

### DIFF
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -2,7 +2,6 @@
 # Licensed under the MIT license.
 
 import logging
-import copy
 from collections import defaultdict
 import torch
 from schema import Schema, And, Or, Optional
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -673,7 +673,7 @@ class DoReFaQuantizer(Quantizer):
         schema.validate(config_list)
 
     def quantize_weight(self, wrapper, **kwargs):
-        weight = wrapper.module.old_weight
+        weight = wrapper.module.weight
         weight_bits = get_bits_length(wrapper.config, 'weight')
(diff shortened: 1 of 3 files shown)


nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -221,15 +220,13 @@ def quantize_input(self, *inputs, wrapper, **kwargs):
             self.record(wrapper, 'input', inputs)
         return inputs
 
-    def quantize_weight(self, wrapper, **kwargs):
+    def quantize_weight(self, weight, wrapper, **kwargs):
```
QuanluZhang, line 223: could you explain more about this change? `weight` can be obtained from `wrapper`, why pass it again?

chenbohua3: It's about the code readability. Since we already pass the weight to be quantized (`new_weight`) to quant_grad [here](https://github.com/microsoft/nni/blob/9cd42f61fbd4ea227b8935353a58b0472ac212ff/nni/compression/pytorch/compressor.py#L562), it is better to directly use it instead of obtaining it from wrapper. The developer can easily know that `quantize_weight` is a big op to simulate quantization and the op takes origin/bn-folded weight as input. I think using `wrapper.weight` will make it difficult to understand to structure of the training graph.

linbinskn:
LGTM. Since this PR also changes `quantize_weight` in DoReFa and BNN quantizer, please also test them together.

linbinskn (approved):

chenbohua3:
Have added a ut about the interface of `quantize_weight`

QuanluZhang (approved):

linbinskn (approved):

J-shang (approved):

PR #4042, opened 2021-08-07
fix shared storage

J-shang:
TODO in next release:
1. validate if local mount point outside code dir
2. check if the mount point is already mounted and give a warning.


examples/trials/mnist-sharedstorage/config_azureblob.yml
```diff
@@ -24,12 +24,14 @@ trainingService:
                             #   C:/Users/USERNAME/.conda/envs/ENVNAME;C:/Users/USERNAME/.conda/envs/ENVNAME/Scripts;C:/Users/USERNAME/.conda/envs/ENVNAME/Library/bin
 sharedStorage:
   storageType: AzureBlob
+  # please set localMountPoint as absolute path and localMountPoint should outside the code directory
+  # because nni will copy user code to localMountPoint
```
QuanluZhang, line 28: suggest to validate the provided path in code, not only put it in comment/doc. we can leave it in next release.

J-shang: Relative path validation is in experiment/config/common.py.

acured (approved):

QuanluZhang (approved):

PR #4047, opened 2021-08-09
[Model Compression] Fix register backward hook for TransformerHeadPruner

xiaowu0162:
When criterion `taylorfo` is used, and if the user attempts to re-instantiate the pruner after speedup, a mismatch in dimension could happen (it seems that the dimension of data collected by the hook is incorrect), with the following warning message. 
`UserWarning: Using a non-full backward hook when the forward contains multiple autograd Nodes is deprecated and will be removed in future versions. This hook will be missing some grad_input. Please use register_full_backward_hook to get the documented behavior.`


xiaowu0162:
Although using `register_full_backward_hook` solves the problem, its effect is not entirely clear. I am still investigating on it. Will make this pr ready if I think the solution is correct.

xiaowu0162:
Another comment: we might need to check user's PyTorch version to decide whether we use `register_backward_hook` or `register_full_backward_hook`

xiaowu0162:
I think this problem is not very urgent, and will not cause errors in most expected use cases.

PR #4086, opened 2021-08-19
Bugfix: wrong device placement and invalid CUDA ordinal when using CGO engine

hzhua:
[x] Use an abstract class `Device` for `GPUDevice` and `CPUDevice`
[x] Fix invalid CUDA ordinal because the CUDA device ID in PyTorch is different from the physical GPU ID when CUDA_VISIBLE_DEVICES are set.
[x] The device placement is wrong since CGOExecutionEngine misses the `placement_constraint` parameter when `send_trial`.
[x] Error thrown since Single-host multi-GPU placement is wrongly identified as multi-host placement, which will throw error.
[ ] CSE is disabled after the first trial finishes, which is not an expected behavior.

### DIFF
--- a/nni/common/device.py
+++ b/nni/common/device.py
@@ -2,6 +2,9 @@
 # Licensed under the MIT license.
 
 from dataclasses import dataclass
+from abc import ABC, abstractmethod
+from os import stat
+
 try:
     from typing import Literal
 except ImportError:
@@ -9,32 +12,60 @@ except ImportError:
 
 
 @dataclass
-class GPUDevice:
+class Device(ABC):
     node_id: str
-    gpu_id: int
     status: Literal['idle', 'busy', 'unknown'] = 'idle'
 
     def __eq__(self, o) -> bool:
+        if type(self) == type(o):
+            return self.node_id == o.node_id
+        else:
+            return False
+
+    def __lt__(self, o) -> bool:
+        return self.node_id < o.node_id
+
+    def set_status(self, status):
+        self.status = status
+
+    def __repr__(self) -> str:
+        return "{Abstract Device %s, Status %s}" % (self.node_id, self.status)
+
+    @abstractmethod
+    def device_repr(self) -> str:
+        pass
+
+
+@dataclass
+class GPUDevice(Device):
+    gpu_id: str = -1
+
+    def __init__(self, node_id, gpu_id, status='idle'):
+        self.node_id = node_id
+        self.gpu_id = gpu_id
+        self.status = status
+
+    def __eq__(self, o: Device) -> bool:
         if isinstance(o, GPUDevice):
             return self.node_id == o.node_id and self.gpu_id == o.gpu_id
         return False
 
-    def __lt__(self, o) -> bool:
+    def __lt__(self, o: Device) -> bool:
         if self.node_id < o.node_id:
             return True
         elif self.node_id > o.node_id:
             return False
         else:
-            return self.gpu_id < o.gpu_id
+            if isinstance(o, GPUDevice):
+                return self.gpu_id < o.gpu_id
+            else:
+                return True
 
     def __repr__(self) -> str:
         return "{Environment %s, GPU %d, Status %s}" % (self.node_id, self.gpu_id, self.status)
 
     def __hash__(self) -> int:
         return hash(self.node_id + '_' + str(self.gpu_id))
-
-    def set_status(self, status):
-        self.status = status
 
     def device_repr(self,):
         return f"cuda:{self.gpu_id}"
--- a/nni/retiarii/execution/cgo_engine.py
+++ b/nni/retiarii/execution/cgo_engine.py
@@ -9,7 +9,7 @@ import time
 import threading
 from typing import Iterable, List, Dict, Tuple
 
-from nni.common.device import GPUDevice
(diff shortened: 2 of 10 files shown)


nni/common/device.py
```diff
@@ -2,39 +2,69 @@
...
+
+
+@dataclass
+class GPUDevice(Device):
+    gpu_id: str = -1
+
+    def __init__(self, node_id, gpu_id, status='idle'):
+        self.node_id = node_id
+        self.gpu_id = gpu_id
+        self.status = status
```
QuanluZhang, line 45: is this typical usage of `@dataclass`?

nni/retiarii/execution/cgo_engine.py
```diff
@@ -120,9 +128,10 @@ def _submit_models_in_batch(self, *models: List[Model]) -> None:
         phy_models_and_placements = self._assemble(logical)
         for model, placement, grouped_models in phy_models_and_placements:
             data = BaseGraphData(codegen.model_to_pytorch_script(model, placement=placement), model.evaluator)
-            trial_id = send_trial(data.dump())
+            placement_constraint = self._extract_placement_constaint(placement)
+            trial_id = send_trial(data.dump(), placement_constraint=placement_constraint)
             # unique non-cpu devices used by the trial
-            self._trial_used_devices[trial_id] = list([_ for _ in set(placement.values()) if isinstance(_, GPUDevice)])
+            self._trial_used_devices[trial_id] = list(set([_ for _ in placement.values() if isinstance(_, GPUDevice)]))
```
QuanluZhang, line 134: you check `GPUDevice` in many places, is there other type of devices that you want to deal with in future?

nni/retiarii/operation_def/torch_op_def.py
```diff
@@ -494,14 +494,25 @@ class ToDevice(PyTorchOperation):
...
         self.src = parameters['src']
         self.dst = parameters['dst']
 
+    def override_device_repr(self, device_repr):
+        self.overridden_device_repr = device_repr
+
     def __repr__(self):
-        return f'to("{self.device}")'
+        if self.overridden_device_repr == None:
+            return f'to("{self.device.device_repr()}")'
```
QuanluZhang, line 506: in what cases that `overridden_device_repr` is None?

nni/retiarii/execution/logical_optimizer/logical_plan.py
```diff
@@ -138,8 +141,8 @@ def _merge_graph(self, from_graph):
             new_tail = id_to_new_node[edge.tail.id]
             Edge((new_head, edge.head_slot), (new_tail, edge.tail_slot), _internal=True)._register()
 
-    def assemble(self, multi_model_placement: Dict[Model, GPUDevice]) \
-            -> Tuple[Model, Dict[Node, Union[GPUDevice, CPUDevice]]]:
+    def assemble(self, multi_model_placement: Dict[Model, Device]) \
+            -> Tuple[Model, Dict[Node, Device]]:
```
QuanluZhang, line 145: better to add docstring for this function

nni/retiarii/execution/cgo_engine.py
QuanluZhang: i see, you also handle cpu device, which is nice!

nni/retiarii/execution/logical_optimizer/logical_plan.py
```diff
@@ -2,20 +2,23 @@
...
 from nni.retiarii.utils import uid
-from nni.common.device import GPUDevice
+from nni.common.device import Device
 
 from ...graph import Cell, Edge, Graph, Model, Node
 from ...operation import Operation, _IOPseudoOperation
 
 
-class CPUDevice:
+class CPUDevice(Device):
```
QuanluZhang, line 14: it is a little strange that why not put `CPUDevice` into device.py? it should also be a `dataclass`?

nni/retiarii/operation_def/torch_op_def.py
hzhua: CUDA `GPUDevice` may remap GPU physical ID to CUDA ID. The device_repr is different from `GPUDevice.device_repr`. 
`override_device_repr` will be called in `pytorch.graph_to_pytorch_model` to replace `device_repr` with the correct
CUDA ID, e.g., when a job uses Physical GPU-1,2, its CUDA ID should be `cuda:0` and `cuda:1`.
`self.device.device_repr()` would return `cuda:1` and `cuda:2`, but `override_device_repr` should be `cuda:0` and
`cuda:1`. 

I just add the comments in code to explain this.

nni/retiarii/execution/logical_optimizer/logical_plan.py
hzhua: Good comment. I have moved `CPUDevice` in nni.common.device and mark it as a dataclass.

nni/retiarii/execution/logical_optimizer/logical_plan.py
hzhua: Done. I have add the docstring for `assemble` in `AbstractLogicalNode`. Also, I add comments in each type of logical node to explain its function and how should they be assembled.

QuanluZhang (approved):

PR #4074, opened 2021-08-16
[Model Compression] update config list key

J-shang:
In iterative mode, `sparsity_per_layer` is hard to satisfy the sparse expression of each iteration if `speed up`. This is because each layer may have different sparsity after `speed up`. Then in the next iteration, sparsity should be specified for each layer, `sparsity_per_layer` can not express this.

This pr keep the user interface but convert `sparsity` and `sparsity_per_layer` to `total_sparsity` inside pruner.

Convert `max_sparsity_per_layer` to `min_retention_numel` (Dict[op_name, int]) ?
Sparsity is element level?

PR #4089, opened 2021-08-20
[Model Compression] Pruning Scheduler

J-shang:
depend on pr #4074


nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
```diff
@@ -0,0 +1,90 @@
...
+
+        # speed up
+        # TODO: speed up only support mask file path as input, maybe we need also support masks directly.
+        tmp_masks = {}
+        for name, mask in masks.items():
+            tmp_masks[name] = {}
+            tmp_masks[name]['weight'] = mask.get('weight_mask')
+            if 'bias' in masks:
+                tmp_masks[name]['bias'] = mask.get('bias_mask')
+        torch.save(tmp_masks, Path('./temp_masks.pth'))
```
xiaowu0162, line 70: Should we clean up this temporary mask?

J-shang: yes, and maybe the better way is to allow `ModelSpeedup`directly using masks object. Can we add an argument like `masks` in `ModelSpeedup`? If it is ok, I will add it in this pr. @zheng-ningxin

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def pruning_one_step(self, model: Module, config_list: List[Dict], masks: Dict[str, Dict[str, Tensor]]) \
+            -> Tuple[Module, Dict[str, Dict[str, Tensor]], float, Dict[str, Dict[str, Tensor]]]:
+        """
+        Pruning the model with config list.
+
+        Parameters
+        ----------
+        model
+            The model under pruning.
```
QuanluZhang, line 48: what is the meaning of "under pruning"?

QuanluZhang: maybe we can say, it is a normal pytorch module. no wrapper, no mask in `model` right?

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+        config_list
+            The config list usually identify the layers that want to prune.
+        masks
+            The masks should be applied on the under pruning model.
```
QuanluZhang, line 52: explain more for `masks`. for example, what `masks` is when the model is speeded up (i.e., compact model), what `masks` is when the model is not speeded up

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+        Parameters
+        ----------
+        task_id
+            The id of the finished task.
+        pruned_model
+            The pruned model after `pruning_one_step`.
+        masks
+            The masks should be applied on the pruned model.
+        score
+            The score of the pruning performance in this task.
```
QuanluZhang, line 34: could explain more for `score`, maybe give an example

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -72,7 +72,7 @@
 }
 
 
-class OneShotPruner(Pruner):
+class ToolBasedPruner(Pruner):
```
QuanluZhang, line 75: what does "tool" refer to for this class? "tool" is a too general word

QuanluZhang: also, add docstring for this class

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: Yes, this model is an unwrapped model. "under pruning" means this model is the model before pruning and will be pruned in this step. will update the docstring.

nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
```diff
@@ -0,0 +1,94 @@
...
+        self.pruner.load_masks(masks)
+        model, masks = self.pruner.compress()
+        up_model_masks = deepcopy(masks)
+
+        # show the pruning effect
+        self.pruner.show_pruned_weights()
+        self.pruner._unwrap_model()
+
+        # speed up
+        # TODO: speed up only support mask file path as input, maybe we need also support masks directly.
```
QuanluZhang, line 66: good point

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: "tool" refer to `DataCollector`, `MetricsCalculator` and `SparsityAllocator`, or rename to `BasicPruner`?

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -440,3 +444,185 @@ def _compress_mask(self, mask: Tensor) -> Tensor:
...
+        """
+        Parameters
+        ----------
+        origin_model
+            The origin model under pruning.
+        origin_config_list
+            The origin config list provided by the user. Note that this config_list is directly config the origin model.
+            This means the sparsity provided by the origin_masks should also be recorded in the origin_config_list.
+        origin_masks
+            The pre masks apply on the origin model.
```
QuanluZhang, line 482: in what type of scenario that this argument is useful?

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -440,3 +444,185 @@ def _compress_mask(self, mask: Tensor) -> Tensor:
...
+        log_dir
+            The log directory use to saving the task generator log.
+        save_result
+            If saving the intermediate result, including intermediate model and masks during each iteration.
```
QuanluZhang, line 486: is there any problem if we make the default value of `save_result` false?

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -440,3 +444,185 @@ def _compress_mask(self, mask: Tensor) -> Tensor:
...
+            if self._best_score is None or score > self._best_score:
+                self._best_score = score
+                self._best_task_id = task_id
+                self._save_best_result(model, task.config_list, masks)
+        if self._save_result:
+            torch.save(model, Path(task.log_dir, 'pruned_model.pth'))
+            torch.save(masks, Path(task.log_dir, 'masks_on_pruned_model.pth'))
+            under_pruning_model, _, _ = self._load_task_data(task_id)
+            torch.save(under_pruning_model, Path(task.log_dir, 'under_pruning_model.pth'))
+            torch.save(up_model_masks, Path(task.log_dir, 'masks_on_under_pruning_model.pth'))
```
QuanluZhang, line 564: a little confusing, you have `model`, `masks`, `under_pruning_model`, `up_model_masks`... and you also load task data from file

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: this is for lottery ticket resetting weight. in each lottery ticket iteration, the remaining weight of the pruned model may reset as the origin model init weight.

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: no, `save_result` can be set False as default, I will modify it.

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+from torch.tensor import Tensor
+
+
+class BasePruningScheduler:
+    def generate_task(self) -> Tuple[int, Module, List[Dict], Dict[str, Dict[str, Tensor]]]:
+        """
+        Returns
+        -------
+        Tuple[int, Module, List[Dict], Dict[str, Dict[str, Tensor]]]
+            Return new task id, model under pruning, config list used in this task and pre-masks.
```
zheng-ningxin, line 16: A little bit concerned about the memory consumption. 
Do we need to deepcopy the module? if so, we may consume twice as much memory ? Besides, if we need save the dependent checkpoint for each task, it will consume more memory?

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: these four things are the task result, the user may not save them if `save_result` is false.
task data is the input of the pruner and saved in another place.

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: Yes, we now use the disk save/load model to interact, but because pruner will reference the previous wrapped pruned model until the next model pass in, we will consume twice memory...
I will check the memory usage and fix this problem.

J-shang: or maybe the `task` is not a good implement, the logic is too complex, a specific scheduler for each scheduling algorithm may be better...

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+        ----------
+        task_id
+            The id of the finished task.
+        pruned_model
+            The pruned model after `pruning_one_step`.
+        masks
+            The masks should be applied on the pruned model.
+        score
+            The score of the pruning performance in this task.
+        up_model_masks
```
zheng-ningxin, line 35: Suggest to change a name. How about `inherited_mask`? I guess this parameter is the masks that inherited from the previous tasks?

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        Used to record the task result.
+
+        Parameters
+        ----------
+        task_id
+            The id of the finished task.
+        pruned_model
+            The pruned model after `pruning_one_step`.
+        masks
```
zheng-ningxin, line 31: Please clarify the difference between the `masks` and the `up_model_masks`?

nni/algorithms/compression/v2/pytorch/base/scheduler.py
xiaowu0162: Also saw people using `prev_mask` or `pre_mask` elsewhere.

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,89 @@
...
+        Returns
+        -------
+        Tuple[Module, Dict[str, Dict[str, Tensor]]]
+            Return the pruned_model and the masks on it in the last iteration.
+        """
+        task_id, model, config_list, pre_masks = self.generate_task()
+        pruned_model, masks = None, None
+
+        while task_id is not None:
+            pruned_model, masks, score, up_model_masks = self.pruning_one_step(model, config_list, pre_masks)
```
xiaowu0162, line 85: Do we want to handle exceptions here? If so, what are the possible scenarios? It seems that `pruning_one_step` can fail for many reasons: the pruner throws an error, the ModelSpeedup fails, or even the disk is full... Do we want to step the entire iterative process for all these scenarios, especially with possible distributed applications in the future in mind?

xiaowu0162: I think it would be impossible to take into account all possibilities. But just to give one instance, maybe we should not fully stop when speedup fails/throws an error in one iteration.

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -440,3 +444,185 @@ def _compress_mask(self, mask: Tensor) -> Tensor:
...
+
+    def _load_task_data(self, task_id: int) -> Tuple[Module, List[Dict], Dict[str, Dict[str, Tensor]]]:
+        task = self._tasks[task_id]
+        model = torch.load(task.model_path)
+        config_list = deepcopy(task.config_list)
+        masks = torch.load(task.masks_path)
```
xiaowu0162, line 570: Is the default device behavior of `torch.load` reliable? Do we need to explicitly check device placement or move the model and masks to the same device?

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
```diff
@@ -0,0 +1,92 @@
...
+        self.current_iteration += 1
+
+        return [task]
+
+    def _generate_config_list(self, target_sparsity: List[Dict], iteration: int, model_based_sparsity: List[Dict]) -> List[Dict]:
+        raise NotImplementedError()
+
+
+class AGPTaskGenerator(FunctionBasedTaskGenerator):
+    def _generate_config_list(self, target_sparsity: List[Dict], iteration: int, model_based_sparsity: List[Dict]) -> List[Dict]:
```
QuanluZhang, line 72: it is a little strange that you override a private method (i.e., with `_` as previous)

nni/algorithms/compression/v2/pytorch/utils/pruning.py
```diff
@@ -82,3 +84,99 @@ def dedupe_config_list(config_list: List[Dict]) -> List[Dict]:
     for idx in sorted(exclude_idxes, reverse=True):
         config_list.pop(idx)
     return config_list
+
+
+def compute_sparsity_compact2origin(origin_model: Module, compact_model: Module, config_list: List[Dict]) -> List[Dict]:
+    """
+    Compare origin model and compact model, return the sparsity of each layer mentioned in config list.
```
QuanluZhang, line 91: the docstring could be more detailed

QuanluZhang:
@J-shang , looks good, better to add necessary tests

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: fix it

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: add docstring

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: add example

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: change to `BasicPruner`

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: modified

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: These four things are packed in a class `TaskResult`

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: add docstring, please refer to `TaskResult` in `scheduler.py`

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: As we discussed, this place will not handle errors right now.

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: maybe we can add an interface in the future

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
J-shang: modified

nni/algorithms/compression/v2/pytorch/utils/pruning.py
J-shang: add more description

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: add gc logic and test with psutil, the memory usage is stable

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -440,3 +441,131 @@ def _compress_mask(self, mask: Tensor) -> Tensor:
...
 
         return (mask != 0).type_as(mask)
+
+
+class TaskGenerator:
+    """
+    This class used to generate config list for pruner in each iteration.
+    """
+    def __init__(self, origin_model: Module, origin_masks: Dict[str, Dict[str, Tensor]] = {},
+                 origin_config_list: List[Dict] = [], log_dir: str = '.', keep_intermidiate_result: bool = False):
```
zheng-ningxin, line 451: what if I run two same pruning tasks at the same time?
Will these two tasks interfere with each other because of the same path? If so suggest adding a unique prefix

nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
zheng-ningxin: Good point! I'll submit a PR to support this feature.

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: I am not sure, do you mean run two task generators at the same time with the same `log_dir`? In this situation, they will interfere with each other, and fixing this problem should use the different `log_dir`.

And two tasks pruning the same model (load from the same model path) will not interfere with each other, because the model file will not be modified after saving, only will be deleted after no more needed.

J-shang:
> @J-shang , looks good, better to add necessary tests

ut will add in another pr #4125.

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
zheng-ningxin: Is there an interface for user to specify a different `log_dir` in the `AutoCompressPruner` or `SimulatedAnnealingPruner`?

J-shang: add subfolder under `log_dir`, format like `2021-09-03-14-19-38-99331`.

xiaowu0162 (approved):

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,183 @@
...
+                    if self._reference_counter[ref] < 0:
+                        _logger.warning('Referance counter error, the number of %s is %d',
+                                        ref, self._reference_counter[ref])
+            self._cleaned = True
+        else:
+            _logger.warning('Already clean up task %d', self.task_id)
+
+
+class TaskResult:
+    def __init__(self, task_id: int, compact_model: Module, compact_model_masks: Dict[str, Dict[str, Tensor]],
```
linbinskn, line 99: It seems `compact_model_masks` will not be used at all since it is empty when model has been speedup or it will be same as old_structure_masks if not. I am curious why not delete it here?

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -27,8 +27,9 @@ def generate_sparsity(self, metrics: Dict[str, Tensor]) -> Dict[str, Dict[str, T
             metric = metrics[name] * self._compress_mask(wrapper.weight_mask)
             prune_num = int(sparsity_rate * metric.numel())
             if prune_num == 0:
-                continue
-            threshold = torch.topk(metric.view(-1), prune_num, largest=False)[0].max()
+                threshold = metric.min() - 1
```
linbinskn, line 30: Why we handle it in this way instead of previous treatment.

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: you are right, `compact_model_mask` is redundant now. In fact, it can be replaced by the boolean flag if the model is speedup. But for the speedup method, some cases can not apply speedup, and I think these cases-related masks should be in the empty `compact_model_masks` in the future. It is a reserved interface for the residue masks after speedup. Maybe users can use this to apply customized speedup.

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
J-shang: Because even though the mask is all one, maybe it is better to be saved in masks.

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,183 @@
...
+                    os.remove(ref)
+                    if self._reference_counter[ref] < 0:
+                        _logger.warning('Referance counter error, the number of %s is %d',
+                                        ref, self._reference_counter[ref])
+            self._cleaned = True
+        else:
+            _logger.warning('Already clean up task %d', self.task_id)
+
+
+class TaskResult:
```
QuanluZhang, line 98: why not merge `TaskResult` into `Task`, or add `TaskResult` instance as a member variable in `Task`?

nni/algorithms/compression/v2/pytorch/base/pruner.py
```diff
@@ -125,54 +128,3 @@ def show_pruned_weights(self, dim: int = 0):
                 sum_idx.remove(dim)
                 index = torch.nonzero(weight_mask.abs().sum(sum_idx) != 0, as_tuple=False).tolist()
             _logger.info(f'simulated prune {wrapper.name} remain/total: {len(index)}/{weight_mask.size(dim)}')
-
-    def export_model(self, model_path, mask_path=None, onnx_path=None, input_shape=None, device=None):
```
QuanluZhang, line 129: export is removed, why?

nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
QuanluZhang: @J-shang , no need to save mask, as @zheng-ningxin has submitted a pr for this and merged

nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
```diff
@@ -0,0 +1,92 @@
...
+    def record_task_result(self, task_result: TaskResult):
+        self.task_generator.receive_task_result(task_result)
+
+    def pruning_one_step(self, task: Task) -> TaskResult:
+        model, masks, config_list = task.load_data()
+
+        # pruning model
+        self.pruner.reset(model, config_list)
+        self.pruner.load_masks(masks)
+        compact_model, old_structure_masks = self.pruner.compress()
```
QuanluZhang, line 59: old_structure_masks -> pruner_generated_masks

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: Because `TaskResult` handles the model directly, it is heavy if add it into `Task`.

nni/algorithms/compression/v2/pytorch/base/pruner.py
J-shang: Because I think the old `export_model()` is useless now because we return masks directly in `compress()`, maybe let users handle export logic is better, or we can support a new `export_model()`.

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -440,3 +441,131 @@ def _compress_mask(self, mask: Tensor) -> Tensor:
...
+        """
+        assert isinstance(origin_model, Module), 'Only support pytorch module.'
+
+        self._log_dir_root = Path(log_dir, datetime.now().strftime('%Y-%m-%d-%H-%M-%S-%f')).absolute()
+        self._log_dir_root.mkdir(parents=True, exist_ok=True)
+        self._keep_intermidiate_result = keep_intermidiate_result
+        self._intermidiate_result_dir = Path(self._log_dir_root, 'intermidiate_result')
+        self._intermidiate_result_dir.mkdir(parents=True, exist_ok=True)
+
+        # save origin data in {log_dir}/intermidiate_model
```
QuanluZhang, line 475: "intermidiate_model"?

J-shang: intermidiate_data?

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
```diff
@@ -0,0 +1,100 @@
...
+        raise NotImplementedError()
+
+
+class AGPTaskGenerator(FunctionBasedTaskGenerator):
+    def generate_config_list(self, target_sparsity: List[Dict], iteration: int, model_based_sparsity: List[Dict]) -> List[Dict]:
+        config_list = []
+        for target, mo in zip(target_sparsity, model_based_sparsity):
+            ori_sparsity = (1 - (1 - iteration / self.total_iteration) ** 3) * target['total_sparsity']
+            sparsity = max(0.0, (ori_sparsity - mo['total_sparsity']) / (1 - mo['total_sparsity']))
+            assert 0 <= sparsity <= 1, 'sparsity: {}, ori_sparsity: {}, model_sparsity: {}'.format(sparsity, ori_sparsity, mo['total_sparsity'])
```
QuanluZhang, line 85: sparsity cannot equal to 1?

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
```diff
@@ -0,0 +1,100 @@
...
+            config_list[-1]['total_sparsity'] = sparsity
+        return config_list
+
+
+class LinearTaskGenerator(FunctionBasedTaskGenerator):
+    def generate_config_list(self, target_sparsity: List[Dict], iteration: int, model_based_sparsity: List[Dict]) -> List[Dict]:
+        config_list = []
+        for target, mo in zip(target_sparsity, model_based_sparsity):
+            ori_sparsity = iteration / self.total_iteration * target['total_sparsity']
+            sparsity = max(0.0, (ori_sparsity - mo['total_sparsity']) / (1 - mo['total_sparsity']))
```
QuanluZhang, line 96: what if the model is not speeded up?

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -65,19 +66,22 @@ def _calculate_threshold(self, group_metric_dict: Dict[str, Tensor]) -> Tuple[fl
...
+            total_weight_num += layer_weight_num
+            expend_times = int(layer_weight_num / metric.numel())
 
             retention_ratio = 1 - max_sparsity_per_layer.get(name, 1)
             retention_numel = math.ceil(retention_ratio * layer_weight_num)
             removed_metric_num = math.ceil(retention_numel / (wrapper.weight_mask.numel() / metric.numel()))
             stay_metric_num = metric.numel() - removed_metric_num
+            if stay_metric_num <= 0:
+                sub_thresholds[name] = metric.min().item() - 1
+                continue
```
QuanluZhang, line 78: mark

nni/algorithms/compression/v2/pytorch/utils/pruning.py
```diff
@@ -82,3 +84,106 @@ def dedupe_config_list(config_list: List[Dict]) -> List[Dict]:
     for idx in sorted(exclude_idxes, reverse=True):
         config_list.pop(idx)
     return config_list
+
+
+def compute_sparsity_compact2origin(origin_model: Module, compact_model: Module, config_list: List[Dict]) -> List[Dict]:
+    """
+    Compare origin model and compact model, return the sparsity of each group mentioned in config list.
+    A group means all layer mentioned in one config.
+    i.e., a linear named 'linear1' and its weight size is [100, 100] in origin model, but in compact model,
```
QuanluZhang, line 93: i.e. -> e.g.

nni/algorithms/compression/v2/pytorch/base/pruner.py
zheng-ningxin: After removing the exporting function, what's the new interface between the pruner and speedup?

nni/algorithms/compression/v2/pytorch/utils/pruning.py
```diff
@@ -82,3 +84,106 @@ def dedupe_config_list(config_list: List[Dict]) -> List[Dict]:
...
+                left_weight_num += module_weight_num
+        mask2compact_sparsity.append(deepcopy(config))
+        mask2compact_sparsity[-1]['total_sparsity'] = 1 - left_weight_num / total_weight_num
+    return mask2compact_sparsity
+
+
+def compute_sparsity(origin_model: Module, compact_model: Module, compact_model_masks: Dict[str, Dict[str, Tensor]],
+                     config_list: List[Dict]) -> Tuple[List[Dict], List[Dict], List[Dict]]:
+    """
+    The current model means the compact model applied the masks. The compact model is the origin model after speed up.
```
QuanluZhang, line 152: where is "current model"?

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
```diff
@@ -400,11 +401,11 @@ def _expand_mask(self, name: str, mask: Tensor) -> Dict[str, Tensor]:
...
-            expand_mask = {'weight_mask': weight_mask.expand(weight_size).clone()}
+            expand_mask = {'weight': weight_mask.expand(weight_size).clone()}
             # NOTE: assume we only mask output, so the mask and bias have a one-to-one correspondence.
             # If we support more kind of masks, this place need refactor.
             if wrapper.bias_mask is not None and weight_mask.size() == wrapper.bias_mask.size():
-                expand_mask['bias_mask'] = weight_mask.clone()
+                expand_mask['bias'] = weight_mask.clone()
         return expand_mask
 
     def _compress_mask(self, mask: Tensor) -> Tensor:
```
QuanluZhang, line 411: add docstring

nni/algorithms/compression/v2/pytorch/utils/pruning.py
QuanluZhang: This function computes how much the origin model has been compressed in the current state. The current state means `compact_model` + `compact_model_masks` (i.e., `compact_model_masks` applied on `compact_model`).

nni/algorithms/compression/v2/pytorch/utils/pruning.py
```diff
@@ -82,3 +84,106 @@ def dedupe_config_list(config_list: List[Dict]) -> List[Dict]:
...
+
+    Returns
+    -------
+    Tuple[List[Dict], List[Dict], List[Dict]]
+        (current2origin_sparsity, compact2origin_sparsity, mask2compact_sparsity).
```
QuanluZhang, line 157: add more description

QuanluZhang:
@J-shang , please add tests for pruning scheduler, maybe in the next pr if you like

nni/algorithms/compression/v2/pytorch/utils/pruning.py
J-shang: done

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: done

nni/algorithms/compression/v2/pytorch/utils/pruning.py
J-shang: done

nni/algorithms/compression/v2/pytorch/utils/pruning.py
J-shang: done

nni/algorithms/compression/v2/pytorch/pruning/tools/base.py
J-shang: done

nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
J-shang: done

nni/algorithms/compression/v2/pytorch/base/pruner.py
J-shang: revert `export_model`, and remove the onnx export.

QuanluZhang (approved):

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,183 @@
...
+        self.masks_path = masks_path
+        self.config_list_path = config_list_path
+
+        self.status: Literal['Pending', 'Running', 'Finished'] = 'Pending'
+        self.score: Optional[float] = None
+
+        self.state = {}
+
+        for ref in self.referenced_paths():
+            self._reference_counter.setdefault(ref, 0)
```
zheng-ningxin, line 45: Is `_reference_counter`  defined in the whole class the same object with `self._reference_counter ` here?

zheng-ningxin: We directly set the reference count to zero here what if `ref` is also used by other threads?

J-shang: yes, all `self._reference_counter` in different instances refer to the same dict `_reference_counter` in class.

it is a good question, maybe we need to use some kind of lock if support multi-threads. for now, we run pruning in a single thread, and I will add comments in this part to remind us, need to refactor this part if we want to support multi-threads.

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,184 @@
...
+class TaskResult:
+    def __init__(self, task_id: int, compact_model: Module, compact_model_masks: Dict[str, Dict[str, Tensor]],
+                 pruner_generated_masks: Dict[str, Dict[str, Tensor]], score: Optional[float]) -> None:
+        """
+        Parameters
+        ----------
+        task_id
+            The unique id of task.
+        compact_model
+            The unwrapped compact pytorch model after pruning. If the compact model has speed up process during pruning,
```
zheng-ningxin, line 108: has been speeduped during the pruning process

nni/algorithms/compression/v2/pytorch/base/scheduler.py
```diff
@@ -0,0 +1,184 @@
...
+            it will have a smaller structure compare with the model before pruning.
+            If the compact model do not speed up, it will have the same structure with the model before pruning.
```
zheng-ningxin, line 110: is not speeduped?
maybe?

zheng-ningxin, line 110: Same here, `do not speed up` -> has not been speeduped?

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: thx, fix it

nni/algorithms/compression/v2/pytorch/base/scheduler.py
J-shang: done

zheng-ningxin (approved):

PR #4093, opened 2021-08-20
fixes #3641 convert data type when reporting results

### LINKED ISSUE #3641: No error raise for wrong type of result
**Describe the issue**:

When parsing value in `torch.tensor` with `dtype=torch.float` to `nni.report_intermediate_result` and `nni.report_final_result`, there is no Exception raised and no result recorded. 

It is desired to raise a `ValueError`, or auto-convert to python float (raise a `ValueError` when cannot be converted). 

**Environment**:
- NNI version:
- Training service (local|remote|pai|aml|etc):
- Client OS:
- Server OS (for remote mode only):
- Python version:
- PyTorch/TensorFlow version:
- Is conda/virtualenv/venv used?:
- Is running in Docker?:

**Configuration**:
 - Experiment config (remember to remove secrets!):
 - Search space:

**Log message**:
 - nnimanager.log:
 - dispatcher.log:
 - nnictl stdout and stderr:
 

**How to reproduce it?**:

ZhiyuanChen:


ultmaster:
We should raise warning for non-standard value format. This includes: int, float, single-value numpy array, and dict with default float. We should not convert the values for users silently, which might cause unexpected behaviors.

QuanluZhang:
@ZhiyuanChen , could you update this pr based on @ultmaster 's suggestion, if you agree with this suggestion.

ZhiyuanChen:
> @ZhiyuanChen , could you update this pr based on @ultmaster 's suggestion, if you agree with this suggestion.

Can I ask why not use json.dumps here? As it will raise an Error as @ultmaster mentioned 
I will update the pr tomorrow

ultmaster:
> > @ZhiyuanChen , could you update this pr based on @ultmaster 's suggestion, if you agree with this suggestion.
> 
> 
> 
> Can I ask why not use json.dumps here? As it will raise an Error as @ultmaster mentioned 
> 
> I will update the pr tomorrow

I think json.dumps can dump quite a lot of different things. But the types that NNI supports are only a subset of those things.

nni/trial.py
```diff
@@ -154,3 +156,22 @@ def report_final_result(metric):
...
+        serializable object.
+    """
+    if not isinstance(metric, (dict, int, float)):
+        raise ValueError('Metric should be a number or a dict with key "default" but got {type(metric)}')
+    if isinstance(metric, dict):
+        if 'default' not in metric.keys():
+            raise ValueError('Metric should have a key "default"')
+        for k, v in metric.items():
+            if not isinstance(v, (int, float)):
+                raise ValueError(f'Metric values should be numbers but got {type(v)} in key {k}')
```
ultmaster, line 177: Can you add some unittests for this function?

It looks quite dangerous. For example, I'm not sure whether it works for types like `numpy.int32`, which I believe should be a valid case.

ultmaster:
Closing this PR as there has been no update for a long time. Feel free to reopen in case of any updates.

PR #4108, opened 2021-08-24
hotfix: fix time format issue

JunweiSUN:
We should not use Date().toLocaleString() to get a formatted time string since it uses the local time format on user's machine and the format is different across platforms (windows, linux, mac) even node versions. This will lead to a failure when starting the rest server. Instead, we construct the time format by ourselves (since using libraries like `strftime` require extra dependences) which should be consistent across different platforms.

BTW, not sure why we add a ' UTC' to a local time and convert it to a UTC time. This will lead to a incorrect time if you are not in +0000.

### DIFF
--- a/ts/nni_manager/common/log.ts
+++ b/ts/nni_manager/common/log.ts
@@ -75,9 +75,8 @@ export class Logger {
             return;
         }
 
-        // `time.toLocaleString('sv')` trick does not work for Windows
-        const isoTime = new Date(new Date().toLocaleString() + ' UTC').toISOString();
-        const time = isoTime.slice(0, 10) + ' ' + isoTime.slice(11, 19);
+        const jsonTime = new Date().toJSON();
+        const time = jsonTime.slice(0, 10) + ' ' + jsonTime.slice(11, 19);
 
         const levelName = levelNames.has(level) ? levelNames.get(level) : level.toString();
 
--- a/ts/nni_manager/common/log.ts
+++ b/ts/nni_manager/common/log.ts
@@ -75,14 +75,17 @@ export class Logger {
             return;
         }
 
-        const jsonTime = new Date().toJSON();
-        const time = jsonTime.slice(0, 10) + ' ' + jsonTime.slice(11, 19);
+        const zeroPad = (num: number) => num.toString().padStart(2, '0');
+        const now = new Date();
+        const date = now.getFullYear() + '-' + zeroPad(now.getMonth() + 1) + '-' + zeroPad(now.getDate());
+        const time = zeroPad(now.getHours()) + ':' + zeroPad(now.getMinutes()) + ':' + zeroPad(now.getSeconds());
+        const datetime = date + ' ' + time;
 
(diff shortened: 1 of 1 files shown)


ts/nni_manager/common/log.ts
```diff
@@ -75,9 +75,8 @@ export class Logger {
             return;
         }
 
-        // `time.toLocaleString('sv')` trick does not work for Windows
-        const isoTime = new Date(new Date().toLocaleString() + ' UTC').toISOString();
-        const time = isoTime.slice(0, 10) + ' ' + isoTime.slice(11, 19);
+        const jsonTime = new Date().toJSON();
```
cruiseliu, line 78: This will print the UTC time. In log we want local time.
The trick above has been tested in en_US and zh_CN locales, but there are hundreds of other locales over the world.
It would be nice if you can find a better method to do the converting.

JunweiSUN: My fault. Didn't notice that toJSON() prints UTC time. But the previous trick seems didn't work on en_US windows machines and zh_CN mac machines. I have updated the pr to calculate a correct time format by our own.

ts/nni_manager/common/log.ts
```diff
@@ -75,15 +75,17 @@ export class Logger {
...
         }
 
-        // `time.toLocaleString('sv')` trick does not work for Windows
-        const isoTime = new Date(new Date().toLocaleString() + ' UTC').toISOString();
-        const time = isoTime.slice(0, 10) + ' ' + isoTime.slice(11, 19);
+        const zeroPad = (num: number): string => num.toString().padStart(2, '0');
+        const now = new Date();
+        const date = now.getFullYear() + '-' + zeroPad(now.getMonth() + 1) + '-' + zeroPad(now.getDate());
+        const time = zeroPad(now.getHours()) + ':' + zeroPad(now.getMinutes()) + ':' + zeroPad(now.getSeconds());
+        const datetime = date + ' ' + time;
```
liuzhe-lz, line 82: Looks really tedious. But I guess there's no better way.

liuzhe-lz (approved):

acured (approved):

liuzhe-lz:
Have you tried `new Date().toLocaleString("sv")` on your machine? The last time with node.js 12.x and windows 10 it does not work for me. With node.js v16.3.0 it seems to work now.

ts/nni_manager/common/log.ts
JunweiSUN: Yes. The only better way may be using a third-party `strftime` library.

JunweiSUN:
> Have you tried `new Date().toLocaleString("sv")` on your machine? The last time with node.js 12.x and windows 10 it does not work for me. With node.js v16.3.0 it seems to work now.

Yes, node 16.x could work while 12.x couldn't. Seems that the format of this function changes a lot across different versions. If we use this function, we must make sure to use the node under `nni_node` dir to start the server.

JunweiSUN:
Shoule we create a hotfix version wheel (like 2.4.1?) to pip to avoid affecting more users? @liuzhe-lz

PR #4116, opened 2021-08-26
[Model Compression] admm pruner

J-shang:


nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -655,3 +654,122 @@ def reset_tools(self):
                 self.sparsity_allocator = Conv2dDependencyAwareAllocator(self, 0, self.dummy_input)
             else:
                 raise NotImplementedError('Only support mode `normal`, `global` and `dependency_aware`')
+
+
+class ADMMPruner(OneShotPruner):
```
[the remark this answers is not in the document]
J-shang: yes, ADMM has scheduler logic, but in fact, it does not generate masks during each iteration. ADMM only reset the elements with small magnitudes to zero, and these elements will also be trained in the following iterations. Only in the last iteration, ADMM will generate masks. So maybe make ADMM a pruner is more reasonable.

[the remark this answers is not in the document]
xiaowu0162: I see. I just read the ADMM paper, and it seems that the iterative elements are for solving the optimization problem instead of iterative pruning.

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -655,3 +654,122 @@ def reset_tools(self):
...
+        self.training_epochs = training_epochs
+        super().__init__(model, config_list)
+
+        self.Z = {name: wrapper.module.weight.data.clone().detach() for name, wrapper in self.get_modules_wrapper().items()}
+        self.U = {name: torch.zeros_like(z).to(z.device) for name, z in self.Z.items()}
+
+    def _validate_config_before_canonical(self, model: Module, config_list: List[Dict]):
+        schema_list = [deepcopy(NORMAL_SCHEMA), deepcopy(INTERNAL_SCHEMA)]
+        for schema in schema_list:
+            schema.update({SchemaOptional('row'): And(float, lambda n: n > 0)})
```
xiaowu0162, line 726: Recommend using 'rho' as 'row' might be confusing. But this involves a change to the original API, so we might don't want to do that.

xiaowu0162: But I do think "row" is very confusing.

J-shang: yes, it's a good suggestion, `rho` is better, I will modify it.

xiaowu0162 (approved):

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -655,3 +654,122 @@ def reset_tools(self):
...
+            The epoch number for training model in each iteration.
+        """
+        self.trainer = trainer
+        self.optimizer = optimizer
+        self.criterion = criterion
+        self.iterations = iterations
+        self.training_epochs = training_epochs
+        super().__init__(model, config_list)
+
+        self.Z = {name: wrapper.module.weight.data.clone().detach() for name, wrapper in self.get_modules_wrapper().items()}
```
zheng-ningxin, line 720: please use a clear name rather than `Z` and `U`.

J-shang: `U` can be named as `scaled_dual_variable`, but I have no idea to name `Z`. The author rewrites the origin problem to 2-block optimization, and `Z` can be seen as another solution of `weight`. I think using `Z` is because in ADMM, the second optimization goal is usually denoted as `z`.
image

And in compression, the problem is
image

zheng-ningxin: Got it.

zheng-ningxin (approved):

QuanluZhang (approved):

PR #4137, opened 2021-09-01
support dtype&scheme customization for QAT quantizer

chenbohua3:


chenbohua3:
I will rebase the master when #4127 is merged into master.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -502,69 +556,103 @@ def quantize_weight(self, wrapper, **kwargs):
...
+        if int(self.bound_model.steps) == 1:
             module.tracked_min_input.copy_(current_min)
             module.tracked_max_input.copy_(current_max)
 
+        tracked_min_input = update_ema(module.tracked_min_input, current_min, ema_decay)
+        tracked_max_input = update_ema(module.tracked_max_input, current_max, ema_decay)
+        module.tracked_min_input.copy_(tracked_min_input)
+        module.tracked_max_input.copy_(tracked_max_input)
+
+        if quant_start_step > int(self.bound_model.steps):
```
linbinskn, line 592: Why delete original logic which can help initiate `tracked_min_input` and `tracked_max_input` here? According to the test result before, keeping them can perform better or converge faster than removing them.

linbinskn: I see, it will be kept whatever `quant_start_step > int(self.bound_model.steps)` or not.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -502,69 +556,103 @@ def quantize_weight(self, wrapper, **kwargs):
...
+            return inputs
+
+        tracked_min_input = update_ema(module.tracked_min_input, current_min, ema_decay)
```
linbinskn, line 595: It seems line 595-598 repeats what line 587-590 does.

nni/compression/pytorch/compressor.py
```diff
@@ -604,10 +607,13 @@ class Quantizer(Compressor):
     def __init__(self, model, config_list, optimizer=None, dummy_input=None):
         if isinstance(model, torch.nn.DataParallel):
             model = model.module
+        model_copied = copy.deepcopy(model)
```
linbinskn, line 610: Why we have to use `deepcopy()` here?

docs/zh_CN/Compression/CustomizeCompressor.rst
```diff
@@ -155,7 +155,7 @@
            grad_output : Tensor
                量化操作输出的梯度
            quant_type : QuantType
```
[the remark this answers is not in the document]
chenbohua3: done

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -376,28 +392,44 @@ def __init__(self, model, config_list, optimizer, dummy_input=None):
...
-            module.register_buffer("zero_point", torch.tensor([0.0]))
-            module.register_buffer("scale", torch.tensor([1.0]))
-            module.register_buffer('ema_decay', torch.tensor([0.99]))
+            name = layer.name
+            # TODO: may relax this limitation?
+            assert name in self.all_shapes, "Could not found shapes for layer {}".format(name)
+            input_shape, output_shape = self.all_shapes[name]
+            layer_quant_setting = LayerQuantSetting(config)
+            layer_quant_setting.ema_decay = 0.99
+            quant_start_step = config.get('quant_start_step', 0)
```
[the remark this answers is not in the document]
chenbohua3: Now `quant_start_step` is a `QAT_Quantizer` specific parameter. I think it is better to set it after `LayerQuantSetting ` initialization

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -426,74 +468,86 @@ def validate_config(self, model, config_list):
...
+        # tensor-wise settings
+        dtype = tensor_quant_setting.quant_dtype
+        scheme = tensor_quant_setting.quant_scheme
+        qmin, qmax = tensor_quant_setting.get_qmin_qmax()
+        bits = tensor_quant_setting.bits
 
+        # In evaluation mode, we only quantize weight without updating statistics
         if not wrapper.training:
+            scale, zero_point = module.weight_scale, module.weight_zero_point
+            weight = self._quantize(weight, scale, zero_point, qmin, qmax)
```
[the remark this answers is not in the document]
chenbohua3: I am afraid that we must can only quantize the weight in each iteration. Because we can not in-place update the weight. So if we use ddp, the update will lost after each iteration.

nni/algorithms/compression/pytorch/quantization/quantizers.py
chenbohua3: have removed them

nni/compression/pytorch/compressor.py
chenbohua3: Because we use `quantizer-wrapped` model to determine which layer's shapes should be recorded. And in the process of recording shape, some hooks would be registered in the model. So I think it is better to use a copied model to do shape recording.

nni/compression/pytorch/compressor.py
```diff
@@ -793,25 +799,54 @@ def find_conv_bn_patterns(self, model, dummy_input):
                     if successor.op_type == 'BatchNorm2d':
                         self.conv_bn_patterns[node_group.name] = successor.name
 
-    def step_with_optimizer(self):
-        pass
+    def record_shape(self, model, dummy_input):
```
linbinskn, line 802: It seems to be an universal function which is not specific for quantizer. Maybe we should put it into other places like utils.py?

nni/compression/pytorch/compressor.py
```diff
@@ -887,28 +922,39 @@ def quant_backward(tensor, grad_output, quant_type, scale, zero_point, qmin, qma
...
-        if hasattr(wrapper.module, 'scale') and hasattr(wrapper.module, 'zero_point'):
+        if hasattr(wrapper.module, "layer_quant_setting"):
+            layer_quant_setting = wrapper.module.layer_quant_setting
+            qmin, qmax = getattr(layer_quant_setting, quant_type).get_qmin_qmax()
+        else:
+            # todo: when dtype/scheme customization is ready for all quantizers, remove this
+            bits = QuantGrad.get_bits_length(wrapper.config, quant_type)
+            qmin, qmax = 0, (1 << bits) - 1
+
+        scale_name, zero_point_name = quant_type.type_to_scale_zero_point_name()
```
linbinskn, line 933: It is strange that we get scale_name and zero_point_name in this way while we register them in module using like 'weight_scale' and `weight_zero_point` directly. So what is the meaning of **type_to_scale_zero_point_name()**?

nni/compression/pytorch/compressor.py
chenbohua3: There exists compressor-specific util function (like `get_modules_to_compress`) and currently it is only used by quantization. Putting this function as a attribute function of quantizer may be a good choice.. :)

nni/compression/pytorch/compressor.py
chenbohua3: Different `quant_type`s correspond to different scale&zero_point names, e.g. `weight_scale`, `input_scale` and `output_scale`. In order to getting them, we must map quantization types to the scale&zero point names. `type_to_scale_zero_point_name` is just for the code simplicity, do you have better ideas about this?

nni/compression/pytorch/quantization/settings.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class TensorQuantSetting(object):
+    def __init__(self, **kwargs):
+        self._fields = {}
+        for k, v in kwargs.items():
+            self._fields[k] = v
+
+    def __setattr__(self, name: str, val: Any) -> None:
+        if name.startswith("_"):
```
J-shang, line 31: is this just for `_fields`?

J-shang:
please fix the conflict

PR #4138, opened 2021-09-01
[Quantization] fix parameter assign problem of weight

linbinskn:
Error will be raised when using observer_quantizer since tensor assigned to parameter.
image

Change `self.module.weight` to `self.module.weight.data` in line 562 can resolve such problem.

### DIFF
--- a/nni/algorithms/compression/pytorch/quantization/quantizers.py
+++ b/nni/algorithms/compression/pytorch/quantization/quantizers.py
@@ -252,6 +252,8 @@ class ObserverQuantizer(Quantizer):
                                             module.weight_zero_point,
                                             module.weight_qmin,
                                             module.weight_qmax)
+                delattr(module, 'weight')
+                module.register_buffer('weight', quantized_weight)
             if "input" in config.get("quant_types", []):
                 scale, zero_point = self.calculate_qparams(layer.name, 'input')
                 module.register_buffer('input_scale', scale.to(self.device))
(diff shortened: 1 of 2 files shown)


chenbohua3:
I think it is better to make `weight` `torch.Tensor` instead of `nn.Parameter`

QuanluZhang:
@chenbohua3 please help review, thanks

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -301,6 +301,14 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
                 calibration_config[name]['tracked_min_weight'] = -val
                 calibration_config[name]['tracked_qmin_weight'] = -127
                 calibration_config[name]['tracked_qmax_weight'] = 127
+                weight = module.weight
+                quantized_weight = self._quantize(weight,
```
J-shang, line 305: why we need `_quantize` again?

chenbohua3: +1

linbinskn: We have to make sure that the type of `module.weight` is `torch.nn.Parameter` in ut [test_compressor_torch](https://github.com/microsoft/nni/blob/master/test/ut/sdk/test_compressor_torch.py#L303). However, this pr register `weight` as buffer in quantizers.py line256 to fix assign weight problem. So we have to convert it as Parameter during `export_model`. And I choose to convert it in this way since it will not influence the correctness.

J-shang (approved):

nni/compression/pytorch/compressor.py
```diff
@@ -559,7 +559,7 @@ def forward(self, *inputs):
                 self.module.weight = new_weight
             else:
                 new_weight = self.module.old_weight
-                self.module.weight.data = new_weight.data
+                self.module.weight = new_weight.data
```
J-shang: why remove `data`?

linbinskn: Because it can cause another error during backward.
image

By the way, i submit this pr to fix the original error which
image
is caused by tensor assigned to parameter since parameter's assignment is invalid. Currently, we resolve this problem by converting parameter to buffer [here](https://github.com/microsoft/nni/blob/c7324918ecbc09cbbfa46e74aa175399de6ea973/nni/algorithms/compression/pytorch/quantization/quantizers.py#L256). And the `self.module.weight = new_weight.data` can be kept, everything works fine right now.

chenbohua3:
LG

acured (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
QuanluZhang: need add comments above this line to explain

PR #4139, opened 2021-09-01
Update amc_pruner.py

### LINKED ISSUE #4117: cudnn.deterministic been set while import, suggestion to avoid this kind of behavior
https://github.com/microsoft/nni/blob/e85f029bd4e4b1bdf3e679893fb6447e4d6b2c79/nni/algorithms/compression/pytorch/pruning/amc/amc_pruner.py#L17

This line of code executes when someone just import `nni.algorithms.compression.pytorch.pruning`, I suggest to move it into the constructor function of `AMCPruner` and reset `cudnn.deterministic` after training because it changed an import setting but does not tell the user.

mstczuo:
fix #4117 :
move `torch.backends.cudnn.deterministic = True` into the internal function of `AMCPruner`.

### DIFF
--- a/nni/algorithms/compression/pytorch/pruning/amc/amc_pruner.py
+++ b/nni/algorithms/compression/pytorch/pruning/amc/amc_pruner.py
@@ -14,8 +14,6 @@ from .channel_pruning_env import ChannelPruningEnv
 from .lib.agent import DDPG
 from .lib.utils import get_output_folder
 
-torch.backends.cudnn.deterministic = True
-
 _logger = logging.getLogger(__name__)
 
 class AMCPruner(Pruner):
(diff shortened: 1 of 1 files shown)


nni/algorithms/compression/pytorch/pruning/amc/amc_pruner.py
```diff
@@ -207,7 +205,10 @@ def __init__(
 
 
     def compress(self):
+        _previous_cudnn_deterministic = torch.backends.cudnn.deterministic
+        torch.backends.cudnn.deterministic = True
```
zheng-ningxin, line 209: Why do we need to set the deterministic here, to reproduce the number in the paper(I guess)? Can we just remove the deterministic?

@chech-ms

J-shang: I check with @chicm-ms , and this line can be removed.

@mstczuo thanks for your finding, could you please remove this line directly?

mstczuo: OK. new PR here: #4161

PR #4148, opened 2021-09-04
[Quantization] support bn-folding for lsq

linbinskn:


nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -1041,12 +1052,46 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
 
         return calibration_config
 
+    def fold_bn(self, *inputs, wrapper):
```
chenbohua3, line 1055: What about moving this function to the superclass, so that QAT_quantizer and LsqQuantizer can both use it

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -1029,6 +1033,13 @@ def export_model(self, model_path, calibration_path=None, onnx_path=None, input_
                 abs_max_input = float(module.input_scale * module.input_qmax)
                 calibration_config[name]['tracked_min_input'] = -abs_max_input
                 calibration_config[name]['tracked_max_input'] = abs_max_input
+                if hasattr(module, BN_FOLD_TAG):
```
chenbohua3, line 1036: It seems that weight is not restored

linbinskn: Done.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Done.

chenbohua3:
LGTM

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -862,7 +828,7 @@ class LsqQuantizer(Quantizer):
        https://arxiv.org/pdf/1902.08153.pdf
     """
 
-    def __init__(self, model, config_list, optimizer):
+    def __init__(self, model, config_list, optimizer, dummy_input=None):
```
J-shang, line 831: It seems we can easily support `fold_bn` by adding `dummy_input` interface? As we discussed before, maybe we need to think about how to support other kinds of pattern fusion simulation in the future. And this pr is good to me.

J-shang (approved):

chenbohua3 (approved):

QuanluZhang (approved):

PR #4149, opened 2021-09-05
Function replacement implementation v1

zheng-ningxin:
In this PR, we implement the framework of the function replacement in compression and the replace function for the view function.


liuzhe-lz:
Deferred to next release.

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -378,6 +389,48 @@ def replace_compressed_modules(self):
         with torch.no_grad():
             for unique_name in self.auto_inferences:
                 self.replace_submodule(unique_name)
+            # the replaced model training state is set to True by default
+            self.bound_model.train(False)
```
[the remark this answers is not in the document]
zheng-ningxin: Yes, we already restore the original training state at the end of function `speedup_model`

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -531,13 +564,90 @@ def replace_layernorm(layernorm, masks):
...
+    # max_norm = embedding.max_norm
+    # norm_type = embedding.norm_type
+    # scale_grad_by_freq = embedding.scale_grad_by_freq
+    # new_embedding = torch.nn.Embedding(num_embeddings, n_remain_dim, padding_idx=padding_idx,
+    #                                    max_norm=max_norm, norm_type=norm_type, scale_grad_by_freq=scale_grad_by_freq)
+    # new_embedding.weight.data = torch.index_select(embedding.weight.data, 1, remained_dim)
+    # return new_embedding
+
+
+def replace_view(trace_model, cpp_node, masks):
```
J-shang, line 600: Do we already fix the view issue in our last discussion?

zheng-ningxin: Nice catch, I forgot this one. Thanks.

J-shang (approved):
Better to add a warning if replacing the view, because output may not align after replacement

Issue #4140, opened 2021-09-01
Let BOHB directly use ConfigSpace file


khituras:
**What would you like to be added**:
The possibility to use a native ConfigSpace configuration file to use with BOHB.

**Why is this needed**:
ConfigSpace is a very sophisticated framework to represent parameter configuration spaces. It supports conditional and forbidden configurations.
The NNI search space does not support conditional parameters directly and, for BOHB, does not support conditionals at all. Forbidden configurations are not supported in any way.

**Without this feature, how does current nni work**：
You use the possibilities that are given via the search space and its translation to ConfigSpace in BOHB which excludes conditionals and forbidden configurations.

**Components that may involve changes**:
* Search space definition
* BOHB advisor initialization

**Brief description of your proposal if any**:
I actually don't see an easy way to tackle this. Two possibilities would be
1. use ConfigSpace to begin with
2. extend NNI's search space definition to include conditionals and and forbidden configurations and translate them appropriately to ConfigSpace within BOHB.

I see that 1. would be a major overhaul that is most probably not desired. Also, ConfigSpace might have properties not desired by the NNI team.
2. Would seem to be quite some work.

For now, I went another, not quite clean, way in a personal NNI fork at https://github.com/khituras/nni:
There, I allow to specify a native ConfigSpace configuration file as a search space parameter like this:
```json
{
    "ConfigSpace": {"_type":"choice", "_value": ["bohb-parameters.pcs"]}
}
```
1. In `launcher.py`, I check if such a parameter is there and, if so, extend the given path with trial code directory to make it absolute
2. In `bohb_advisor.py`, I check again for the `ConfigSpace` parameter and, if found, load the given file directly.

I could create a PR from that. I didn't until now because I don't expect this solution to be favorable for your framework. But it was a quick fix that seems to work fine for me.

ultmaster:
Could you make the "bohb-parameter.pcs" an optional argument for BOHB advisor init function. When this pcs is available, it becomes the search space for tuner and directly ignores the search space json from NNI. I think this is more elegantly than defining a new search space format.

khituras:
Thank you for the response! I would like the optional argument. However, the init function is called in a generic way, as far as I understand the code, so I wouldn't know how to actually pass the .pcs file (or the ConfigSpace .json file; ConfigSpace understands multiple file formats).
Any hints appreciated, I would love to make a useful contribution here.

ultmaster:
Closing as PR has been merged.

PR #4153, opened 2021-09-06
Pass ConfigSpace definition file directly to BOHB

khituras:
This PR show a possible way to pass a path to a ConfigSpace file to BOHB so that it can read the file directly.
Thus, the translation from the NNI search space to the ConfigSpace format is circumvented.
Here, I introduced the special search space parameter named `ConfigSpace` that whose value is the path to the ConfigSpace file. `bohb_advisor.py` checks for this parameter and loads the file from the given path, if present.
The modification in `launcher.py` extends a relative ConfigSpace definition file path to an absolute path, if necessary.

I do not expect this PR to be accepted as-is. It is rather here for reference for the issue #4140 where I requested a similar functionality as a new feature in NNI.


ultmaster:
As I mentioned in the comments, my suggestion is that the pcs file should be passed as an init argument.

For how to support an optional argument, please see: https://nni.readthedocs.io/en/stable/Tutorial/InstallCustomizedAlgos.html#optional-create-a-validator-to-validate-classargs

khituras:
I did add an optional classArgs parameter named `config_space`. To make it optional, I had to add a method to `ClassArgsValidator` because there previously was no method to check for arbitrary strings. But we need this since we want to be able to pass a file name or a path. Thus, I added the `keyType()` method that just checks if the parameter has the right type, and added a check to the `bohb_advisor` for `config_space` being a string.

Now you can specify the ConfigSpace file like so (note the new `config_space` property at the end):

```yaml
advisor:
  builtinAdvisorName: BOHB
  classArgs:
    optimize_mode: maximize
    min_budget: 1
    max_budget: 9
    eta: 3
    min_points_in_model: 7
    top_n_percent: 15
    num_samples: 64
    random_fraction: 0.33
    bandwidth_factor: 3.0
    min_bandwidth: 0.001
    config_space: bohb-parameters.pcs
```

When doing this, the contents of the actual search space file are unused. The file can thus just contain `{}` to be valid JSON.

If this is acceptable to you, I could add the respective documentation to the BOHB docs.

nni/utils.py
```diff
@@ -305,3 +305,6 @@ def range(self, key, keyType, start, end):
             And(keyType, error='%s should be %s type!' % (key, keyType.__name__)),
             And(lambda n: start <= n <= end, error='%s should be in range of (%s, %s)!' % (key, start, end))
         )
+
+    def keyType(self, key, keyType):
```
ultmaster, line 309:
```python
def path(self, key, keyType):
   # check whether this path is valid and exists.
```

nni/tools/nnictl/launcher.py
```diff
@@ -421,6 +421,18 @@ def launch_experiment(args, experiment_config, mode, experiment_id, config_versi
         else:
             experiment_config['searchSpace'] = ''
 
+    # Only for BOHB: The search space may contain the path to another search space file in
+    # ConfigSpace format by defining the parameter 'ConfigSpace' as a 'choice' parameter
+    # with '_value' being the path to the configuration file. If the path is relative, we
+    # need to expand it
+    if experiment_config.get('advisor') and experiment_config['advisor'].get('name') == 'BOHB'\
```
ultmaster, line 428: Why do we still need this?

ultmaster (commented):
Design looks good. Might need to resolve another few detailed comments.

nni/utils.py
khituras: Will make this change.

nni/tools/nnictl/launcher.py
khituras: To make the relative path to the ConfigSpace file an absolute one. With this code, the configSpace file can just reside in the `trialCodeDirectory` and be referenced by name. Without it, one would have to define the complete path.
The comment is deprecated, though, will change it.

khituras: I have now moved that part to `launcher_utils.py` because this is where other paths are handled as well.

ultmaster (approved):
LGTM. Zhe please take a look. :)

nni/algorithms/hpo/bohb_advisor/bohb_advisor.py
```diff
@@ -468,48 +471,56 @@ def handle_update_search_space(self, data):
...
-                cs.add_hyperparameter(CSH.NormalFloatHyperparameter(
-                    var, mu=search_space[var]["_value"][1], sigma=search_space[var]["_value"][2],
-                    q=search_space[var]["_value"][3], log=True))
-            else:
-                raise ValueError(
-                    'unrecognized type in search_space, type is {}'.format(_type))
+        cs = None
+        logger.debug(f'Received data: {data}')
+        if self.config_space:
+            logger.debug(f'Got a ConfigSpace file path, parsing the search space directly from {self.config_space}.\
```
liuzhe-lz, line 477: Suggest raise log level to info.

liuzhe-lz (approved):

ultmaster:
Reopen to re-trigger pipeline.

QuanluZhang (approved):

PR #4163, opened 2021-09-07
[Quantization] support load_calibration_config

linbinskn:
This pr aims to help quantizer set quantization parameters by loading from a calibration_config which is exported by other quantizer or itself. It mainly has two usages:

-  Help quantize aware training quantizer set appropriate initial parameters so that the training process will be much more flexible and converges quickly. 
- What's more, it can also enable quantizer resume quantization model by loading parameters from config.


linbinskn:
@chenbohua3 Please help review this pr, thanks :)

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -546,6 +546,23 @@ def quantize_output(self, output, wrapper, **kwargs):
         out = self._dequantize(module, out)
         return out
 
+    def load_calibration_config(self, calibration_config):
```
chenbohua3, line 549: I think it is better to use 
```
modules_to_compress = self.get_modules_to_compress()
for layer, config in modules_to_compress:
   load calibration config...
```

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -995,6 +1012,22 @@ def quantize_input(self, inputs, wrapper, **kwargs):
...
+        for name, module in self.bound_model.named_modules():
+            if name not in calibration_config:
+                if hasattr(module, 'weight_bits') or hasattr(module, 'output_bits'):
+                    logger.warning(f"Can not find module {name}'s parameter in input config.")
+                continue
+            if hasattr(module, 'weight_bits'):
+                assert calibration_config[name]['weight_bits'] == module.weight_bits, f"weight bits of module {name} fail to match"
+            if hasattr(module, 'input_bits'):
+                assert calibration_config[name]['input_bits'] == module.input_bits, f"input bits of module {name} fail to match"
+                module.input_scale = float(calibration_config[name]['tracked_max_input'] / module.input_qmax)
```
chenbohua3, line 1025: `input_scale` is` torch.nn.Parameter` while the right operand is `float`

test/ut/sdk/test_compressor_torch.py
```diff
@@ -418,6 +418,39 @@ def test_torch_quantizer_export(self):
...
 
+    def test_quantizer_load_calibration_config(self):
+        configure_list = [{
+            'quant_types': ['weight', 'input'],
+            'quant_bits': {'weight': 8, 'input': 8},
+            'op_names': ['conv1', 'conv2']
+        }, {
+            'quant_types': ['output', 'weight', 'input'],
+            'quant_bits': {'output': 8, 'weight': 8, 'input': 8},
+            'op_names': ['fc1'],
```
chenbohua3, line 429: fc1 and fc2 can be combined

chenbohua3:
Add codes for enable this function？

test/ut/sdk/test_compressor_torch.py
```diff
@@ -418,6 +418,39 @@ def test_torch_quantizer_export(self):
...
+        }]
+        quantize_algorithm_set = [torch_quantizer.ObserverQuantizer, torch_quantizer.QAT_Quantizer, torch_quantizer.LsqQuantizer]
+        calibration_config = None
+        for quantize_algorithm in quantize_algorithm_set:
+            model = TorchModel().eval()
+            model.relu = torch.nn.ReLU()
+            optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.5)
+            quantizer = quantize_algorithm(model, configure_list, optimizer)
+            quantizer.compress()
+            if calibration_config is not None:
```
chenbohua3, line 443: I am not sure it is better to load config after instantiating the quantizer or during instantiate the quantizer

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: Done.

test/ut/sdk/test_compressor_torch.py
linbinskn: Have merged.

nni/algorithms/compression/pytorch/quantization/quantizers.py
linbinskn: exactly==, have modified.

test/ut/sdk/test_compressor_torch.py
linbinskn: Your concern is valid. Currently, I think we can keep this design because we need an `load_calibration_config` function anyhow and we can discuss this problem while consider about refactoring.

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -546,6 +546,28 @@ def quantize_output(self, output, wrapper, **kwargs):
         out = self._dequantize(module, out)
         return out
 
+    def load_calibration_config(self, calibration_config):
+        modules_to_compress = self.get_modules_to_compress()
+        for layer, config in modules_to_compress:
+            name, module = layer.name, layer.module
+            if name not in calibration_config:
+                if hasattr(module, 'weight_bits') or hasattr(module, 'output_bits'):
```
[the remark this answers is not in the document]
linbinskn: Because `input` quantize is always combined with `weight` quantize. Add `intput_bits` here is more clear and may avoid some corner cases. Have added.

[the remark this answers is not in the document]
J-shang: got it, what will happen if the user only sets 'input_bits' in config?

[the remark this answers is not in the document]
linbinskn: No thing will happen during simulation process. However, if we deploy it on inference framework, computation of this op will be converted to 32bit and warning will be raised since input tensor and weight tensor have different bit-width.

J-shang (approved):

chenbohua3 (approved):

nni/algorithms/compression/pytorch/quantization/quantizers.py
```diff
@@ -546,6 +546,28 @@ def quantize_output(self, output, wrapper, **kwargs):
         out = self._dequantize(module, out)
         return out
 
+    def load_calibration_config(self, calibration_config):
+        modules_to_compress = self.get_modules_to_compress()
+        for layer, config in modules_to_compress:
+            name, module = layer.name, layer.module
+            if name not in calibration_config:
```
chenbohua3, line 576: I think it is better to separate `weight` with `input`. For example, Given layer like `conv1-bn-relu-conv2`, The input of `conv2` is equal to `relu`'s output. And we should only set once for this tensor in the configure list. If it is set as the output of `relu`, we do not need to configure the quantization of the input of the `conv2`

linbinskn: I agree with the conclusion that we should separate weight and input. The reason why I only judge `weight_bits` and `output_bits` while ignore `input_bits` is that input quantization will always appear with weight or output quantization, it will not appear alone. Currently I add `input_bits` to avoid misunderstanding.

QuanluZhang (approved):

PR #4172, opened 2021-09-10
General serialization infrastructure

ultmaster:
This is generally ready for review.

Will migrate all the serialization used in NNI to the new infrastructure, and add more tests in a follow-up PR. Otherwise the PR will be too large for review.

### DIFF
--- a/nni/common/serializer.py
+++ b/nni/common/serializer.py
@@ -44,12 +44,15 @@ def trace(cls_or_func: Union[Type, Callable]) -> Union[Type, Callable]:
         return _trace_func(cls_or_func)
 
 
-def dump(obj: Any, fp: Optional[Any] = None, use_trace: bool = True, **json_tricks_kwargs) -> Union[str, bytes]:
+def dump(obj: Any, fp: Optional[Any] = None, use_trace: bool = True, pickle_size_limit: int = 4096,
+         **json_tricks_kwargs) -> Union[str, bytes]:
     """
     Parameters
     ----------
     fp : file handler or path
         File to write to. Keep it none if you want to dump a string.
+    pickle_size_limit : int
+        This is set to avoid too long serialization result. Set to -1 to disable size check.
     json_tricks_kwargs : dict
         Other keyword arguments passed to json tricks (backend), e.g., indent=2.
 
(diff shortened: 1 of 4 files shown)


nni/common/serializer.py
```diff
@@ -0,0 +1,369 @@
...
+        This is the one that should be used when you want to "mutate" a serializable object.
+        """
+        return SerializableObject(
+            self._get_nni_attr('symbol'),
+            self._get_nni_attr('args'),
+            self._get_nni_attr('kwargs')
+        )
+
+    def __json_encode__(self):
+        ret = {'__symbol__': _get_hybrid_cls_or_func_name(self._get_nni_attr('symbol'))}
```
QuanluZhang, line 61: if falling back to cloudpickle, args and kwargs are useless and cannot be mutated, right?

nni/common/serializer.py
```diff
@@ -0,0 +1,369 @@
...
+    """
+
+    def __init__(self, symbol: T, args: List[Any], kwargs: Dict[str, Any],
+                 _self_contained: bool = False):
+        # use dict to avoid conflicts with user's getattr and setattr
+        self.__dict__['_nni_symbol'] = symbol
+        self.__dict__['_nni_args'] = args
+        self.__dict__['_nni_kwargs'] = kwargs
+
+        self.__dict__['_nni_self_contained'] = _self_contained
```
QuanluZhang, line 29: what `_self_contained` mainly used for?

nni/common/serializer.py
QuanluZhang: does cloudpickle work across different OSes?

nni/common/serializer.py
```diff
@@ -0,0 +1,369 @@
...
+
+    # if we're being called as @trace()
+    if cls_or_func is None:
+        return wrap
+
+    # if we are called without parentheses
+    return wrap(cls_or_func)
+
+
+def dump(obj: Any, fp: Optional[Any] = None, use_trace: bool = True, pickle_size_limit: int = 4096,
```
QuanluZhang, line 135: what are `dump` and `load` used for?

nni/common/serializer.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def _get_hybrid_cls_or_func_name(cls_or_func: Any, pickle_size_limit: int = 4096) -> str:
+    try:
+        name = _get_cls_or_func_name(cls_or_func)
+        # import success, use a path format
+        return 'path:' + name
+    except ImportError:
+        b = cloudpickle.dumps(cls_or_func)
+        if len(b) > pickle_size_limit:
```
QuanluZhang, line 293: how to avoid serializing data in dataloader?

nni/common/serializer.py
ultmaster: Yes.

nni/common/serializer.py
ultmaster: Serializable class instances are self-contained, meaning that we don't have to use `.get()` to retrieve the original instance. It behaves naturally like an original one.

nni/common/serializer.py
ultmaster: Same semantics as dump and load in json-tricks.

nni/common/serializer.py
ultmaster: You can serialize a factory of dataloader.

QuanluZhang (approved):

PR #4173, opened 2021-09-10
Error code for Speedup Module

zheng-ningxin:
Add Error code for speedup module.

### DIFF
--- a/nni/compression/pytorch/speedup/error_code.py
+++ b/nni/compression/pytorch/speedup/error_code.py
@@ -1,0 +1,31 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+# Error Code of the speedup
+class SpeedupError(Exception):
+    def __init__(self, msg):
+        self.msg = msg
+
+    def __str__(self):
+        return str(self.msg)
+
+class EmptyLayerError(SpeedupError):
+    def __init__(self):
+        super(EmptyLayerError, self)("Pruning a Layer to empty is not legal")
+
+class ShapeMisMatchError(SpeedupError):
+    def __init__(self):
+        super(ShapeMisMatchError, self)("Shape mismatch!")
+
+class InputsNumberError(SpeedupError):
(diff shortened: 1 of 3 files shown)


nni/compression/pytorch/speedup/error_code.py
```diff
@@ -0,0 +1,31 @@
...
+class SpeedupError(Exception):
+    def __init__(self, msg):
+        self.msg = msg
+
+    def __str__(self):
+        return str(self.msg)
+
+class EmptyLayerError(SpeedupError):
+    def __init__(self):
+        super(EmptyLayerError, self).__init__("Pruning a Layer to empty is not legal")
```
J-shang, line 13: In this situation, should we suggest the user add this layer in `exclude`?

nni/compression/pytorch/speedup/error_code.py
```diff
@@ -0,0 +1,31 @@
...
+
+class ShapeMisMatchError(SpeedupError):
+    def __init__(self):
+        super(ShapeMisMatchError, self).__init__("Shape mismatch!")
```
J-shang, line 17: maybe we can also add layer name in the error message for debugging?

zheng-ningxin: Good suggestion, but I cannot get the `op_name` in the module replace functions under current interfaces. To keep the replacement function interface clean, I suggest to just report the error type now.

nni/compression/pytorch/speedup/error_code.py
zheng-ningxin: Not exactly the same,  there may be two reasons that a layer is pruned to empty: (1) it's sparsity ratio equals to 1.0 (2) it's output is useless after the mask propagation.

J-shang: got it

J-shang (approved):

QuanluZhang (approved):

PR #4178, opened 2021-09-13
[Model Compression] Add more Task Generator

J-shang:
- [x] Simulated Annealing
- [x] Lottery Ticket

### DIFF
--- a/nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
@@ -35,8 +35,7 @@ class PruningScheduler(BasePruningScheduler):
             Evaluate the pruned model and give a score.
             If evaluator is None, the best result refers to the latest result.
         reset_weight
-            Reset model weight to the begining before return in one step.
-            If set True, means the final compressed model weight is same as the corresponding position in the origin model.
+            If set True, the model weight will reset to the origin model weight at the end of each iteration step.
         """
         self.pruner = pruner
         self.task_generator = task_generator
--- a/nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
@@ -15,7 +15,8 @@ from .tools import TaskGenerator
 
 class PruningScheduler(BasePruningScheduler):
     def __init__(self, pruner: Pruner, task_generator: TaskGenerator, finetuner: Callable[[Module], None] = None,
-                 speed_up: bool = False, dummy_input: Tensor = None, evaluator: Optional[Callable[[Module], float]] = None):
+                 speed_up: bool = False, dummy_input: Tensor = None, evaluator: Optional[Callable[[Module], float]] = None,
+                 reset_weight: bool = False):
         """
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
```diff
@@ -33,26 +34,33 @@ def __init__(self, pruner: Pruner, task_generator: TaskGenerator, finetuner: Cal
         evaluator
             Evaluate the pruned model and give a score.
             If evaluator is None, the best result refers to the latest result.
+        reset_weight
```
QuanluZhang, line 37: reset_weight is True means every iteration uses original model weights, right?

J-shang: yes, finetune stage is placed at the beginning in each `step`, and the weight will be reset to the weight before finetuning then speedup.

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
```diff
@@ -89,12 +111,223 @@ def generate_config_list(self, target_sparsity: List[Dict], iteration: int, mode
...
+
+        scale = target_sparsity / (total_weights_pruned / total_weights)
+
+        # rescale the sparsity
+        sparsity = np.asarray(sparsity) * scale
+        return sparsity
+
+    def _sparsity_to_config_list(self, sparsity: List, config: Dict) -> List[Dict]:
+        sparsity = sorted(sparsity)
+        op_names = [k for k, _ in sorted(self.weights_numel.items(), key=lambda item: item[1]) if k in config['op_names']]
```
QuanluZhang, line 254: what is the meaning of this line?

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
```diff
@@ -89,12 +111,223 @@ def generate_config_list(self, target_sparsity: List[Dict], iteration: int, mode
...
+                pruned_weight_numel = 0
+                for name in op_names:
+                    remaining_weight_numel += self.weights_numel[name]
+                    if name in self.masked_rate:
+                        pruned_weight_numel += 1 / (1 / self.masked_rate[name] - 1) * self.weights_numel[name]
+                config['total_sparsity'] = max(0, sparsity - pruned_weight_numel / (pruned_weight_numel + remaining_weight_numel))
+
+    def _init_temp_config_list(self):
+        self._temp_config_list = []
+        self._temp_sparsity_list = []
```
QuanluZhang, line 208: what is `_temp_sparsity_list` used for?

nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
J-shang: sort the `op_name` in `config['op_names']` with the number of op weight elements.

QuanluZhang (approved):

nni/algorithms/compression/v2/pytorch/pruning/basic_scheduler.py
```diff
@@ -75,12 +83,71 @@ def pruning_one_step(self, task: Task) -> TaskResult:
...
+            score = None
 
         # clear model references
         self.pruner.clear_model_references()
 
         return TaskResult(task.task_id, compact_model, compact_model_masks, pruner_generated_masks, score)
 
+    def pruning_one_step_reset_weight(self, task: Task) -> TaskResult:
+        """
+        finetune -> generate masks -> reset weight -> speed up -> evaluate
```
zheng-ningxin, line 103: Why the `normal` way and the `reset` way have two different workflow?

J-shang: because in `lottery ticket` paper, the workflow is finetuning at first, then pruning and reset the weight to the origin, the final output can be seen as a structure and don't care about the weight values.

In other pruning paper, usually pruning at first then finetuning the pruned model to recover the acc.

zheng-ningxin (approved):

PR #4180, opened 2021-09-14
[Model Compression] transformer pruner

J-shang:
migrate transformer pruner to v2

The test is heavy, will not add it.


J-shang:
For the latest result of measurement, we found (QK pruning in the same position, V free pruning) is better than (QKV pruning in the same position) in most cases. Maybe this implementation should redesign.

liuzhe-lz:
Deferred to next release.

PR #4199, opened 2021-09-18
Fix aml stop environment logic

SparkSnail:


ts/nni_manager/config/aml/amlUtil.py
```diff
@@ -44,7 +44,16 @@
...
+            loop_count = 0
+            status = run.get_status()
+            # wait until the run is canceled
+            while status != 'Canceled':
+                if loop_count > 5:
+                    print('stop_result:failed')
+                    exit(0)
+                loop_count += 1
+                time.sleep(500)
+            print('stop_result:success')
```
J-shang, line 56: do we need `exit(0)` in this place?

J-shang (approved):

ts/nni_manager/config/aml/amlUtil.py
SparkSnail: sure, generally it will exit automatically when main process stop, add exit here to avoid corner case.

acured (approved):

Issue #4198, opened 2021-09-18
SPOS  hardware-aware  nas  multi_trial.py  don't work


GeorgeW9527:
**Describe the issue**:

**Environment**: ubutu
- NNI version:2.4
- Training service (local|remote|pai|aml|etc):
- Client OS:
- Server OS (for remote mode only):
- Python version:3.7
- PyTorch/TensorFlow version:PyTorch
- Is conda/virtualenv/venv used?:conda
- Is running in Docker?:

**Configuration**:
 - Experiment config (remember to remove secrets!):
 - Search space:

**Log message**:
 - nnimanager.log:
 - dispatcher.log:
 - nnictl stdout and stderr:
 
/opt/conda/envs/pytorch1.9/lib/python3.7/site-packages/deprecate/deprecation.py:115: LightningDeprecationWarning: The `Accuracy` was deprecated since v1.3.0 in favor of `torchmetrics.classification.accuracy.Accuracy`. It will be removed in v1.5.0.
  stream(template_mgs % msg_args)
Traceback (most recent call last):
  File "/opt/conda/envs/pytorch1.9/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/opt/conda/envs/pytorch1.9/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/notebook/code/personal/gzh/compression_toolkit/nni_v24/nni/nni/retiarii/trial_entry.py", line 25, in <module>
    engine.trial_execute_graph()
  File "/home/notebook/code/personal/gzh/compression_toolkit/nni_v24/nni/nni/retiarii/execution/base.py", line 128, in trial_execute_graph
    graph_data.evaluator._execute(model_cls)
  File "/home/notebook/code/personal/gzh/compression_toolkit/nni_v24/nni/nni/retiarii/evaluator/pytorch/lightning.py", line 100, in _execute
    return self.fit(model_cls)
  File "/home/notebook/code/personal/gzh/compression_toolkit/nni_v24/nni/nni/retiarii/evaluator/pytorch/lightning.py", line 128, in fit
    self.module.set_model(model)
  File "/home/notebook/code/personal/gzh/compression_toolkit/nni_v24/nni/nni/retiarii/evaluator/pytorch/lightning.py", line 37, in set_model
    self.model = model()
[...]

PR #4214, opened 2021-09-26
Fix: refine shape attribute

JiahangXu:
To resolve issue #4198


nni/retiarii/graph.py
```diff
@@ -532,9 +536,12 @@ def __init__(self, graph, node_id, name, operation, _internal=False):
         # maybe we should copy it here or make Operation class immutable, in next release
         self.operation: Operation = operation
         self.label: Optional[str] = None
+        self.input_shape: Optional[list] = None
+        self.output_shape: Optional[list] = None
```
ultmaster, line 540: Can we attach `input_shape` and `output_shape` onto `self.operation` as `self.operation.attrs['input_shape']`?

JiahangXu: I don't think `input_shape` and `output_shape` are operation's parameters as it could change along with different data size for the same operation. Actually this PR is to fix a bug in codegen. In codegen, all operation's parameter will be converted to the input argument when initializing the operation (https://github.com/microsoft/nni/blob/cb57730cd3f82b769f857c1e2f2f9a737084f7c5/nni/retiarii/operation.py#L136). This may cause the generated code has wrong input parameter.

nni/retiarii/graph.py
```diff
@@ -532,9 +536,12 @@ def __init__(self, graph, node_id, name, operation, _internal=False):
         # maybe we should copy it here or make Operation class immutable, in next release
         self.operation: Operation = operation
         self.label: Optional[str] = None
+        self.input_shape: Optional[list] = None
```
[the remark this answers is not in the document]
JiahangXu: Thanks. Already fixed.

PR #4184, opened 2021-09-15
Support 'op_partial_names' in config_list

Fiascolsy:
You now could use 'op_partial_names' instead of 'op_names' in config_list

PR #4225, opened 2021-09-28
Add annotations for 'op_partial_names'

Fiascolsy:
`op_partial_names` is an auxiliary field to match corresponding `op_names` in model. It will transfer to op_names after finishing its responsibility. This PR is related to #4184


Fiascolsy:
> It's ok for me, do we have an example of how to use it?

Not yet, but I could add one ASAP : )

PR #4243, opened 2021-10-11
Dev: add python name as Node attribute of graph_gen

JiahangXu:
Usage:
1. tracking the python module name in graph generate (for SPOS)
2. support weight inherit in the future


nni/retiarii/graph.py
```diff
@@ -535,13 +566,14 @@ def __init__(self, graph, node_id, name, operation, _internal=False):
         self.graph: Graph = graph
         self.id: int = node_id
         self.name: str = name or f'_generated_{node_id}'
+        self.python_name: Optional[str] = None
```
ultmaster, line 569: Suggest adding a comment line to state that " `python_name` is None by default. It should be set after initialization if it is needed."

ultmaster (approved):

liuzhe-lz (approved):

nni/retiarii/graph.py
JiahangXu: Thanks for the comment. Have fixed it.

PR #4252, opened 2021-10-14
[Quantization] fix QAT export param

linbinskn:
Since #4137 moves `weight_bits`, 'input_bits', 'output_bits' to `layer_quant_setting`, function `export_model` can't work correctly. This pr aims to fix this problem.


linbinskn:
@chenbohua3 @J-shang Please help review this pr, thanks~

PR #4265, opened 2021-10-21
Fix [v2 sparsity_allocator] the way tensor expands

Fiascolsy:
This enables `mode: 'global'` in `TaylorFOWeightPruner`.


J-shang:
please detach the Retiarii example, we base on branch `v2.5` now

PR #4275, opened 2021-10-26
[Draft] Create V2 model compression simple test examples

Fiascolsy:
Add a simple example for slim pruning.
There will be more pruning examples in both respective and integrate ways.

### DIFF
--- a/examples/model_compress/pruning/v2/slim_pruning_example.py
+++ b/examples/model_compress/pruning/v2/slim_pruning_example.py
@@ -3,7 +3,7 @@
 
 '''
 NNI example for supported slim pruning algorithms.
-In this example, we show the end-to-end pruning process: pre-training -> pruning -> fine-tuning.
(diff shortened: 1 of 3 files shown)


examples/model_compress/pruning/v2/slim_pruning_example.py
```diff
@@ -0,0 +1,129 @@
...
+    g_epoch = 0
+
+    # Start to prune and speedup
+    print('\n' + '=' * 50 + ' START TO PRUNE THE BEST ACCURACY PRETRAINED MODEL ' + '=' * 50)
+    config_list = [{
+            'total_sparsity': 0.5,
+            'op_types': ['BatchNorm2d'],
+    }]
+
+    optimizer, scheduler = optimizer_schedular_generator(model)
```
J-shang, line 109: `optimizer, _= optimizer_schedular_generator(model)` maybe better

J-shang:
LGTM

PR #4286, opened 2021-11-01
Fix bug where numpy array is used as a truth value

khituras:
The numpy error "The truth value of an array with more than one element is ambiguous" appears as soon as the `valid_indices` array has more than one entry.
The original BOHB code uses `len(valid_indices)` which has been removed in NNI for some reason.


liuzhe-lz:
Seems we modified it to pass pylint.
And since pylint does not complain for it now wo do should add it back.

liuzhe-lz (approved):

ultmaster:
It looks like we still have got many "truth value" complaints in unittests, and I have no idea where that comes from.

ultmaster (approved):

PR #4303, opened 2021-11-10
config refactor stage 0 - new base class

liuzhe-lz:
### Description ###

The first version of `ConfigBase` was too magic that nobody can understand how to use it.
This refactored version has 4 major improvements:
  1. The API is much more intuitive
  2. Support validating data type
  3. Handle base path transparently, so it no longer pollutes `__init__()`'s args
  4. Automatically unpack nested config, so most subclasses won't need to override `__init__()`

Note that this PR is merging to `dev-config` branch, not `master`.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###
Should be tested together with other configure refactor PRs.


nni/experiment/__init__.py
```diff
@@ -2,5 +2,5 @@
 # Licensed under the MIT license.
 
 from .config import *
-from .experiment import Experiment
-from .data import *
+#from .experiment import Experiment
```
SparkSnail, line 5: remove the line if it is not used anymore, or add annotation to annotate why it should keep.

J-shang (approved):

nni/experiment/__init__.py
liuzhe-lz: It will be added back in following stages. It's temporarily removed to run UT for the new module, or otherwise the test case will raise import error.

PR #4304, opened 2021-11-10
[retiarii] support visualize model space with the hpo chart on webui

QuanluZhang:
### Description ###
support the hyper-parameter chart for retiarii experiment.

### Checklist ###
  - [x] search space generated by retiarii engine is stored to db through nnimanager
  - [x] add a sub dict in trial configuration for visualization
  - [x] modify webui for correct visualization of the hyper parameter chart
  - [ ] test case
  - [x] doc: no doc should be updated

### How to test ###
run the multi-trial nas example and view the hyper-parameter chart


nni/runtime/protocol.py
```diff
@@ -27,6 +27,7 @@ class CommandType(Enum):
     SendTrialJobParameter = b'SP'
     NoMoreTrialJobs = b'NO'
     KillTrialJob = b'KI'
+    ReportSearchSpace = b'RS'
```
liuzhe-lz, line 30: I thought the visual search space could be directly embedded into original search space, rather than requiring a new channel.

QuanluZhang: Because we should add a new transmission direction, i.e., sending search space from tuner/advisor to nnimanager. Previously, search space is sent from nnimanager to tuner/advisor

nni/retiarii/execution/python.py
```diff
@@ -3,23 +3,25 @@
...
 from ..integration_api import receive_trial_parameters
 from ..utils import ContextStack, import_, get_importable_name
-from .base import BaseExecutionEngine
+from .base import BaseExecutionEngine, extract_visual_hp
 
 
 class PythonGraphData:
     def __init__(self, class_name: str, init_parameters: Dict[str, Any],
-                 mutation: Dict[str, Any], evaluator: Evaluator) -> None:
+                 mutation: Dict[str, Any], evaluator: Evaluator, visual_hp: dict = None) -> None:
```
ultmaster, line 11: `visual_hp` can be inferred from `mutation` I think.

nni/retiarii/execution/python.py
```diff
@@ -3,23 +3,25 @@
...
+        self.visual_hp = visual_hp
 
     def dump(self) -> dict:
         return {
             'class_name': self.class_name,
             'init_parameters': self.init_parameters,
             'mutation': self.mutation,
-            'evaluator': self.evaluator
+            'evaluator': self.evaluator,
+            '_visual_hyper_params_': self.visual_hp
```
ultmaster, line 24: I don't think `visual_hyper_params` is so special that it needs to be quoted with underline.

nni/retiarii/execution/base.py
```diff
@@ -10,20 +10,31 @@
...
+        if len(mut.samples) == 1:
+            hp_config[mut.mutator.label] = mut.samples[0]
+        else:
+            for i, sample in enumerate(mut.samples):
+                hp_config[f'{mut.mutator.label}_{i}'] = sample
+    return hp_config
 
 class BaseGraphData:
-    def __init__(self, model_script: str, evaluator: Evaluator) -> None:
+    def __init__(self, model_script: str, evaluator: Evaluator, visual_hp: dict = None) -> None:
```
ultmaster, line 28: Suggest renaming it into something like `mutation_summary`. It's not for visualization purpose only.

nni/retiarii/integration_api.py
```diff
@@ -34,6 +34,13 @@ def send_trial(parameters: dict, placement_constraint=None) -> int:
     """
     return get_advisor().send_trial(parameters, placement_constraint)
 
+def report_search_space(search_space: dict) -> None:
+    """
+    Execution engine uses this API to report the search space extracted by
+    strategy to nnimanager which stores the search space into db,
+    so that webui can access the search space like HPO
```
ultmaster, line 41: Missing period.

nni/retiarii/strategy/base.py
```diff
@@ -4,12 +4,38 @@
...
+        search_space = {}
+        auto_label_seq = 0
+        new_model = base_model
+        for mutator in applied_mutators:
+            recorded_candidates, new_model = mutator.dry_run(new_model)
+            # TODO: check label is never None
+            if mutator.label is not None:
+                label = mutator.label
+            else:
+                label = f'auto_label_{auto_label_seq}'
```
ultmaster, line 29: Where does `auto_label` come from? I think this logic is already done in `dry_run_for_search_space`, and `self` is never used in this function. Suggest refact it into a util.

nni/retiarii/execution/python.py
liuzhe-lz: I think that depends on whether `_visual_hyper_params_` is designed to be Retiarii only.
In HPO scenario all other stuffs are meant for tuning algorithm, but this field is for web UI.

nni/runtime/protocol.py
liuzhe-lz: We already have update search space in REST API. See `Experiment.update_search_space()` for an example implementation.
I think Retiarii should report search space use this method.

nni/retiarii/execution/base.py
QuanluZhang: good point

nni/retiarii/execution/python.py
QuanluZhang: some other stuffs are meant for trial. in hpo, those stuffs are visualizable, but in retiarii they are not. that is why we need a new field. "_visual_hyper_params_" is special, because it delivers redundant data

nni/runtime/protocol.py
QuanluZhang: update search space is from nnimanager to advisor, not advisor to nnimanager

nni/retiarii/strategy/base.py
QuanluZhang: moved to util. `dry_run_for_search_space` cannot be used as the generated search space should follow our search space format

nni/runtime/protocol.py
liuzhe-lz: `Experiment.update_search_space()` is from experiment launcher (`RetiariiExperiment`) to NNI manager, and then to dispatcher.

QuanluZhang: updated to using this function

nni/retiarii/execution/base.py
```diff
@@ -8,27 +8,29 @@
...
 
 _logger = logging.getLogger(__name__)
 
-
 class BaseGraphData:
-    def __init__(self, model_script: str, evaluator: Evaluator) -> None:
+    def __init__(self, model_script: str, evaluator: Evaluator, mutation_summary: dict) -> None:
         self.model_script = model_script
         self.evaluator = evaluator
+        self.mutation_summary = mutation_summary
```
cruiseliu, line 22: This introduces a new term so it needs doc to describe what is "mutation summary".

nni/retiarii/strategy/base.py
```diff
@@ -6,10 +6,19 @@
 
 from ..graph import Model
 from ..mutator import Mutator
+from .utils import dry_run_for_formated_search_space
 
 
 class BaseStrategy(abc.ABC):
 
+    @classmethod
+    def get_model_space(self, base_model: Model, applied_mutators: List[Mutator]) -> None:
```
matluster, line 15: This looks really weird. BaseStrategy is an interface, yet you put concrete implementations into it.

nni/retiarii/experiment/pytorch.py
```diff
@@ -193,6 +193,8 @@ def _start_strategy(self):
         )
 
         _logger.info('Start strategy...')
+        search_space = BaseStrategy.get_model_space(base_model_ir, self.applied_mutators)
```
matluster, line 196: Should call `dry_run_for_formated_search_space` here. Not sure whether every search space can be successfully formatted though.

ts/webui/src/static/model/trial.ts
```diff
@@ -27,13 +28,14 @@ import {
  * @returns Parsed structured parameters and unexpected entries
  */
 function inferTrialParameters(
-    paramObj: object,
+    paramObj: RetiariiParameter,
```
ultmaster, line 31: Why is this `RetiariiParameter`? How about normal trials?

QuanluZhang: @Lijiaoa

ultmaster: I think the changes above are not needed. In L35, we can put:

```typescript
'mutation_summary' in paramObj ? (paramObj as any).mutation_summary : paramObj;
```

nni/retiarii/execution/base.py
QuanluZhang: added

ts/webui/src/static/model/trial.ts
Lijiaoa: @ultmaster @QuanluZhang  also fix this in PR #4314

nni/retiarii/execution/base.py
```diff
@@ -8,27 +8,39 @@
...
 from ..graph import Model, ModelStatus, MetricData, Evaluator
 from ..integration_api import send_trial, receive_trial_parameters, get_advisor
 
 _logger = logging.getLogger(__name__)
 
-
 class BaseGraphData:
-    def __init__(self, model_script: str, evaluator: Evaluator) -> None:
+    """
+    Parameters
```
ultmaster, line 20: Attributes~~Parameters~~

ultmaster (approved):

liuzhe-lz (approved):

PR #4307, opened 2021-11-11
[Compression v2] bugfix & improvement

J-shang:
### Description ###
The criterion patch in SlimPruner V2 is invalid, fix it.
No need to clone in the data collector, if needed, do it outside.

### How to test ###
Run SlimPruner and find if the BatchNorm weight is sparsified during training.


nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -413,7 +413,7 @@ def criterion_patch(self, criterion: Callable[[Tensor, Tensor], Tensor]) -> Call
         def patched_criterion(input_tensor: Tensor, target: Tensor):
             sum_l1 = 0
             for _, wrapper in self.get_modules_wrapper().items():
-                sum_l1 += torch.norm(wrapper.module.weight.data, p=1)
+                sum_l1 += torch.norm(wrapper.module.weight, p=1)
```
Fiascolsy, line 416: Does this matter final result?
I ran the experiment with SlimPruner in two ways (use `.data` or not) and found almost no difference : (

QuanluZhang: what is the difference between "weight.data" and "weight"?

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -141,7 +141,7 @@ def generate_sparsity(self, metrics: Dict) -> Dict[str, Dict[str, Tensor]]:
             group_mask = torch.cat(group_mask, dim=0)
 
             for name, metric in group_metric_dict.items():
-                metric = (metric - metric.min()) * group_mask
+                metric = metric * group_mask
```
QuanluZhang, line 144: why the previous version minus `metric.min()`?

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: image

`weight` has `grad_fn` but `weight.data` hasn't.

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
J-shang: Because we want to support the metric with negative values before, but I found this is not a good way, so remove it. We don't have a negative metric except the `movement pruning score` right now.

QuanluZhang (approved):

Fiascolsy (approved):

liuzhe-lz (approved):

PR #4313, opened 2021-11-15
config refactor stage 1 - ExperimentConfig and LocalConfig

liuzhe-lz:
### Description ###

Port `ExperimentConfig` to the new base class.

Now the canonicalize and validation logic is shorter, and should be far more readable.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###

Should be tested together with other configure refactor PRs.


nni/experiment/config/exp_config.py
```diff
@@ -0,0 +1,136 @@
...
+        super()._validate_canonical()
+
+        space_cnt = (self.search_space is not None) + (self.search_space_file is not None)
+        if self.use_annotation and space_cnt != 0:
+            raise ValueError('ExperimentConfig: search space must not be set when annotation is enabled')
+        if not self.use_annotation and space_cnt != 1:
+            raise ValueError('ExperimentConfig: search_space and search_space_file must be set one')
+
+        if self.search_space_file is not None:
+            self.search_space = yaml.safe_load(open(self.search_space_file))
```
SparkSnail, line 116: does the file need close()?

nni/experiment/config/utils/internal.py
```diff
@@ -108,17 +112,44 @@ def guess_list_config_type(objs, type_hint, _hint_list_item=False):
...
     subclasses = set(cls.__subclasses__())
     return subclasses.union(*[_all_subclasses(subclass) for subclass in subclasses])
+
+## training service factory ##
+
+def training_service_config_factory(platform):
+    from ..training_service import TrainingServiceConfig  # avoid circular import
+
+    # import all custom config classes so they can be found in TrainingServiceConfig.__subclasses__()
+    custom_ts_config_path = nni.runtime.config.get_config_file('training_services.json')
```
SparkSnail, line 146: What is content of `training_service.json` file? Does the file has path validation?

nni/experiment/config/training_services/local.py
```diff
@@ -0,0 +1,44 @@
...
+    platform: str = 'local'
+    use_active_gpu: Optional[bool] = None
+    max_trial_number_per_gpu: int = 1
+    gpu_indices: Union[List[int], int, str, None] = None
+    reuse_mode: bool = False
+
+    def _canonicalize(self, parents):
+        super()._canonicalize(parents)
+        self.gpu_indices = canonical_gpu_indices(self.gpu_indices)
+        self.nni_manager_ip = None
```
J-shang, line 36: do we have `nni_manager_ip` in the training service config before?

nni/experiment/config/exp_config.py
liuzhe-lz: Well I'm pretty sure it won't cause problem to open a file read-only and let GC to close it, but you are right it's better to use `with`.

nni/experiment/config/utils/internal.py
liuzhe-lz: This is a config file for custom training services, similar to the config file for registered algorithms.
The availability is guaranteed by the `config` module.

nni/experiment/config/training_services/local.py
liuzhe-lz: `nni_manager_ip`, `trial_command`, `trial_gpu_number`, etc were only in the top level config and not in training service configs.
But it was decided several releases ago, according to discussion, that training service config should contain all things needed by training service. And then these fields in top level config will work as "shortcuts" for convenience and for compatibility.

J-shang (approved):

SparkSnail (approved):

PR #4322, opened 2021-11-19
Evaluate current SPOS support

JiahangXu:
All changes:

- [x] Remove flop-related module in `network.py` and `blocks.py`, since all flop-related part depends on previous api of `nni.nas`;
- [x] Change dataloader in `supernet.py` and `scratch.py`.
- [x] Change "load state dict" in  `supernet.py` because the module have changed to adapt to nni graph engine, and the checkpoint file cannot be directly loaded to supernet. 
- [x] Delete `tester.py`, `tunner.py` and `config_search.yml` due to their out of date implementation; 
- [x] Change the name of `multi_trial.py` to `search.py`, change name of `scratch.py` to `evaluation.py` to be consistant with original repo (https://github.com/megvii-model/SinglePathOneShot/tree/master/src)
- [x] Adapt `strategy.RegularizedEvolution` in `multi_trial.py` and add `model_filter` param in evolution strategy.
- [x] Change api for `fixed_arch` in `scratch.py` and debug.
- [x] Update docs, including `spos.rst` and `hardwareAwareNAS.rst`.


examples/nas/oneshot/spos/architecture_final.json
```diff
@@ -1,22 +1,22 @@
...
-  "LayerChoice13": [true, false, false, false],
-  "LayerChoice14": [true, false, false, false],
-  "LayerChoice15": [true, false, false, false],
-  "LayerChoice16": [true, false, false, false],
-  "LayerChoice17": [false, false, false, true],
-  "LayerChoice18": [false, false, true, false],
-  "LayerChoice19": [false, false, false, true],
-  "LayerChoice20": [false, false, false, true]
-}
+    "_mutation_1": "2",
```
ultmaster, line 2: Suggest use customized labels (`label="my_own_label"`) as the default label format has changed in the latest version

nni/retiarii/strategy/evolution.py
```diff
@@ -127,7 +130,8 @@ def run(self, base_model, applied_mutators):
     def _submit_config(self, config, base_model, mutators):
         _logger.debug('Model submitted to running queue: %s', config)
         model = get_targeted_model(base_model, mutators, config)
-        submit_models(model)
+        if filter_model(self.filter, model):
```
ultmaster, line 133: If a model is ignored here, should `_running_models` have this model?

docs/en_US/NAS/SPOS.rst
```diff
@@ -106,6 +100,7 @@ Known Limitations
 
 
 * Block search only. Channel search is not supported yet.
+* Weight inheriting while search is not supported yet.
 * Only GPU version is provided here.
```
ultmaster, line 104: This is no longer a limitation.

docs/en_US/NAS/SPOS.rst
```diff
@@ -106,6 +100,7 @@ Known Limitations
 
 
 * Block search only. Channel search is not supported yet.
+* Weight inheriting while search is not supported yet.
```
ultmaster, line 103: In the search phase, training from scratch is required. Inheriting weights from super-net is not supported yet.

docs/en_US/NAS/SPOS.rst
```diff
@@ -61,15 +57,13 @@ NOTE: The data loading used in the official repo is `slightly different from usu
 Step 2. Evolution Search
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-Single Path One-Shot leverages evolution algorithm to search for the best architecture. The tester, which is responsible for testing the sampled architecture, recalculates all the batch norm for a subset of training images, and evaluates the architecture on the full validation set.
-
-In order to make the tuner aware of the flops limit and have the ability to calculate the flops, we created a new tuner called ``EvolutionWithFlops`` in ``tuner.py``\ , inheriting the tuner in SDK.
+Single Path One-Shot leverages evolution algorithm to search for the best architecture. The search module, which is responsible for testing the sampled architecture, recalculates all the batch norm for a subset of training images, and evaluates the architecture on the full validation set.
```
ultmaster, line 60: We should note here that this is what has been done in paper, not what we have implemented.

docs/en_US/NAS/SPOS.rst
JiahangXu: I think it would be "training from the pretrained checkpoint"?

examples/nas/oneshot/spos/architecture_final.json
JiahangXu: I have changed all label as `"default_xxx"`, as in our implementation of supernet the labels of layerchoice name like that.

ultmaster: Have you tested it? This doesn't sound correct to me.
Why is `LayerChoice(..., label='my_own_label')` not favored?

JiahangXu: Have updated the script with setting label as "LayerChoice1".

nni/retiarii/strategy/evolution.py
JiahangXu: I think `_running_models` should also include the filtered out model to avoid excessive running times beyond cycles and population size.

ultmaster: Then you are adding a model to `_running_models` that will never be popped out.

ultmaster: I think this is correct for `on_failure="ignore"`.

For `on_failure="worst"`, we can just mark the model as "a worst model", so that the algo can somehow learn to avoid selecting such model.

ultmaster (approved):
Although there are unresolved comments, I think this PR is generally ready to merge (to catch up with code freeze). We can discuss further on those minor details later.

nni/retiarii/strategy/evolution.py
JiahangXu: I agree with you. I have modified the code in the new commit. When the model is filtered out and `on_failure="worst"`, the model will be appended to `_running_models` and marked as `Failed`, while will not be submitted. When the model is filtered out and `on_failure="ignore"`, the module will do nothing. Is this solution feasible?

examples/nas/oneshot/spos/network.py
```diff
@@ -25,18 +25,19 @@ def __init__(self, input_size=224, first_conv_channels=16, last_conv_channels=10
         super().__init__()
 
         assert input_size % 32 == 0
-        with open(os.path.join(os.path.dirname(__file__), op_flops_path), "rb") as fp:
-            self._op_flops_dict = pickle.load(fp)
+        # with open(os.path.join(os.path.dirname(__file__), op_flops_path), "rb") as fp:
+        #     self._op_flops_dict = pickle.load(fp)
```
QuanluZhang, line 29: why some lines are commented out rather than removing?

QuanluZhang (approved):

PR #4330, opened 2021-11-23
config refactor stage 3 - nnictl & Experiment

liuzhe-lz:
### Description ###
Rewrite nnictl create/resume/view to use `nni.Experiment`.

The old version is renamed to `legacy_launcher.py`. It is keeped for ADL training service.
The renaming causes +600/-600 lines of code, so the real PR size is about +400/-200.

Annotation support will be added in upcoming PR.

### Checklist ###
  - [ ] ~~test case~~ (no new public api)
  - [x] doc

### How to test ###
Should be tested together with other PRs.


nni/tools/nnictl/launcher.py
```diff
@@ -30,13 +30,20 @@ def create_experiment(args):
...
+            print(Fore.RED + 'ERROR: You are using legacy config file, please update it to latest format.' + Fore.RESET)
+            print(Fore.RED + 'Reference: https://nni.readthedocs.io/en/stable/reference/experiment_config.html' + Fore.RESET)
+            exit(1)
+        print(Fore.YELLOW + f'WARNING: You are using legacy config file, please update it to latest format:' + Fore.RESET)
+        print(Fore.YELLOW + '=' * 80 + Fore.RESET)
+        print(yaml.dump(v2_config).strip())
+        print(Fore.YELLOW + '=' * 80 + Fore.RESET)
+        print(Fore.YELLOW + 'Reference: https://nni.readthedocs.io/en/stable/reference/experiment_config.html' + Fore.RESET)
         config = ExperimentConfig(**v2_config)
+        print(config)
```
SparkSnail, line 45: why print config?

liuzhe-lz: oops, debug code

nni/experiment/config/exp_config.py
```diff
@@ -129,7 +127,7 @@ def _validate_canonical(self):
         space_cnt = (self.search_space is not None) + (self.search_space_file is not None)
         if self.use_annotation and space_cnt != 0:
             raise ValueError('ExperimentConfig: search space must not be set when annotation is enabled')
-        if not self.use_annotation and space_cnt != 1:
+        if not self.use_annotation and space_cnt < 1:
```
J-shang, line 130: why we allow `space_cnt=2`?

liuzhe-lz: When user specifies `seachSpaceFile`, NNI converts it to `searchSpace` (required by NNI manager). In previous version NNI removes `searchSpaceFile` during conversion, and in this implementation it gets kept. When resuming an experiment, the search space has already been converted and the check will fail.
So there are 3 approaches:
  1. don't store `searchSpaceFile` (user can't view the path in web UI)
  2. don't check it (use can set both `searchSpace` and `searchSpaceFile` and the latter is silently ignored)
  3. don't validate config for resumed experiment (the logic is more complex and might introduce potential problem)
I think it is very unlikely for users to set both `searchSpace` and `searchSpaceFile`, and therefore prefer the second approach.

J-shang: got it

J-shang (approved):

SparkSnail (approved):

PR #4337, opened 2021-11-26
Serialization infrastructure V2

ultmaster:
### Description ###

This PR refines the serialization infrastructure as introduced in the previous release, and integrates it with Retiarii.

Adopt `nni.load` and `nni.dump` in most cases where serialization is used in NNI.

I think this PR is big enough. I'll leave the work of `evaluator_fn` (to support evaluator that doesn't need to wrap with serialize) in a future PR.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###

See `test/ut/sdk/test_serializer.py`. Needs to test NAS multi-trial examples and integration tests.


J-shang (approved):
LGTM

liuzhe-lz (approved):

nni/common/serializer.py
```diff
@@ -98,16 +205,16 @@ def trace(cls_or_func: T = None, *, kw_only: bool = True) -> Union[T, Serializab
...
-    To get the original object, you should use ``obj.get()`` to retrieve. The retrieved object can be used
-    like the original one, but there are still subtle differences in implementation.
-
-    Note that when using the result from a trace in another trace-able function/class, ``.get()`` is automatically
-    called, so that you don't have to worry about type-converting.
+    One exception is that if your function returns None, it will return an empty SerializableObject instead,
+    which should raise your attention when you want to check whether the None ``is None``.
 
-    Also it records extra information about where this object comes from. That's why it's called "trace".
+    When parameters of functions are received, it is first stored, and then a shallow copy will be passed to inner function.
```
QuanluZhang, line 211: what is the meaning of "inner function"? is there example for it?

nni/common/serializer.py
```diff
@@ -308,40 +511,47 @@ def _import_cls_or_func_from_hybrid_name(s: str) -> Any:
...
 
     # do nothing to instance that is not a serializable object and do not use trace
-    if not use_trace or not isinstance(obj, SerializableObject):
+    if not use_trace or not is_traceable(obj):
         return obj
 
-    return obj.__json_encode__()
+    if isinstance(obj.trace_symbol, property):
+        # commonly made mistake when users forget to call the traced function/class.
+        warnings.warn(f'The symbol of {obj} is found to be a property. Did you forget to create the instance with ``xx(...)``?')
```
QuanluZhang, line 539: this is only a warning, not an error?

nni/retiarii/utils.py
```diff
@@ -107,5 +111,46 @@ def top(cls, key: str) -> Any:
...
+    """
+
+    def __init__(self, key: str = _DEFAULT_MODEL_NAMESPACE):
+        # for example, key: "model_wrapper"
+        self.key = key
+
+    def __enter__(self):
+        # For example, currently the top of stack is [1, 2, 2], and [1, 2, 2, 3] is used,
+        # the next thing up is [1, 2, 2, 4].
+        # `reset_uid` to count from zero for "model_wrapper_1_2_2_4"
```
QuanluZhang, line 126: what is the relation between this label and user provided label?

QuanluZhang:
@ultmaster when should users use `nni.trace`, in what cases? do we have example or doc for the usage, not how to use it but when to use it?

nni/retiarii/utils.py
matluster: If user provides a label, it will always override this label.

matluster:
> @ultmaster when should users use `nni.trace`, in what cases? do we have example or doc for the usage, not how to use it but when to use it?

Users should never use `nni.trace`, except when they are explicitly instructed to use it (e.g., in NAS and compression cases).

I don't know about compression. For NAS, the planned user interfaces that implicitly use `nni.trace` are:

1. `@basic_unit` and `@model_wrapper` -- already public API.
2. `@serialize` and @serialize_cls` -- will stay in the docs for another release. Users will get a deprecation warning that such API will soon be replaced by `@nni.trace` though. The usage of `@nni.trace` should be exactly same as `@serialize` and `@serialize_cls`.

So, how is `@nni.trace` different from `@serialize`?

1. `@nni.trace` is much more complicated and should be a lot more robust than `@serialize`.
2. Thanks to the new serializer (dump and load), we can now ask the user to write `evaluator_fn` instead of recursively initializing all the objects used in evaluator with `@nni.trace`. This is not implemented yet, but should be the next work item coming up after this PR gets merged.

nni/common/serializer.py
matluster: We can't be sure. In rare cases, they might actually want this as expected behavior.

nni/common/serializer.py
matluster: The inner function means the wrapped function/class. Will update to clarify this.

J-shang:
This PR will be merged, and @QuanluZhang comments could be fixed in the following PR if they are required.

PR #4361, opened 2021-12-02
[retiarii] use fixed seeds for unit test of CGO engine

hzhua:
The validation accuracy for the unit test may be lower than 0.8 with some initialized weights. Use fixed seeds instead.


ultmaster (approved):
In Pytroch-lightning they have an API called `seed_everything`

test/ut/retiarii/test_cgo_engine.py
```diff
@@ -5,6 +5,9 @@
 import time
 import torch
 import torch.nn as nn
+import numpy as np
+import random
```
liuzhe-lz, line 9: Is this required?

hzhua: removed.

liuzhe-lz (approved):

PR #4365, opened 2021-12-03
V2 Model Compression experiment result

Fiascolsy:
### Description ###
Some simple tests to measure the performance of v2 pruner.

### Checklist ###
  - [ ] test case
  - [ ] doc


docs/en_US/Compression/v2_pruning_algo.rst
```diff
@@ -55,6 +55,29 @@ User configuration for Level Pruner
 
 .. autoclass:: nni.algorithms.compression.v2.pytorch.pruning.LevelPruner
 
+Performance Test
```
QuanluZhang, line 58: all these performance numbers are obtained from our example code?

Fiascolsy: Yes, the seed was set,  and some hyperparameters may change.

QuanluZhang: is it better to also provide the hyper-parameters that you use?

docs/en_US/Compression/v2_pruning_algo.rst
```diff
@@ -55,6 +55,29 @@ User configuration for Level Pruner
...
+     - Sparsity
+   * - VGG-16
+     - 93.86%
+     - 14.98M
+     - 313.46M
+     - 
+   * - Pruned VGG-16
+     - 93.74%
+     - 14.98M
+     - 313.46M
```
QuanluZhang, line 78: why parameters and flops do not change in the pruned model?

docs/en_US/Compression/v2_pruning_algo.rst
```diff
@@ -200,6 +309,29 @@ User configuration for Activation APoZ Rank Pruner
...
+     - FLOPs
+     - Sparsity
+   * - VGG-16
+     - 93.86%
+     - 14.98M
+     - 313.46M
+     - 
+   * - Pruned VGG-16
+     - 87.24%
+     - 0.64M
```
QuanluZhang, line 331: why parameters are reduce so much with 80% sparsity?

docs/en_US/Compression/v2_pruning_algo.rst
```diff
@@ -401,6 +579,29 @@ User configuration for AGP Pruner
...
+     - Sparsity
+   * - VGG-16
+     - 93.86%
+     - 14.98M
+     - 313.46M
+     - 
+   * - Pruned VGG-16
+     - 76.94%
+     - 
+     -
```
QuanluZhang, line 602: why the two numbers are missing?

docs/en_US/Compression/v2_pruning_algo.rst
```diff
@@ -478,6 +702,29 @@ User configuration for Simulated Annealing Pruner
...
+     - Sparsity
+   * - VGG-16
+     - 93.86%
+     - 14.98M
+     - 313.46M
+     - 
+   * - Pruned VGG-16
+     - 33.6% (wo FT)
+     - 
+     -
```
QuanluZhang, line 725: what is the meaning of "wo FT", and there are also two missing numbers

docs/en_US/Compression/v2_pruning_algo.rst
Fiascolsy: Now this has been provided.

docs/en_US/Compression/v2_pruning_algo.rst
Fiascolsy: This is fine-grained algorithm, it has no speedup procedure.

docs/en_US/Compression/v2_pruning_algo.rst
Fiascolsy: The pruning behavior may differ according to various default `mode` setting. It has three types: 'normal', 'global' and 'dependency_aware'. In the intermediate `config_list` we can see it does prune nearly 0.8 in each layer.

docs/en_US/Compression/v2_pruning_algo.rst
Fiascolsy: Because the result was not saved at that time, now it has been added.

docs/en_US/Compression/v2_pruning_algo.rst
Fiascolsy: At first I want the result 'without finetune', but this time it has been normal.
Just like the workflow: `prune -> speedup -> finetune`

QuanluZhang (approved):

J-shang (approved):

J-shang:
These results have already been updated. But the result has little difference, I think maybe we need a more significant benchmark...

PR #4367, opened 2021-12-06
Automatically update builtin algorithm

liuzhe-lz:
### Description ###

This fixes a bug introduced by new Random & TPE.

NNI gets metainfo of built-in algorithms from config file `registered_algorithms.yml`. This file has been modified in this release to let Random/TPE point to their new class.
If the user has installed a previous version of NNI before, the config file will not get updated and Random/TPE will point to hyperopt class.

### Checklist ###

  - [x] test case
  - ~~doc~~ (no need)

### How to test ###

Upgrade from old version:
1. remove `~/.config/nni`
2. install an old version
3. register a custom tuner
4. upgrade to latest version
5. test Random/TPE with new args (e.g. seed)
6. test custom tuner

Fresh install:
1. remove `~/.config/nni`
2. install latest version
3. register custom tuner
4. test Random/TPE and custom tuner


nni/runtime/config.py
```diff
@@ -1,13 +1,18 @@
...
 import os
 from pathlib import Path
 import shutil
 import sys
 
 import nni
 
+# the version when default config got modified
+# don't need to update if default config does not change
+latest_config_version = '2.6'
```
ultmaster, line 14: Suggest reading `registered_algorithms.yml` and putting its md5sum here.

liuzhe-lz: `registered_algorithms.yml` can be modified by user with `nnictl algo register`.
This is only relevant to `default_config/registered_algorithms.yml` committed into our code repo.

ultmaster: I meant, why can't we automatically calculate and check the checksum here instead of a string `2.6`?

liuzhe-lz: Well...... The config can be updated with `nnictl algo register`, so we do not except `~/.config/nni/registered_algorithms.yml` to be identical to `nni/runtime/default_configs/registered_algorithms.yml`.

ultmaster: We store a checksum in `nni/runtime.config.py`. When users access their own `registered_algorithms.yml`, we check the checksum which was stored as a field in the yaml,

1. If a match -- no problem
2. Otherwise -- we update his/her yaml, as well as the checksum in the yaml.

As for updating from more than one old formats, seemingly we might as well use `nni.__version__` as checksum. But my concern is that `nni.__version__` might be unreliable in develop mode. Why don't we keep another field called schema version?

I'm against writing the version **manually** here, because people **forget**.

QuanluZhang: I agree, manually writing version number here is not a good practice

Issue #4298, opened 2021-11-10
Chinese doc discussion


liuzhe-lz:
The table below shows monthly viewed times for top 30 doc pages, ordered by readthedocs catalogue.

These pages should be translated in this release: (15 in total)
- [x] Readme (index) @scarlett2018 
- [ ] Installation @SparkSnail 
- [ ] 　　Linux & Mac @SparkSnail 
- [ ] 　　Windows @SparkSnail 
- [x] QuickStart @JiahangXu 
- [ ] Auto (Hyper-parameter) Tuning @liuzhe-lz 
- [ ] 　　Write Trial @QuanluZhang 
- [ ] 　　Tuners (merge with Overview) @liuzhe-lz 
- [x] 　　WebUI @Lijiaoa
- [x] Neural Architecture Search @ultmaster 
- [x] 　　Quick Start @ultmaster 
- [x] Model Compression @Fiascolsy 
- [ ] 　　Overview @J-shang
- [ ] 　　Quick Start @linbinskn

liuzhe-lz:
| | English | Chinese |
| - | - | - |
| [**(index)**](https://nni.readthedocs.io/en/latest/index.html) | 1904 | 766 |
| [**Overview**](https://nni.readthedocs.io/en/latest/Overview.html) | 783 | 571 |
| [**Installation**](https://nni.readthedocs.io/en/latest/installation.html) | 661 | 460 |
| 　　[**Linux & Mac**](https://nni.readthedocs.io/en/latest/Tutorial/InstallationLinux.html) | 436 | 307 |
| 　　[**Windows**](https://nni.readthedocs.io/en/latest/Tutorial/InstallationWin.html) | 297 | |
| 　　Use Docker | | |
| [**QuickStart**](https://nni.readthedocs.io/en/latest/Tutorial/QuickStart.html) | 965 | 862 |
| [**Auto (Hyper-parameter) Tuning**](https://nni.readthedocs.io/en/latest/hyperparameter_tune.html) | 800 | 772 |
|　　 [**Write Trial**](https://nni.readthedocs.io/en/latest/TrialExample/Trials.html) | 403 | 473 |
|　　 [**Tuners**](https://nni.readthedocs.io/en/latest/builtin_tuner.html) | | 266 |
|　　　　 [**Overview**](https://nni.readthedocs.io/en/latest/Tuner/BuiltinTuner.html) | 538 | 383 |
|　　　　 ... | | |
|　　 Assessors... | | |
[...]

PR #4376, opened 2021-12-08
AutoFormer2NNI

penghouwen:
## Description

AutoFormer is a new one-shot architecture search framework dedicated to vision transformer search. It entangles the weights of different vision transformer blocks in the same layers during supernet training. Benefiting from the strategy, the trained supernet allows thousands of subnets to be well-trained. Specifically, the performance of these subnets with weights inherited from the supernet is comparable to those retrained from scratch.

This work has been published in [ICCV2021](https://openaccess.thecvf.com/content/ICCV2021/papers/Chen_AutoFormer_Searching_Transformers_for_Visual_Recognition_ICCV_2021_paper.pdf).

## Checklist

- [x] test case
- [x] doc

## How to test

The Step.0 is necessary for test. For quick test, you could download our trained models and jump to Step. 3 directly. Otherwise, follow the complete steps to search the optimal model and test it. 

Notably, these steps are detailly described in the [documentation](https://nni.readthedocs.io/en/latest/NAS/Autoformer.html). Please check it if necessary.

### 0.  Prepare the data and build environments.

You need to first download the [ImageNet-2012](http://www.image-net.org/) to the folder `./data/imagenet` and move the validation set to the subfolder `./data/imagenet/val`. To move the validation set, you cloud use the following [script](https://raw.githubusercontent.com/soumith/imagenetloader.torch/master/valprep.sh).

The directory structure is the standard layout as following.

```
/path/to/imagenet/
  train/
    class1/
      img1.jpeg
    class2/
      img2.jpeg
  val/
    class1/
      img3.jpeg
    class/2
      img4.jpeg
```

In order to build the environments, run the following commands:

```
cd /PATH/TO/NNI/examples/nas/oneshot/autoformer
pip install -r ./requirements
```

### 1. Supernet Train

To train the supernet-T/S/B, we provided the corresponding supernet configuration files in `/experiments/supernet/`. For example, to train the supernet-B, you can run the following command. The default output path is `./`, you can specify the path with argument `--output`.

```
python -m torch.distributed.launch --nproc_per_node=8 --use_env supernet_train.py --data-path /PATH/TO/IMAGENT --gp --change_qk --relative_position --mode super --dist-eval --cfg ./experiments/supernet/supernet-B.yaml --epochs 500 --warmup-epochs 20 --output /OUTPUT_PATH --batch-size 128
```

### 2. Search

We run our evolution search on part of the ImageNet training dataset and use the validation set of ImageNet as the test set for fair comparison. To generate the subImagenet in `/PATH/TO/IMAGENET`, you could simply run:

```
python ./lib/subImageNet.py --data-path /PATH/TO/IMAGENT
```

After obtaining the subImageNet and training of the supernet. We could perform the evolution search using below command. Please remember to config the specific constraint in this evolution search using `--min-param-limits` and `--param-limits`: 

```
python -m torch.distributed.launch --nproc_per_node=8 --use_env evolution.py --data-path /PATH/TO/IMAGENT --gp --change_qk --relative_position --dist-eval --cfg ./experiments/supernet/supernet-B.yaml --resume /PATH/TO/CHECKPOINT --min-param-limits YOUR/CONFIG --param-limits YOUR/CONFIG --data-set EVO_IMNET
```

### 3. Test

To test the trained models, you need to put the model in `/PATH/TO/CHECKPOINT`. After that you could use the following command to test the model (Please change your config file and model checkpoint according to different models. Here we use the AutoFormer-B as an example).

```
python -m torch.distributed.launch --nproc_per_node=8 --use_env supernet_train.py --data-path /PATH/TO/IMAGENT --gp --change_qk --relative_position --mode retrain --dist-eval --cfg ./experiments/subnet/AutoFormer-B.yaml --resume /PATH/TO/CHECKPOINT --eval 
```


ultmaster:
Thanks for the contribution. It's nice to have one of the state-of-the-art algorithms in NNI and the instructions to use it are clear. Personally, there are several suggestions for discussion.

1. It's not clear "when" users should use Auto-Former. The code seems to have provided an instruction to reproduce paper results. If users write a customized search space, for their own task (e.g., text classification), will AutoFormer be able to empower such scenario?
2. AutoFormer is still implemented with `nni.nas` interface, which is deprecated now and we will migrate to `nni.retiarii`. I understand this is kind of confusing, but I will appreciate it if you have also integrate the trainer as a Retiarii one-shot trainer. The interface of the two frameworks look very similar, and it shouldn't be much effort.
3. The building blocks like `LinearSuper` and `LayerNormSuper` looks like an **implementation** of `LayerChoice` and `ValueChoice` (https://nni.readthedocs.io/en/stable/NAS/MutationPrimitives.html) induced by AutoFormerTrainer. Users who write the search space (as in examples/nas/oneshot/autoformer/model/supernet_transformer.py) should not involve such details. Instead, they should write pure APIs like `LayerChoice`, and the algorithm (AutoFormerTrainer) should figure out the rest, and replace the necessary modules with `XXXSuper`.
4. The evolution phase should be implemented with the [multi-trial framework](https://nni.readthedocs.io/en/stable/NAS/multi_trial_nas.html). This might be a more tricky case. We can discuss deeper once we have the SPOS example refactoring ready.

Another minor comment: Per discussion #4298, we no longer keep chinese docs other than the docs maintained in a specific whitelist. We thank you for the effort to translate the AutoFormer documentation, but to cut down the maintainence effort, we can't ready have this translation checked in. Sorry for the confusion and inconvenience.

penghouwen:
> Thanks for the contribution. It's nice to have one of the state-of-the-art algorithms in NNI and the instructions to use it are clear. Personally, there are several suggestions for discussion.
> 
> 1. It's not clear "when" users should use Auto-Former. The code seems to have provided an instruction to reproduce paper results. If users write a customized search space, for their own task (e.g., text classification), will AutoFormer be able to empower such scenario?
> 2. AutoFormer is still implemented with `nni.nas` interface, which is deprecated now and we will migrate to `nni.retiarii`. I understand this is kind of confusing, but I will appreciate it if you have also integrate the trainer as a Retiarii one-shot trainer. The interface of the two frameworks look very similar, and it shouldn't be much effort.
> 3. The building blocks like `LinearSuper` and `LayerNormSuper` looks like an **implementation** of `LayerChoice` and `ValueChoice` (https://nni.readthedocs.io/en/stable/NAS/MutationPrimitives.html) induced by AutoFormerTrainer. Users who write the search space (as in examples/nas/oneshot/autoformer/model/supernet_transformer.py) should not involve such details. Instead, they should write pure APIs like `LayerChoice`, and the algorithm (AutoFormerTrainer) should figure out the rest, and replace the necessary modules with `XXXSuper`.
> 4. The evolution phase should be implemented with the [multi-trial framework](https://nni.readthedocs.io/en/stable/NAS/multi_trial_nas.html). This might be a more tricky case. We can discuss deeper once we have the SPOS example refactoring ready.
> 
> Another minor comment: Per discussion #4298, we no longer keep chinese docs other than the docs maintained in a specific whitelist. We thank you for the effort to translate the AutoFormer documentation, but to cut down the maintainence effort, we can't ready have this translation checked in. Sorry for the confusion and inconvenience.

Thanks @ultmaster for your careful review and valuable suggestions. We are honored to be able to make continuous contributions to NNI community and very pleased to keep collaborations with your talents. The following are our responses 

1. Done. Autoformer is designed for vision transformer backbone search, such as image classification. We think it can be extended to other tasks and domains, such as natural language processing, however, our current version does not support this feature at current time. We can discuss it in the next steps. Users can collect their own vision datasets and organize them following our instructions. We provide a detailed introcudtion in the updated documentation(`docs/en_US/NAS/Autoformer.rst`).

2. Done. Per your suggestion and the new version of NNI, we have changed the interface to `nni.retrarii`. Thanks :)

3. Done. Actually, the `XXXSuper` is not an implementation of `LayerChoice` and `ValueChoice`. Unlike those works where candidates in the same layer are independent, such as SPOS and ProxylessNAS, AutoFormer entangles the weights of different candidates in the same layers. Only the candidate with the maximum number of parameters is built in each layer during training. Consequently, the mutation primitives `LayerChoice` and `ValueChoice` may not be suitable for Autoformer. Autoformer supports searching the ratio of MLP, the number of heads, the depth of network and the embedding dimension. Users can easily define their search space through the config file(`experiments\supernet\xx.yaml`). We have emphasized it in the updated documentation(`docs/en_US/NAS/Autoformer.rst`).

4. To be discussed. We can not implement the evolution phase with the multi-trial framework. It will be fixed immediately once the SPOS example is ready. We are terribly sorry for this inconvenience.

4. Let's discuss the implementation of evolution using the multi-trial framework after the SPOS example being ready. Pls give us a notice when it is ready. Thanks.

About docs:  We have provided an English updated version for the introduction of Autoformer. More details about dataset and search space are added in this new version. Thanks for your maintenance efforts :)

ultmaster:
I disagree with 3. My major concern is that AutoFormer is an end-to-end research solution, and most users can barely benefit from this example, because they either don't want its search space, or don't like its search strategy (let's admit that few users are actually interested in reproducing the paper). But having one another powerful research work that incorporates and uses NNI help strengthening our brand (and possibly attracts more researchers to try to use NNI).

To trade off, if we are to merge this PR, there are two options.

The first one: let's divide current examples into two categories: general examples and research works.
- *General examples* usually function as tutorials, where users can learn from the example about how to use NNI. Specifically, how to write their own search space, how to apply a specific strategy to their search space. The target users are general NNI users. Its role model is [pytorch tutorials](https://github.com/pytorch/examples). There is nothing SOTA here, but solutions targetting at diverse scenarios.
- *Research works* are usually end-to-end solutions that serve as a reproduction of some frontier research papers. The target users are usually researchers, and those who are eager to try out the frontier works. Users don't expect them to be plugged nad played. Notably, those works might not be stable and high-quality enough, but they are usually self-contained, meaning that they might have their own dependency, code-base. The only requirement is that they need to be related to NNI. Its role model is [google-research](https://github.com/google-research/google-research), which is a collections of paper code.

Admittedly, there are useful tricks/ideas in research works, such as search space, training tricks, which as time goes, more and more people are willing to try. Recently I've been working on `nashub`, which decouples the ideas in research works, and are implemented in the way that NNI works. For example, I have a dedicated folder called search space, which collects all the popular search space in the past few years, nicely refactored and reformatted with NNI's API.
I believe we can do the same to AutoFormer. We can put it into the category of *research works*, and then, we can break it down. For example, we can put its transformer search space into `nashub`, incorporate its weight entanglement trick into one-shot trainers. 

The second one: we prove here that AutoFormer works in general cases. To do this, we make AutoFormer trainer works with general LayerChoice and InputChoice APIs. We write unittests and integration tests. We test it on general cases. In the end, AutoFormer trainer is just another trainer like DARTS and ProxylessNAS. We can have an example (AutoFormer's search space in paper) to demonstrate its power, but it's not essential.

Judging from the current code, I believe the second option requires huge efforts (could even require helping hands from NNI dev). Therefore I recommend option 1.

Over the years, we've been adding new solutions (either papers or feature demos) **blindly**, unaware of users' needs. Actually it only causes confusions, as more and more examples only make users lost and not sure where to look at. It's time we make some changes.

@QuanluZhang @scarlett2018 @penghouwen We can discuss deeper about in a teams meeting.

PR #4383, opened 2021-12-13
update interim patch

J-shang:
### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/interim_patch.py
+++ b/interim_patch.py
@@ -1,3 +1,4 @@
+# fix mnist dataset download
 import os
 import torchvision.datasets.mnist as mnist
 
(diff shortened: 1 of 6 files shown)


ultmaster (commented):
Maybe this patch is no longer needed?

J-shang:
> Maybe this patch is no longer needed?

I will try if the `torchvision` patch is needed.
This pr is for `setuptools.distutils` has some issues in the latest version and will cause `torch.utils.tensorboard` break.

https://msrasrg.visualstudio.com/NNIOpenSource/_build/results?buildId=30308&view=logs&j=597b6a97-1552-58b1-ca74-8e543bcf8377&t=35dc6439-9fc2-5498-5aa9-345bd49c2643

interim_patch.py
```diff
@@ -0,0 +1,17 @@
+# fix setuptools.distutils import in torch
```
ultmaster, line 1: Can we move this file to `tests/vso_tools`?

J-shang: moved

ultmaster (approved):

liuzhe-lz (approved):

PR #4390, opened 2021-12-14
Serialization improvements (docs, examples, bug fixes)

ultmaster:
### Description ###

1. Documentation improvements: promote functional evaluator to recommended.
2. Update MNIST example to comply with documentation.
3. Disable pickle size limit (raise a warning instead) and fixed a bug
4. Add tests for multiprocessing. There has been reports that traced MNIST doesn't work well with DataLoader. ~Need to test with py3.6 on pipeline to verify.~ Need to fix for Windows and MacOS.

done -- I'll refresh the Chinese documentation when the review of this PR is almost done.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###

Run mnist/search.py


examples/nas/multi-trial/mnist/search.py
```diff
@@ -40,21 +49,76 @@ def forward(self, x):
         x = F.max_pool2d(self.conv2(x), 2)
         x = torch.flatten(self.dropout1(x), 1)
         x = self.fc2(self.dropout2(F.relu(self.fc1(x))))
-        output = F.log_softmax(x, dim=1)
```
liuzhe-lz, line 43: Then where is the softmax?

ultmaster: It's already in `CrossEntropyLoss`.

liuzhe-lz: OK, so the change is due to the migration from lightning to custom training loop.
Why do we abandon lightning? Seems CGO only supports lightning trainer.

ultmaster: We didn't abandon lightning.
I only updated the beginner example here so that users don't have to learn lightning first before using NAS. I hope this change will smoothen the learning curve.

For advanced features like CGO, lightning is still recommended (actually required).

docs/en_US/NAS/ModelEvaluators.rst
```diff
@@ -1,22 +1,45 @@
...
 
-Usage of Model Evaluator
-------------------------
+Customize Evaluator with Any Function
+-------------------------------------
 
-In multi-trial NAS, a sampled model should be able to be executed on a remote machine or a training platform (e.g., AzureML, OpenPAI). Thus, both the model and its model evaluator should be correctly serialized. To make NNI correctly serialize model evaluator, users should apply ``serialize`` on some of their functions and objects.
+The simplest way to customize a new evaluator is with functional APIs, which is very easy when training code is already available. Users only need to write a fit function that wraps everything, which usually includes training, validating and testing of a single model.. This function takes one positional arguments (``model_cls``) and possible keyword arguments. The keyword arguments (other than ``model_cls``) are fed to FunctionEvaluator as its initialization parameters (note that they will be `serialized <./Serialization.rst>`__). In this way, users get everything under their control, but expose less information to the framework and thus fewer opportunities for possible optimization. An example is as belows:
```
QuanluZhang, line 9: .. -> .

QuanluZhang: what is the meaning of "fewer opportunities"?

docs/en_US/NAS/ModelEvaluators.rst
```diff
@@ -1,22 +1,45 @@
...
+        acc = test(model, dataloader)
+        nni.report_final_result(acc)
+
+    # The dataloader will be serialized, thus ``nni.trace`` is needed here.
+    # See serialization tutorial for more details.
+    evaluator = FunctionalEvaluator(fit, dataloader=nni.trace(DataLoader)(foo, bar))
+    experiment = RetiariiExperiment(base_model, evaluator, mutators, strategy)
 
-`serialize <./ApiReference.rst#utilities>`__ enables re-instantiation of model evaluator in another process or machine. It is implemented by recording the initialization parameters of user instantiated evaluator.
+.. tip:: When using customized evaluators, if you want to visualize models, you need to export your model and save it into ``$NNI_OUTPUT_DIR/model.onnx`` in your evaluator.
```
QuanluZhang, line 27: would be better to add more description or link on "visualize models"

docs/en_US/NAS/ModelEvaluators.rst
```diff
@@ -27,17 +50,85 @@ Below is an example, ``transforms.Compose``, ``transforms.Normalize``, and ``MNI
   transform = serialize(transforms.Compose, [serialize(transforms.ToTensor()), serialize(transforms.Normalize, (0.1307,), (0.3081,))])
   train_dataset = serialize(MNIST, root='data/mnist', train=True, download=True, transform=transform)
   test_dataset = serialize(MNIST, root='data/mnist', train=False, download=True, transform=transform)
```
QuanluZhang, line 52: it is a little confusion that we use both `nni.trace` and `serialize`, what is the difference between them?

docs/en_US/NAS/ModelEvaluators.rst
```diff
@@ -27,17 +50,85 @@ Below is an example, ``transforms.Compose``, ``transforms.Normalize``, and ``MNI
...
+.. code-block:: python
+
+    import nni.retiarii.evaluator.pytorch.lightning as pl
+    from nni.retiarii.experiment.pytorch import RetiariiExperiment
+
+    lightning = pl.Lightning(AutoEncoder(),
+                             pl.Trainer(max_epochs=10),
+                             train_dataloader=pl.DataLoader(train_dataset, batch_size=100),
+                             val_dataloaders=pl.DataLoader(test_dataset, batch_size=100))
+    experiment = RetiariiExperiment(base_model, lightning, mutators, strategy)
```
QuanluZhang, line 134: better to add a link to a complete example code, so that users can better understand what the end2end code looks like

docs/en_US/NAS/QuickStart.rst
```diff
@@ -52,6 +52,8 @@ Below is a very simple example of defining a base model.
       output = F.log_softmax(x, dim=1)
       return output
 
+.. tip:: Always keep in mind that you should use ``import nni.retiarii.nn.pytorch as nn`` and ``@model_wrapper``. Many mistakes are a result of forgetting one of those. Also, please use ``torch.nn`` for submodules of ``nn.init``, e.g., ``torch.nn.init`` instead of ``nn.init``.
```
QuanluZhang, line 55: what is the difference between `@module_wrapper` and `@nni.trace`?

docs/en_US/NAS/Serialization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Serialization is implemented as a combination of `json-tricks <https://json-tricks.readthedocs.io/en/latest/>`_ and `cloudpickle <https://github.com/cloudpipe/cloudpickle>`_. Essentially, it is json-tricks, that is a enhanced version of Python JSON, enabling handling of serialization of numpy arrays, date/times, decimal, fraction and etc. The difference lies in the handling of class instances. Json-tricks deals with class instances with ``__dict__`` and ``__class__``, which in most of our cases are not reliable (e.g., datasets, dataloaders). Rather, our serialization deals with class instances with two methods:
+
+1. If the class / factory that creates the object is decorated with ``nni.trace``, we can serialize the type / factory function, along with the parameters, such that the instance can be re-instantiated.
```
QuanluZhang, line 8: type / -> class /

docs/en_US/NAS/Serialization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+To trace a function or class, users can use decorator like,
+
+.. code-block:: python
+
+    @nni.trace
+    class MyClass:
+        ...
+
+Inline trace that traces instantly before being used is also acceptable: ``nni.trace(MyClass)(parameters)``.
```
QuanluZhang, line 27: before being used -> on the object instantiation or function invoke

QuanluZhang, line 27: what will happen if I both decorate class with `@nni.trace` and use `nni.trace()` on instantiation?

docs/en_US/NAS/Serialization.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Assuming a class ``cls`` is already traced, when it is serialized, its class type along with initialization parameters will be dumped. As the parameters are possibly class instances (if not primitive types like ``int`` and ``str``), their serialization will be a similar problem. We recommend decorate them with ``nni.trace`` as well. In other words, ``nni.trace`` should be applied recursively if necessary.
```
QuanluZhang, line 29: it would be interesting to automatically add `nni.trace` on necessary objects, possibly using syntax checking technique as plugin :)

nni/retiarii/integration.py
```diff
@@ -121,7 +122,19 @@ def send_trial(self, parameters, placement_constraint=None):
             'placement_constraint': placement_constraint
         }
         _logger.debug('New trial sent: %s', new_trial)
-        send(CommandType.NewTrialJob, nni.dump(new_trial))
+
+        send_payload = nni.dump(new_trial, pickle_size_limit=-1)
+        if len(send_payload) > 50 * 1024:
```
QuanluZhang, line 127: 50KB is too small? how about 1MB or 10MB?

liuzhe-lz (approved):

docs/en_US/NAS/ModelEvaluators.rst
matluster: Oops, forgot to change `serialize` to `nni.trace`.
They are the same. I deprecated the `serialize` API (but it can still be used for backward compatibility).

docs/en_US/NAS/ModelEvaluators.rst
matluster: I'll add an example for a real-world task in the upcoming release.

docs/en_US/NAS/QuickStart.rst
matluster: I mentioned nothing about `nni.trace` here.

`@model_wrapper` used `@nni.trace` as its underlying implementation, but they are totally different APIs.
`@model_wrapper` marks a module as a search space (which will reset label counting and do many other stuffs). `@nni.trace` is to record parameters and that's all.

docs/en_US/NAS/Serialization.rst
matluster: The second time `nni.trace` is called, it will detect that it is already `traceable`, so it does nothing.

nni/retiarii/integration.py
matluster: I'd rather play safe. In most of my tests, the whole payload is actually less than 2KB.
Judging from the pipe limit is 1MB (maybe my memory goes wrong), Maybe 256KB is still acceptable?

ultmaster:
@QuanluZhang Comments are mostly resolved. Still working on a "hard" test case for Windows.

QuanluZhang (approved):

ultmaster:
Link known issue #4434

PR #4399, opened 2021-12-17
Support hyper-parameter tuning in Retiarii

ultmaster:
### Description ###

Along with docs and tests updates.

Won't go with v2.6.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###

```python
# used in evaluator
def train_and_evaluate(model_cls, learning_rate):
  ...

self.evaluator = FunctionalEvaluator(train_and_evaluate, learning_rate=nn.ValueChoice([1e-3, 1e-2, 1e-1]))
```

PR #4400, opened 2021-12-17
Fix deepcopy of mutables

ultmaster:
### Description ###

This can be merged into v2.6.

Fix issues that prevents layer choices from being deep-copied.

### Checklist ###
  - [x] test case
  - [x] doc (in #4399)

### How to test ###

```python
self.blocks = nn.Repeat(nn.LayerChoice([...]), (1, 3))
```

### DIFF
--- a/nni/retiarii/nn/pytorch/api.py
+++ b/nni/retiarii/nn/pytorch/api.py
@@ -3,6 +3,7 @@
 
 import copy
 import warnings
+from abc import ABC, abstractclassmethod
 from typing import Any, List, Union, Dict, Optional
 
 import torch
@@ -10,14 +11,13 @@ import torch.nn as nn
 
 from nni.common.serializer import Translatable
 from nni.retiarii.serializer import basic_unit
-from nni.retiarii.utils import NoContextError
-from .utils import generate_new_label, get_fixed_value
+from .utils import Mutable, generate_new_label, get_fixed_value
 
 
 __all__ = ['LayerChoice', 'InputChoice', 'ValueChoice', 'Placeholder', 'ChosenInputs']
 
 
-class LayerChoice(nn.Module):
+class LayerChoice(Mutable):
     """
     Layer choice selects one of the ``candidates``, then apply it on inputs and return results.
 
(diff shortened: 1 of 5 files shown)


liuzhe-lz (approved):

nni/retiarii/nn/pytorch/utils.py
```diff
@@ -1,6 +1,35 @@
-from typing import Any, Optional, Tuple
+from abc import ABC, abstractclassmethod
+from typing import Any, Optional, Tuple, Union
 
-from nni.retiarii.utils import ModelNamespace, get_current_context
+import torch.nn as nn
+from nni.retiarii.utils import NoContextError, ModelNamespace, get_current_context
+
+
+class Mutable(ABC, nn.Module):
```
liuzhe-lz, line 8: I'm not sure but to my understanding `ABC` is not designed for this scenario. The PEP states it's mainly for overriding `isinstance`, and I believe we don't want to treat all objects with `create_fixed_module` method as mutable. To force subclass overriding a virtual method, `NotImplementedError` should be enough.

Anyway, I guess inheriting `ABC` won't cause problem.

ultmaster: I might want to refine and make `Mutable` a more clear interface, which defines what a "high-level API" should do.

For example, each "mutable" class should have a corresponding mutator, and a dedicated approach to preprocess the "mutable" into base class + mutator. For graph engine, this might even include the part where the model is parsed into graph.

On the other hand, I concern that such design is over complex, and we should have another "Adapter" module that handles all the complex logics there.

I haven't made the decision yet. Let's keep it for now and I'd like to have more thoughts about it.

liuzhe-lz: Seems we are talking about different things...
I mean, although it's an "abstract base class" (in English), it does not need to be `ABC` (in Python).
To my understanding inheriting `ABC` implies that the user does not need to inherit this class. Instead they only need to provide required methods.
For example a class with `__iter__` method is considered `issubclass(MyContainer, Iterable)`, even though it does not actually inherit `Iterable`.

nni/retiarii/nn/pytorch/api.py
```diff
@@ -10,14 +10,13 @@
 
 from nni.common.serializer import Translatable
 from nni.retiarii.serializer import basic_unit
-from nni.retiarii.utils import NoContextError
-from .utils import generate_new_label, get_fixed_value
+from .utils import Mutable, generate_new_label, get_fixed_value
```
QuanluZhang, line 13: why `Mutable` is put into utils?

ultmaster: Because it's currently an implementation trick.

Will formalize it into an interface when necessary.

nni/retiarii/nn/pytorch/utils.py
ultmaster: Read the PEP and some comments and SO, seems that I'm abusing ABC.
I'll consider removing it.

nni/retiarii/nn/pytorch/component.py
```diff
@@ -69,7 +68,7 @@ def _replicate_and_instantiate(blocks, repeat):
         assert repeat <= len(blocks), f'Not enough blocks to be used. {repeat} expected, only found {len(blocks)}.'
         blocks = blocks[:repeat]
         if not isinstance(blocks[0], nn.Module):
-            blocks = [b() for b in blocks]
+            blocks = [b(i) for i, b in enumerate(blocks)]
```
QuanluZhang, line 71: how can you make sure `i` is the input argument of a block?

ultmaster: The function is provided by user, and user should make sure of it.
Yes, this is an API change. But I think this should not be a problem because the old version doesn't work at all, and no one actually complained about it.

QuanluZhang: so, the document has not been updated in that pr, right? i cannot find the explanation of how to write a block

QuanluZhang (approved):

nni/retiarii/nn/pytorch/component.py
ultmaster: The guide to use APIs has already been rewritten in #4399.

Originally, I think these changes should also go with #4399. But I split them into this PR because I consider them as bug fixes, thus should be merged into this release.

PR #4411, opened 2021-12-21
Add loose-end support for Cell API

ultmaster:
### Description ###

Support `merge_op="loose_end"` in Cell API. Note that this feature is not supported for graph engine. Won't merge into v2.6.

Fix a bug in graph converter that doesn't support `/` in node name.

### Checklist ###
  - [x] test case
  - [ ] doc: will be done after #4399.

### How to test ###

Try cell API with `merge_op="loose_end"`.


nni/retiarii/codegen/pytorch.py
```diff
@@ -63,42 +67,37 @@ def _format_inputs(node: Node) -> Tuple[List[str], List[Any]]:
...
-        for name in names:
-            if name.startswith(graph_name):
-                converted_names.append(name[len(graph_name):])
-            else:
-                converted_names.append(name)
-        return converted_names
-    else:
-        return names[len(graph_name):] if names.startswith(graph_name) else names
+    name = name[len(graph_name):] if name.startswith(graph_name) else name
+    name = name.replace('/', '__')
```
QuanluZhang, line 97: why adding this line, seems there is no such logic in the previous version

test/ut/retiarii/test_highlevel_apis.py
```diff
@@ -568,6 +568,45 @@ def forward(self, x):
...
+                self.linear = nn.Linear(3, 16)
+
+            def forward(self, x: List[torch.Tensor]) -> List[torch.Tensor]:
+                return [self.linear(x[0]), x[1]]
+
+        class Postprocessor(nn.Module):
+            def forward(self, this: torch.Tensor, prev: List[torch.Tensor]) -> Tuple[torch.Tensor, torch.Tensor]:
+                return prev[-1], this
+
+        @self.get_serializer()
```
QuanluZhang, line 586: use model_wrapper here?

nni/retiarii/codegen/pytorch.py
ultmaster: Previously, nodes in cell are named as `cell0__op1` because `/` is not supported in graph converter.
I changed them to `cell0/op1` and added support for `/` in converter here.

test/ut/retiarii/test_highlevel_apis.py
ultmaster: ok

nni/retiarii/codegen/pytorch.py
QuanluZhang: good choice, but where did you do this? not in this pr?

QuanluZhang (approved):

nni/retiarii/codegen/pytorch.py
ultmaster: image

image

test/ut/retiarii/test_highlevel_apis.py
JiahangXu: why replace @self.get_serializer() by model_wrapper?

ultmaster: In the old past, putting `@model_wrapper` on models used by graph engine makes the engine crash, while python engine needs `@model_wrapper`. As you see, `self.get_serializer()` have different implementations between super-class and sub-class.

But this bug has been fixed. We can always apply `@model_wrapper` now.

JiahangXu (approved):

PR #4412, opened 2021-12-21
fix pruning examples & pruner memory usage optimize

J-shang:
### Description ###
Optimize memory usage in TaylorFO Pruner and Activation-base Pruner.
Fix example bug.

### DIFF
--- a/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
@@ -564,7 +564,7 @@ class ActivationPruner(BasicPruner):
 
 class ActivationAPoZRankPruner(ActivationPruner):
     def _activation_trans(self, output: Tensor) -> Tensor:
-        # return the sum of zero counts of the same activation position cross each batch.
+        # return a matrix that the position of zero in `output` is one, others is zero.
         return torch.eq(self._activation(output.detach()), torch.zeros_like(output)).type_as(output)
 
     def _get_metrics_calculator(self) -> MetricsCalculator:
@@ -573,7 +573,7 @@ class ActivationAPoZRankPruner(ActivationPruner):
 
 class ActivationMeanRankPruner(ActivationPruner):
     def _activation_trans(self, output: Tensor) -> Tensor:
-        # return the sum of zero counts of the same activation position cross each batch.
+        # return the activation of `output` directly.
         return self._activation(output.detach())
 
     def _get_metrics_calculator(self) -> MetricsCalculator:
--- a/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
@@ -524,10 +524,7 @@ class ActivationPruner(BasicPruner):
             raise 'Unsupported activatoin {}'.format(activation)
 
     def _collector(self, buffer: List) -> Callable[[Module, Tensor, Tensor], None]:
-        def collect_activation(_module: Module, _input: Tensor, output: Tensor):
-            if len(buffer) < self.training_batches:
-                buffer.append(self._activation(output.detach()))
-        return collect_activation
+        raise NotImplementedError()
 
     def reset_tools(self):
         collector_info = HookCollectorInfo([layer_info for layer_info, _ in self._detect_modules_to_compress()], 'forward', self._collector)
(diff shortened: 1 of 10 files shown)


nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -551,11 +548,35 @@ def _get_metrics_calculator(self) -> MetricsCalculator:
 
 
 class ActivationAPoZRankPruner(ActivationPruner):
+    def _collector(self, buffer: List) -> Callable[[Module, Tensor, Tensor], None]:
+        assert len(buffer) == 0, 'Buffer pass to activation pruner collector is not empty.'
```
ultmaster, line 552: Better to add comment here: what are buffer[0] and buffer[1].

J-shang: added

ultmaster (approved):

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -551,11 +548,43 @@ def _get_metrics_calculator(self) -> MetricsCalculator:
...
+        # The length of the buffer used in this pruner will always be 2.
+        # buffer[0] is the number of how many batches are counted in buffer[1].
+        # buffer[1] is a tensor and the size of buffer[1] is same as the activation,
+        # each element in buffer[1] means the sum of the same activation position cross each batch.
+        buffer.append(0)
+
+        def collect_activation(_module: Module, _input: Tensor, output: Tensor):
+            if len(buffer) == 1:
```
zheng-ningxin, line 581: can we initialize the buffer at line-578? so that we do not need check the length of the buffer?

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -551,11 +548,43 @@ def _get_metrics_calculator(self) -> MetricsCalculator:
...
 class ActivationAPoZRankPruner(ActivationPruner):
+    def _collector(self, buffer: List) -> Callable[[Module, Tensor, Tensor], None]:
+        assert len(buffer) == 0, 'Buffer pass to activation pruner collector is not empty.'
+        # The length of the buffer used in this pruner will always be 2.
+        # buffer[0] is the number of how many batches are counted in buffer[1].
+        # buffer[1] is a tensor and the size of buffer[1] is same as the activation,
+        # each element in buffer[1] means the sum of zero counts of the same activation position cross each batch.
+        buffer.append(0)
+
+        def collect_activation(_module: Module, _input: Tensor, output: Tensor):
```
zheng-ningxin, line 559: APOZ's _collector looks exactly the same with MeanRank? why not put the common functions into `ActivationPruner`?

J-shang: Yes, it is a good idea, I will merge these two functions

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: I think it is hard because we need the size of the output to initialize the buffer[1].

zheng-ningxin: Get it!

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: fix it

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -551,11 +563,19 @@ def _get_metrics_calculator(self) -> MetricsCalculator:
 
 
 class ActivationAPoZRankPruner(ActivationPruner):
+    def _activation_trans(self, output: Tensor) -> Tensor:
+        # return the sum of zero counts of the same activation position cross each batch.
```
QuanluZhang, line 567: seems this comment is not what this function does?

QuanluZhang: BTW, where is `self._activation` defined?

J-shang: updated the comment, `self._activation` is defined in line 507.

zheng-ningxin (approved):

PR #4417, opened 2021-12-22
Fix retiarii experiment

liuzhe-lz:


nni/retiarii/experiment/pytorch.py
```diff
@@ -81,7 +81,7 @@ def __setattr__(self, key, value):
         if key in fixed_attrs and fixed_attrs[key] != value:
             raise AttributeError(f'{key} is not supposed to be set in Retiarii mode by users!')
         # 'trial_code_directory' is handled differently because the path will be converted to absolute path by us
-        if key == 'trial_code_directory' and not (value == Path('.') or os.path.isabs(value)):
+        if key == 'trial_code_directory' and not (str(value) == '.' or os.path.isabs(value)):
```
QuanluZhang, line 84: what is the difference?

nni/retiarii/experiment/pytorch.py
```diff
@@ -183,6 +183,8 @@ def __init__(self, base_model: nn.Module, trainer: Union[Evaluator, BaseOneShotT
         self._proc: Optional[Popen] = None
         self._pipe: Optional[Pipe] = None
 
+        self.url_prefix = None
```
QuanluZhang, line 186: what is it used for?

ultmaster (approved):

nni/retiarii/experiment/pytorch.py
liuzhe-lz: In init, it's a string. So the check failed.

nni/retiarii/experiment/pytorch.py
liuzhe-lz: NNI has supported web UI / REST API URL prefix since v2.3. In this release the feature was added to Python API.
This field is required to make it in sync with nni.Experiment.

QuanluZhang (approved):

PR #4423, opened 2021-12-24
[Bugbash] update example import path

J-shang:
### Description ###
~1. update the error message for user-friendly~
1. update the compression example models import

### DIFF
--- a/examples/model_compress/pruning/amc/amc_train.py
+++ b/examples/model_compress/pruning/amc/amc_train.py
@@ -23,7 +23,6 @@ from data import get_dataset
 from utils import AverageMeter, accuracy, progress_bar
 
 from pathlib import Path
-import sys
 sys.path.append(str(Path(__file__).absolute().parents[2] / 'models'))
 from mobilenet import MobileNet
 from mobilenet_v2 import MobileNetV2
--- a/examples/model_compress/pruning/amc/amc_train.py
+++ b/examples/model_compress/pruning/amc/amc_train.py
@@ -22,14 +22,11 @@ from nni.compression.pytorch import ModelSpeedup
 from data import get_dataset
 from utils import AverageMeter, accuracy, progress_bar
 
-sys.path.append('../../models')
-try:
-    from mobilenet import MobileNet
-    from mobilenet_v2 import MobileNetV2
-except ModuleNotFoundError as e:
-    print(e)
-    print('Please check whether the current working path is the path where this file is located.')
-    exit()
+from pathlib import Path
+import sys
+sys.path.append(str(Path(__file__).absolute().parents[2] / 'models'))
+from mobilenet import MobileNet
+from mobilenet_v2 import MobileNetV2
 
 def parse_args():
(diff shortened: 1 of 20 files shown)


examples/model_compress/pruning/amc/amc_train.py
```diff
@@ -23,8 +23,13 @@
...
 
 sys.path.append('../../models')
-from mobilenet import MobileNet
-from mobilenet_v2 import MobileNetV2
+try:
+    from mobilenet import MobileNet
+    from mobilenet_v2 import MobileNetV2
+except ModuleNotFoundError as e:
+    print(e)
+    print('Please check whether the current working path is the path where this file is located.')
```
liuzhe-lz, line 31: Then I would suggest `sys.path.append(Path(__file__).parents[2]`.

J-shang: Good suggestion, fix it

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -36,6 +36,8 @@ def generate_sparsity(self, metrics: Dict[str, Tensor]) -> Dict[str, Dict[str, T
                 threshold = torch.topk(metric.view(-1), prune_num, largest=False)[0].max()
             mask = torch.gt(metric, threshold).type_as(metric)
             masks[name] = self._expand_mask(name, mask)
+            if self.continuous_mask:
```
zheng-ningxin, line 39: what's this for?

J-shang: e.g., if we set fine-grained pre-masks to a channel-wise pruner, then the generated masks will be channel-wise masks and may not inherit the fine-grained pre-masks. This line is for inheriting the pre-maks.

zheng-ningxin: Looks like it is not related to the `import path`? If so, suggest to split into two PRs, so that the comments can be much clear when merged.

JiahangXu (approved):
Suggest to add the folder information of running the code in docs as well.

QuanluZhang:
> 

for the second point, it might be better to also print warning message

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
J-shang: split to pr #4428

J-shang:
> Suggest to add the folder information of running the code in docs as well.

For the current import way, we don't need to cd the file folder, we can run the example anywhere now😃

zheng-ningxin (approved):

JiahangXu:
> > Suggest to add the folder information of running the code in docs as well.
> 
> For the current import way, we don't need to cd the file folder, we can run the example anywhere now😃

That will be great~

PR #4428, opened 2021-12-27
[Bugbash] fix pre-masks inherit

J-shang:
### Description ###
Split from pr #4423 

e.g., if we set fine-grained pre-masks to a channel-wise pruner, then the generated masks will be channel-wise masks and may not inherit the fine-grained pre-masks. This pr is for inheriting the pre-maks.


J-shang:
> for the second point, it might be better to also print warning message

Hi @QuanluZhang , this is the pr split from #4423 point 2nd.
What warning message should we add for users?

QuanluZhang (approved):

zheng-ningxin (approved):

PR #4437, opened 2021-12-30
Retiarii serializer user experience improvements

ultmaster:
### Description ###

1. Check `model_wrapped` in Retiarii experiment.
2. Throw error when pickling size is too large and update related document.
3. Fix issues in serialization of evaluator to suppress the warning: "Callstack exhausted but main module not found".

### Checklist ###
  - ~test case~
  - [x] doc

### How to test ###

Same as #4390.

### DIFF
--- a/nni/retiarii/evaluator/functional.py
+++ b/nni/retiarii/evaluator/functional.py
@@ -26,6 +26,7 @@ class FunctionalEvaluator(Evaluator):
 
     def _dump(self):
         return {
+            'type_name': self.__class__,
             'function': self.function,
             'arguments': self.arguments
         }
--- a/nni/retiarii/evaluator/functional.py
+++ b/nni/retiarii/evaluator/functional.py
@@ -26,7 +26,7 @@ class FunctionalEvaluator(Evaluator):
 
     def _dump(self):
         return {
-            'type_name': self.__class__,
+            'type': self.__class__,
             'function': self.function,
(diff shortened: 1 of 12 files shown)


nni/retiarii/evaluator/functional.py
```diff
@@ -26,6 +26,7 @@ def _load(ir):
 
     def _dump(self):
         return {
+            'type_name': self.__class__,
```
QuanluZhang, line 29: what is the type of `self.__class__`? is it safe to dump this `self.__class__` object and load it in another environment (e.g., another machine)?

ultmaster: Its type is "type". `nni.dump` and `nni.load` will handle the serialization of a "type" object.

1. If we can get its path, we dump it into a path. e.g., `torch.nn.Conv2d`.
2. Otherwise, we dump it into a binary. e.g., class defined in a jupyter notebook.

QuanluZhang (approved):

nni/retiarii/evaluator/functional.py
liuzhe-lz: Since it's a class object, why the key is called "type name"?

ultmaster: Good question. I copied from `operation.py`. Let me fix it.

liuzhe-lz (approved):

Issue #4398, opened 2021-12-16
NNI resume does not create and run new trials


alpeq:
**Describe the issue**:
NNI unable to resume. I have access to a cluster with limited time, I would like to stop when my job time has been reached and resume again. I work with latest version NNI, in the log I provide the job has been interrupted causing FAILED trials, when I resume the experiment while the status change to RUNNING, there is no new trials created ( no new RUNNING trials are created). 
In other workarounds, I tried to restart the failed jobs as other issues suggested () but I cannot use their solution since I don't have access the webUI and I use command line as a mean of communication (no command line to rerun failed jobs in the docs). 
I am not as much interested in the restart of the failed jobs but in resuming and continue with the experiment.

**Environment**:
- NNI version: 2.5
- Training service (local|remote|pai|aml|etc): local
- Client OS: ubuntu20.04
- Server OS (for remote mode only):
- Python version: 3
- PyTorch/TensorFlow version: torch==1.8.1+cu111 torchvision==0.9.1+cu111 torchaudio==0.8.1 
- Is conda/virtualenv/venv used?: no
- Is running in Docker?: docker based container ( Singularity ) 

**Configuration**:
 - Experiment config (remember to remove secrets!):
 
```
experimentName: SpyTorch
trialConcurrency: 1
maxExecDuration: 20d
maxTrialNum: 1000
trainingServicePlatform: local
searchSpacePath: ../searchspaces/nni_SearchSpace_SpyTorch_events.json
useAnnotation: false
tuner:
  builtinTunerName: Anneal
  classArgs:
    optimize_mode: maximize
trial:
  command: python3 nni_SpyTorch.py
  codeDir: ../experiments/
  gpuNum: 1
localConfig:
  gpuIndices: '0,1,2,3'
  maxTrialNumPerGpu: 5
  useActiveGpu: true
```

 - Search space:

```
{
  "scale": {
    "_type": "quniform",
    "_value": [
      5,
      50,
      5
    ]
  },
  "time_bin_size": {
    "_type": "quniform",
    "_value": [
      1,
      10,
      1
    ]
  },
  "nb_input_copies": {
    "_type": "choice",
    "_value": [
      1,
      2,
      4,
      8
    ]
  },
  "tau_mem": {
    "_type": "quniform",
    "_value": [
      0.010,
      0.100,
      0.010
    ]
  },
  "tau_ratio": {
    "_type": "choice",
    "_value": [
      1,
      2,
      5,
      10
    ]
  },
[...]

PR #4438, opened 2021-12-31
[bug fix] fix the bug in update search space

QuanluZhang:
### Description ###
Fix the bug in updating search space, the bug is incorrect json dump, details in #4398 

### Checklist ###
  - [ ] test case
  - [x] doc, no doc update needed

### How to test ###

run mnist pytorch example with tpe tuner, update search space to a new one during experiment running, update search space is successfully updated, stop the experiment. Then resume this experiment to check whether the resume is successful and whether new trial jobs are created.

as this pr is also related to retiarii example, need to run a multi-trial nas example to check whether it works.

### DIFF
--- a/nni/tools/nnictl/updater.py
+++ b/nni/tools/nnictl/updater.py
@@ -41,7 +41,7 @@ def validate_dispatcher(args):
 
 def load_search_space(path):
     '''load search space content'''
-    content = json.dumps(get_json_content(path))
+    content = get_json_content(path)
     if not content:
         raise ValueError('searchSpace file should not be empty')
     return content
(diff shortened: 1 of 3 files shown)


liuzhe-lz (approved):

nni/tools/nnictl/updater.py
```diff
@@ -112,11 +112,18 @@ def update_trialnum(args):
...
         print_error('Update %s failed!' % 'trialnum')
 
+def load_imported_data(path):
+    '''load the trial data that will be imported'''
+    content = json.dumps(get_json_content(path))
+    if not content:
+        raise ValueError('Imported data should not be empty')
+    return content
+
 def import_data(args):
```
ultmaster, line 122: Better to have a UT for this.

QuanluZhang: yes, add UT will become a follow-up task after the upcoming release, because @liuzhe-lz said that the time budget for UT is very tight.

Issue #4452, opened 2022-01-07
Can't use powershell to run config with my own code


DoveEchoing:
**Describe the issue**:
I can use the source code to run nni on my computer,but when I was trying to modify the code to meet my needs,it occurred some mistakes. I referred some examples to modify my code(like regression/classification of sklearn,mnist-keras)，because the framework is different,when I first try to modify my code and use powershell to run it,it occurred a lot mistakes.
Today I modified my code once again,it could run successfully in pycharm,but it failed in powershell.
crying,I think I need some help.

**Environment**:
- NNI version:2.5
- Training service (local|remote|pai|aml|etc):local
- Client OS:Windows 10
- Server OS (for remote mode only):None
- Python version:3.6
- PyTorch/TensorFlow version:1.10.1/2.6.2
- Is conda/virtualenv/venv used?:conda
- Is running in Docker?:no

**code**

```
LOG = logging.getLogger('ANN Example')

def load_data(train_path='./data/data712pc_train.csv', test_path='./data/data712pc_test.csv'):
    InputIndex = ['X1', 'X2', 'X3', 'X4', 'X5', 'X6', 'X7', 'X8', 'X9', 'X10',
                  'X11', 'X12', 'X13', 'X14', 'X15']
    OutputIndex = ['Y']
    data_train = pd.read_csv(train_path)
    train_X = data_train[InputIndex]
    train_Y = data_train[OutputIndex]
    data_test = pd.read_csv(test_path)
    test_X = data_test[InputIndex]
    test_Y = data_test[OutputIndex]
    minmax = MinMaxScaler(feature_range=(0, 1))
    X_train = minmax.fit_transform(train_X)
    X_test = minmax.transform(test_X)
    Y_train = minmax.fit_transform(train_Y)
    Y_test = minmax.transform(test_Y)
    return X_train, X_test, Y_train, Y_test

def default_params():
    params = {
        'activation': 'relu',
        'optimizer': 'Adam',
        'learning_rate': 0.001,
        'momentum': 0.9
    }
    return params

def create_model(PARAMS):
    layers = [
        Dense(8, activation=PARAMS.get('activation')),
        Dense(4, activation=PARAMS.get('activation')),
        Dense(1, activation=PARAMS.get('activation'))
    ]
    model = Sequential(layers)

    if PARAMS.get('optimizer') == 'Adam':
        optimizer = Adam(lr = PARAMS.get('learning_rate'))
    elif PARAMS.get('optimizer') == 'SGD':
[...]

Issue #4450, opened 2022-01-06
Exponential Notation is not allowed in search space


Lionelsy:
**Describe the issue**:
The Exponential Notation is not working in search space.
Similar to the [previous issue](https://github.com/microsoft/nni/issues/4080), I just found the expression of Exponential Notation in search space would lead to an error after I updated to nni2.5. 
{
    "lr": {"_type": "uniform", "_value": [1e-6, 1e-3]}
}
But it works normally when I set the search space in decimals
{
    "lr": {"_type": "uniform", "_value": [0.000001, 0.001]}
}

Considering the [previous issue](https://github.com/microsoft/nni/issues/4080), I test it on pyyaml== 6.0 and 5.4.1. The Exponential Notation expression is not working in both cases.

**Environment**:
- NNI version: 2.5
- Training service (local|remote|pai|aml|etc): local
- Client OS: ubuntu
- Server OS (for remote mode only):
- Python version: 3.8
- PyTorch/TensorFlow version: PyTorch 1.9.1
- Is conda/virtualenv/venv used?: conda
- Is running in Docker?: no

**Configuration**:
 - Experiment config (remember to remove secrets!):
 - Search space:
{
    "lr": {"_type": "uniform", "_value": [1e-6, 1e-3]}
}

**Log message**:
 - dispatcher.log:
 
[2022-01-06 15:00:37] INFO (nni.runtime.msg_dispatcher_base/MainThread) Dispatcher started
[2022-01-06 15:00:37] ERROR (nni.common.hpo_utils.validation/Thread-1) search space "lr" (uniform) must have float values : OrderedDict([('_type', 'uniform'), ('_value', ['1e-6', '1e-3'])])
[2022-01-06 15:00:37] INFO (hyperopt.tpe/Thread-1) tpe_transform took 0.000977 seconds
[2022-01-06 15:00:37] INFO (hyperopt.tpe/Thread-1) TPE using 0 trials
[2022-01-06 15:00:37] ERROR (nni.runtime.msg_dispatcher_base/Thread-1) must be real number, not str
Traceback (most recent call last):
  File "/home/shuyu/anaconda3/envs/trajectory/lib/python3.8/site-packages/nni/runtime/msg_dispatcher_base.py", line 88, in command_queue_worker
    self.process_command(command, data)
  File "/home/shuyu/anaconda3/envs/trajectory/lib/python3.8/site-packages/nni/runtime/msg_dispatcher_base.py", line 147, in process_command
[...]

PR #4455, opened 2022-01-10
Fix search space compatibility with JSON

liuzhe-lz:
### Description ###

As of now the latest version of PyYAML (6.0) [does not support YAML 1.2](https://github.com/yaml/pyyaml/issues/116), which means it is not fully compatible with JSON.
This has caused multiple problems for NNI: #4450 #4452

The PR partially fixed this bug by trying YAML and JSON loaders consecutively. However if a user uses comment and scientific notation at same time, NNI will still fail.
Since PyYAML has already been working on [YAML 1.2 support](https://github.com/yaml/pyyaml/pull/555), I prefer to wait for its update for a perfect fix.

As a reminder, ruamel.yaml supports YAML 1.2, but according to our previous experience it's pip package and conda package have inconsistent APIs, which caused even more problems.

### Checklist ###
  - [x] test case
  - ~~doc~~

### How to test ###
Create tricky JSON search space file and make sure it's loaded correctly.

### DIFF
--- a/nni/experiment/config/experiment_config.py
+++ b/nni/experiment/config/experiment_config.py
@@ -116,8 +116,6 @@ class ExperimentConfig(ConfigBase):
         super()._canonicalize([self])
 
         if self.search_space_file is not None:
-            if self.search_space is not None:
-                raise ValueError('ExperimentConfig: search_space and search_space_file cannot be set at same time')
             yaml_error = None
             try:
                 self.search_space = _load_search_space_file(self.search_space_file)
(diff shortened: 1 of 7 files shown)


nni/experiment/config/experiment_config.py
```diff
@@ -156,3 +164,13 @@ def _validate_canonical(self):
...
+
+def _load_search_space_file(search_space_path):
+    # FIXME
+    # we need this because PyYAML 6.0 does not support YAML 1.2,
+    # which means it is not fully compatible with JSON
+    content = Path(search_space_path).read_text()
+    try:
+        return json.loads(content)
+    except Exception:
+        return yaml.safe_load(content)
```
QuanluZhang, line 176: so the yaml format search space still does not support scientific notation, right?

QuanluZhang (approved):

nni/experiment/config/experiment_config.py
cruiseliu: It doesn't.

ultmaster (approved):

PR #4457, opened 2022-01-10
fix SA task generator bug

J-shang:
### Description ###
e.g. `origin_masks` of `op_name_1` is [1, 1, 1, 0, 0, 0], and
`config_list = [{'op_names': ['op_name_1'], 'total_sparsity': 0.5}, ...]`,
then the generated `config_list` will not contain `op_name_1`, because `op_name_1` have already masked 50%.

But it is not true because we define the `config_list` should cover the preset sparsity(masks), we expect the `config_list` to contain `op_name_1` with `sparsity=0.5`

### DIFF
--- a/nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
@@ -227,7 +227,7 @@ class SimulatedAnnealingTaskGenerator(TaskGenerator):
         op_names = config['op_names']
 
         if target_sparsity == 0:
-            return [], []
+            return [{'total_sparsity': 0, 'op_names': [op_name]} for op_name in op_names], []
 
         low_limit = 0
         while True:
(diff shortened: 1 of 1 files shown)


nni/algorithms/compression/v2/pytorch/pruning/tools/task_generator.py
```diff
@@ -227,7 +227,10 @@ def _init_config_sparsity(self, config: Dict) -> Tuple[List[Dict], List]:
         op_names = config['op_names']
 
         if target_sparsity == 0:
```
QuanluZhang, line 229: in which case `target_sparsity` is 0?

J-shang: as in description example, if `origin_masks` already masks 50%, and sparsity set in `config_list` is 50%, then the target_sparsity is 0.

QuanluZhang (approved):

zheng-ningxin (approved):

PR #4475, opened 2022-01-18
Bump Python version to 3.7~3.9

liuzhe-lz:
### Description ###

Python 3.6 is dead.
PyTorch has not yet supported Python 3.10.

Need to upgrade Python on k8s IT client node.

Version resolving of Python packages on ubuntu-legacy is a complete mess now.
This will be fixed with [cache refactor](https://github.com/microsoft/nni/tree/dev-pipeline-cache).

### Checklist ###
  - ~~test case~~
  - ~~doc~~

### How to test ###

No need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ Within the following table, we summarized the current NNI capabilities, we are g
 
 ### **Install**
 
-NNI supports and is tested on Ubuntu >= 18.04, Windows 21H2, and macOS >= 11.
+NNI supports and is tested on Ubuntu >= 18.04, Windows 10 >= 21H2, and macOS >= 11.
 Simply run the following `pip install` in an environment that has `python 64-bit >= 3.7`.
 
 Linux or macOS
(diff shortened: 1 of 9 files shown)


ultmaster:
I think the pipeline issues are related to cache, which only concerns ubuntu version.

cruiseliu:
> I think the pipeline issues are related to cache, which only concerns ubuntu version.

It's about numpy version.
Before the last attempt, the problem was that when ConfigSpaceNNI is built, numpy version was 1.22. But later, a new package requires numpy < 1.22 and pip silently downgraded it. Then ConfigSpaceNNI has an incompatible numpy and therefore cannot work.
The last attempt met a different problem. Seems pip is trying to prepare all packages before installing any of them. So when ConfigSpaceNNI gets built, numpy has not yet been installed.

ultmaster (approved):

J-shang (approved):

PR #4476, opened 2022-01-18
Create soft link "_latest" in nni-experiments

liuzhe-lz:
### Description ###

A quality-of-life improvement for developers.

It creates/updates a soft link `~/nni-experiments/_latest` pointing to the latest experiment directory.
So you can do something like `cat _latest/log/nni_manager.log` instead of re-typing experiment ID every time.

On Windows [developer mode](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) must be turned on to use symlink. Or otherwise the patch silently does nothing.

### Checklist ###
  - ~~test case~~ no need because it is mostly in a catch block and is harmless on failure
  - ~~doc~~ no need because no one has to care about this

### How to test ###

Create an experiment and check `~/nni-experiments/_latest` points to the experiment directory.
Then create another experiment and check it points to the new directory.

### Release Note ###

Don't mention.


J-shang (approved):

nni/experiment/launcher.py
```diff
@@ -109,15 +109,23 @@ def start_experiment(action, exp_id, config, port, debug, run_mode, url_prefix):
...
-        return proc
-
     except Exception as e:
         _logger.error('Create experiment failed')
         if proc is not None:
             with contextlib.suppress(Exception):
                 proc.kill()
         raise e
 
+    link = Path(config.experiment_working_directory, 'latest')
```
ultmaster, line 119: Suggest `.latest` so that it's not flooded by tons of folders in `nni-experiments`, and people can find it with a first glance of `ls -a`.

ultmaster (approved):

PR #4481, opened 2022-01-19
[Model Compression] Add bank pruning for level pruner

linbinskn:
### Description ###
This pr aims to support level pruner with bank feature. Bank pruning means finegrained pruning will be done in granularity of bank. For instance, if user set number of values in per bank to 4 and sparsity to 50%, level pruner will generate the same sparse pattern with [NVIDIA Sparse TensorCore](https://developer.nvidia.com/blog/exploiting-ampere-structured-sparsity-with-cusparselt/).

### Checklist ###
  - [x] test case
  - [x] doc

### DIFF
--- a/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
@@ -148,7 +148,8 @@ class LevelPruner(BasicPruner):
         acceleration. Please refer to releated paper for further information 'Balanced Sparsity for 
         Efficient DNN Inference on GPU'(https://arxiv.org/pdf/1811.00206.pdf).
     balance_gran : list
-        Balance_gran is for special sparse pattern balanced sparsity, Default value is None which means pruning without awaring balance, namely normal finegrained pruning.
+        Balance_gran is for special sparse pattern balanced sparsity, Default value is None which means pruning 
+        without awaring balance, namely normal finegrained pruning.
         If passing list of int, LevelPruner will prune the model in the granularity of multi-dimension block.
         Attention that the length of balance_gran should be smaller than tensor dimension.
         For instance, in Linear operation, length of balance_gran should be equal or smaller than two since
@@ -157,6 +158,9 @@ class LevelPruner(BasicPruner):
         and 10 values would be kept after pruning. Finegrained pruning is applied in the granularity of block 
         so that each block will kept same number of non-zero values after pruning. Such pruning method "balance" 
         the non-zero value in tensor which create chance for better hardware acceleration.
+
+        Note: If length of given balance_gran smaller than length of pruning tensor shape, it will be made up
+              in right align(such as example 1).
 
             example 1:
                 operation: Linear
@@ -180,7 +184,7 @@ class LevelPruner(BasicPruner):
                 
     """
 
-    def __init__(self, model: Module, config_list: List[Dict], mode: str = "normal", balance_gran: list = None):
+    def __init__(self, model: Module, config_list: List[Dict], mode: str = "normal", balance_gran: Optional[List] = None):
         self.mode = mode
         self.balance_gran = balance_gran
         super().__init__(model, config_list)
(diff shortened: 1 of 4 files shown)


linbinskn:
@J-shang @littlelittlehorse
Please help review this pr, thanks.

J-shang:
It is a good feature and I think maybe it could be more powerful. `SparsityAllocator` always corresponds to a `mode`, maybe we can add a `bank` mode here, and bank maybe can be any block size, not only one dimension, https://nni.readthedocs.io/en/stable/Compression/v2_pruning_algo.html#nni.algorithms.compression.v2.pytorch.pruning.TaylorFOWeightPruner

linbinskn:
> It is a good feature and I think maybe it could be more powerful. `SparsityAllocator` always corresponds to a `mode`, maybe we can add a `bank` mode here, and bank maybe can be any block size, not only one dimension, https://nni.readthedocs.io/en/stable/Compression/v2_pruning_algo.html#nni.algorithms.compression.v2.pytorch.pruning.TaylorFOWeightPruner

Have updated, please help review again：）

J-shang:
please fix ut, better to use a sparsity range

linbinskn:
> please fix ut, better to use a sparsity range

Have fixed.
I think keeping using specific sparsity value instead of range here is better since it can help validate correctness of `balance` mode implementation (balance_gran = [5], sparsity = 0.7 -> remain 3 value per block, finally 0.6 sparsity).

J-shang:
> > please fix ut, better to use a sparsity range
> 
> Have fixed. I think keeping using specific sparsity value instead of range here is better since it can help validate correctness of `balance` mode implementation (balance_gran = [5], sparsity = 0.7 -> remain 3 value per block, finally 0.6 sparsity).

Got it

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -137,9 +138,51 @@ class LevelPruner(BasicPruner):
...
+                sparsity: 25%
+                balance_gran: [32, 32]
+
+                pruning result: Weight tensor whose shape is [64, 64] will be split into 4 [32, 32] sub blocks.
+                                Each sub block will be pruned 256 values.
+                
     """
 
-    def __init__(self, model: Module, config_list: List[Dict]):
+    def __init__(self, model: Module, config_list: List[Dict], mode: str = "normal", balance_gran: list = None):
```
J-shang, line 183: `balance_gran: Optional[List] = None`

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -40,6 +41,60 @@ def generate_sparsity(self, metrics: Dict[str, Tensor]) -> Dict[str, Dict[str, T
...
+                sub_mask_str =  "mask[{}] = mask_bank".format(index_str)
+                metric_bank = eval(sub_metric_str)
+                prune_num = int(sparsity_rate * metric_bank.numel())
+                if prune_num == 0:
+                    threshold = metric_bank.min() -1
+                else:
+                    threshold = torch.topk(metric_bank.reshape(-1), prune_num, largest=False)[0].max()
+                # mask_bank will be used in exec(sub_mask_str)
+                mask_bank = torch.gt(metric_bank, threshold).type_as(metric_bank)
+                exec(sub_mask_str)
```
J-shang, line 92: Amazing loop! 😆

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -137,9 +138,51 @@ class LevelPruner(BasicPruner):
...
+
+            example 1:
+                operation: Linear
+                pruning tensor: weight
+                pruning tensor shape: [32, 32]
+                sparsity: 50%
+                balance_gran: [4]
+
+                pruning result: Weight tensor whose shape is [32, 32] will be split into 256 [1, 4] sub blocks.
+                                Each sub block will be pruned 2 values.
```
J-shang, line 169: we should note we will align from the back.

J-shang (approved):

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
linbinskn: Have modified.

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
linbinskn: Good point, added.

littlelittlehorse (approved):

PR #4488, opened 2022-01-20
[Model Compression] Pruning Wrapper Refactor

J-shang:
### Description ###
Refactor the pruner wrapper to preserve masked weight values ​​during pruning.

### Checklist ###
  - [ ] test case
  - [ ] doc

### How to test ###
Run any v2 pruners and check if they can work. Suggest `TaylorFOWeightPruner`.


nni/algorithms/compression/v2/pytorch/pruning/movement_pruner.py
```diff
@@ -38,56 +38,12 @@ class PrunerScoredModuleWrapper(Module):
...
         self.weight_score = Parameter(torch.empty(self.weight.size()))
         torch.nn.init.constant_(self.weight_score, val=0.0)
 
-        # register buffer for mask
-        self.register_buffer("weight_mask", torch.ones(self.module.weight.shape))
-        if hasattr(self.module, 'bias') and self.module.bias is not None:
-            self.register_buffer("bias_mask", torch.ones(self.module.bias.shape))
-            self.bias = Parameter(torch.empty(self.module.bias.size()))
-        else:
-            self.register_buffer("bias_mask", None)
```
QuanluZhang, line 63: so where are masks are generated then?

J-shang: all these logic are moved to its parent's class `PrunerModuleWrapper`

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -699,7 +699,7 @@ def _calculate_taylor_expansion(self, weight_tensor: Tensor, grad: Tensor) -> Te
         return (weight_tensor.detach() * grad.detach()).data.pow(2)
 
     def reset_tools(self):
-        hook_targets = {layer_info.name: layer_info.module.weight for layer_info, _ in self._detect_modules_to_compress()}
+        hook_targets = {name: wrapper.weight for name, wrapper in self.get_modules_wrapper().items()}
```
QuanluZhang, line 702: so, weight as a member variable in wrapper always exists? no matter wrapped or unwrapped?

QuanluZhang:
better to add ut to verify that masked weights are indeed preserved

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: only wrapped wrapper has `weight`, the `weight` will be del when unwrapped.

J-shang: Yes, the weight will always exists... Sometimes we need to `_unwrap_model()` to get the model graph, then `_wrap_model()`, so reserving `weight` after unwrapped is necessary.

J-shang:
> @J-shang better to add ut to verify that masked weights are indeed preserved

Good suggestion, I will add it.

nni/algorithms/compression/v2/pytorch/pruning/movement_pruner.py
QuanluZhang: got it

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
QuanluZhang: got it

J-shang:
> @J-shang better to add ut to verify that masked weights are indeed preserved

add it

QuanluZhang (approved):

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
zheng-ningxin: I'm just curious, what's the specific benefit of this(allocating the weight and bias as  the member variable) compared to the previous way? Performance or cost less memory?

zheng-ningxin (approved):

J-shang:
This pr will merge after doc refactor.

Issue #4454, opened 2022-01-10
NNI v2.7 Iteration Plan


liuzhe-lz:
# Release Plan for upcoming release
- Release manager: @J-shang
- Doc detail items: 2.9
- Feature freeze date: ~2.18~ 3.11
- Code freeze and demo date: ~2.25~
- Bugbash: 4.11
- Branch cut and next release planning date: 4.12
- Release date: ~4.13~ 4.18

### NAS (discuss on 1/17)
- [ ] P0 - Darts NAS example support (Process: training (visualize), model export, retrain, record execution results)
  - [ ] DARTS search space refactor @ dong fan
  - [ ] DARTS NAS alog refactor, and all the core NAS function @ultmaster 
  - [ ] P0 - DARTS end to end owner @JiahangXu (discussion: 1.25) pr: 2.16 (pending for ...)

NAS Search Space Hub and more - @ultmaster and dong fang DL: 2.11
- [ ]  P0 - As a user, when using NNI NAS, I can use Search Space from SearchSpaceHub on a (next step: more, ie. AutoFormer) certain stratgy (tbd by yuge) as a base search space. (能跑，跑对，灵活用）

#
- [x]  P1 - weight inheritance between multi-trial and one-shot @ultmaster
  - [x] - evaluator interface @JiahangXu DL: 2.23
- [ ]  P0 - merging layer-choices into super-kernel.

#
- [ ]  P0 - [NAS Experiment User Story][@QuanluZhang ] As an NNI user, I would like NNI to provide more user-friendly programming interface for designing and tuning DNN models, including more clear APIs, debuggability, functionality (e.g., export best model), properly combining different search modes. (design review: ~1.26~ ~2.9~ 2.14)

### Model Compression
- [ ]  P0 - transformer pruning support @J-shang  (discussion 1.18) (DL: 1.28) #4512 
- [ ]  P0 - peer comparation with competitors (internal summary first) 同比 nnPruning pruning performance, capbility, and results
- [ ] P0 - Support more modules using speedup @littlelittlehorse (pr: 1.24) #4492 
- [ ] P1 - Support pattern-based transformer pruning @linbinskn  #4481
- [ ]  P1 - rethink autoCompression https://nni.readthedocs.io/en/latest/Compression/AutoCompression.html#provide-user-model-for-compression; including: metric, scalar, schedular (discussion 1.18)
[...]

PR #4495, opened 2022-01-22
[pipeline] More aggressive cache

liuzhe-lz:
### Checklist ###
  - [ ] test case
  - [ ] doc


pipelines/templates/install-dependencies.yml
```diff
@@ -0,0 +1,56 @@
...
+    set -e
+    brew install 'swig@3'
+    rm /usr/local/bin/swig
+    ln -s '/usr/local/opt/swig@3/bin/swig' /usr/local/bin/swig
+  displayName: (macOS) Downgrade swig
+  condition: and(succeeded(), contains('${{ parameters.platform }}', 'macos'))
+
+- task: UniversalPackages@0
+  inputs:
+    vstsFeed: NNIOpenSource/dependencies
```
ultmaster, line 28: I think if a PR tries to upgrade ubuntu-latest to, let's say, Python 3.10. When it runs the pipeline check, it will get the wrong binary, which causes pipeline failure.

But on the other hand, this should not happen often. And since dependencies expire every day, that should not take long to heal itself.

ultmaster (approved):
Looks like a clever way to do it! 👍

J-shang (approved):

pipelines/templates/install-dependencies.yml
cruiseliu: Upgrading to new minor version is save, because the prefix will change and the cache will do nothing.

I don't have confidence about micro/patch version though.
There was a micro version upgrade during the development and caused a problem.
I have fixed that specific problem but I don't know if there are more.

dependencies/required_extra.txt
```diff
@@ -7,7 +7,7 @@ smac4nni
 
 # BOHB
 ConfigSpace>=0.4.17
-statsmodels==0.12.0
+statsmodels>=0.12.0
```
paiforsyth, line 10: @liuzhe-lz  Glad to see the requirement on statsmodels is no longer pinned.  This was causing issues for us.  Just want to know if there are plans for a pypi release in the near future that would include this change.  That would be very helpful

ultmaster: link #4454

PR #4512, opened 2022-01-26
[Model Compression] attention mode

J-shang:
### Description ###
Add a new sparsity allocator. Add mode `attention` to `L1NormPruner`, `L2NormPruner`, `TaylorFOWeightPruner`.

This allocator is for pruning Q K V projection. Q K V need to be implemented as `torch.nn.Linear` layers, and their weights should have the same size. Then this allocator will sum their weight metric and apply the same masks to these three layers in one attention head.

### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
+++ b/nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
@@ -184,6 +184,11 @@ class NormPruner(BasicPruner):
         harvest the speed benefit from the pruned model. Note that, if set 'dependency_aware'
         , the dummy_input cannot be None, because the pruner needs a dummy input to trace the
         dependency between the conv layers.
+
+        If prune the model in an attention way, make sure you have (self-attention-like, QKV projection) structure in your model,
+        and they are implemented with `torch.nn.Linear` (like Bert implemented in hugging face).
+        In this mode, NNI will find these three Linear layers in each attention,
+        then apply the mask on the same position among three layers as possible.
     dummy_input : Optional[torch.Tensor]
         The dummy input to analyze the topology constraints. Note that, the dummy_input
         should on the same device with the model.
@@ -245,6 +250,11 @@ class L1NormPruner(NormPruner):
         harvest the speed benefit from the pruned model. Note that, if set 'dependency_aware'
         , the dummy_input cannot be None, because the pruner needs a dummy input to trace the
         dependency between the conv layers.
+
+        If prune the model in an attention way, make sure you have (self-attention-like, QKV projection) structure in your model,
+        and they are implemented with `torch.nn.Linear` (like Bert implemented in hugging face).
+        In this mode, NNI will find these three Linear layers in each attention,
+        then apply the mask on the same position among three layers as possible.
     dummy_input : Optional[torch.Tensor]
         The dummy input to analyze the topology constraints. Note that, the dummy_input
         should on the same device with the model.
@@ -278,6 +288,11 @@ class L2NormPruner(NormPruner):
         harvest the speed benefit from the pruned model. Note that, if set 'dependency_aware'
         , the dummy_input cannot be None, because the pruner needs a dummy input to trace the
         dependency between the conv layers.
+
+        If prune the model in an attention way, make sure you have (self-attention-like, QKV projection) structure in your model,
+        and they are implemented with `torch.nn.Linear` (like Bert implemented in hugging face).
+        In this mode, NNI will find these three Linear layers in each attention,
+        then apply the mask on the same position among three layers as possible.
     dummy_input : Optional[torch.Tensor]
         The dummy input to analyze the topology constraints. Note that, the dummy_input
         should on the same device with the model.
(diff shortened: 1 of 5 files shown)


nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -176,3 +177,108 @@ def _group_metric_calculate(self, group_metrics: Union[Dict[str, Tensor], List[T
...
+
+    def _get_dependency(self):
+        """
+        Populate self.attention_name_groups by running inference on the module graph.
+        Currently, the group inferred AttentionWeightDependency is limited to a set of four weights, with the first
+        three corresponding to :math:`Q_{proj}`, :math:`K_{proj}`, :math:`V_{proj}` (in any order) and the last one being :math:`{Output}_{proj}`.
+        """
+        try:
+            graph = self.pruner.generate_graph(dummy_input=self.dummy_input)
+            dependency_tracer = AttentionWeightDependency(traced_model=graph.trace)
```
zheng-ningxin, line 201: The logic of `AttentionWeightDependency` is not robust enough, maybe we can implement a function to find the exact same subgraph with `SelfAttention`?

J-shang: you are right, I will check if `SelfAttention` implementation is similar in other transformers, and find a better way to match the attention.

linbinskn: What is the constraint of `AttentionWeightDependency`, only support specific module type (such as only torch.nn.Module, not support F function) or also not support other transformer implementation (such as only support Huggingface)? May be we should provide more notes to help user notice these constraints.

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -176,3 +177,108 @@ def _group_metric_calculate(self, group_metrics: Union[Dict[str, Tensor], List[T
...
+        # Group the QKV metrics in one head.
+        for idx, names in enumerate(self.attention_name_groups):
+            group_metric_dict = OrderedDict()
+            for name in names[0: 3]:
+                # If the Linear will be pruned, add it to this group.
+                if name in metrics:
+                    group_metric_dict[name] = metrics[name]
+                    if self.continuous_mask:
+                        group_metric_dict[name] *= self._compress_mask(self.pruner.get_modules_wrapper()[name].weight_mask)
+            grouped_metrics[idx] = group_metric_dict
```
linbinskn, line 233: It seems that we don't support `op_names` in attention mode. If module is pruned by setting `op_names` in config_list while `self.attention_name_groups` recording all name of Q, K, V, error may be raised on line 237 since some value of `grouped_metrics` will be empty.

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
```diff
@@ -176,3 +177,108 @@ def _group_metric_calculate(self, group_metrics: Union[Dict[str, Tensor], List[T
...
+        # Calculate group metric.
+        for idx, group_metric_dict in grouped_metrics.items():
+            group_metric = self._group_metric_calculate(group_metric_dict)
+            for name in group_metric_dict.keys():
+                metrics[name] = group_metric
+
+            # Special rule based output layer mask generation.
+            output_layer_name = self.attention_name_groups[idx][-1]
+            if output_layer_name in metrics and self.dim and len(self.dim) == 1 and self.dim[0] == 0:
+                # Don't use colculated metric, use group metric.
```
linbinskn, line 244: colculated->calculated

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
```diff
@@ -635,7 +653,7 @@ def trainer(model: Module, optimizer: Optimizer, criterion: Callable[[Tensor, Te
     training_batches : int
         The batch number used to collect activations.
     mode : str
-        'normal', 'dependency_aware' or 'global'.
+        'normal', 'dependency_aware', 'global', 'attention'.
```
QuanluZhang, line 656: is 'attention' qualified to be one mode?

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
J-shang: For checking several Huggingface transformer implementations, `bert`, `roberta`, `gpt_neo`, `bart` `bait`...have similar `Q K V Projection Structure`.
And implementations from other repos or even in Huggingface may not have the similar structure, so I think this mode can only support huggingface implementations right now.
I will add more information to explain these constraints.

nni/algorithms/compression/v2/pytorch/pruning/tools/sparsity_allocator.py
J-shang: you are right, this place will raise an error, I will fix it.

nni/algorithms/compression/v2/pytorch/pruning/basic_pruner.py
J-shang: I think it is a pruning mode because it will prune multiple layers as a whole

PR #4569, opened 2022-02-22
Retry logic for getting experiment metadata failed

J-shang:
### Description ###
Sometimes, `nnictl stop --all` failed in the unit test.

https://msrasrg.visualstudio.com/NNIOpenSource/_build/results?buildId=32629&view=logs&jobId=95b93358-fda6-5c72-3eb7-ee58b0085220&j=95b93358-fda6-5c72-3eb7-ee58b0085220&t=6d99b3d4-eb6f-53b7-dead-77c8a6940a44&s=859b8d9a-8fd6-5a5c-6f5e-f84f1990894e

Two reasons may cause getting experiment metadata failed from `~/nni-experiments/.experiment`.
1. metadata does not exist
2. read `.experiment` failed

This issue is hard to reproduce, so just retry for now.

### How to test ###
Create several experiments and `nnictl stop --all`.

### DIFF
--- a/nni/tools/nnictl/nnictl_utils.py
+++ b/nni/tools/nnictl/nnictl_utils.py
@@ -222,6 +222,7 @@ def stop_experiment(args):
     if experiment_id_list:
         for experiment_id in experiment_id_list:
             print_normal('Stopping experiment %s' % experiment_id)
+            # FIXME: Retry should be placed to `Experiments`, need review both python and ts code.
             # retry up to 10 times to get the experiment metadata
             for i in range(1, 11):
                 experiments_dict = Experiments().get_all_experiments()
--- a/nni/tools/nnictl/nnictl_utils.py
+++ b/nni/tools/nnictl/nnictl_utils.py
@@ -222,9 +222,18 @@ def stop_experiment(args):
     if experiment_id_list:
         for experiment_id in experiment_id_list:
             print_normal('Stopping experiment %s' % experiment_id)
-            experiments_config = Experiments()
-            experiments_dict = experiments_config.get_all_experiments()
-            rest_pid = experiments_dict.get(experiment_id).get('pid')
+            # retry up to 10 times to get the experiment metadata
+            for i in range(1, 11):
+                experiments_dict = Experiments().get_all_experiments()
+                experiment_info = experiments_dict.get(experiment_id)
+                if experiment_info is None:
+                    print_warning('Get experiment {} metadata failed, {} time retry...'.format(experiment_id, i))
+                else:
+                    break
+            if experiment_info is None:
+                print_error('Experiment {} metadata getting failed, please manually check its status in `~/nni-experiments/.experiment`.')
+                exit(1)
+            rest_pid = experiment_info.get('pid')
             if rest_pid:
                 kill_command(rest_pid)
             print_normal('Stop experiment success.')
--- a/nni/tools/nnictl/nnictl_utils.py
+++ b/nni/tools/nnictl/nnictl_utils.py
@@ -228,10 +228,13 @@ def stop_experiment(args):
                 experiment_info = experiments_dict.get(experiment_id)
                 if experiment_info is None:
                     print_warning('Get experiment {} metadata failed, {} time retry...'.format(experiment_id, i))
+                    time.sleep(0.5)
                 else:
                     break
             if experiment_info is None:
-                print_error('Experiment {} metadata getting failed, please manually check its status in `~/nni-experiments/.experiment`.')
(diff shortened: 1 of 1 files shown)


nni/tools/nnictl/nnictl_utils.py
```diff
@@ -222,9 +222,18 @@ def stop_experiment(args):
...
             print_normal('Stopping experiment %s' % experiment_id)
-            experiments_config = Experiments()
-            experiments_dict = experiments_config.get_all_experiments()
-            rest_pid = experiments_dict.get(experiment_id).get('pid')
+            # retry up to 10 times to get the experiment metadata
+            for i in range(1, 11):
+                experiments_dict = Experiments().get_all_experiments()
+                experiment_info = experiments_dict.get(experiment_id)
+                if experiment_info is None:
+                    print_warning('Get experiment {} metadata failed, {} time retry...'.format(experiment_id, i))
```
ultmaster, line 230: time.sleep

nni/tools/nnictl/nnictl_utils.py
```diff
@@ -222,9 +222,18 @@ def stop_experiment(args):
     if experiment_id_list:
         for experiment_id in experiment_id_list:
             print_normal('Stopping experiment %s' % experiment_id)
-            experiments_config = Experiments()
-            experiments_dict = experiments_config.get_all_experiments()
-            rest_pid = experiments_dict.get(experiment_id).get('pid')
+            # retry up to 10 times to get the experiment metadata
+            for i in range(1, 11):
+                experiments_dict = Experiments().get_all_experiments()
+                experiment_info = experiments_dict.get(experiment_id)
```
ultmaster, line 228: Could it be possible that the experiment doesn't exist at all? Maybe `experiment_id` is wrong.

nni/tools/nnictl/nnictl_utils.py
```diff
@@ -222,9 +222,18 @@ def stop_experiment(args):
...
+                if experiment_info is None:
+                    print_warning('Get experiment {} metadata failed, {} time retry...'.format(experiment_id, i))
+                else:
+                    break
+            if experiment_info is None:
+                print_error('Experiment {} metadata getting failed, please manually check its status in `~/nni-experiments/.experiment`.')
```
ultmaster, line 234: Suggest putting debugging info here so that we can do a diagnosis when similar issues appears on pipeline.

nni/tools/nnictl/nnictl_utils.py
```diff
@@ -222,9 +222,18 @@ def stop_experiment(args):
     if experiment_id_list:
         for experiment_id in experiment_id_list:
             print_normal('Stopping experiment %s' % experiment_id)
-            experiments_config = Experiments()
-            experiments_dict = experiments_config.get_all_experiments()
-            rest_pid = experiments_dict.get(experiment_id).get('pid')
+            # retry up to 10 times to get the experiment metadata
```
ultmaster, line 225: Only stop has this problem?

nni/tools/nnictl/nnictl_utils.py
J-shang: `experiment_id` should be in `.experiemnt`, because we get this `experiment_id` from `.experiemnt`, but yes, maybe `experiment_id` is deleted from this file before we read it, or the whole `.experiemnt` file is deleted...

nni/tools/nnictl/nnictl_utils.py
J-shang: No, a better way is to put this retry into `Experiments`, but in this way, we need to align both `python` and `ts` code.

nni/tools/nnictl/nnictl_utils.py
J-shang: I think this may mainly cause by the file being modified when we read the file... the lock will be expired in 2s, but the action on file is over 2s, I'm not sure.

nni/tools/nnictl/nnictl_utils.py
J-shang: add it

nni/tools/nnictl/nnictl_utils.py
J-shang: print information in `.experiment`

nni/tools/nnictl/nnictl_utils.py
ultmaster: Well, in that case, add a FIXME in code as reminder.

J-shang: add it

ultmaster (approved):

QuanluZhang (approved):

PR #4593, opened 2022-02-26
Bug: reduce liveEnvironmentsCount after environment is stopped

weidankong:
### Description ###
liveEnvironmentsCount failed to reduce after existing live environment is stopped.

### Checklist ###
  - [ ] test case
  - [ ] doc

### How to test ###
Create a reuseable training service and set the reuseMode=False.
Environments will be created and stop after use, new environment will be created.

### DIFF
--- a/ts/nni_manager/training_service/reusable/trialDispatcher.ts
+++ b/ts/nni_manager/training_service/reusable/trialDispatcher.ts
@@ -507,6 +507,7 @@ class TrialDispatcher implements TrainingService {
                                 throw new Error(`${environment.id} does not has environment service!`);
                             }
                             await environment.environmentService.stopEnvironment(environment);
+                            liveEnvironmentsCount--;
                             continue;
                         }
 


ts/nni_manager/training_service/reusable/trialDispatcher.ts
```diff
@@ -507,6 +507,7 @@ class TrialDispatcher implements TrainingService {
                                 throw new Error(`${environment.id} does not has environment service!`);
                             }
                             await environment.environmentService.stopEnvironment(environment);
+                            liveEnvironmentsCount--;
```
liuzhe-lz, line 510: I'm not really sure whether this is correct.
Suggest to merge first and I'll test it later.

liuzhe-lz (approved):

ts/nni_manager/training_service/reusable/trialDispatcher.ts
J-shang: got it.

J-shang (approved):

PR #4551, opened 2022-02-14
Autoformer search space

nbl97:
### Description ###
We build the search space of Autoformer in `nni/examples/nas/hub/searchspace/autoformer.py`.

There are 4 searchable variables, depth, embedding dimension, heads number and MLP ratio. 
-	Depth, the number of layers of the architecture.
-	Embedding dimension, the embedding dimension in each layer. Note that it is shared in each layer which means the architecture is non-hierarchical. 
-	Heads number, the number of heads in self-attn in each layer. The heads number in each layer is independent, like the kernel size in traditional CNN search space.
-	MLP ratio, the ratio in FFN in each layer. Like the heads number, the MLP ratio is also independent in each layer.

For depth and embedding dimension, we use nn.ValueChoice() to select the values and pass the embedding dimension as a parameter to each layer to keep the dimension consistent.
For the heads number and MLP ratio, we use nn.LayerChoice() to select from the candidates in each layer.

The weight entanglement technique is not supported in the current version. We will follow the latest NNI framework and update the search space in time.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###
We used the coming features of nn.ValueChoice(), such as multiplying the value with a constant or passing the value to nn.Parameter(). We will test the search space once the new features are supported. I look forward to more discussion about Autoformer_Space2NNI.

PR #4598, opened 2022-03-01
Support ValueChoice as depth in Repeat

ultmaster:
### Description ###

Repeat can previously accept a fixed depth, or a range of depth.
Added in this PR is a support of ValueChoice as depth.

One prerequisite is to support ``all_options()`` for value choice, which is a iterable of all the possibilities (combinations) of a ValueChoice.

### Checklist ###
  - [x] test case
  - [x] doc


QuanluZhang:
seems not necessary, though maybe better to have

nni/retiarii/converter/graph_gen.py
```diff
@@ -660,8 +660,9 @@ def _convert_module(self, script_module, module, module_name, module_python_name
             attrs = {
                 'mutation': 'repeat',
                 'label': module.label,
+                'depth': module.depth_choice,
+                'max_depth': module.max_depth,
```
QuanluZhang, line 664: what is the difference between depth and max_depth?

ultmaster: `depth` is the chosen depth, `max_depth` and `min_depth` is a depth range.

nni/retiarii/nn/pytorch/component.py
```diff
@@ -57,15 +69,32 @@ def __init__(self,
...
+            self.min_depth = depth if isinstance(depth, int) else depth[0]
+            self.max_depth = depth if isinstance(depth, int) else depth[1]
+            self.depth_choice = ValueChoice(list(range(self.min_depth, self.max_depth + 1)), label=label)
+        elif isinstance(depth, int):
+            self.min_depth = self.max_depth = depth
+            self.depth_choice = depth
+        else:
+            raise TypeError(f'Unsupported "depth" type: {type(depth)}')
         assert self.max_depth >= self.min_depth > 0
         self.blocks = nn.ModuleList(self._replicate_and_instantiate(blocks, self.max_depth))
```
QuanluZhang, line 93: by the way, why must be max_depth here?

nni/retiarii/nn/pytorch/mutator.py
```diff
@@ -144,14 +144,13 @@ def _retrieve_chain_from_graph(self, graph: Graph) -> List[Node]:
...
-        max_depth = self.nodes[0].operation.parameters['max_depth']
-        if min_depth < max_depth:
-            chosen_depth = self.choice(list(range(min_depth, max_depth + 1)))
         for node in self.nodes:
             # the logic here is similar to layer choice. We find cell attached to each node.
             target: Graph = model.graphs[node.operation.cell_name]
             chain = self._retrieve_chain_from_graph(target)
+            # and we get the chosen depth (by value choice)
+            node_in_model = model.get_node_by_name(node.name)
+            chosen_depth = node_in_model.operation.parameters['depth']
```
QuanluZhang, line 153: what is the type of `chosen_depth`? int or valuechoice?

QuanluZhang: which part decides the chosen depth?

test/ut/retiarii/test_highlevel_apis.py
```diff
@@ -578,14 +578,39 @@ def forward(self, x):
...
+    def test_repeat_static(self):
+        class AddOne(nn.Module):
+            def forward(self, x):
+                return x + 1
+
+        @model_wrapper
+        class Net(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.block = nn.Repeat(lambda index: nn.LayerChoice([AddOne(), nn.Identity()]), 4)
```
QuanluZhang, line 596: is there valuechoice tested?

matluster:
> seems not necessary, though maybe better to have

There have been already use cases in #4524 and #4551.

What I'm trying to say, is that, I didn't "imagine" that people could be using value choice in repeat. They are actually using it "for real".

test/ut/retiarii/test_highlevel_apis.py
matluster: oops, missed a test.

Thanks for pointing this out. It also helped me catch a bug. :(

nni/retiarii/nn/pytorch/component.py
matluster: `max_depth` is used to initialize "enough" blocks when super-net is created. Useful in weight sharing.

nni/retiarii/nn/pytorch/mutator.py
matluster: It's originally a ValueChoice, but it has already been mutated by a ParameterChoiceMutator here.

QuanluZhang (approved):

liuzhe-lz (approved):

PR #4609, opened 2022-03-03
Hyper-parameter Choice in Retiarii

ultmaster:
### Description ###

**Why?** Currently ValueChoice can be only used as arguments of `basic_unit`. There are so many other use cases it can't handle. For example, arguments of `torch.zeros`, arguments of functions used in forward, or arbitrary for-loop and if-condition. Note that these cases can be simply implemented (with config) if users are NOT using NNI). That means, Retiarii is becoming a framework burden when users are trying to implement these search spaces.

**How?** Proposed in this PR is a new API called HyperParameterChoice. It's just like a config of model. Users only need to declare it. It automatically registers itself into the model wrapper, and they can use it just like a normal python value.

**Limitations.** It only supports multi-trial NAS with Python execution engine. For more powerful support from Retiarii, other APIs like ValueChoice, Repeat are still highly recommended.

### Checklist ###
  - [x] test case
  - [ ] doc


nni/retiarii/utils.py
```diff
@@ -115,43 +117,78 @@ def top(cls, key: str) -> Any:
...
         # For example, currently the top of stack is [1, 2, 2], and [1, 2, 2, 3] is used,
         # the next thing up is [1, 2, 2, 4].
         # `reset_uid` to count from zero for "model_wrapper_1_2_2_4"
         try:
-            current_context = ContextStack.top(self.key)
-            next_uid = uid(self._simple_name(self.key, current_context))
-            ContextStack.push(self.key, current_context + [next_uid])
-            reset_uid(self._simple_name(self.key, current_context + [next_uid]))
+            parent_context: 'ModelNamespace' = ModelNamespace.current_context(self.key)
+            next_uid = uid(parent_context._simple_name())
```
QuanluZhang, line 149: why using uid rather than string (e.g., module name)

QuanluZhang (approved):

nni/retiarii/utils.py
matluster: We can't get module name because they are assigned in the parent module. At least not without hacks like `inspect.get_frame`.

Instead, uid is simple and globally unique.

liuzhe-lz (approved):

PR #4612, opened 2022-03-03
[doc] experiment refactor

QuanluZhang:


docs/source/tutorials/hello_nas.rst
```diff
@@ -496,7 +496,7 @@ Visualize the Experiment
 
 Users can visualize their experiment in the same way as visualizing a normal hyper-parameter tuning experiment.
 For example, open ``localhost:8081`` in your browser, 8081 is the port that you set in ``exp.run``.
-Please refer to :doc:`here </experiment/webportal>` for details.
+Please refer to :doc:`here </experiment/webui>` for details.
```
ultmaster, line 499: Please don't modify this file directly. Instead, modify the examples python code, and rerun `make html` to build the rst.

QuanluZhang: got it. it is pretty easy to make this mistake, we should come up with an approach to avoid such mistake

Issue #4592, opened 2022-02-25
examples/nas/multi-trial/mnasnet/search.py cannot run


Tonyhao96:
**Describe the issue**:

Cannot run multi-trial nas example for mnasnet. I never modified the model code.  Error in trail `stderr` in nni-experiment folder:
```bash
  File "/home/xxx/miniconda3/lib/python3.9/site-packages/pytorch_lightning/core/lightning.py", line 1434, in backward
    loss.backward(*args, **kwargs)
  File "/home/xxx/miniconda3/lib/python3.9/site-packages/torch/_tensor.py", line 307, in backward
    torch.autograd.backward(self, gradient, retain_graph, create_graph, inputs=inputs)
  File "/home/xxx/miniconda3/lib/python3.9/site-packages/torch/autograd/__init__.py", line 154, in backward
    Variable._execution_engine.run_backward(
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.FloatTensor [100, 1280]], which is output 0 of ReluBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

**Environment**:
- NNI version: 2.6.1
- Training service (local|remote|pai|aml|etc):local, with 4 RTX 3090
- Client OS: Ubuntu 20.04
- Server OS (for remote mode only):
- Python version: 3.9.7
- PyTorch/TensorFlow version:PyTorch 1.10.2
- Is conda/virtualenv/venv used?: Yes, use conda
- Is running in Docker?: No

**Configuration**:
 - Experiment config (remember to remove secrets!):
 - Search space:

I only add `exp_config.trial_gpu_number = 1` and ` gpus=1` as below:

```python
trainer = pl.Classification(
        train_dataloader=pl.DataLoader(train_dataset, batch_size=100),
        val_dataloaders=pl.DataLoader(test_dataset, batch_size=100),
        max_epochs=2,
        gpus=1,
    )
```

**How to reproduce it?**:
Run `python search.py`

ultmaster:
@QuanluZhang Looks like a graph-engine issue. Would you please take a look?

JiahangXu:
Hi, it seems the back propagation is interrupted by the dropout layer in classifier. You can remove `inplace=True` in line 184 from `./base_mnasnet.py` and run `python search.py` again.
``` python
[...]

PR #4614, opened 2022-03-04
Fix bug Issue4592

JiahangXu:
### Description ###

fix bug in issue #4592 

### Checklist ###
  - [ ] test case
  - [ ] doc


examples/nas/multi-trial/mnasnet/base_mnasnet.py
```diff
@@ -4,8 +4,8 @@
 
 import torch
 import torch.nn as torch_nn
-from torchvision.models.utils import load_state_dict_from_url
 import torch.nn.functional as F
+from nni.retiarii.serializer import model_wrapper
```
ultmaster, line 8:
```python
from nni.retiarii import model_wrapper
```

examples/nas/multi-trial/mnasnet/base_mnasnet.py
```diff
@@ -180,7 +180,7 @@ def __init__(self, alpha, depths, convops, kernel_sizes, num_layers,
             nn.ReLU(inplace=True),
         ]
         self.layers = nn.Sequential(*layers)
-        self.classifier = nn.Sequential(nn.Dropout(p=dropout, inplace=True),
```
QuanluZhang, line 183: why adding `inplace=True` induces problem?

JiahangXu: Error info says 'one of the variables needed for gradient computation has been modified by an inplace operation.'
The new version of PyTorch does not support inplace operation, so that the back propagation is interrupted by the `inplace` operation in dropout layer. Here is a similar issue in pytorch: https://discuss.pytorch.org/t/help-runtimeerror-one-of-the-variables-needed-for-gradient-computation-has-been-modified-by-an-inplace-operation/107030/2.

QuanluZhang: so this is an issue of pytorch not nni, right?

QuanluZhang (approved):

ultmaster (approved):

PR #4618, opened 2022-03-04
[doc] Full reference TOC

ultmaster:
### Description ###

Add a reference TOC with auto-summary.

Some files need ``__all__`` to make pipeline happy. However, it doesn't mean that other files with `__all__` don't need one. Ideally, they all should have `__all__` to avoid situations like this:

image

There are many classes not written by NNI got rendered. I'll try to solve it by patching autosummary. If it doesn't work, we have to examine `__all__` of all the files.

### Checklist ###
  - [x] test case - no need
  - [x] doc


ultmaster:
Some changes are from #4613

PR #4669, opened 2022-03-21
Typehint and copyright header

liuzhe-lz:
### Description ###

Add type check to pipeline.

Currently NAS and compression are not verified.

In `pyrightconfig.json`:

```json
{
    "ignore": [
        "nni/algorithms",
        "nni/common/device.py",
        "nni/common/graph_utils.py",
        "nni/common/serializer.py",
        "nni/compression",
        "nni/nas",
        "nni/retiarii",
        "nni/smartparam.py",
        "nni/tools"
    ]
}
```

### Checklist ###
  - ~~test case~~
  - ~~doc~~

### How to test ###

No need.


nni/typehint.py
```diff
@@ -0,0 +1,34 @@
...
+    TrialMetric: TypeAlias = float
+
+    class TrialRecord(TypedDict):
+        parameter: Parameters
+        value: TrialMetric
+
+else:
+    from typing import Any
+
+    Literal = Any
```
matluster, line 28: from typing_extension import?

ultmaster (approved):

nni/common/hpo_utils/validation.py
```diff
@@ -1,8 +1,10 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from __future__ import annotations
+
 import logging
-from typing import Any, List, Optional
+from typing import Any
```
J-shang, line 7: we recommended using `list` not `List`, and `| None` not `Optional`?

J-shang (approved):

nni/common/hpo_utils/validation.py
cruiseliu: I personally prefer the new way. It think it looks much cleaner, and is closer to TypeScript style, which saves mental context switch overhead for me.

cruiseliu: If other modules choose to use 3.7 grammar, I think it's pretty okay and no need to change.
Notably `nni.experiment.config` will keep 3.7 grammar for years. Because it needs runtime checking and I'm afraid new grammar will not work.

PR #4670, opened 2022-03-21
Try to solve conflicts

liuzhe-lz:
### Checklist ###
  - [ ] test case
  - [ ] doc


J-shang:
@ultmaster pls help to check serializer.py

matluster:
Accepting both change: Change `_check_wrapped(cls)` to `if _check_wrapped(cls, ...): return cls`.

PR #4679, opened 2022-03-22
Deduplicate generated parameters for TPE and Random

liuzhe-lz:
### Description ###

Users have complained about new TPE generating duplicate parameters.

This PR ensures that TPE and Random will never generate same params. When the search space is exhausted, they raise no more trial.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###

Run HPO experiment with TPE and Random tuner on small search space. Check the parameters are different.

### DIFF
--- a/nni/common/hpo_utils/dedup.py
+++ b/nni/common/hpo_utils/dedup.py
@@ -1,0 +1,66 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+Deduplicate repeated parameters.
+"""
+
+import logging
+
+import nni
+from .formatting import deformat_parameters
+
+_logger = logging.getLogger(__name__)
+
+# TODO:
+# Move main logic of basic tuners (random and grid search) into SDK,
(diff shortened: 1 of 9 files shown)


ultmaster (approved):

nni/common/hpo_utils/dedup.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if self._grid_search is None:
+            _logger.info(f'Tuning algorithm generated duplicate parameter: {formatted_parameters}')
+            _logger.info(f'Use grid search for deduplication.')
+            self._init_grid_search()
+
+        while True:
+            new = self._grid_search._suggest()
+            if new is None:
+                raise nni.NoMoreTrialError()
```
J-shang, line 57: gridsearch is a good way to check if it is `NoMoreTrial`, but why directly return a `new` not a `params` re-generated by the original tuner?

J-shang: got it

PR #4691, opened 2022-03-24
[Compression] add channel pruning mode for admm pruner & optimize movement pruning performance

J-shang:
### Description ###
AutoCompress using admm channel-wise pruning in their paper, so support channel pruning for admm.

For the movement pruning part, I don't know why training gets slower and slower if pass `buffer` to an autograd function without `.clone().detach_()`...

### How to test ###
pass `granularity='coarse-grained'` to admm to test it.


examples/model_compress/pruning/v2/movement_pruning_glue.py
```diff
@@ -113,7 +116,7 @@ def preprocess_function(examples):
     # make sure you have used nni.trace to wrap the optimizer class before initialize
     traced_optimizer = nni.trace(Adam)(model.parameters(), lr=2e-5)
     pruner = MovementPruner(model, config_list, p_trainer, traced_optimizer, criterion, training_epochs=10,
-                            warm_up_step=3000, cool_down_beginning_step=27000)
+                            warm_up_step=12272, cool_down_beginning_step=110448)
```
QuanluZhang, line 119: what is the reason of doing this change?

J-shang: just because I change the data-size-per-optimized-step from 128 to 32, for aligning the paper, then the corresponding `step` magnified four times...

QuanluZhang (approved):

nni/algorithms/compression/v2/pytorch/pruning/tools/metrics_calculator.py
```diff
@@ -19,7 +19,7 @@ class StraightMetricsCalculator(MetricsCalculator):
     def calculate_metrics(self, data: Dict[str, Tensor]) -> Dict[str, Tensor]:
         metrics = {}
         for name, tensor in data.items():
-            metrics[name] = tensor.clone().detach()
+            metrics[name] = tensor.clone().detach_()
```
zheng-ningxin, line 22: Could you please add more comments to explain why `detach_` rather that `detach`?

J-shang: add comments, in fact, just avoid creating a new tensor

zheng-ningxin (approved):

PR #4700, opened 2022-03-25
Fix typehint

liuzhe-lz:
### Description ###

#4679 was created before enabling type checking in pipeline.

### DIFF
--- a/nni/common/hpo_utils/dedup.py
+++ b/nni/common/hpo_utils/dedup.py
@@ -7,16 +7,21 @@ Deduplicate repeated parameters.
 No guarantee for forward-compatibility.
 """
 
+from __future__ import annotations
+
 import logging
+import typing
 
 import nni
 from .formatting import FormattedParameters, FormattedSearchSpace, ParameterSpec, deformat_parameters
 
-_logger = logging.getLogger(__name__)
-
 # TODO:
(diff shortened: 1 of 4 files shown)


nni/common/hpo_utils/dedup.py
```diff
@@ -52,29 +57,29 @@ def __call__(self, formatted_parameters):
...
         self._grid_search = GridSearchTuner()
         self._grid_search.history = self._history
         self._grid_search.space = self._space
         self._grid_search._init_grid()
 
-    def _not_dup(self, formatted_parameters):
+    def _not_dup(self, formatted_parameters: FormattedParameters) -> bool:
         params = deformat_parameters(formatted_parameters, self._space)
-        params_str = nni.dump(params, sort_keys=True)
+        params_str = typing.cast(str, nni.dump(params, sort_keys=True))
```
cruiseliu, line 75: @ultmaster I think it would be better to divide "dump to string" and "dump to bytes".

J-shang (approved):

nni/common/hpo_utils/dedup.py
ultmaster: I'll do that in another PR.

ultmaster (approved):

PR #4703, opened 2022-03-26
Refactor NNI manager globals (step 3) - globals module

liuzhe-lz:
### Description

Introduce a `globals` module to collect all global variables in `global.nni`.

This is essential for 3rd-party training service to become practical.

Unlike Python, each Node.js package has its own dependency directory. If package `A` and `B` both depends on package `C`, there will be 2 copies of `C`. This causes problems when `C` has singletons like logger stream, port listener, etc.

The simplest solution is to utilize `global`, the special singleton offered by Node.js runtime. This namespace object is shared by all modules, no matter where they are. So we will use it to share singletons and constant information between NNI manager core and training services.

This is the first PR and it initiates the infrastructure. Now the globals module contains command line args and experiment paths. With following up PRs, the module will add logger stream, shutdown notifier, and express router register.

### Checklist

- [x] test case
- [x] doc

### How to Test

Run ordinary experiments.

### Release Note

No need to mention.

### DIFF
--- a/ts/nni_manager/common/globals/index.ts
+++ b/ts/nni_manager/common/globals/index.ts
@@ -24,7 +24,7 @@ export { NniManagerArgs, NniPaths };
 /**
  *  Collection of global objects.
  *
- *  It can be obtained with `import globals from '...'` or `globals.nni`.
+ *  It can be obtained with `import globals from 'common/globals'` or `global.nni`.
  *  The former is preferred because it exposes less underlying implementations.
  **/
 export interface NniGlobals {
@@ -37,13 +37,17 @@ declare global {
     var nni: NniGlobals;  // eslint-disable-line
 }
 
+// prepare the namespace object and export it
 if (global.nni === undefined) {
     global.nni = {} as NniGlobals;
 }
 const globals: NniGlobals = global.nni;
 export default globals;
 
-// Must and must only be invoked once in "main.ts".
+/**
+ *  Initialize globals.
+ *  Must and must only be invoked once in "main.ts".
+ **/
 export function initGlobals(): void {
     assert.deepEqual(global.nni, {});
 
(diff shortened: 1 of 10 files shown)


ultmaster (approved):

ts/nni_manager/common/globals/index.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ *  Collection of global objects.
+ *
+ *  Although global is anit-pattern in OOP, there are two scenarios NNI uses globals.
```
matluster, line 7: anti-pattern

ultmaster (approved):

ts/nni_manager/common/globals/unittest.ts
```diff
@@ -0,0 +1,65 @@
...
+type Mutable<Type> = {
+    -readonly [Property in keyof Type]: Type[Property];
+};
+
+export interface MutableGlobals {
+    args: Mutable<NniManagerArgs>;
+    paths: Mutable<NniPaths>;
+}
+
+export function resetGlobals(): void {
```
QuanluZhang, line 32: why this file is named unittest?

ts/nni_manager/rest_server/index.ts
```diff
@@ -41,7 +40,6 @@ import { createRestHandler } from './restHandler';
  *  RestServer must be initialized with start() after NNI manager constructing, but not necessarily after initializing.
  *  This is because RestServer needs NNI manager instance to register API handlers.
  **/
-@Singleton
```
QuanluZhang, line 44: why removing this annotation?

QuanluZhang (approved):

PR #4705, opened 2022-03-27
Refactor NNI manager globals (step 4) - logging

liuzhe-lz:
### Description

Make logging system global and robust.

It also changed a behavior that when using `foreground=True`, logs will be written to stdout AND nnimanager.log.
Previously in foreground mode it does not write to log file.

### Checklist

- [x] test case
- [x] doc

### How to Test

Run ordinary experiments in foreground and background mode, then check log file.

### Release Note

- [Experiment] Foreground mode now saves `nnimanager.log`.


ts/nni_manager/common/globals/log_stream.ts
```diff
@@ -0,0 +1,68 @@
...
+    private toConsole: boolean;
+
+    constructor(logFile: string, toConsole: boolean) {
+        this.logFileFd = fs.openSync(logFile, 'a');
+        this.toConsole = toConsole;
+    }
+
+    public writeLine(line: string): void {
+        this.buffer.push(line);
+        this.flush();
```
matluster, line 37: How do you catch errors raised from flush?

ts/nni_manager/common/log.ts
```diff
@@ -1,141 +1,116 @@
...
-    [ERROR, 'ERROR'],
-    [WARNING, 'WARNING'],
-    [INFO, 'INFO'],
-    [DEBUG, 'DEBUG'],
-    [TRACE, 'TRACE'],
-]);
+const levelNameToValue = { trace: 0, debug: 10, info: 20, warning: 30, error: 40, critical: 50 } as const;
 
-/* global_ states */
+const loggers: Record<string, Logger> = {};
```
matluster, line 21: Didn't use globals here. Means that there can be multiple `loggers` available, right?

ts/nni_manager/common/globals/unittest.ts
```diff
@@ -41,14 +45,18 @@ export function resetGlobals(): void {
...
     const paths = createPaths(args);
+    const logStream = {
+        writeLine: (_line: string): void => { /* dummy */ },
+        close: (): void => { /* dummy */ }
+    };
 
-    const globals = { args, paths };
-    if (global.nni === undefined) {
-        global.nni = globals;
+    const globalAsAny = global as any;
```
QuanluZhang, line 54: where is `global` defined?

ts/nni_manager/common/globals/log_stream.ts
cruiseliu: Not handled yet.
Seems we have discussed it in the review meeting. But I forgot to fix it.

ts/nni_manager/common/log.ts
cruiseliu: Yes. `Logger` is simply a formatter. It's safe to have multiple loggers.

ts/nni_manager/common/globals/unittest.ts
cruiseliu: It's node.js built-in.

ts/nni_manager/common/globals/log_stream.ts
cruiseliu: Done. I decided to use synchronous write for robust logger. In critical area speed is not a thing.

ts/nni_manager/common/log.ts
```diff
@@ -1,141 +1,119 @@
...
+    }
+    return loggers[name];
+}
 
-/* major api */
+/**
+ *  A special logger prints to stderr when the logging system has problems.
+ *  For modules that are responsible for handling logger errors.
+ **/
+export function getRobustLogger(name: string): Logger {
```
matluster, line 34: I didn't find where this is used.

ts/nni_manager/common/log.ts
```diff
@@ -1,141 +1,119 @@
...
         }
     }
-}
 
-export function getLogger(name: string = 'root'): Logger {
-    let logger = loggers.get(name);
-    if (logger === undefined) {
-        logger = new Logger(name);
-        loggers.set(name, logger);
+    private logAfterError(levelName: string, args: any[]): void {
```
matluster, line 101: Are there any tests for `logAfterError`? Making sure this method is correct is important, as it's the bottomline.

ts/nni_manager/common/log.ts
cruiseliu: In shutdown manager.
This is a split PR.

ts/nni_manager/common/log.ts
cruiseliu: Added.

ultmaster (approved):

QuanluZhang (approved):

PR #4706, opened 2022-03-27
Refactor NNI manager globals (step 5) - shutdown

liuzhe-lz:
### Description

Add a global shutdown manager.

### Checklist

- [x] test case
- [x] doc

### How to Test

Run experiments and stop them in different ways.
Then check the end of nnimanager.log to see if there is anything strange.

### Release Note

No need to mention.

### DIFF
--- a/ts/nni_manager/common/globals/shutdown.ts
+++ b/ts/nni_manager/common/globals/shutdown.ts
@@ -1,0 +1,109 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ *  Shutdown manager.
+ *
+ *  Each module should register its clean up method with:
+ *
+ *      globals.shutdown.register('MyModuleName', async () => { this.cleanUp(); });
+ *
+ *  And can request for shutdown when unrecoverable error occurs:
+ *
+ *      try {
+ *          this.doSomethingMustSuccess();
+ *      } catch (error) {
+ *          globals.shutdown.criticalError('MyModuleName', error);
+ *      }
+ *
+ *  Note that when a module invokes `criticalError()`, its own registered callback will not get called.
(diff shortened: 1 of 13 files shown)


ultmaster (approved):

ts/nni_manager/common/globals/shutdown.ts
```diff
@@ -0,0 +1,110 @@
...
+            this.processStatus = 'running';
+        } else {
+            logger.error('notifyInitializeComplete() invoked in status', this.processStatus);
+        }
+    }
+
+    private shutdown(): void {
+        this.processStatus = 'stopping';
+
+        const promises = Array.from(this.modules).map(async ([moduleName, callback]) => {
```
QuanluZhang, line 73: does the order of calling modules' callback matter?

cruiseliu: Good question.
The key is, if one module _owns_ another, the two modules should have composition relationship and only one of them should be exposed to public. For example `NNIManager` owns `NNIDataStore`, so `NNIDataStore` should be a private member of `NNIManager` and therefore should not be visible to any other modules, including `main.ts` and `globals.shutdown`. `NNIManager` is responsible for invoking `NNIDataStore`s constructor and destructor.
And if a module is a _global service_, like `LogStream`, it should be inside `globals` module. In this case, shutdown manager is logically aware of its existence and will destruct it after all callbacks.
This design does have limitations, as it does not support complex module relationships that cannot fit into above two cases. However, since NNI manager's task is relatively simple, it would be a sign of over-engineering if it required dependency resolving. So I decided to use a "flat" shutdown callback manager, rather than a more powerful "component manager".

QuanluZhang (approved):

PR #4738, opened 2022-04-06
Enable trial version check in Retiarii

ultmaster:
### Description ###

Sometimes, inconsistent version between manager side and trial side causes problems that are not straightforward to debug.
This PR adds a warning message for inconsistent package versions detected.

### Checklist ###
  - [x] test case
  - [x] doc

### How to test ###

Try to make NNI manager run in a different env from the trial.

### DIFF
--- a/nni/common/version.py
+++ b/nni/common/version.py
@@ -1,10 +1,82 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from __future__ import annotations
+
 import logging
+import sys
+import warnings
+
+import cloudpickle
+import json_tricks
+import numpy
+import yaml
+
+import nni
+
+def _minor_version_tuple(version_str: str) -> tuple[int, int]:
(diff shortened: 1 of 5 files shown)


liuzhe-lz (approved):

nni/common/version.py
```diff
@@ -1,10 +1,83 @@
...
+        'json_tricks': JSON_TRICKS_VERSION,
+        'pyyaml': PYYAML_VERSION,
+        'nni': NNI_VERSION
+    }
+
+def version_check(expect: dict, raise_error: bool = False) -> None:
+    current_ver = version_dump()
+    for package in expect:
+        # version could be list due to serialization
+        exp_version: tuple | None = tuple(expect[package]) if expect[package] else None
```
QuanluZhang, line 70: why put `tuple` here?

QuanluZhang (approved):

nni/common/version.py
matluster: In dump & load, tuple could be converted to a list.

PR #4739, opened 2022-04-06
Bug fix of mutating architectures and hparams simultaneously

ultmaster:
### Description ###

There's a bug when mutating parameters on architectures and hyper-parameters (evaluators) simultaneously. I made a mistake of writing `mutator.name` rather than `mutator.label`, and tests didn't cover this part.

This PR also adds a recursive inspection of evaluator arguments, so that you can do this:

```python
FunctionalEvaluator(
    foo,
    t=Foo(some_argument=ValueChoice([1, 2, 3], label='t')),
    v=ValueChoice([1, 2, 3], label='t') + ValueChoice([10, 20, 30])
)
```

### Checklist ###
  - [x] test case
  - [x] doc

### DIFF
--- a/nni/retiarii/experiment/pytorch.py
+++ b/nni/retiarii/experiment/pytorch.py
@@ -144,6 +144,10 @@ def preprocess_model(base_model, trainer, applied_mutators, full_ir=True, dummy_
                            'do not use mutators when you use LayerChoice/InputChoice')
     if mutators is not None:
         applied_mutators = mutators
+
+    # Add mutations on evaluators
+    applied_mutators += process_evaluator_mutations(trainer, applied_mutators)
+
     return base_model_ir, applied_mutators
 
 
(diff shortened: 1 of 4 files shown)


nni/retiarii/experiment/pytorch.py
```diff
@@ -144,6 +144,10 @@ def preprocess_model(base_model, trainer, applied_mutators, full_ir=True, dummy_
                            'do not use mutators when you use LayerChoice/InputChoice')
     if mutators is not None:
         applied_mutators = mutators
+
+    # Add mutations on evaluators
+    applied_mutators += process_evaluator_mutations(trainer, applied_mutators)
+
```
QuanluZhang, line 150: I remember some mutators should be put at the end of applied_mutators. could you double check this logic?

matluster: That's not relevant to this part. Those belong to model mutators. These are evaluator mutators.

And we don't support sharing choices between model and evaluator.

liuzhe-lz (approved):

QuanluZhang (approved):

PR #4751, opened 2022-04-11
Bug fix of Retiarii hyperparameter mutation

ultmaster:
This is a collection of bug fixes in v2.7 bug bash.

1. Calling `get()` is needed after `trace_copy()` because sometimes we need functions from the original class.
2. Fix the case when ValueChoice only appears in the deep recursion. In this case, all ancester objects need to be updated.
3. Add a warning about Python version when cloudpickle fails to de-serialize.
4. Make error message more informative when execution engine is set multiple times.

### DIFF
--- a/nni/common/serializer.py
+++ b/nni/common/serializer.py
@@ -826,5 +826,5 @@ def _wrapped_cloudpickle_loads(b: bytes) -> Any:
         return cloudpickle.loads(b)
     except TypeError:
         warnings.warn('TypeError encountered during deserializing object. This could be caused by '
-                      'consistency between Python versions where dump and load happens.')
+                      'inconsistency between Python versions where dump and load happens.')
         raise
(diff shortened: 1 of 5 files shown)


nni/common/serializer.py
```diff
@@ -800,5 +817,14 @@ def _json_tricks_any_object_decode(obj: Dict[str, Any]) -> Any:
...
+        return _wrapped_cloudpickle_loads(b)
     return obj
+
+
+def _wrapped_cloudpickle_loads(b: bytes) -> Any:
+    try:
+        return cloudpickle.loads(b)
+    except TypeError:
+        warnings.warn('TypeError encountered during deserializing object. This could be caused by '
+                      'consistency between Python versions where dump and load happens.')
```
QuanluZhang, line 829: consistency -> inconsistency

PR #4754, opened 2022-04-11
Update doc Makefile

liuzhe-lz:
Commands:

  - `make`: Build English doc to `build/html` and Chinese doc to `build/html_zh`.
  - `make en`: Build English doc only.
  - `make zh`: Build Chinese doc only.
  - `make clean`: Remove built docs.

PR #4759, opened 2022-04-13
i18n toolchain based on sphinx-intl

ultmaster:
The per-file translation toolchain is not easy to use, because in case that one sentence is changed within a file (which is the most common case), it's difficult to find the particular sentence and update its translation. This is a problem especially for tutorials, where lots of information (e.g., line number, execution time) are auto-generated, which could differ between versions, even if there are no changes in the primary texts.

This PR introduces a new toolchain, recommended by [sphinx-intl](https://www.sphinx-doc.org/en/master/usage/advanced/intl.html). It automatically extracts text messages from built documents, and builds message catalogs for translation. To make the tool only cover important documents but nothing else as [previously discussed](https://github.com/microsoft/nni/issues/4298), I wrote a `getpartialtext` which reads the config and only writes messages from selected documents into the catalog.

From now on, developers need to write the translation in `docs/source/locales`. Pipeline will still complain if the English documents are changed, but translation is not updated.

The original toolchain based on `_zh.rst` will stay until we migrate all the translations to the new system.


docs/source/notes/contributing.rst
```diff
@@ -305,16 +305,31 @@ To contribute a new tutorial, here are the steps to follow:
...
+
+
+To build the translated documentation (for example Chinese documentation), please run:
+
+.. code-block:: bash
 
-1. Add a ``xxx_zh.rst`` in the same folder where ``xxx.rst`` exists. 
-2. Run ``python tools/chineselink.py`` under ``docs`` folder, to generate a hash string in your created ``xxx_zh.rst``.
-3. Don't delete the hash string, add your translation after it.
+   make -e SPHINXOPTS="-D language='zh'" html
```
J-shang, line 330: a reminder for #4754

J-shang, line 330: a reminder for #4754

liuzhe-lz: Let this PR merge first.

J-shang (approved):

QuanluZhang (approved):

PR #4773, opened 2022-04-18
Use "tuner" to config advisor

liuzhe-lz:
I believe it has little benefit to distinguish advisor from tuner in public API. In end users' perspective, it's just a tuning algorithm that handles early stopping. And on the other hand, non-standard terminologies take much efforts to learn an can cause confusion. So this PR removes term "advisor" from public APIs and docs except "advanced usage".

In addtion, this PR also does following things:

1. Rename tuners like "PPOTuner" to "PPO" to align with other tuners, while keeping old name as alias.
2. Make tuner name case insensitive.
3. Partially add type annotation for nni.tools.package_utils.


nni/__main__.py
```diff
@@ -29,82 +29,50 @@ def main():
...
-            raise
-
-
-def _run_advisor(exp_params):
-    if exp_params.get('advisor').get('name'):
-        dispatcher = create_builtin_class_instance(
-            exp_params['advisor']['name'],
-            exp_params['advisor'].get('classArgs'),
-            'advisors')
+            logger.error('Tuner {type(tuner).__name__} does not support early stopping. Assessor is ignored.')
```
QuanluZhang, line 46: not that precise. maybe "Tuner XXX does not support additional early stopping logic of Assessor. Assessor is ignored."

nni/experiment/config/experiment_config.py
```diff
@@ -113,6 +113,11 @@ def _canonicalize(self, _parents):
             if algo is not None and algo.name == '_none_':
                 setattr(self, algo_type, None)
 
+        if self.advisor is not None:
+            assert self.tuner is None
```
QuanluZhang, line 117: add a message here? for example, "advisor is not allowed when tuner is set."

nni/__main__.py
```diff
@@ -29,82 +29,50 @@ def main():
...
+        tuner.run()
```
J-shang, line 47: just curious about the different between `tuner` and `advisor`, why can't adviser be a tuner?

nni/__main__.py
```diff
@@ -29,82 +29,50 @@ def main():
...
-
-def _create_assessor(exp_params):
-    if exp_params['assessor'].get('name'):
-        assessor = create_builtin_class_instance(
-            exp_params['assessor']['name'],
-            exp_params['assessor'].get('classArgs'),
-            'assessors')
+def _create_algo(algo_config, algo_type):
+    if algo_config.get('name'):
+        algo = create_builtin_class_instance(algo_config['name'], algo_config.get('classArgs'), algo_type + 's')
```
J-shang, line 70: `+ 's'` is ok, but a little strange...

nni/tools/nnictl/config_schema.py
```diff
@@ -73,16 +69,13 @@ def __init__(self, algo_type):
         }
         self.builtin_name_schema = {}
         for k, n in self.builtin_keys.items():
-            self.builtin_name_schema[k] = {Optional(n): setChoice(n, *get_all_builtin_names(k+'s'))}
+            self.builtin_name_schema[k] = {Optional(n): setType(n, str)}
```
QuanluZhang, line 72: why converting from `setChoice` to `setType`?

nni/__main__.py
liuzhe-lz: The root cause is that `Tuner` API is not powerful enough (because it designed for simplicity) and therefore some algorithms have to use lower level APIs. Tuning algorithms which uses low level APIs are (formerly) called "advisor".

nni/tools/nnictl/config_schema.py
```diff
@@ -73,16 +69,13 @@ def __init__(self, algo_type):
...
 
         self.customized_keys = set(['codeDir', 'classFileName', 'className'])
 
     def validate_class_args(self, class_args, algo_type, builtin_name):
         if not builtin_name or not class_args:
             return
-        meta = get_registered_algo_meta(builtin_name, algo_type+'s')
-        if meta and 'acceptClassArgs' in meta and meta['acceptClassArgs'] == False:
```
QuanluZhang, line 84: what is the meaning of `acceptClassArgs`?

nni/tools/package_utils/config_manager.py
```diff
@@ -9,24 +11,26 @@
...
     for algo in get_all_algo_meta():
-        if algo.name == name:
+        if algo.name.lower() == name:
+            return algo
+        if algo.alias is not None and algo.alias.lower() == name:
             return algo
     return None
 
-def get_all_algo_meta() -> List[AlgoMeta]:
+def get_all_algo_meta() -> list[AlgoMeta]:
```
J-shang, line 33: when should we use `List` or never use it?

nni/__main__.py
liuzhe-lz: I totally agree. This is historical burden.
In experiment config they are called "tuner" and "assessor", but in algorithm manifest the keys are "tuners" and "assessors".
We don't want to change the file format until 3.0.

nni/tools/nnictl/config_schema.py
liuzhe-lz: Because `get_all_builtin_names` no longer works. And since this is V1 config only, I believe it's not worthy to fix it.

nni/tools/nnictl/config_schema.py
liuzhe-lz: Random tuner (and maybe some other tuners) does not accept any arguments. This checks no args are passed to such tuner.
Again this is V1 only and I don't want to fix.

nni/tools/package_utils/config_manager.py
liuzhe-lz: It's just style preference.
I personally prefer `list` so modules maintained by me will always use `list`.

J-shang (approved):

nni/__main__.py
liuzhe-lz: done

nni/experiment/config/experiment_config.py
liuzhe-lz: done

QuanluZhang (approved):

PR #4774, opened 2022-04-18
Add license header and typehints for NAS

ultmaster:
This PR would make Retiarii type-safe, except:

- CGO: I tried, but there are too much code that I don't fully understand. I'll ask @hzhua for help.
- Oneshot: I'll handle them in a dedicated PR to avoid blocking P0 works.


nni/typehint.py
```diff
@@ -18,6 +18,8 @@
 else:
     from typing_extensions import Literal, TypedDict
 
+from typing_extensions import TypeGuard
```
cruiseliu, line 21: In "required.txt": `typing_extensions >= 4.0.0 ; python_version < "3.8"`
Should we bump it to python 3.10?

ultmaster: I think typing-extensions is already in the dependency tree. We can just make it required regardless of python versions.

cruiseliu: Seems both pytorch and tensorflow depends on typing extensions. Then I'm OK.

liuzhe-lz (approved):

nni/typehint.py
liuzhe-lz: I'm not familiar with typing_extensions. Is there advantage to use like `typing.Literal` over `typing_extensions.Literal`?
If the answer is no, seems we should stop using `nni.typehint.Literal` and use `typing_extensions.Literal` instead.

nni/common/serializer.py
```diff
@@ -115,7 +115,7 @@ def is_wrapped_with_trace(cls_or_func: Any) -> bool:
     )
 
 
-class SerializableObject(Traceable):
+class SerializableObject(Generic[T], Traceable):
```
QuanluZhang, line 118: what is the application scope of `T`, only this file? or the whole code base?

ultmaster: For this case, this class.

nni/retiarii/hub/pytorch/nasnet.py
```diff
@@ -394,30 +398,31 @@ def __init__(self,
...
-        num_cells_per_stage = [i * num_cells // 3 - (i - 1) * num_cells // 3 for i in range(3)]
+            self.num_cells = nn.ValueChoice(list(num_cells), label='depth')
+        num_cells_per_stage = [i * self.num_cells // 3 - (i - 1) * self.num_cells // 3 for i in range(3)]
 
         # auxiliary head is different for network targetted at different datasets
         if dataset == 'imagenet':
             self.stem0 = nn.Sequential(
-                nn.Conv2d(3, C // 2, kernel_size=3, stride=2, padding=1, bias=False),
-                nn.BatchNorm2d(C // 2),
+                nn.Conv2d(3, cast(int, C // 2), kernel_size=3, stride=2, padding=1, bias=False),
```
QuanluZhang, line 409: do we need to use `cast` here? because it is obvious that `//` can only be applied on integer

nni/retiarii/nn/pytorch/api.py
```diff
@@ -377,8 +365,14 @@ def _valuechoice_codegen(*, _internal: bool = False):
                 print(binary_r_template.format(op=op, opt=opt, sym=sym) + '\n')
 
 
-def _valuechoice_staticmethod_helper(orig_func):
-    orig_func.__doc__ += """
+_func = TypeVar('_func')
+_cand = TypeVar('_cand')
+_value = TypeVar('_value')
```
QuanluZhang, line 370: when should we use TypeVar? for example, '_func', can I use 'Callable' instead?

QuanluZhang (approved):

nni/retiarii/hub/pytorch/nasnet.py
ultmaster: C can be ValueChoice.

nni/retiarii/nn/pytorch/api.py
ultmaster: TypeVar is useful in generics.

https://docs.python.org/3/library/typing.html#typing.TypeVar

PR #4781, opened 2022-04-20
Bump dev tools version

liuzhe-lz:
### Description ###

According to postmortem of v2.6 release, we bump dev tools at the start of each release cycle.

We should also bump Python latest to 3.10, but onnx failed to build.

### Release Note ###

Do not mention.


ultmaster (approved):
I didn't recall that postmortem. Would you add it in the release management note?

QuanluZhang (approved):

Issue #4499, opened 2022-01-24
Log mechanism discussion


cruiseliu:
## Current Log Behavior ##

| Component                     | Log Behavior                         | Log Level        | Detected by    | Note                 |
| ----------------------------- | ------------------------------------ | ---------------- | -------------- | -------------------- |
| Experiment management         | Log → stdout & (bug) dispatcher.log | Aware `debug`    | Explicit setup | `nni.*` are colorful |
| nnictl (create, resume, view) | Log → stdout & dispatcher.log       | Aware `debug`    | Explicit setup | Same as Experiment   |
| nnictl (Other commands)       | Log → stdout                        | INFO             | Default        | Never used           |
| Trials (local)                | Log & stdout → trial.log            | INFO             | NNI_PLATFORM   |                      |
| Trials (reuse)                | Log → stdout                        | INFO             | REUSE_MODE     | Same as default      |
| Standalone (debug trial)      | Log → stdout                        | INFO             | Default        |                      |
| Tuner                         | Log → dispatcher.log                | Aware `logLevel` | SDK_PROCESS    |                      |
| NAS main process              | Everything → stdout                 | INFO             | Default        |                      |
| NAS multi-trial trials        | Same as HPO trials                   | INFO             | NNI_PLATFORM   |                      |
| Compression                   | Everything → stdout                 | INFO             | Default        |                      |

liuzhe-lz:
TODO:
- [ ] Check current usage. Whether they can access experiment info.

PR #4792, opened 2022-04-21
Logging refactor (step 1) - experiment handlers

liuzhe-lz:
### Description ###

Refactor logging system according to plan #4499

### DIFF
--- a/nni/experiment/experiment.py
+++ b/nni/experiment/experiment.py
@@ -10,13 +10,12 @@ from pathlib import Path
 import socket
 from subprocess import Popen
 import time
-from typing import Any
+from typing import Any, cast
 
-import colorama
 import psutil
 from typing_extensions import Literal
 
-import nni.runtime.log
+from nni.runtime.log import start_experiment_logging, stop_experiment_logging
 
 from .config import ExperimentConfig
 from .data import TrialJob, TrialMetricData, TrialResult
@@ -73,8 +72,6 @@ class Experiment:
     """
 
     def __init__(self, config_or_platform: ExperimentConfig | str | list[str] | None):
-        nni.runtime.log.init_logger_for_command_line()
-
         self.config: ExperimentConfig | None = None
         self.id: str = management.generate_experiment_id()
         self.port: int | None = None
@@ -109,11 +106,10 @@ class Experiment:
         if config.use_annotation:
             raise RuntimeError('NNI annotation is not supported by Python experiment API.')
 
-        if config.experiment_working_directory is not None:
-            log_dir = Path(config.experiment_working_directory, self.id, 'log')
-        else:  # this should never happen in latest version, keep it until v2.7 for potential compatibility
-            log_dir = Path.home() / f'nni-experiments/{self.id}/log'
-        nni.runtime.log.start_experiment_log(self.id, log_dir, debug)
+        log_file = Path(config.experiment_working_directory, self.id, 'log', 'experiment.log')
+        log_file.parent.mkdir(parents=True, exist_ok=True)
+        log_level = 'debug' if (debug or config.log_level == 'trace') else config.log_level
+        start_experiment_logging(self.id, log_file, cast(str, log_level))
 
         self._proc = launcher.start_experiment(self._action, self.id, config, port, debug, run_mode, self.url_prefix)
         assert self._proc is not None
@@ -126,7 +122,7 @@ class Experiment:
                 if interface.family == socket.AF_INET:
                     ips.append(interface.address)
         ips = [f'http://{ip}:{port}' for ip in ips if ip]
-        msg = 'Web portal URLs: ' + colorama.Fore.CYAN + ' '.join(ips) + colorama.Style.RESET_ALL
+        msg = 'Web portal URLs: ${CYAN}' + ' '.join(ips)
         _logger.info(msg)
 
     def stop(self) -> None:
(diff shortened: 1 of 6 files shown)


nni/runtime/log.py
```diff
@@ -7,14 +7,149 @@
...
 from .env_vars import dispatcher_env_vars, trial_env_vars
 
+_global_logging_enabled: bool = False
+
+_root_logger: Logger = logging.getLogger()
+logging.getLogger('nni').setLevel(logging.INFO)
+
+_handlers: dict[str, Handler] = {}
+
+colorama.init()
```
matluster, line 26: Does colorama init have side effect? What if user doesn't want colorama?

nni/experiment/experiment.py
```diff
@@ -136,7 +132,7 @@ def stop(self) -> None:
         _logger.info('Stopping experiment, please wait...')
         atexit.unregister(self.stop)
 
-        nni.runtime.log.stop_experiment_log(self.id)
+        stop_experiment_logging(self.id)
```
matluster, line 135: What if I resume the experiment after stop?

I think either we have an API to explicitly re-enable it, or we postpone this stop to `__del__`.

nni/runtime/log.py
```diff
@@ -7,14 +7,149 @@
...
+    It is safe to call it multiple times.
+    """
+    if '_stdout_' in _handlers:
+        return
+
+    handler = StreamHandler(sys.stdout)
+    handler.addFilter(_nni_filter)
+    handler.setFormatter(_StdoutFormatter())
+
+    _handlers['_stdout_'] = handler
```
matluster, line 45: This is actually `_nni_stdout_`.

nni/runtime/log.py
```diff
@@ -7,14 +7,149 @@
...
+    _root_logger.addHandler(handler)
```
matluster, line 46: I think adding a handler to `logging.getLogger('nni')` enables log-handling on every logger prefixed with NNI. There's no need to manipulate the root logger. Maybe I'm wrong.

nni/runtime/log.py
```diff
@@ -45,41 +180,7 @@ def init_logger() -> None:
         _init_logger_trial()
```
matluster, line 180: There's a colorama.init() in `init_logger`.

nni/runtime/log.py
```diff
@@ -45,41 +180,7 @@ def init_logger() -> None:
...
-    log_path = _prepare_log_dir(log_directory) / 'dispatcher.log'
-    log_level = logging.DEBUG if debug else logging.INFO
-    _register_handler(FileHandler(log_path), log_level, experiment_id)
-
-def stop_experiment_log(experiment_id: str) -> None:
-    if experiment_id in handlers:
-        handler = handlers.pop(experiment_id, None)
-        if handler is not None:
-            logging.getLogger().removeHandler(handler)
+    start_stdout_logging()
```
matluster, line 183: What's the condition in which this will be triggered?

nni/runtime/log.py
```diff
@@ -7,14 +7,149 @@
...
+    """
+    Register the log file handler for an experiment's ``experiment.log``.
+
+    This function should be invoked on starting experiment.
+    We don't want to create the experiment folder if the user does not launch it.
+
+    If there are multiple experiments running concurrently,
+    log messages will be written to all running experiments' log file.
+
+    It is safe to call it multiple times.
```
matluster, line 69: Suggest adding some docstrings here to explain how to properly persist logs in log files, for first-come NNI users.

nni/experiment/experiment.py
J-shang: `resume` also use `start()` to resume the experiment, so it is ok?

nni/runtime/log.py
cruiseliu: I assumed `colorama.init()` was setting `colorama.Fore.RED`'s value. But I was wrong.
After checking the source I found `colorama.Fore.RED` is platform independent constant and `colorama.init()` overwrites `sys.stdout`.

So yes you are right. It should not be here.

nni/runtime/log.py
cruiseliu: `_handlers` is a local variable. So I think no bother.

nni/runtime/log.py
cruiseliu: The filter was introduced by previous design. The main reason I kept it is `enable_global_logging()`. In current implementation it only needs to set a switch variable. Otherwise it will have considerably complex logic.
If you think it's bad to use root logger + custom filter, I can change it anyway.

nni/runtime/log.py
cruiseliu: On importing `nni`. We assume this has no effect "outside" NNI.

nni/runtime/log.py
cruiseliu: OK. Previous how-to-use guide was removed with `nni.get_logger()` functions.

nni/runtime/log.py
```diff
@@ -7,14 +7,149 @@
...
+    """
+    if experiment_id in _handlers:
+        return
```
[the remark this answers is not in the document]
cruiseliu: It "should" not happen but we want robustness for edge cases.
For example if a user tries to create experiment in interactive Python and `start()` raises exception, it may leave uncleaned handlers.

J-shang (approved):

ultmaster (approved):

PR #4802, opened 2022-04-25
Replace nni.typehint with typing_extensions

liuzhe-lz:
### Description ###

We have made `typing_extensions` required dependency, so `nni.typehint` does not need to help importing annotation types any more.

### How to test ###

No need.
No release note.

### DIFF
--- a/nni/typehint.py
+++ b/nni/typehint.py
@@ -5,18 +5,12 @@
 Types for static checking.
 """
 
-__all__ = [
-    'Literal', 'TypedDict',
-    'Parameters', 'SearchSpace', 'TrialMetric', 'TrialRecord',
-]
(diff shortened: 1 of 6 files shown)


ultmaster (approved):

nni/typehint.py
```diff
@@ -5,18 +5,11 @@
...
 
-__all__ = [
-    'Literal', 'TypedDict',
-    'Parameters', 'SearchSpace', 'TrialMetric', 'TrialRecord',
-]
-
-import sys
-from typing import Any, Dict, List, TYPE_CHECKING
-
-if TYPE_CHECKING or sys.version_info >= (3, 8):
```
matluster, line 16: By the way, I think this doesn't work on Python 3.7.

Maybe it should be ``not TYPE_CHECKING``.

J-shang (approved):

PR #4806, opened 2022-04-25
WebSocket (step 1) - Python client

liuzhe-lz:
### Description ###

This PR implements the WebSocket client at Python side, but does not actually use it.

`legacy.py` is copy-pasted from old `protocol.py`.

Planned steps:

1. [x] Python side client
2. [x] NNI manager server
3. [ ] Migrate experiment to new protocol
4. [ ] Migrate old UTs and remove old code

### Checklist ###
  - [x] test case
  - [x] doc

### DIFF
--- a/nni/runtime/tuner_command_channel/shim.py
+++ b/nni/runtime/tuner_command_channel/shim.py
@@ -1,0 +1,33 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+Compatibility layer for old protocol APIs.
+
+We are working on more semantic new APIs.
+"""
+
+from __future__ import annotations
+
+from .command_type import CommandType
+from .websocket import WebSocket
+
+_ws: WebSocket = None  # type: ignore
+
+def connect(url: str) -> None:
+    global _ws
+    _ws = WebSocket(url)
+    _ws.connect()
+
+def send(command_type: CommandType, data: str) -> None:
+    command = command_type.value.decode() + data
+    _ws.send(command)
+
+def receive() -> tuple[CommandType, str]:
+    command = _ws.receive()
+    if command is None:
+        raise RuntimeError('NNI manager closed connection')
+    command_type = CommandType(command[:2].encode())
+    if command_type is CommandType.Terminate:
+        _ws.disconnect()
+    return command_type, command[2:]
--- a/nni/runtime/tuner_command_channel/websocket.py
+++ b/nni/runtime/tuner_command_channel/websocket.py
@@ -23,6 +23,7 @@ _logger = logging.getLogger(__name__)
 # the singleton event loop
 _event_loop: asyncio.AbstractEventLoop = None  # type: ignore
 _event_loop_lock: Lock = Lock()
+_event_loop_refcnt: int = 0  # number of connected websockets
 
 class WebSocket:
     """
@@ -43,8 +44,9 @@ class WebSocket:
         self._ws: Any = None  # the library does not provide type hints
 
     def connect(self) -> None:
-        global _event_loop
+        global _event_loop, _event_loop_refcnt
         with _event_loop_lock:
+            _event_loop_refcnt += 1
             if _event_loop is None:
                 _logger.debug('Starting event loop.')
                 # following line must be outside _run_event_loop
@@ -61,12 +63,20 @@ class WebSocket:
         if self._ws is None:
             _logger.debug('disconnect: No connection.')
             return
+
         try:
             _wait(self._ws.close())
             _logger.debug('Connection closed by client.')
         except Exception as e:
             _logger.warning(f'Failed to close connection: {repr(e)}')
         self._ws = None
+
+        global _event_loop, _event_loop_refcnt
+        with _event_loop_lock:
+            _event_loop_refcnt -= 1
+            if _event_loop_refcnt == 0:
+                _event_loop.call_soon_threadsafe(_event_loop.stop)
+                _event_loop = None
 
     def send(self, message: str) -> None:
         _logger.debug(f'Sending {message}')
--- a/nni/runtime/tuner_command_channel/websocket.py
+++ b/nni/runtime/tuner_command_channel/websocket.py
@@ -54,7 +54,7 @@ class WebSocket:
         _logger.debug(f'Connecting to {self._url}')
         self._ws = _wait(_connect_async(self._url))
         _logger.debug(f'Connected.')
-    
+
     def disconnect(self) -> None:
         if self._ws is None:
             _logger.debug('disconnect: No connection.')
@@ -65,11 +65,11 @@ class WebSocket:
         except Exception as e:
             _logger.warning(f'Failed to close connection: {repr(e)}')
         self._ws = None
-    
+
     def send(self, message: str) -> None:
         _logger.debug(f'Sending {message}')
         _wait(self._ws.send(message))
-    
+
     def receive(self) -> str | None:
         """
         Return received message;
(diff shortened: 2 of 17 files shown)


test/ut/sdk/test_tuner_command_channel.py
```diff
@@ -0,0 +1,85 @@
...
+_server = None
+_client = None
+
+_command1 = 'T_hello world'
+_command2 = 'T_你好'
+
+## test cases ##
+
+def test_connect():
+    global _client
```
matluster, line 26: Does this work for pytest?

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -0,0 +1,121 @@
...
+            _wait(self._ws.close())
+            _logger.debug('Connection closed by client.')
+        except Exception as e:
+            _logger.warning(f'Failed to close connection: {repr(e)}')
+        self._ws = None
+
+        global _event_loop, _event_loop_refcnt
+        with _event_loop_lock:
+            _event_loop_refcnt -= 1
+            if _event_loop_refcnt == 0:
```
matluster, line 77: So if I forgot to disconnect, and the program exits / crashes for reason, is it possible that the event loop run forever? Will the program hang in this case?

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -0,0 +1,121 @@
...
+                _event_loop.call_soon_threadsafe(_event_loop.stop)
+                _event_loop = None  # type: ignore
+
+    def send(self, message: str) -> None:
+        _logger.debug(f'Sending {message}')
+        _wait(self._ws.send(message))
```
matluster, line 83: If the message sent is very large, what will happen here?

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -0,0 +1,121 @@
...
+    def receive(self) -> str | None:
+        """
+        Return received message;
+        or return ``None`` if the connection has been closed by peer.
+        """
+        try:
+            msg = _wait(self._ws.recv())
+            _logger.debug(f'Received {msg}')
+        except websockets.ConnectionClosed:  # type: ignore
+            _logger.debug('Connection closed by server.')
```
matluster, line 94: Looks like a warning or critical.

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -0,0 +1,121 @@
...
+        The WebSocket URL.
+        For tuner command channel it should be something like ``ws://localhost:8080/tuner``.
+    """
+    def __init__(self, url: str):
+        self._url: str = url
+        self._ws: Any = None  # the library does not provide type hints
+
+    def connect(self) -> None:
+        global _event_loop, _event_loop_refcnt
+        with _event_loop_lock:
```
matluster, line 48: Does this imply that `/tuner` can serve multiple clients simultaneously?

nni/runtime/tuner_command_channel/websocket.py
liuzhe-lz: The event loop thread has `daemon=True`, so no it does not block exiting.

nni/runtime/tuner_command_channel/websocket.py
liuzhe-lz: I have set message size limit to "no limit" in python and 4gb in nni manager.
The message can be split under the hood, and a WebSocket implementation must reconstruct split messages before reporting to user API.

nni/runtime/tuner_command_channel/websocket.py
liuzhe-lz: The error log is in its caller.
I plan to reuse `websocket.py` in some other modules in future, and they might not require a steady connection.

nni/runtime/tuner_command_channel/websocket.py
liuzhe-lz: No. But a python process can have multiple experiment and each experiment serves a client.

test/ut/sdk/test_tuner_command_channel.py
liuzhe-lz: Theoretically yes, but Windows test still have problem...

ultmaster (approved):

nni/runtime/tuner_command_channel/shim.py
```diff
@@ -0,0 +1,33 @@
...
+
+We are working on more semantic new APIs.
+"""
+
+from __future__ import annotations
+
+from .command_type import CommandType
+from .websocket import WebSocket
+
+_ws: WebSocket = None  # type: ignore
```
QuanluZhang, line 15: we do not support running two experiments in one python process, right? as they share the same `_ws`.

cruiseliu: We do.
This file is the temporary compatibility layer to smooth the migration, not the final API.

pipelines/fast-test.yml
```diff
@@ -218,30 +220,30 @@ stages:
...
+
+  #  - script: |
+  #      cd test
+  #      python -m pytest ut
+  #    displayName: Python unit test
+
+  #  - script: |
+  #      cd test
+  #      python nni_test/nnitest/run_tests.py --config config/pr_tests.yml
+  #    displayName: Simple integration test
```
QuanluZhang, line 249: why removing all the tests?

test/ut/sdk/test_protocol.py
```diff
@@ -5,17 +5,20 @@
...
+    out_file = BytesIO()
+    protocol._set_out_file(out_file)
+    return out_file
 
 def _prepare_receive(data):
-    protocol._in_file = BytesIO(data)
+    protocol._set_in_file(BytesIO(data))
 
 
+@skipIf(sys.platform =='win32', 'debug')
```
QuanluZhang, line 21: why skip win32?

pipelines/fast-test.yml
cruiseliu: debugging

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -0,0 +1,121 @@
...
+import logging
+from threading import Lock, Thread
+from typing import Any
+
+import websockets
+
+_logger = logging.getLogger(__name__)
+
+# the singleton event loop
+_event_loop: asyncio.AbstractEventLoop = None  # type: ignore
```
QuanluZhang, line 24: why `_event_loop` should be a singleton？

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        Return received message;
+        or return ``None`` if the connection has been closed by peer.
+        """
+        try:
+            msg = _wait(self._ws.recv())
+            _logger.debug(f'Received {msg}')
+        except websockets.ConnectionClosed:  # type: ignore
+            _logger.debug('Connection closed by server.')
+            self._ws = None
```
QuanluZhang, line 95: do we need to do the clear up here? for example, `_event_loop_refcnt -= 1`?

pipelines/fast-test.yml
QuanluZhang: since you have created this pr, why not uncomment them?

nni/runtime/tuner_command_channel/websocket.py
cruiseliu: Because that is how asyncio designed to work.
In fact there should be only one global event loop thread, not one per module. But since nni does not use asyncio elsewhere (except trial keeper), for simplicity we put the event loop in websocket module for now.

pipelines/fast-test.yml
cruiseliu: The problem only appears on azure pipeline, so I had to debug inside the PR.
It should have finished now.

PR #4808, opened 2022-04-26
WebSocket (step 2) - TS server 

liuzhe-lz:
### Description ###

This PR implements the WebSocket server at NNI manager side, but does not actually use it.

It does **not** depends on step 1.

`legacy.ts` is copy-pasted from old `ipcInterface.ts`.

Planned steps:

1. [ ] Python side client
2. [x] NNI manager server
3. [ ] Migrate experiment to new protocol
4. [ ] Migrate old UTs and remove old code

### Checklist ###
  - [x] test case
  - [x] doc


ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+ *  The callback to serve WebSocket connection request. Used by REST server module.
+ *  It should only be invoked once, or an error will be raised.
+ *
+ *  Type hint of express-ws is somewhat problematic. Don't want to waste time on it so use `any`.
+ **/
+export function serveWebSocket(ws: WebSocket): void {
+    channelSingleton.setWebSocket(ws);
+}
+
+class WebSocketChannelImpl {
```
matluster, line 49: What's the interface of this "impl"?

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+    private deferredInit: Deferred<void> = new Deferred<void>();
+    private emitter: EventEmitter = new EventEmitter();
```
matluster, line 51: Who's listening this emitter? If it's only this object itself, why not use the callbacks directly?

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+            this.handleError(new Error('tuner_command_channel: Tuner loses responsive'));
+        }
+
+        this.waitingPong = true;
+        this.ws.ping();
+    }
+
+    private receive(data: Buffer, _isBinary: boolean): void {
+        logger.debug('Received', data);
+        this.emitter.emit('command', data.toString());
```
matluster, line 122: `_isBinary` is not used?

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private handleError(error: Error): void {
+        if (!this.serving) {
```
matluster, line 126: If this is not serving any more (due to some glitch or whatever reason), should we set `this.ws` to undefined?

In that way, maybe tuner could set a request later, and re-setup a ws, pretending that nothing happened.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
```diff
@@ -0,0 +1,146 @@
...
+        this.emitter.removeAllListeners();
+    }
+
+    public sendCommand(command: string): void {
+        assert.ok(this.ws !== undefined);
+
+        logger.debug('Sending', command);
+        this.ws.send(command);
+
+        if (this.ws.bufferedAmount > command.length + 1000) {
```
matluster, line 97: Is this 1000 a magic number?

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: `WebSocketChannel`. I forgot `implements`.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: I can't understand its behavior. Sometimes it holds unexpected value. So I always call `toString()`.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: This field is only used for one case here:
An `error` event will followed the by a `close` event, and in normal cases `close` will trigger on error handler. `serving` will ensure that the handler is not invoked twice.
If we set `this.ws` to `undefined` before the `close` event, I don't know if there will be problems. It requires experiments and is not an easy work. So I don't want to deal with it now.
Anyway the OS guarantees loopback connection is always reliable.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
matluster: Better to add comments (maybe a SO link) about this.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: Yes. It's only for a warning message. Any value above zero is acceptable.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: `EventEmitter` is used as a "callback container". It deals with multiple handlers on one event.
Without it, we need either to maintain callback sets, or to make assertions and add docs. I think it is simpler to use a `EventEmitter`.

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: Oh, it's Python part's problem. According to the type hint node API always provides a `Buffer` instead of `string`, so `toString()` is required anyway.

liuzhe-lz: To my understanding, WebSocket protocol send a bunch of plain data and attach a header indicating whether its binary or text. Node's WebSocket implementation simply passes the data and header to users as is. So we don't need the header when reading message.

QuanluZhang (approved):

ultmaster (approved):

PR #4821, opened 2022-04-28
WebSocket (step 3) - HPO experiment

liuzhe-lz:
### Description ###

Migrate HPO experiment to WebSocket.

Following steps:
 4. Migrate NAS experiment
 5. Migrate old UTs and remove old code

### Checklist ###
  - [x] test case
  - [x] doc


ultmaster (approved):

ts/nni_manager/test/core/nnimanager.test.ts
```diff
@@ -135,7 +137,10 @@ describe('Unit test for nnimanager', function () {
 
     after(async () => {
         // FIXME
```
matluster, line 139: What is this FIXME

liuzhe-lz: It is not a full clean up. A full clean up will kill mocha.

ts/nni_manager/core/nnimanager.ts
```diff
@@ -476,19 +476,27 @@ class NNIManager implements Manager {
...
             NNI_LOG_LEVEL: getLogLevel(),
             NNI_INCLUDE_INTERMEDIATE_RESULTS: includeIntermediateResultsEnv,
+            NNI_TUNER_COMMAND_CHANNEL: tunerWs,
             CUDA_VISIBLE_DEVICES: toCudaVisibleDevices(this.experimentProfile.params.tunerGpuIndices)
         };
         const newEnv = Object.assign({}, process.env, nniEnv);
         const tunerProc: ChildProcess = getTunerProc(command, stdio, newCwd, newEnv);
         this.dispatcherPid = tunerProc.pid!;
-        this.dispatcher = await createDispatcherInterface(tunerProc);
+        this.dispatcher = await createDispatcherInterface();
```
QuanluZhang, line 499: so the interfaces of ws and pipe are not the same, right?

liuzhe-lz: Yes. There will be no `pipe` anymore.
I plan to use a new command line arg `tuner_command_channel`.
Similar to `dispatcher_pipe`, if `tuner_command_channel` is set, NNI manager will not call `setupTuner()`. Instead, it will only invoke `createDispatcherInterface()` which listens on `ws://localhost:8080/tuner`.
Once NNI manager is initialized, Python experiment code can connect to the URL and starts tuner/strategy logic.

PR #4826, opened 2022-04-30
enable customizing the replace function

zheng-ningxin:
This pr enable customizing the replace function accordingly.

### DIFF
--- a/nni/compression/pytorch/speedup/compressor.py
+++ b/nni/compression/pytorch/speedup/compressor.py
@@ -48,11 +48,12 @@ class ModelSpeedup:
         is the replace function of corresponding opertor. The replace function should take
         two input parameters, one is the original module, the second input parameter is tuple
         of the input mask, output mask and weight mask. This replace function should prune the module
-        accordingly. Here is an example of the replace function(more examples can refer to compress_modules.py):
-        def example_replace(ori_module, masks):
-            in_mask, out_mask, weight_mask = masks
-            # prune the ori_module to a new smaller module according to the mask
-            return new_small_module
+        accordingly. Here is an example of the replace function(more examples can refer to compress_modules.py)::
+
+            def example_replace(ori_module, masks):
+                in_mask, out_mask, weight_mask = masks
+                # prune the ori_module to a new smaller module according to the mask
+                return new_small_module
 
     """
 
--- a/nni/compression/pytorch/speedup/compressor.py
+++ b/nni/compression/pytorch/speedup/compressor.py
@@ -43,7 +43,7 @@ class ModelSpeedup:
     confidence: the confidence coefficient of the sparsity inference. This value is
         actually used as the batchsize of the dummy_input.
     customized_replace_func: None/Dict
-        If the parameter is not None, then we will use the given function to replace the
+        If `customized_replace_func` is not None, then we will use the given function to replace the
         corresponding modules. The `key` of the dict is the opertor types and the `value`
         is the replace function of corresponding opertor. The replace function should take
         two input parameters, one is the original module, the second input parameter is tuple
--- a/nni/compression/pytorch/speedup/compressor.py
+++ b/nni/compression/pytorch/speedup/compressor.py
@@ -42,10 +42,14 @@ class ModelSpeedup:
         the index of batch dimension in the dummy_input
     confidence: the confidence coefficient of the sparsity inference. This value is
         actually used as the batchsize of the dummy_input.
+    customized_replace_func: None/Dict
+        If the parameter is not None, then we will use the given function to replace the
+        corresponding modules. The `key` of the dict is the opertor types and the `value`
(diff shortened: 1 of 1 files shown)


nni/compression/pytorch/speedup/compressor.py
```diff
@@ -441,12 +450,15 @@ def forward(self, x):
             super_module, leaf_module = get_module_by_name(
                 self.bound_model, g_node.name)
             m_type = g_node.op_type
-            if not m_type in replace_module:
+            if not m_type in replace_module and (self.customized_replace_func is not None and m_type not in self.customized_replace_func):
```
QuanluZhang, line 453: suggest to set `self.customized_replace_func` to {} if it is None, then no need to check `self.customized_replace_func` any more.

zheng-ningxin: Using a mutable default parameter is dangerous. FYI, https://betterprogramming.pub/mutable-default-arguments-in-python-643ae2583e00

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -42,18 +42,23 @@ class ModelSpeedup:
         the index of batch dimension in the dummy_input
     confidence: the confidence coefficient of the sparsity inference. This value is
         actually used as the batchsize of the dummy_input.
+    customized_replace_func: None/Dict
+        If the parameter is not None, then we will use the given function to replace the
+        corresponding modules. The `key` of the dict is the opertor types and the `value`
+        is the replace function of corresponding opertor.
```
J-shang, line 48: maybe users don't know how to use it if they didn't see the source code, do we need a simple example to show how to write `customized_replace_func`?

J-shang, line 48: maybe users don't know how to use it if they didn't see the source code, do we need a simple example to show how to write `customized_replace_func`?

QuanluZhang: good suggestion!

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -71,10 +76,12 @@ def __init__(self, model, dummy_input, masks_file, map_location=None,
         elif isinstance(masks_file, dict):
             self.masks = masks_file
         else:
-            raise Exception('Please provide the mask or the path of the mask file')
+            raise Exception(
+                'Please provide the mask or the path of the mask file')
```
QuanluZhang, line 80: no need to have a new line

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -71,10 +76,12 @@ def __init__(self, model, dummy_input, masks_file, map_location=None,
...
         self.constant = {}
         # self.internal_result save the internal output of the submodules
         self.internal_result = {}
+        self.customized_replace_func = customized_replace_func
```
QuanluZhang, line 84: maybe: `self.customized_replace_func = customized_replace_func if customized_replace_func is not None else {}`. then at the bottom, no need to check self.customized_replace_func any more

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: good suggestion!

nni/compression/pytorch/speedup/compressor.py
zheng-ningxin: good suggestion!

J-shang (approved):

nni/compression/pytorch/speedup/compressor.py
J-shang: need `::` when generating doc.
```
    customized_replace_func: None/Dict
        If the parameter is not None, then we will use the given function to replace the
        corresponding modules. The `key` of the dict is the opertor types and the `value`
        is the replace function of corresponding opertor. The replace function should take
        two input parameters, one is the original module, the second input parameter is tuple
        of the input mask, output mask and weight mask. This replace function should prune the module
        accordingly.
        Here is an example of the replace function(more examples can refer to compress_modules.py)::

            def example_replace(ori_module, masks):
                in_mask, out_mask, weight_mask = masks
                # prune the ori_module to a new smaller module according to the mask
                return new_small_module
```

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -42,18 +42,32 @@ class ModelSpeedup:
         the index of batch dimension in the dummy_input
     confidence: the confidence coefficient of the sparsity inference. This value is
         actually used as the batchsize of the dummy_input.
+    customized_replace_func: None/Dict
+        If the parameter is not None, then we will use the given function to replace the
```
QuanluZhang, line 46: "parameter" is a little misleading, could use "customized_replace_func"

QuanluZhang (approved):

PR #4835, opened 2022-05-04
Support Repeat and Cell in One-shot NAS

ultmaster:
## Support Repeat in one-shot NAS:

- For path-sampling: randomly sample a depth and use a prefix.
- For differentiable: compute a weighted sum with several prefixes.

## Support Cell in one-shot NAS:

Only two types of cell need ad-hoc support in weight-sharing:

- Operation factory used: When the created operation is dependant on node index, and selected input's index (which is unknown before InputChoice are fixed), the static-created operation might not work for dynamically-changing input shape. (NOTE: operation dependant on op index is not supported)
- `loose_end`: The cell's output shape is dynamic. This causes confusion to the supernet.

These two types of cells need to be replaced with SuperNetCell, which I follow the [implementation of DARTS](https://github.com/quark0/darts/blob/master/cnn/model_search.py). Specifically,

- A cartesian product of LayerChoice and InputChoice is created: each node is connected with every predecessing nodes with every possible operations. For differentiable cases, a set of architecture weight is created on every connection.
- For path-sampling, they sample from space spec and choose the create path to do the forward.
- For differentiable, they compute on the full graph. When export, [all architecture weights across different edges are compared](https://github.com/quark0/darts/blob/f276dd346a09ae3160f8e3aca5c7b193fda1da37/cnn/model_search.py#L135).

To reduce the coding effort, this PR also adjusts the implementation of ``nn.Cell`` by a little bit.


nni/retiarii/oneshot/pytorch/base_lightning.py
```diff
@@ -58,7 +58,7 @@ def traverse_and_mutate_submodules(
     module_list = []
 
     def apply(m):
-        for name, child in m.named_children():
+        for name, child in list(m.named_children()):
```
liuzhe-lz, line 61: why?

liuzhe-lz: oh it's a generator and this loop body may affect the generator's behavior?
very fancy

matluster: Yes. setattr might replace some of the "children" in-place.
The original version doesn't work at all.

liuzhe-lz (approved):

QuanluZhang (approved):

PR #4841, opened 2022-05-05
[retiarii] refactor of nas experiment

QuanluZhang:
### Description ###
refactor the retiarii (nas) experiment to make it more aligned with HPO experiment, making HPO, NAS, and Compression experiments more consistent. Also, this refactor is to make retiarii experiment more stable and user friendly.

in this refactor, users need to explicitly call exp.stop() to stop a finished experiment, otherwise the python script will not exit.

### Checklist ###
  - [x] refactor of retiarii experiment to leverage functionality of basic hpo experiment
  - [ ] refactor of nas and hpo experiment config to make them share a new base config
  - [ ] provide a new experimental nas experiment class to make its interface more aligned with hpo experiment
  - [ ] greatly improve debuggability
  - [ ] test case
  - [ ] doc


nni/retiarii/experiment/pytorch.py
```diff
@@ -409,81 +303,42 @@ def run(self, config: Optional[RetiariiExeConfig] = None, port: int = 8080, debu
...
+            return
 
         if config is None:
-            warnings.warn('config = None is deprecate in future. If you are running a one-shot experiment, '
-                          'please consider creating a config and set execution engine to `oneshot`.', DeprecationWarning)
-            config = RetiariiExeConfig()
-            config.execution_engine = 'oneshot'
-
-        if config.execution_engine == 'oneshot':
+            self.config = RetiariiExeConfig()
```
matluster, line 309: Where is the warning?

nni/retiarii/experiment/config/experiment_config.py
```diff
@@ -0,0 +1,64 @@
...
+    def __setattr__(self, key, value):
+        #TODO: tuner settings can also be blocked here
+        fixed_attrs = {'search_space': '',
+                       'trial_command': '_reserved'}
+        if key in fixed_attrs and fixed_attrs[key] != value:
+            raise AttributeError(f'{key} is not supposed to be set in Retiarii mode by users!')
+        # 'trial_code_directory' is handled differently because the path will be converted to absolute path by us
+        if key == 'trial_code_directory' and not (str(value) == '.' or os.path.isabs(value)):
+            raise AttributeError(f'{key} is not supposed to be set in Retiarii mode by users!')
+        #if key == 'execution_engine':
```
matluster, line 58: Delete this.

nni/experiment/experiment.py
```diff
@@ -114,13 +101,10 @@ def start(self, port: int = 8080, debug: bool = False, run_mode: RunMode = RunMo
         else:  # this should never happen in latest version, keep it until v2.7 for potential compatibility
             log_dir = Path.home() / f'nni-experiments/{self.id}/log'
         nni.runtime.log.start_experiment_log(self.id, log_dir, debug)
+        return config
```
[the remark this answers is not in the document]
liuzhe-lz: I would suggest
```python
class Base:
    def start(self):
        self._start()
    def _start(self, **launcher_args)
        ...
        launcher.start_experiment(..., **launcher_args)
        ...

class Derived:
    def start(self):
        self._start(tags=['xxx'], tuner_command_channel='xxx')

File: nni/experiment/config/experiment_config.py [hunk 4]
```diff
@@ -141,7 +141,7 @@ def _canonicalize(self, _parents):
                 msg = f'nni_manager_ip is not set, please make sure {ip} is accessible from training machines'
                 logging.getLogger('nni.experiment.config').warning(msg)
 
-    def _validate_canonical(self):
+    def _validate_canonical(self, validate_tuner: bool = True): # FIXME: remove validate_tuner
```
liuzhe-lz, line 144: Don't violate base class' signature.
If you need a temp hack, use `type(self)`.

nni/retiarii/experiment/config/experiment_config.py
```diff
@@ -0,0 +1,64 @@
...
+    return None
+
+@dataclass(init=False)
+class RetiariiExeConfig(ExperimentConfig):
+    # FIXME: refactor this class to inherit from a new common base class with HPO config
+    search_space: Any = ''
+    trial_code_directory: utils.PathLike = '.'
+    trial_command: str = '_reserved'
+    # new config field for NAS
+    execution_engine: ExecutionEngineConfig = PyEngineConfig()
```
liuzhe-lz, line 35: This is incorrect. All `RetiariiExeConfig` will share one execution engine.
This is similar to `def foo(arg={}): ...`

nni/retiarii/experiment/config/experiment_config.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, training_service_platform: str | None = None,
+                 execution_engine: str | ExecutionEngineConfig = PyEngineConfig(),
```
liuzhe-lz, line 38: Same as above, the execution engine object will be shared.

nni/experiment/experiment.py
```diff
@@ -129,11 +113,32 @@ def start(self, port: int = 8080, debug: bool = False, run_mode: RunMode = RunMo
...
+
+        self._proc = launcher.start_experiment(self._action, self.id, config, port, debug, run_mode, self.url_prefix)
+        assert self._proc is not None
+
+        self.port = port  # port will be None if start up failed
+
+        self._start_end(port, config.nni_manager_ip)
+
+    def _stop(self) -> None:
         atexit.unregister(self.stop)
```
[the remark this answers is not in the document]
liuzhe-lz: I think it's okay. `atexit.register()` is in fact `add_at_exit_handler()`. Multiple hooks will be fine.

nni/retiarii/experiment/pytorch.py
```diff
@@ -263,37 +193,31 @@ def __init__(self, base_model: nn.Module, evaluator: Union[BaseOneShotTrainer, E
         if evaluator is None:
             raise ValueError('Evaluator should not be none.')
 
-        # TODO: The current design of init interface of Retiarii experiment needs to be reviewed.
+        self.id: str = management.generate_experiment_id()
```
liuzhe-lz, line 196: This fields are derived. Use `super`.

nni/retiarii/experiment/config/experiment_config.py
```diff
@@ -0,0 +1,64 @@
...
+        super().__init__(training_service_platform, **kwargs)
+
+        if isinstance(execution_engine, str):
+            self.execution_engine = execution_engine_config_factory(execution_engine)
+        else:
+            self.execution_engine = execution_engine
+
+        self.__dict__['trial_command'] = 'python3 -m nni.retiarii.trial_entry ' + self.execution_engine.name
+
+    def __setattr__(self, key, value):
```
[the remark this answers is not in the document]
QuanluZhang: updated

nni/retiarii/experiment/config/experiment_config.py
QuanluZhang: deleted

nni/retiarii/experiment/config/experiment_config.py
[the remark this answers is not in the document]
QuanluZhang: it is not easy to do the checking in `_validate_canonical`, because the function `_canonicalize` may be executed multiple times accordingly. thus, still a little hacky, please look at the latest version.

examples/nas/multi-trial/mnist/search.py
```diff
@@ -131,15 +131,16 @@ def evaluate_model(model_cls):
...
     # uncomment this for graph-based execution engine
     # exp_config.execution_engine = 'base'
     # export_formatter = 'code'
 
-    exp.run(exp_config, 8080)
+    exp.run(exp_config, 8090)
     print('Final model:')
     for model_code in exp.export_top_models(formatter=export_formatter):
         print(model_code)
+    exp.stop()
```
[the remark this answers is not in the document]
QuanluZhang: yes. Another option is putting this behavior in the new NasExperiment interface while keeping the previous behavior in RetiariiExperiment. Explicitly stopping an experiment is a good style.

nni/experiment/experiment.py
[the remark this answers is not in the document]
QuanluZhang: good suggestions, let's put these changes to the refactor of HPO experiment, i.e., extracting a common base class

nni/retiarii/execution/base.py
```diff
@@ -54,12 +57,22 @@ class BaseExecutionEngine(AbstractExecutionEngine):
     Resource management is implemented in this class.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, rest_port: int | None = None, rest_url_prefix: str | None = None) -> None:
```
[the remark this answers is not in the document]
QuanluZhang: because it is prefix

nni/retiarii/execution/base.py
```diff
@@ -123,8 +136,8 @@ def query_available_resource(self) -> int:
         return self.resources
 
     def budget_exhausted(self) -> bool:
-        advisor = get_advisor()
-        return advisor.stopping
+        resp = rest.get(self.port, '/check-status', self.url_prefix)
```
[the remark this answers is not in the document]
QuanluZhang: maybe we can, but it would change a lot. let's improve this part step by step.
this functionality is provided by execution engine. for now, advisor is only the channel to send trial configurations.

nni/retiarii/experiment/pytorch.py
```diff
@@ -263,37 +193,31 @@ def __init__(self, base_model: nn.Module, evaluator: Union[BaseOneShotTrainer, E
...
-        self.port: Optional[int] = None
 
         self.base_model = base_model
         self.evaluator: Union[Evaluator, BaseOneShotTrainer] = evaluator
         self.applied_mutators = applied_mutators
         self.strategy = strategy
 
-        from nni.retiarii.oneshot.pytorch.strategy import OneShotStrategy
-        if not isinstance(strategy, OneShotStrategy):
-            self._dispatcher = RetiariiAdvisor()
```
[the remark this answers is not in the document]
QuanluZhang: no, only for py base and cgo. these are removed lines.

nni/retiarii/experiment/pytorch.py
```diff
@@ -502,7 +357,8 @@ def export_top_models(self, top_k: int = 1, optimize_mode: str = 'maximize', for
             If ``dict``, the mutation history will be returned.
         """
         if formatter == 'code':
-            assert self.config.execution_engine != 'py', 'You should use `dict` formatter when using Python execution engine.'
+            assert not isinstance(self.config.execution_engine, PyEngineConfig), \
```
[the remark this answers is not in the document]
QuanluZhang: yes, we can. just wondering why execution engine are imported in `_create_execution_engine` not at the top this this file? if we check engine here, the import should be moved to the top. any concern about this?

nni/experiment/experiment.py
```diff
@@ -148,8 +153,24 @@ def stop(self) -> None:
         self.id = None  # type: ignore
         self.port = None
         self._proc = None
+
+    def stop(self) -> None:
+        """
+        Stop the experiment.
+        """
+        _logger.info('Stopping experiment, please wait...')
+        self._stop()
```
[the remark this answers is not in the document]
QuanluZhang: let's leave it to the refactor of HPO experiment. Your suggest is good, when there is a base experiment, all subclasses implement its `stop_impl` (or `_stop_impl`).

ts/nni_manager/core/nnimanager.ts
```diff
@@ -303,8 +303,8 @@ class NNIManager implements Manager {
         }
 
         this.trainingService.removeTrialJobMetricListener(this.trialJobMetricListener);
+        this.dispatcher.sendCommand(TERMINATE);
```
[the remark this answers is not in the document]
QuanluZhang: because there is no `dispatcherPid` (i.e., =0) in retiarii experiment where advisor process is started before nnimanager

nni/retiarii/experiment/pytorch.py
QuanluZhang: added back:)

nni/retiarii/experiment/config/experiment_config.py
QuanluZhang: updated

nni/retiarii/experiment/config/experiment_config.py
QuanluZhang: updated

nni/experiment/config/experiment_config.py
QuanluZhang: then this base class should import its subclass?

nni/retiarii/experiment/pytorch.py
QuanluZhang: how? the super().__init__() also has other logic that should not be included in RetiariiExperiment. Maybe leave it in the refactor of HPO experiment

nni/experiment/experiment.py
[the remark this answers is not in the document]
liuzhe-lz: > good suggestions, let's put these changes to the refactor of HPO experiment, i.e., extracting a common base class

No I can't bear `_start_begin()` and `_start_end()` in "experiment.py". They are not meaningful functions.
I can accept this kind of TODO or FIXME in experimental features, but `nni.experiment` is a stable module.

nni/experiment/config/experiment_config.py
liuzhe-lz: Use `type(self).__name__` or delayed import.
It's a hack, and hack should be as "local" as possible.

nni/retiarii/experiment/pytorch.py
liuzhe-lz: I don't see anything HPO specific if the argument is a config object or None.

```python
    def __init__(self, config_or_platform: ExperimentConfig | str | list[str] | None):
        nni.runtime.log.init_logger_for_command_line()

        self.config: ExperimentConfig | None = None
        self.id: str = management.generate_experiment_id()
        self.port: int | None = None
        self._proc: Popen | psutil.Process | None = None
        self._action: Literal['create', 'resume', 'view'] = 'create'
        self.url_prefix: str | None = None

        if isinstance(config_or_platform, (str, list)):
            self.config = ExperimentConfig(config_or_platform)
        else:
            self.config = config_or_platform
```

examples/nas/multi-trial/mnist/search.py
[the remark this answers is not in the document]
J-shang: If I understand correctly, `run = start + stop`?

nni/retiarii/execution/cgo_engine.py
```diff
@@ -41,24 +42,35 @@ class CGOExecutionEngine(AbstractExecutionEngine):
...
         Seconds to wait for each batch of trial submission.
         The trials within one batch could apply cross-graph optimization.
+    rest_port
+        The port of the experiment's rest server
+    rest_url_prefix
+        The url prefix of the experiment's rest entry
     """
 
-    def __init__(self, devices: List[Device] = None,
+    def __init__(self, training_service: RemoteConfig,
```
J-shang, line 58: Why only remote is supported?

liuzhe-lz: CGO is a researching optimization. It only supports remote.

examples/nas/multi-trial/mnist/search.py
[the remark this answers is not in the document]
liuzhe-lz: Seems I haven't made it clear.
In v2.7 HPO experiment, `stop()` is recommended but not enforced. `run()` waits for completion but does not invoke `stop()`.

Consider 2 use cases:

###  Case A

```python
exp = Experiment(...)
exp.run()
# end of code
```

The experiment will stop when done. This is (effectively) the same as v2.6.
At the moment when `exp.run()` returns, the experiment is still serving web portal. But right after that, Python finds there is no more code to run and starts exiting. Then Python triggers `atexit` hooks, which will invoke `exp.stop()` to stop NNI manager.

### Case B

```python
exp = Experiment(...)
exp.run()
input()
```

The experiment will serve web portal forever until the user presses enter. This is different from v2.6.
In v2.6 the experiment stops serving web portal when `exp.run()` returns. So the user could not view web portal, or use REST API functions to query experiment result, even though the Python program is still running.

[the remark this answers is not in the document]
J-shang: got it!

nni/experiment/experiment.py
```diff
@@ -87,20 +87,10 @@ def __init__(self, config_or_platform: ExperimentConfig | str | list[str] | None
...
-
-        Parameters
-        ----------
-        port
-            The port of web UI.
-        debug
-            Whether to start in debug mode.
-        """
+    def _start_impl(self, port: int, debug: bool, run_mode: RunMode,
+                    url_prefix: str | None,
```
liuzhe-lz, line 91: I think `url_prefix` is a field `self.url_prefix`.

nni/experiment/experiment.py
```diff
@@ -128,12 +119,27 @@ def start(self, port: int = 8080, debug: bool = False, run_mode: RunMode = RunMo
         ips = [f'http://{ip}:{port}' for ip in ips if ip]
         msg = 'Web portal URLs: ' + colorama.Fore.CYAN + ' '.join(ips) + colorama.Style.RESET_ALL
         _logger.info(msg)
+        return config
```
liuzhe-lz, line 122: Is the return value ever used?
If no I suggest to remove it because I will not expect a function named "start" to return a config.
We can add it back any time when we really need it.

QuanluZhang: this is not `start`, it is `_start_impl`. they are different. `start` may include more logic than `_start_impl`, the canonicalize is called in `_start_impl`, it would be better for it to return the canonicalized config for the following usage. It is used in RetiariiExperiment

nni/retiarii/experiment/pytorch.py
```diff
@@ -303,103 +218,49 @@ def _start_strategy(self):
...
+            assert config.execution_engine.benchmark is not None, \
+                '"benchmark" must be set when benchmark execution engine is used.'
+            engine = BenchmarkExecutionEngine(config.execution_engine.benchmark)
         else:
-            raise ValueError(f'Unsupported engine type: {self.config.execution_engine}')
+            raise ValueError(f'Unsupported engine type: {config.execution_engine}')
         set_execution_engine(engine)
 
-        self.id = management.generate_experiment_id()
+    def start(self, port: int = 8080, debug: bool = False, run_mode: RunMode = RunMode.Background) -> None:
```
liuzhe-lz, line 252: Suggest the parameters to be `*args, **kwargs`

nni/retiarii/execution/api.py
```diff
@@ -21,9 +22,10 @@ def set_execution_engine(engine: AbstractExecutionEngine) -> None:
     if _execution_engine is None:
         _execution_engine = engine
     else:
-        raise RuntimeError('Execution engine is already set. '
-                           'You should avoid instantiating RetiariiExperiment twice in one process. '
-                           'If you are running in a Jupyter notebook, please restart the kernel.')
+        warnings.warn('Execution engine is already set. '
```
matluster, line 25: Have you tested the behavior of running a notebook cell twice? Does it work, or not?

nni/retiarii/experiment/config/experiment_config.py
```diff
@@ -0,0 +1,61 @@
...
+        self.execution_engine = execution_engine
+
+    def _canonicalize(self, _parents):
+        msg = '{} is not supposed to be set in Retiarii experiment by users, your config is {}.'
+        if self.search_space != '':
+            raise ValueError(msg.format('search_space', self.search_space))
+        if str(self.trial_code_directory) != '.' and not os.path.isabs(self.trial_code_directory):
+            raise ValueError(msg.format('trial_code_directory', self.trial_code_directory))
+        if self.trial_command != '_reserved' and \
+            not self.trial_command.startswith('python3 -m nni.retiarii.trial_entry '):
```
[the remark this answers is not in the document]
liuzhe-lz: Not really. `sys.executable` might not work on training services. We need to rethink the design here.
Mentioned above in compression experiment's review.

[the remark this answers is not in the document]
matluster: Just a reminder, python3 doesn't work on some of our pipelines.

examples/nas/multi-trial/mnist/search.py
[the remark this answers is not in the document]
QuanluZhang: it is needed, otherwise the python script won't exit. because different from hpo experiment, retiarii experiment starts advisor in a thread which waits for nnimanager to be stopped

nni/experiment/experiment.py
QuanluZhang: do we have unit test for url_prefix?

QuanluZhang: in the previous version of retiarii experiment, url_prefix is None? BTW, how to configure this field?

nni/experiment/experiment.py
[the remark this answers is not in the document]
QuanluZhang: leave this part to @liuzhe-lz

nni/experiment/experiment.py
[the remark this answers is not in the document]
QuanluZhang: updated

nni/retiarii/experiment/config/experiment_config.py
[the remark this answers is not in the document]
QuanluZhang: leave it to next release

nni/retiarii/experiment/pytorch.py
QuanluZhang: why? this is the signature of the hpo experiment start function

nni/retiarii/experiment/pytorch.py
```diff
@@ -411,81 +272,55 @@ def run(self, config: Optional[RetiariiExeConfig] = None, port: int = 8080, debu
...
+        self._stop()
+        if self._dispatcher_thread:
+            self._dispatcher_thread.join()
         self._dispatcher = cast(RetiariiAdvisor, None)
         self._dispatcher_thread = None
         _logger.info('Experiment stopped')
 
     def export_top_models(self, top_k: int = 1, optimize_mode: str = 'maximize', formatter: str = 'dict') -> Any:
         """
+        TODO: the base class may also need this method
```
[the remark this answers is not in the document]
QuanluZhang: moved

ts/nni_manager/core/nnimanager.ts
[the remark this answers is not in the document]
QuanluZhang: yes, a type of kill. it is a soft signal. actually, this sub-thread (advisor) can be seen as the interface of nnimanager. as strategy needs to easily call this interface, this sub-thread is launched in the strategy process.

nni/retiarii/execution/api.py
QuanluZhang: checked, and fixed bugs

nni/experiment/config/experiment_config.py
QuanluZhang: too me, it is more urgly, because base class has the logic of sub-class. anyway, updated

nni/experiment/experiment.py
liuzhe-lz: You can set `self.url_prefix` to `None`. This is an undocumented feature in Python API, but is documented in `nnictl`.
In NNI manager it does have unit test. But I cannot remember whether it has integration test. Since not documented in Python API it should be tested with `nnictl create`, and the first version was not implemented by me.

examples/nas/multi-trial/mnist/search.py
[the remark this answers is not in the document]
ultmaster: Why atextit can't register a stop?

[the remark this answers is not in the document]
QuanluZhang: stop already registered, but the main script does not exit due to a running sub-thread (advisor)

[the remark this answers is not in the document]
ultmaster: atexit triggers stop. Stop kills nnimanager, advisor detects the termination of nnimanager and exits.

nni/retiarii/experiment/pytorch.py
liuzhe-lz: But this is not HPO experiment class. You have totally different `run()` signature, so there is no reason assume identical `start()` signature.

ts/nni_manager/core/nnimanager.ts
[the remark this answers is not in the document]
matluster: Suggest adding comments in the code if you insist on doing this.

nni/runtime/msg_dispatcher_base.py
```diff
@@ -18,15 +18,32 @@
...
         self._channel = TunerCommandChannel(command_channel_url)
+        # NOTE: `connect()` should be put in __init__. First, this `connect()` affects nnimanager's
+        # starting process, without `connect()` nnimanager is blocked in `dispatcher.init()`.
+        # Second, nas experiment uses a thread to execute `run()` of this class, thus, there is
+        # no way to know when the websocket between nnimanager and dispatcher is built. The following
+        # logic may crash is websocket is not built. One example is updating search space. If updating
+        # search space too soon, as the websocket has not been built, the rest api of updating search
+        # space will timeout.
+        # FIXME: this is making unittest happy
+        if not command_channel_url.startswith('ws://_unittest_'):
```
matluster, line 45: Why can't we have a flag in `__init__`, e.g., `auto_connect`. When it's true, the channel connects automatically in init.

nni/retiarii/experiment/pytorch.py
```diff
@@ -411,75 +272,47 @@ def run(self, config: Optional[RetiariiExeConfig] = None, port: int = 8080, debu
...
+            canonicalized_config = cast(RetiariiExeConfig, canonicalized_config)
+            self._dispatcher = RetiariiAdvisor(ws_url)
+            self._dispatcher_thread = Thread(target=self._dispatcher.run, daemon=True)
+            self._dispatcher_thread.start()
+            # FIXME: engine cannot be created twice
+            self._create_execution_engine(canonicalized_config)
+            try:
+                self._run_strategy(canonicalized_config)
+                # FIXME: move this logic to strategy with a new API provided by execution engine
+                self._wait_completion()
```
matluster, line 302: Maybe you can check the return value here and raise some warning / error if the status is `ERROR`.

nni/runtime/msg_dispatcher_base.py
QuanluZhang: `auto_connect` means we should also provide the option that `auto_connect` is false in dispatcher, which is dangerous because we do not know whether it works except unit test. it is safe to leave it to the refactor of dispatcher.

nni/retiarii/experiment/pytorch.py
QuanluZhang: @liuzhe-lz what do you think? if you agree, I will update both hpo and nas experiment according to @matluster 's suggestion.

ts/nni_manager/core/nnimanager.ts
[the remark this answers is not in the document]
QuanluZhang: added

ultmaster (approved):

nni/retiarii/execution/base.py
```diff
@@ -1,12 +1,15 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+from __future__ import annotations
```
liuzhe-lz, line 3: Add an empty line before this.
https://peps.python.org/pep-0008/#module-level-dunder-names

nni/retiarii/experiment/pytorch.py
liuzhe-lz: I suggest to keep current behavior (return `False` on error) before 3.0.

liuzhe-lz (approved):

nni/experiment/config/experiment_config.py
```diff
@@ -164,10 +164,11 @@ def _validate_canonical(self):
         # currently I have only seen one issue of this kind
         #Path(self.experiment_working_directory).mkdir(parents=True, exist_ok=True)
 
-        utils.validate_gpu_indices(self.tuner_gpu_indices)
+        if type(self).__name__ != 'RetiariiExeConfig':
```
J-shang, line 167: maybe we should place this logic in somewhere like `HPOConfig` in the future

hzhua (approved):

nni/retiarii/experiment/config/engine_config.py
```diff
@@ -0,0 +1,41 @@
...
+    # input used in GraphConverterWithShape. Currently support shape tuple only.
+    dummy_input: Optional[List[int]] = None
+
+@dataclass(init=False)
+class CgoEngineConfig(ExecutionEngineConfig):
+    name: str = 'cgo'
+    max_concurrency_cgo: Optional[int] = None
+    batch_waiting_time: Optional[int] = None
+    # input used in GraphConverterWithShape. Currently support shape tuple only.
+    dummy_input: Optional[List[int]] = None
```
J-shang, line 36: maybe `dummy_input_shape`, but it is fine for me if we will support tensor or other types of input in the future

nni/retiarii/experiment/config/experiment_config.py
```diff
@@ -0,0 +1,60 @@
...
+        # TODO: maybe we should also allow users to specify trial_code_directory
+        if str(self.trial_code_directory) != '.' and not os.path.isabs(self.trial_code_directory):
+            raise ValueError(msg.format('trial_code_directory', self.trial_code_directory))
+        if self.trial_command != '_reserved' and \
+            not self.trial_command.startswith('python3 -m nni.retiarii.trial_entry '):
+            raise ValueError(msg.format('trial_command', self.trial_command))
+
+        if isinstance(self.execution_engine, str):
+            self.execution_engine = execution_engine_config_factory(self.execution_engine)
+        if self.execution_engine.name in ('py', 'base', 'cgo'):
```
J-shang, line 55: what's the trial command if `benchmark` or `oneshot`? we don't need a command under these engines?

J-shang (approved):

PR #4855, opened 2022-05-10
Support multiple HPO experiments in one process

liuzhe-lz:
### Description ###

Support creating and running multiple HPO experiments in one Python process, either sequentially or concurrently.

Don't want to add test until new IT framework complete.

### Checklist ###
  - [ ] test case
  - ~~doc~~

### How to test ###

```python
exp1 = Experiment(...)
exp1.start()
exp2 = Experiment(...)
exp3 = Experiment(...)
exp2.start()
exp3.start()
wait_completion(exp1)  # unbelievably we do not have this API
exp1.stop()
exp4 = Experiment(...)
exp4.start()
# ctrl-c
```


nni/runtime/protocol.py
```diff
@@ -7,26 +7,12 @@
...
-    global _use_ws
-    _use_ws = True
-    shim.connect(url)
 
 def send(command: CommandType, data: str) -> None:
-    if _use_ws:
-        shim.send(command, data)
-    else:
-        legacy.send(command, data)
+    legacy.send(command, data)
```
J-shang, line 12: do we have somewhere still use `protocal.send` or `protocal.receive`?

J-shang (approved):

nni/runtime/protocol.py
```diff
@@ -7,26 +7,12 @@
 
 from .tuner_command_channel.command_type import CommandType
 from .tuner_command_channel import legacy
-from .tuner_command_channel import shim
```
matluster, line 10: This file is abandoned?

nni/runtime/protocol.py
liuzhe-lz: Only in legacy UT.

nni/runtime/protocol.py
liuzhe-lz: Yes. The migration is nearly complete.

matluster: Well, in that case, I suggest adding a deprecated warning when someone tries to import this file.

ultmaster (approved):

nni/runtime/protocol.py
liuzhe-lz: Removed.

nni/tools/nnictl/launcher.py
```diff
@@ -88,11 +88,14 @@ def create_experiment(args):
...
 
-    _logger.info(f'To stop experiment run "nnictl stop {exp.id}" or "nnictl stop --all"')
-    _logger.info('Reference: https://nni.readthedocs.io/en/stable/reference/nnictl.html')
+    if foreground:
+        exp.run(port, debug=debug)
+
+    else:
+        exp.start(port, debug, RunMode.Detach)
+        _logger.info(f'To stop experiment run "nnictl stop {exp.id}" or "nnictl stop --all"')
+        _logger.info('Reference: https://nni.readthedocs.io/en/stable/reference/nnictl.html')
```
QuanluZhang, line 98: these lines imply that the only different between `exp.start` and `exp.run` is that `exp.run` will wait until experiment completes, right? better to put this information in docstring.

liuzhe-lz: If I remember correctly, end users should always use `run()` according a previous discussion. `run()` has carefully designed and reviewed parameters, while `start()` has kept its "experimental" signature.
We never mention `start()` in tutorial.
I have once thought about renaming `start()` to `_start()`, but gave up due to compatibility concern.

Back to the code. I have to admit line 93 is incorrect. It should be `exp.start(..., Foreground) ; exp.wait_completion()`. I don't know why I forgot about that...

Issue #4839, opened 2022-05-04
Hyper-parameter overview top trials order


TimSchim:
In the WebUI under: Trial details -> Hyper-parameter one can choose to show the top xx% trials.
By default this shows the trials with the highest scores.
Is there a way to change this to showing the trials with the lowest scores, e.g. if my metric is MSE?
So far my workaround is to just to multiply my score by (-1).

QuanluZhang:
@Lijiaoa  please help follow this issue.

Lijiaoa:
@TimSchim Why not set `optimize_mode` to `minimize` in config? If you set it to `minimize`, webui will filter best trials as small final metric.

![image](https://user-images.githubusercontent.com/61399850/166855567-43b7794d-ecdb-4980-81f5-a598d18cd589.png)

TimSchim:
This seems to only work when using a tuner that uses this option. When I use Random instead of TPE I get
`TypeError: __init__() got an unexpected keyword argument 'optimize_mode'`

Tried these two options:

- experiment.config.tuner.class_args = {'optimize_mode': 'minimize'}

- experiment.config.tuner.class_args['optimize_mode'] = 'minimize'

when setting the config from Python like [here](https://nni.readthedocs.io/en/stable/tutorials/hpo_quickstart_pytorch/main.html).

Lijiaoa:
It seems that random tuner doesn't support this arg by [doc](https://nni.readthedocs.io/en/stable/reference/hpo.html?highlight=Random#random-tuner).

TimSchim:
Any other suggestions or should this be a new feature for future versions of the Web Portal?

Lijiaoa:
You could change [this line](https://github.com/microsoft/nni/blob/master/ts/webui/src/components/experiment/trialdetail/chart/Para.tsx#L203) into `convertedTrials.sort((a, b) => (a[k] - b[k]));` and delete the `EXPERIMENT` from [this line](https://github.com/microsoft/nni/blob/master/ts/webui/src/components/experiment/trialdetail/chart/Para.tsx#L7). And you could follow the [Readme](https://github.com/microsoft/nni/tree/master/ts/webui) to format the code.

We will discuss this feature in the scrum. I'll let you know right away if there's any progress on this matter.

ThePhoenixCoding:
I'm having the same problem and can add some information: I use RegularizedEvolution which supports optimization_mode = 'minimize'. 
[...]

PR #4856, opened 2022-05-10
Add args optimize_mode to random and grid search tuner

liuzhe-lz:
### Description ###

Add optional arguments `optimize_mode` to random tuner and grid search tuner.

The value will be used by web portal to sort trials: #4839

### Checklist ###
  - ~~test case~~
  - ~~doc~~

### How to test ###

Create experiment with random tuner in different optimize modes. Check that web portal sorts the trials in correct order.

### DIFF
--- a/nni/algorithms/hpo/gridsearch_tuner.py
+++ b/nni/algorithms/hpo/gridsearch_tuner.py
@@ -88,7 +88,7 @@ class GridSearchTuner(Tuner):
         config.tuner.name = 'GridSearch'
     """
 
-    def __init__(self):
+    def __init__(self, optimize_mode=None):
         self.space = None
 
         # the grid to search in this epoch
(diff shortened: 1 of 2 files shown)


Lijiaoa:
Maybe update the [doc](https://nni.readthedocs.io/en/stable/reference/hpo.html?highlight=Random#random-tuner) is more friendly for user.

nni/algorithms/hpo/gridsearch_tuner.py
```diff
@@ -88,7 +88,7 @@ class GridSearchTuner(Tuner):
         config.tuner.name = 'GridSearch'
     """
 
-    def __init__(self):
+    def __init__(self, optimize_mode=None):
```
J-shang, line 91: a little strange but it's ok for me, maybe let user config this on web portal is better (Top max 10 or Top min 10, default visualization is `optimize_mode` if it has)...

J-shang (approved):

nni/algorithms/hpo/random_tuner.py
```diff
@@ -42,13 +42,15 @@ class RandomTuner(Tuner):
         The random seed.
     """
 
-    def __init__(self, seed: int | None = None):
+    def __init__(self, seed: int | None = None, optimize_mode: str | None = None):
```
ultmaster, line 45: Literal[maximize, minimize]

ultmaster (approved):

PR #4858, opened 2022-05-12
Azure pipeline filter

sunqinzheng:
### Description ###

Tools for conditional azure pipeline execution.

### Checklist ###
  - [ ] test case
  - [ ] doc


matluster:
It seems to be a general tool that can benefit many repos rather than NNI only.

Is it possible to maintain it in one of your personal repos, and NNI just use it as an external plugin?

PR #4859, opened 2022-05-13
Add APIs to communicate with NNI manager

yuhuixii:
### Description ###
Add low level APIs for algorithms to communicate with NNI manager.

  - [ ] TunerCommandChannel implementation.
### Checklist ###
  - [ ] test case
  - [ ] doc


nni/runtime/tuner_command_channel/channel.py
```diff
@@ -49,6 +50,18 @@ def disconnect(self) -> None:
     # def receive(self) -> Command | None:
     #     ...
 
+    def send(self, command: KillTrialJob) -> None:
```
liuzhe-lz, line 53: There should be a base class. Python does not support overloading.

nni/runtime/tuner_command_channel/semantic_command.py
```diff
@@ -0,0 +1,51 @@
...
+
+class BasicCommand:
+    def __init__(self, trial_job_id, parameter_id, parameter_index, parameters, parameter_source):
+        self.trial_job_id = trial_job_id
+        self.parameter_id = parameter_id
+        self.parameter_index = parameter_index
+        self.parameters = parameters
+        self.parameter_source = parameter_source
+
+class ReportMetricData:
```
liuzhe-lz, line 12: Command classes should be [`dataclass`](https://docs.python.org/3/library/dataclasses.html)

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -48,6 +56,35 @@ def disconnect(self) -> None:
...
+        command_dict = command.__dict__
+        command_json = json.dumps(command_dict)
+        self._channel.send(command_json)
+
+    def receive(self) -> Command | None:
+        command_json = self._channel.receive()
+        if command_json is None:
+            raise RuntimeError('NNI manager closed connection')
+        command = json.loads(command_json)
+        if command['command_type'] == 'ReportMetricData':
```
liuzhe-lz, line 69: There can be a factory function to avoid if-else.
https://github.com/microsoft/nni/blob/master/nni/experiment/config/utils/internal.py#L171

nni/runtime/tuner_command_channel/semantic_command.py
```diff
@@ -0,0 +1,61 @@
...
+    trial_job_id: str = ''
+    parameter_id: int = 0
+    parameter_index: int = 0
+    parameters: dict = {}
+    parameter_source: str = ''
+
+@dataclass
+class ReportMetricData:
+    command_type: str = 'ReportMetricData'
+    trial_job_id: str = ''
```
liuzhe-lz, line 20: There is no need to have default value here.

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -6,11 +6,19 @@
...
 
 __all__ = ['TunerCommandChannel']
 
 from .command_type import CommandType
 from .websocket import WebSocket
+from .semantic_command import ReportMetricData, UpdateSearchSpace
+from .semantic_command import ImportData, TrialEnd
+from .semantic_command import NewTrialJob, SendTrialJobParameter
+from .semantic_command import NoMoreTrialJobs, KillTrialJob
+Command = Union[ReportMetricData, UpdateSearchSpace,
```
liuzhe-lz, line 20: This should be a base class instead of type hint Union.
I expect the base class to include common logic of converting to JSON and from JSON. And if a specific command has special converting logic different from others, it can override the method.

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -6,11 +6,19 @@
...
 
 from __future__ import annotations
+import json
+from typing import Union
 
 __all__ = ['TunerCommandChannel']
 
 from .command_type import CommandType
 from .websocket import WebSocket
+from .semantic_command import ReportMetricData, UpdateSearchSpace
```
liuzhe-lz, line 16: Imported identifiers can be a tuple and a tuple can have line wrap.
`from x import (a, b, c, d, e, f, g)

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -48,6 +50,19 @@ def disconnect(self) -> None:
     #     ...
     # def receive(self) -> Command | None:
     #     ...
+    def send(self, command: BaseCommand) -> None:
+        command_json = command.dump()
+        self._channel.send(command_json)
```
liuzhe-lz, line 55: Please add a temporary compatibility layer. For example `self._channel.send(command._to_legacy_command_type(), command_json)`.

liuzhe-lz:
It looks good to me in general.
Now please add validation to the command classes, update `msg_dispatcher.py` to use the new commands, and run experiments to check the fields are correct.

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -48,6 +50,19 @@ def disconnect(self) -> None:
...
+
+    def receive(self) -> BaseCommand | None:
+        command_json = self._channel.receive()
```
liuzhe-lz, line 58:
```python
old_command_type, command_json = self._channel._receive()
command_json['command_type'] = old_to_new[old_command_type]
...
command = cls.load(command_json)
command.validate()
return command
```

PR #4863, opened 2022-05-16
Migrate remote IT to scale-sets

ultmaster:


ts/nni_manager/training_service/reusable/aml/amlClient.ts
```diff
@@ -135,11 +135,19 @@ export class AMLClient {
...
+            // FIXME: The error will only appear in console.
+            // Still need to find a way to put them into logs.
+            console.error(`Python process stderr: ${error}`);
+        });
         pythonShellClient.on('error', function (error: any) {
-            deferred.reject(error);
+            console.error(`Python process fires error: ${error}`);
+            deferred.reject(error || new Error('AML client Python process unknown error.'));
         });
         pythonShellClient.on('close', function (error: any) {
```
cruiseliu, line 147: `on(event: "close", listener: () => void): this;`
https://github.com/extrabacon/python-shell/blob/master/index.ts#L449

This event has no argument at all.

ts/nni_manager/training_service/reusable/aml/amlClient.ts
```diff
@@ -135,11 +135,19 @@ export class AMLClient {
     
     // Monitor error information in aml python shell client
     private monitorError(pythonShellClient: PythonShell, deferred: Deferred<any>): void {
+        pythonShellClient.on('stderr', function (error: any) {
+            // FIXME: The error will only appear in console.
+            // Still need to find a way to put them into logs.
+            console.error(`Python process stderr: ${error}`);
+        });
         pythonShellClient.on('error', function (error: any) {
```
cruiseliu, line 143: This guarantees to report an `Error`.
https://github.com/extrabacon/python-shell/blob/master/index.ts#L456
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/globals.d.ts#L193

ts/nni_manager/training_service/reusable/aml/amlClient.ts
```diff
@@ -135,11 +135,19 @@ export class AMLClient {
     
     // Monitor error information in aml python shell client
     private monitorError(pythonShellClient: PythonShell, deferred: Deferred<any>): void {
+        pythonShellClient.on('stderr', function (error: any) {
```
cruiseliu, line 138: The argument is not "error". It's stderr's content.
https://github.com/extrabacon/python-shell/blob/master/index.ts#L442

ts/nni_manager/training_service/reusable/aml/amlClient.ts
```diff
@@ -135,11 +135,19 @@ export class AMLClient {
...
+            // FIXME: The error will only appear in console.
+            // Still need to find a way to put them into logs.
+            console.error(`Python process stderr: ${error}`);
```
cruiseliu, line 141:
```typescript
import { getLogger } from 'common/log';
getLogger('AmlClient').error('Python process stderr:', error);
```
Prefer comma separated parameters over template string. The logger uses built-in `inspect()` util to keep as much info as possible, while the default string converting commonly generates meaningless `[Object object]`.

matluster:
Won't fix in this work:

1. TBD: Schedules of pipelines.
2. I won't enable GPU on remote (both w&l), local w, and hybrid, because,
  a. I didn't install nvidia-docker on linux with GPUs.
  b. There are no GPU tests on remote and hybrid's remote part.
  c. (for windows) Singapore's GPUs has been sold out. The VM builder (packer) can't even create a single VM with GPUs. When the image is built upon a machine without GPU, the driver fails to install.
  d. (for windows) We can't install driver on-the-fly because there is no administrative rights.
3. I won't fix the three full tests (HPO / NAS / Compression)

ts/nni_manager/training_service/reusable/aml/amlClient.ts
matluster: I plan to split this PR into smaller ones. Let's discuss later.

PR #4866, opened 2022-05-17
Fix issue 4810

liuzhe-lz:
### Description ###

Issue 4810 reported a bug. The user said it's about TPE nested search space.

It was in fact caused by parameter deserialization. For some reason the trial code requires the parameter to be a plain dict, but in current version it will be `OrderedDict`.

During investigating this problem, however, I did find a bug in TPE nested search space. 🤔

### Checklist ###
  - ~~test case~~
  - ~~doc~~

### How to test ###

Check the search space and trial code in issue 4810.

### DIFF
--- a/nni/common/serializer.py
+++ b/nni/common/serializer.py
@@ -393,7 +393,7 @@ def _dump(*, obj: Any, fp: Optional[Any], use_trace: bool, pickle_size_limit: in
 
 
 def load(string: Optional[str] = None, *, fp: Optional[Any] = None,
-         preserve_order=False, ignore_comments: bool = True, **json_tricks_kwargs) -> Any:
+         preserve_order: bool = False, ignore_comments: bool = True, **json_tricks_kwargs) -> Any:
     """
     Load the string or from file, and convert it to a complex data structure.
     At least one of string or fp has to be not none.
@@ -404,7 +404,7 @@ def load(string: Optional[str] = None, *, fp: Optional[Any] = None,
         JSON string to parse. Can be set to none if fp is used.
     fp : str
         File path to load JSON from. Can be set to none if string is used.
-    preserve_order: bool
+    preserve_order : bool
         Use ``OrderedDict`` instead of ``dict``. Note that the order is always preserved even when this is False.
     ignore_comments : bool
         Remove comments (starting with ``#`` or ``//``). Default is true.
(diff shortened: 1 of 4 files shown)


J-shang (approved):

nni/runtime/platform/local.py
```diff
@@ -53,7 +53,7 @@ def get_next_parameter():
     while not (os.path.isfile(params_filepath) and os.path.getsize(params_filepath) > 0):
         time.sleep(3)
     params_file = open(params_filepath, 'r')
-    params = load(fp=params_file)
+    params = load(fp=params_file, preserve_order=False)
```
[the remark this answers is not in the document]
cruiseliu: I think there's no reason to do so. Start from 3.7 `dict` is guaranteed to keep the order.
Do you suggest to move this parameter into `nni.load()`?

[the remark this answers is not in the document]
liuzhe-lz: Updated.

nni/common/serializer.py
```diff
@@ -403,6 +404,8 @@ def load(string: Optional[str] = None, *, fp: Optional[Any] = None, ignore_comme
         JSON string to parse. Can be set to none if fp is used.
     fp : str
         File path to load JSON from. Can be set to none if string is used.
+    preserve_order: bool
+        Use ``OrderedDict`` instead of ``dict``. Note that the order is always preserved even when this is False.
```
J-shang, line 408: maybe a little confused when I use `dump`, `load` to serialize models, what does `preserve_order` mean

nni/common/serializer.py
```diff
@@ -392,7 +392,8 @@ def _dump(*, obj: Any, fp: Optional[Any], use_trace: bool, pickle_size_limit: in
         return json_tricks.dumps(obj, obj_encoders=encoders, **json_tricks_kwargs)
 
 
-def load(string: Optional[str] = None, *, fp: Optional[Any] = None, ignore_comments: bool = True, **json_tricks_kwargs) -> Any:
+def load(string: Optional[str] = None, *, fp: Optional[Any] = None,
+         preserve_order=False, ignore_comments: bool = True, **json_tricks_kwargs) -> Any:
```
ultmaster, line 396: missing typehint

nni/common/serializer.py
liuzhe-lz: @ultmaster Any idea?

ultmaster: Add a link to json-tricks.

Users will not use it if they don't know what that means.

PR #4892, opened 2022-05-30
Refactor integration test (step 1) - wait for kill

ultmaster:
There is a long-known issue that the node could be still on even after experiment stops / nnictl stop exits. This causes many issues including port not available, experiment status incorrect (could be STOPPING).

This PR attempts to fix this issue by adding a wait of 60 seconds, for the kill command to be successfully executed. It also adds a twice-Ctrl-C to refrain users from killing the cleanup process by accident.

### DIFF
--- a/nni/tools/nnictl/command_utils.py
+++ b/nni/tools/nnictl/command_utils.py
@@ -131,7 +131,6 @@ def _check_pid_running(pid):
     else:
         try:
             indicator, _ = os.waitpid(pid, os.WNOHANG)
-            print(time.time(), 'check pid running', indicator, _)
             return indicator == 0
         except ChildProcessError:
             # One of the reasons we reach here is: pid may be not a child process.
--- a/nni/tools/nnictl/command_utils.py
+++ b/nni/tools/nnictl/command_utils.py
@@ -4,9 +4,10 @@
 from subprocess import call, check_output
 import sys
 import os
+import time
 import signal
 import psutil
-from .common_utils import print_error
+from .common_utils import print_error, print_warning
 
 
 def check_output_command(file_path, head=None, tail=None):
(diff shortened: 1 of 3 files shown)


test/ut/tools/nnictl/test_kill_command.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def test_kill_process_slow_no_patience():
+    process = subprocess.Popen([sys.executable, __file__, '--mode', 'kill_slow'])
+    time.sleep(1)  # wait 1 second for the process to launch and register hooks
+    start_time = time.time()
+    kill_command(process.pid, timeout=1)  # didn't wait long enough
+    end_time = time.time()
+    assert 0.5 < end_time - start_time < 2
+    if sys.platform == 'linux':  # FIXME: on non-linux, seems that the time of termination can't be controlled
```
J-shang, line 56: does this mean the handler in `signal.signal` will not register successfully on non-linux?

J-shang (approved):

test/ut/tools/nnictl/test_kill_command.py
matluster: I tested this on pipeline. Sometimes, the assert is successful. In other times, it's not. It's very strange.

I don't think it's a critical problem now: IT didn't complain; developers seldom use Windows.

nni/tools/nnictl/command_utils.py
```diff
@@ -31,14 +32,25 @@ def check_output_command(file_path, head=None, tail=None):
...
+    """Kill the process of pid (with a terminate signal).
+    Waiting up to 60 seconds until the process is killed.
+    """
+    # TODO: The input argument should better be Popen rather than pid.
     if sys.platform == 'win32':
-        process = psutil.Process(pid=pid)
-        process.send_signal(signal.CTRL_BREAK_EVENT)
+        try:
+            process = psutil.Process(pid=pid)
+            process.send_signal(signal.CTRL_BREAK_EVENT)
```
liuzhe-lz, line 43: I don't know why not using `process.terminate()` or `process.kill()` here.
But well, not your fault.

nni/tools/nnictl/command_utils.py
```diff
@@ -65,6 +77,72 @@ def install_requirements_command(requirements_path):
...
+    keyboard_interrupted = False
+    time_count = 0
+    # Usually, a process is killed very quickly.
+    # This little nap will save 1 second.
+    time.sleep(0.01)
+    while True:
+        try:
+            # Implementation of waiting
+            while time_count < timeout:
+                pid_running = _check_pid_running(pid)
```
liuzhe-lz, line 90: Ah, this time is your fault.
Use `process.is_running()`.
https://psutil.readthedocs.io/en/latest/#psutil.Process.is_running

test/ut/tools/nnictl/test_kill_command.py
```diff
@@ -0,0 +1,121 @@
...
+import os
+import subprocess
+import signal
+import sys
+import signal
+import time
+
+import pytest
+
+from nni.tools.nnictl.command_utils import kill_command, _check_pid_running
```
liuzhe-lz, line 12: Is this case testing the `kill_command()` function?
I think it's not a public API, and NNI experiments are designed to be stopped with REST API.

liuzhe-lz:
I think we need a review meeting for this series of PRs.
Maybe using the meeting this afternoon?

nni/tools/nnictl/command_utils.py
matluster: I don't know either why this is pid here instead of process.

I noticed the usage of psutil is limited to Windows. It reminds that someone once told me that psutil is something binary, and we plan to drop it in future.

test/ut/tools/nnictl/test_kill_command.py
matluster: `kill_command()` means: kill a process and wait for the kill to finish.

liuzhe-lz:
I will "approve" all these to avoid RTT of discussion...

liuzhe-lz (approved):

PR #4896, opened 2022-05-30
Refactor integration test (step 5) - VM image builder

ultmaster:
Add a VM image builder, that builds Linux and Windows VM images, used in VMSS pipeline agents.

All technical details can be found in the scripts. I think it's well-commented.

GPU support is not complete. As a result, in follow-up PRs, I won't enable GPU on remote (both w&l), local w, and hybrid, because,

1. I didn't install nvidia-docker on linux with GPUs.
2. There are no GPU tests on remote and hybrid's remote part.
3. (for windows) Singapore's GPUs has been sold out. The VM builder (packer) can't even create a single VM with GPUs. When the image is built upon a machine without GPU, the driver fails to install.
4. (for windows) We can't install driver on-the-fly because there is no administrative rights.

### DIFF
--- a/test/vso_tools/build_vm/change_ssh_port.sh
+++ b/test/vso_tools/build_vm/change_ssh_port.sh
@@ -1,19 +1,9 @@
 #!/bin/bash
 
+# Microsoft internal subscription has a firewall (network security group),
+# to deny traffic from low ports (maybe less than 10000).
+# We have to change the port at VM provision, so that the VM can be connected and build scripts can run.
+
 CONFIG_PATH=/etc/ssh/sshd_config
-# CONFIG_PATH=sshd_config
 sudo sed -i -e "s/#Port 22/Port 10022/g" $CONFIG_PATH
 sudo service ssh restart
-
-# {
-#   "type": "shell-local",
-#   "inline": [
-#     "az network nsg rule create -g <resource_group> --nsg-name <network_security_group> -n {{ build `PackerRunUUID` }} \\",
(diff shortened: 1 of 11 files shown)


test/vso_tools/build_vm/change_ssh_port.sh
```diff
@@ -0,0 +1,19 @@
...
+#   "inline": [
+#     "az network nsg rule create -g <resource_group> --nsg-name <network_security_group> -n {{ build `PackerRunUUID` }} \\",
+#     "  --priority 100 --source-address-prefixes $(curl ifconfig.me) --source-port-ranges '*' \\",
+#     "  --destination-address-prefixes '*' --destination-port-ranges '*' --access Allow --protocol '*' \\",
+#     "  --description 'Disposable rule for packer build {{ build `PackerRunUUID` }}'",
+#     "export NIC=$(az network nic list -g <resource_group> --query [].'name' --output tsv | grep pkrni | head -1)",
+#     "echo \"NIC Found: ${NIC}\"",
+#     "az network nic update -n ${NIC} -g <resource_group> --network-security-group <network_security_group>"
+#   ]
+# },
```
J-shang, line 19: remove?

test/vso_tools/build_vm/change_ssh_port.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
cruiseliu, line 1: Suggest adding a readme in build_vm so future maintainers will know when should they run this.

liuzhe-lz (approved):

PR #4897, opened 2022-05-30
Refactor integration test (step 6) - pipeline templates

ultmaster:
This PR modifies and adds several pipeline templates, which are the common reusable utilities for pipeline YAMLs.

* Multiple ways to configure python: no configure, default, and install+configure.
* AzureML dependencies install.
* Install customized tuner.
* Support GPU platform.
* Support wheel building as one of installation method.
* Experiment artifact uploading in case of failure.
* Agent-setup is no longer required because we have a VM image now.


pipelines/templates/config-version.yml
```diff
@@ -1,19 +1,56 @@
 parameters:
 - name: platform
   type: string
+- name: python
```
cruiseliu, line 4: Maybe `python_env`? This name seems too generic.

pipelines/templates/config-version.yml
```diff
@@ -1,19 +1,56 @@
...
+      touch x64.complete
+    displayName: Create Python 3.7 venv
+  ${{ else }}:
+    script: |
+      set -e
+      mkdir -p $(Agent.ToolsDirectory)/Python
+      cd $(Agent.ToolsDirectory)/Python
+      PY39_VER=$(python3.9 -c "import sys; print('.'.join([f'{val}' for val in sys.version_info[:3]]))")
+      mkdir $PY39_VER
+      ln -s $PY39_VER 3.9
```
cruiseliu, line 39: If I upgrade latest to Python 3.10, do I need to update the vmss image?

liuzhe-lz (approved):

pipelines/templates/config-version.yml
matluster: Yes. We didn't install python 3.10 in the image, because I thought we only need python 3.7 and 3.9.

New python versions are installed from deadsnakes/ppa. So it's just a new line of apt-get.

pipelines/templates/install-customized-tuner.yml
```diff
@@ -0,0 +1,14 @@
...
+      cd examples/tuners/customized_tuner
+      python setup.py develop --user
+      nnictl algo register --meta meta_file.yml
+  ${{ else }}:
+    script: |
+      set -e
+      cd examples/tuners/customized_tuner
+      python setup.py develop
+      nnictl algo register --meta meta_file.yml
+  displayName: Install customized tuner
```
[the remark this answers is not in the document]
matluster: I respect, as much as possible, the original design of each test.

PR #4975, opened 2022-06-30
upgrade react router to v6

Lijiaoa:
### Description ###

- Route highlighting while completing the default route `/oview`

  
### Release Note ###
   None

### How to test ###
-  start url prefix exp
-  start none prefix exp


ts/webui/src/index.tsx
```diff
@@ -13,24 +13,24 @@ import * as serviceWorker from './serviceWorker';
...
-        </Suspense>
-    </Router>,
+    <Suspense
+        fallback={
+            <div className='loading'>
+                <img title='loading-graph' src={(path || '') + '/loading.gif'} />
+            </div>
+        }
+    >
+        <Router basename={path === undefined ? '' : path}>
```
liuzhe-lz, line 23: `path ?? ''`

liuzhe-lz:
After v2.9 release I plan to bump all dependencies to latest.
It includes nodejs 18, react 18, etc.

liuzhe-lz (approved):

ts/webui/src/index.tsx
Lijiaoa: What's the difference between `??` and `||`

liuzhe-lz: > What's the difference between `??` and `||`

`??` checks `undefined` and `null`, while `||` checks all falsy value (false, zero, empty string, etc).
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

Lijiaoa: Thanks, get it!

ultmaster (approved):

PR #4976, opened 2022-06-30
Prepare for multi-framework support in NAS

ultmaster:
### Description ###

Add a shortcut for `nni.retiarii.nn.pytorch`. Now we can `from nni.retiarii.nn import LayerChoice`.

Fix an issue that prevents Retiarii from being imported when PyTorch is not installed.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [x] test case
  - [x] doc


nni/common/framework.py
```diff
@@ -0,0 +1,93 @@
...
+    return DEFAULT_FRAMEWORK
+
+
+def shortcut_module(current: str, target: str, package: Optional[str] = None) -> None:
+    """Make ``current`` module an alias of ``target`` module in ``package``."""
+    # Reference: https://github.com/dmlc/dgl/blob/d70a362dba8d46fd9838c79d76998a5e33f22cb7/python/dgl/nn/__init__.py#L27
+    mod = importlib.import_module(target, package)
+    thismod = sys.modules[current]
+    for api, obj in mod.__dict__.items():
+        setattr(thismod, api, obj)
```
J-shang, line 86: could IDE (like VSCode) auto prompt shortcut like `nni.retiarii.codegen.xxx` if I `nni.set_default_framework('pytorch')` 
and want to import `nni.retiarii.codegen.pytorch.xxx`

test/ut/nas/test_import_nodep.py
```diff
@@ -0,0 +1,42 @@
...
+import argparse
+import subprocess
+import sys
+
+import pytest
+
+
+def import_related(mask_out):
+    import nni
+    nni.set_default_framework(mask_out)
```
J-shang, line 12: should we prevent the user from calling this function repeatedly?

nni/common/framework.py
ultmaster: I'm afraid not. We could investigate whether we can implement a plugin for language servers like Pylance however.

test/ut/nas/test_import_nodep.py
ultmaster: We don't. In case they change the framework in the middle of a script, modules already imported will keep the same. Newly imported modules will respect the new setting.

nni/common/framework.py
J-shang: A plugin is cool, and I think this plugin is important for users to use the shortcut.

J-shang (approved):

nni/common/framework.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import importlib
+import os
+import sys
+from typing import Optional, cast
+
+from nni.typehint import Literal
```
liuzhe-lz, line 9: I remember we should prefer `from typing_extensions` now?

nni/common/framework.py
```diff
@@ -0,0 +1,93 @@
...
+
+__all__ = ['set_default_framework', 'get_default_framework', 'shortcut_module', 'shortcut_framework']
```
liuzhe-lz, line 11: PEP 8 suggest to place this before `import`.

nni/common/framework.py
liuzhe-lz: Maybe a `.pyi` file can help?

liuzhe-lz (approved):

nni/common/framework.py
ultmaster: Yeah... really counter-intuitive.

Issue #4974, opened 2022-06-30
Error: Dispatcher stream error, tuner may have crashed.


Sunny-Island:
**Describe the issue**:

I have searched previous issue. dispatcher.log reports that **one of the lower bounds is greater than an upper bound**, but my search sapce is:
```
search_space_xgboost = {    'max_depth': {'_type': 'quniform', '_value': [3, 10, 1]},    'gamma': {'_type': 'uniform', '_value': [0,0.2]},
    'min_child_weight': {'_type': 'quniform', '_value': [0,6,1]},    'eta': {'_type': 'uniform', '_value': [0.01,0.3]},}
```

nnimanager.log:
```
[2022-06-29 20:01:19] ERROR (tuner_command_channel.WebSocketChannel) Error: Error: tuner_command_channel: Tuner closed connection    at WebSocket.<anonymous> (/root/miniconda3/envs/automl/lib/python3.7/site-packages/nni_node/core/tuner_command_channel/websocket_channel.js:41:49)    at WebSocket.emit (node:events:538:35)
    at WebSocket.emitClose (/root/miniconda3/envs/automl/lib/python3.7/site-packages/nni_node/node_modules/express-ws/node_modules/ws/lib/websocket.js:246:10)
    at Socket.socketOnClose (/root/miniconda3/envs/automl/lib/python3.7/site-packages/nni_node/node_modules/express-ws/node_modules/ws/lib/websocket.js:1127:15)    at Socket.emit (node:events:526:28)    at TCP.<anonymous> (node:net:687:12)
[2022-06-29 20:01:19] ERROR (NNIManager) Dispatcher error: tuner_command_channel: Tuner closed connection[2022-06-29 20:01:19] ERROR (NNIManager) Error: Dispatcher stream error, tuner may have crashed.    at EventEmitter.<anonymous> (/root/miniconda3/envs/automl/lib/python3.7/site-packages/nni_node/core/nnimanager.js:647:32)
    at EventEmitter.emit (node:events:526:28)    at WebSocketChannelImpl.handleError (/root/miniconda3/envs/automl/lib/python3.7/site-packages/nni_node/core/tuner_command_channel/websocket_channel.js:107:22)
    at WebSocket.<anonymous> (/root/miniconda3/envs/automl/lib/python3.7/site-packages/nni_node/core/tuner_command_channel/websocket_channel.js:41:37)
    at WebSocket.emit (node:events:538:35)
[...]

PR #4977, opened 2022-06-30
Add condition control when passing parameters bounds to scipy.minimize.

Sunny-Island:
### Description ###
See [Issue](https://github.com/microsoft/nni/issues/4974). When using GP tuner, it will pass wrong search space bound to scipy, which will fail bound check. Previous code is:
```
    bounds_minmax = np.array(
        [[bound['_value'][0], bound['_value'][-1]] for bound in bounds])
```
But GP tuner should support both quniform and qloguniform, which is [low, high, q]. `bound['_value'][-1]` will visit q instead of higher bound.

### DIFF
--- a/nni/algorithms/hpo/gp_tuner/util.py
+++ b/nni/algorithms/hpo/gp_tuner/util.py
@@ -89,7 +89,8 @@ def acq_max(f_acq, gp, y_max, bounds, space, num_warmup, num_starting_points):
     x_seeds = [space.random_sample() for _ in range(int(num_starting_points))]
 
     bounds_minmax = np.array(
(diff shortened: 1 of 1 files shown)


Sunny-Island:
Please review this pull request. @liuzhe-lz @ultmaster @J-shang @suiguoxin @leckie-chn

Sunny-Island:
This PR fails pylint test, but the problem is caused by #4883. It seems it will be fixed later.

suiguoxin (approved):

ultmaster (approved):
On behalf of @suiguoxin

PR #4990, opened 2022-07-07
Bug fix: after clicking compare button, the dropdown button (for selecting different keys) is not shown

Lijiaoa:
### Description ###

- fix title bug
  - rename `Compare.tsx` to `CompareIndex.tsx`
  - delete compareIndex props: showDetails, because it could be got by its `title` property
  - delete compareIndex props: intermediateKeyList , deal with it in CompareIndex file

- about Trial class changes:
   - delete `final` and `finalAcc` field
   - delete `description` field (so change some other files)
   - add `parameter`(obj) field (so change some other files)
- about trialManager.ts 
   - add `allTrialsIntermediateChart` function (to filter Intermediate result all line chart data)
- Refactor Intermediate all line chart with the function: `allTrialsIntermediateChart `
- some little changes
  - hyper-parameter chart only render `succeed` trials
  - duration chart only render not `waiting` trials
     - update duration file `trialId` type: it's `number[]` rather than `string[]`
- about `TableList.tsx` file
  - get ret by `trial.tableRecord` 
- delete some interface: TableObj, Intermedia, Parameters, ErrorParameter
- delete interface:TableRecord's `accuracy` and `accDictionary` field because it's unuseful

#### Test Options ####
  - [x] test mnist-tfv2 experiment
  - [x] test final result is dict experiment
  - [x] test intermediate result is dict experiment
  - [ ] test nested search space experiment

Issue #4969, opened 2022-06-28
detail page empty with tensorflow tutorial code because of the "None"


osayes:
![temp](https://user-images.githubusercontent.com/18533904/176320472-c1099be4-693b-4e0e-9fae-0a5c78d9911c.jpg)
tutorial link:
https://nni.readthedocs.io/en/stable/tutorials/hpo_quickstart_tensorflow/main.html
https://nni.readthedocs.io/zh/stable/tutorials/hpo_quickstart_tensorflow/main.html
No one would have thought the problem was here

platform: win10
nni version: 2.8
tensorflow version: 2.7.0
python version: 3.9.7

ultmaster:
It's a WebUI bug. @Lijiaoa

PR #5011, opened 2022-07-21
detail page empty with tensorflow tutorial code because of the "None"

Lijiaoa:
### Description ###
**should be merged after the PR #4990 !!!** 

**fix issue:** #4969 
**background:** for hpo experiment: its search space has "None"(in the yaml file) value and trial parameters have `null`(in ts) value
**cause bug:** webui hyper-parameter chart will broke down because the `null`
**resolve solution:** convert the `null` to string 'null' in code

### How to test ###
as the issue to run an hpo experiment, and when having succeeded trials to view hyper-parameter chart. If the chart is normal and the page is not broken down, the fix PR worked fine~


ultmaster (approved):

ts/webui/src/static/function.ts
```diff
@@ -254,7 +254,12 @@ function formatComplexTypeValue(value: any): string | number {
     if (['number', 'string'].includes(typeof value)) {
         return value;
     } else {
-        return value.toString();
+        // for hpo experiment: search space choice value is None, and it shows null
+        if (Object.is(null, value)) {
```
liuzhe-lz, line 258: I think `String(value)` is more general and robust.

Lijiaoa: ok! string() is better than toString

ts/webui/src/static/model/trial.ts
```diff
@@ -234,6 +234,12 @@ class Trial implements TableObj {
         } else {
             const tempHyper = this.info.hyperParameters;
             let params = JSON.parse(tempHyper[tempHyper.length - 1]).parameters;
+            // for hpo experiment: search space choice value is None, and it shows null
+            for (const key in params) {
```
liuzhe-lz, line 238: Use for-of if there is not special reason.

Lijiaoa: yes!

liuzhe-lz (approved):

Issue #5026, opened 2022-07-27
frameworkcontroller reuseMode


N-Kingsley:
**Describe the issue**: How to set 'reuseMode=false'? It seems that when reuseMode is true, the code in reusable is executed, and when reuseMode is false, the code in kubernetes is executed.
![image](https://user-images.githubusercontent.com/38311217/181188776-4be408af-c3f8-4cc3-9b2d-008aa4a15d58.png)

If the code in reusable is executed, the error as below will be got:
![image](https://user-images.githubusercontent.com/38311217/181189012-9d541905-e8df-48b8-b2e3-f03105bf92da.png)

**config.yml**:
experimentName: example_mnist_pytorch
trialConcurrency: 1
maxExecDuration: 1h
maxTrialNum: 2
debug: false
reuse: false
#nniManagerIp: 
#choice: local, remote, pai, kubeflow
trainingServicePlatform: frameworkcontroller
searchSpacePath: search_space.json
#choice: true, false
useAnnotation: false
tuner:
  #choice: TPE, Random, Anneal, Evolution, BatchTuner, MetisTuner, GPTuner
  builtinTunerName: TPE
  classArgs:
    #choice: maximize, minimize
    optimize_mode: maximize
trial:
  codeDir: .
  taskRoles:
    - name: worker
      taskNum: 1
      command: python3 model.py
      gpuNum: 0
      cpuNum: 1
      memoryMB: 8192
      image: frameworkcontroller/nni:v1.0
      frameworkAttemptCompletionPolicy:
        minFailedTaskCount: 1
        minSucceededTaskCount: 1
frameworkcontrollerConfig:
  storage: nfs
  serviceAccountName: frameworkcontroller
  nfs:
    # Your NFS server IP, like 10.10.10.10
    server: <my nfs server ip>
    # Your NFS server export path, like /var/nfs/nni
    path: /nfs/data

**Environment**:
- NNI version: 2.8
- Training service (local|remote|pai|aml|etc):  k8s frameworkcontroller

Issue #4874, opened 2022-05-19
Error about running with framecontroller mode


N-Kingsley:
**config.yml**

experimentName: example_mnist_pytorch
trialConcurrency: 1
maxExecDuration: 1h
maxTrialNum: 2
debug: false
nniManagerIp: 172.16.40.155
#choice: local, remote, pai, kubeflow
trainingServicePlatform: frameworkcontroller
searchSpacePath: search_space.json
#choice: true, false
useAnnotation: false
tuner:
  #choice: TPE, Random, Anneal, Evolution, BatchTuner, MetisTuner, GPTuner
  builtinTunerName: TPE
  classArgs:
    #choice: maximize, minimize
    optimize_mode: maximize
trial:
  codeDir: .
  taskRoles:
    - name: worker
      taskNum: 1
      command: python3 model.py
      gpuNum: 0
      cpuNum: 1
      memoryMB: 8192
      image: frameworkcontroller/nni:v1.0
      securityContext:
        privileged: true
      frameworkAttemptCompletionPolicy:
        minFailedTaskCount: 1
        minSucceededTaskCount: 1
frameworkcontrollerConfig:
  storage: nfs
  serviceAccountName: frameworkcontroller
  nfs:
    # Your NFS server IP, like 10.10.10.10
    server: <my nfs server ip>
    # Your NFS server export path, like /var/nfs/nni
    path: /nfs/nni
    readOnly: false

nfs server: /etc/exports
/nfs/nni *(rw,no_root_squash,sync,insecure,no_subtree_check)

**k8s logs**

kubectl describe pod nniexpywn5cftxenvma3hh-worker-0:
![image](https://user-images.githubusercontent.com/38311217/169243547-73b7641d-0265-41b0-b5fc-38af2df6e0a0.png)

How to solve this error?

N-Kingsley:
I find a question:
the dictionary nni/ywn5cftx does not exist in the shared dictionary<nfs server>:nfs/nni :
![image](https://user-images.githubusercontent.com/38311217/169490862-af816841-0219-4062-8a05-bee8f5dc9f1d.png)

N-Kingsley:
I find a question:
the dictionary nni/ywn5cftx does not exist in the shared dictionary <nfs server>:nfs/nni :
image

N-Kingsley:
I find a question:
the dictionary nni/ywn5cftx does not exist in the shared dictionary :nfs/nni :
[...]

Issue #4588, opened 2022-02-25
Having trouble nni with frameworkcontroller on k8s


juniroc:
**Describe the issue**:
When I tried nni with frameworkcontroller on k8s, I used these yaml files

- I tried nfs 

for nni config
`config_framework.yml`
```
authorName: default
experimentName: example_mnist_pytorch
trialConcurrency: 1
maxExecDuration: 1h
maxTrialNum: 10
#choice: local, remote, pai, kubeflow
trainingServicePlatform: frameworkcontroller
searchSpacePath: search_space.json
#choice: true, false
useAnnotation: false
tuner:
  #choice: TPE, Random, Anneal, Evolution, BatchTuner, MetisTuner, GPTuner
  builtinTunerName: TPE
  classArgs:
    #choice: maximize, minimize
    optimize_mode: maximize
assessor:
  builtinAssessorName: Medianstop
  classArgs:
    optimize_mode: maximize
trial:
  codeDir: .
  taskRoles:
    - name: worker
      taskNum: 1
      command: python3 mnist.py
      gpuNum: 0
      cpuNum: 1
      memoryMB: 8192
      image: msranni/nni:latest
      frameworkAttemptCompletionPolicy:
        minFailedTaskCount: 3
        minSucceededTaskCount: 1
frameworkcontrollerConfig:
  storage: nfs
  nfs:
    # Your NFS server IP, like 10.10.10.10
    server: 192.168.1.106
    # Your NFS server export path, like /var/nfs/nni
    path: /home/mj_lee/mount
  serviceAccountName: frameworkcontroller
```

and for frameworkcontroller Statefulset
`frameworkcontroller-with-default-config.yaml`
```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: frameworkcontroller
  namespace: default
spec:
  serviceName: frameworkcontroller
  selector:
    matchLabels:
      app: frameworkcontroller
  replicas: 1
  template:
    metadata:
      labels:
        app: frameworkcontroller
    spec:
      # Using the ServiceAccount with granted permission
      # if the k8s cluster enforces authorization.
      serviceAccountName: frameworkcontroller
      containers:
      - name: frameworkcontroller
        image: frameworkcontroller/frameworkcontroller
        # Using k8s inClusterConfig, so usually, no need to specify
        # KUBE_APISERVER_ADDRESS or KUBECONFIG
        env:
        #- name: KUBE_APISERVER_ADDRESS
[...]

Issue #4954, opened 2022-06-22
NNI v2.9 Iteration Plan


ultmaster:
- Release manager: @ultmaster
- Release start date: 6.22
- Feature freeze date (at most 30 days): ~7.22~ 8.1
- First test package (1 week since feature freeze): ~7.29~ ~8.12~ 8.16
- Release date (3 weeks since feature freeze): ~8.12~ 9.7

## Top selling points

- Official announcement of NAS search space hub
- Support training / search of AutoFormer (BigNAS-styled one-shot NAS)
- Support pruning of transformer: User experience and performance preliminary optimization (masks generation/movement pruning)

### Other nominations

- Pythonic HPO search space factory
- Sklearn and Framework-independent tutorial
- Compression supports Lightning: no need to rewrite code when users use lightning in the first place

### NAS

- [ ] P0 - Official announcement of search space hub @ultmaster
  - [x] Setup cluster
  - [x] A table summarizing the performance of popular one-shot strategies ~7.8~ ~7.29~ 8.1 #5034
  - [x] Documentation 8.1 #5035
  - [x] #5051
  - [ ] #5049 
  - [x] #5050 
  - [ ] #5052 
  - [ ] #5053
- [ ] P0 - FLOPs estimator for NAS @ultmaster
  - [x] design review ~7.11~ 7.18
- [x] P0 - Support training (PAO TONG) of AutoFormer (BigNAS-styled one-shot NAS) Renjie ~#4965~ #4987
- [ ] P0 - Support evolution search of AutoFormer ~7.27~ ~8.1~ 8.5 #5054
- [ ] P0 - Promote nni.retiarii to nni.nas @ultmaster ~7.1~ 7.8
  - [x] #4976 
  - [x] Move retiarii to NAS #5014
  - [ ] NasExperiment interface (deferred)
- [ ] P1 - Refactor of nni.nas.nn @ultmaster
  - [ ] design review (3rd round) 
- [ ] P1 - Tutorial for one-shot NAS. (depends on 4760) #4509 @ultmaster  @JiahangXu ~4.27~ ~5.6~ ??
- [ ] P2 - Strategy refactor: strategies respect resource constraints
- [ ] P0 - Refactor experiment @QuanluZhang 7.8
  - [ ] Support view of NAS experiment ~7.1~ #4985
  - [ ] Support resume
  - [ ] Support new interface of experiment

### Compression @J-shang

- [x] P0 - Support Block Sparse #4932
[...]

PR #5045, opened 2022-08-03
Fix reusable k8s training service bug

amznero:
### Description ###

For now, the reusable k8s training service(framework controller and kubeflow) will save the entry point file(run.sh) and env dir at the root path of NFS, but the content of run.sh declare it should be located at "envs/exp_id". So, the "cannot find xxxx" error will be raised when the worker pod runs this entry point. 

Second, when users submit multi-experiments, different experiments will affect each other(due to all contents are stored at the root of NFS).

And the entry-point will be overwritten by different environments when the trialConcurrency is more than 1, so each environment will actually run the same(the latest generated) run.sh script, and https://github.com/microsoft/nni/blob/v2.8/nni/tools/trial_tool/trial_runner.py#L164 will use dir name as runner id, finally, it will raise an error.

More discussion can be found at https://github.com/microsoft/nni/issues/4588

---

- [x] fix upload dir bug,  use a separate working directory for each experiment
- [x] fix trialConcurrency bug, use ${envId}_run.sh to replace run.sh

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### How to test ###

Related Issues: https://github.com/microsoft/frameworkcontroller/issues/75, https://github.com/microsoft/nni/issues/4588, https://github.com/microsoft/nni/issues/4874, https://github.com/microsoft/nni/issues/5026.


amznero:
The commit(0c9c151c325caf9f4b5bd809b7f76e59df7dd431) has a side effect when the storage is not nfs(like azure). To be refined.

liuzhe-lz (approved):

ultmaster (approved):

ultmaster:
Need to be tested on K8S pipline. Tracked in #4954.

PR #5063, opened 2022-08-12
[TSv3] Step 0 - Compat layer for using v3 training service as v1

liuzhe-lz:
### Description ###

This is the preparation of training service v3.

It contains the new interface and one-side compatibility layer.

The compat layer allows to use a v3 training service as a v1 training service, so we can test new training services without modifying NNI manager.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/ts/nni_manager/common/training_service_v3.ts
+++ b/ts/nni_manager/common/training_service_v3.ts
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ *  TODO
+ *  This type is planned to be an interface.
+ *  To minimize the change, it is currently identical to TrialJobMetric.data
+ **/
+export type Metric = string;
+
+/**
+ *  TODO
+ *  This type is planned to be an interface.
+ *  To minimize the change, it is currently identical to HyperParameters.value
+ **/
+export type Parameter = string;
+
+export interface EnvironmentInfo {
+    id: string;
+}
+
+export interface TrainingServiceV3 {
+    init(): Promise<void>;
+    start(): Promise<EnvironmentInfo[]>;
+    stop(): Promise<void>;
+
+    uploadDirectory(directoryName: string, path: string): Promise<void>;
+    createTrial(environmentId: string, trialCommand: string, directoryName: string): Promise<string | null>;
+    stopTrial(trialId: string): Promise<void>;
+    sendParameter(trialId: string, parameter: Parameter): Promise<void>;
+
+    onRequestParameter(callback: (trialId: string) => Promise<void>): void;
+    onMetric(callback: (trialId: string, metric: Metric) => Promise<void>): void;
+    onTrialEnd(callback: (trialId: string, timestamp: number, exitCode: number | null) => Promise<void>): void;
+    onTrialStart(callback: (trialId: string, timestamp: number) => Promise<void>): void;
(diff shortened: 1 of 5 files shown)


ts/nni_manager/common/training_service_v3.ts
```diff
@@ -0,0 +1,38 @@
...
+    init(): Promise<void>;
+    start(): Promise<EnvironmentInfo[]>;
+    stop(): Promise<void>;
+
+    uploadDirectory(directoryName: string, path: string): Promise<void>;
+    createTrial(environmentId: string, trialCommand: string, directoryName: string): Promise<string | null>;
+    stopTrial(trialId: string): Promise<void>;
+    sendParameter(trialId: string, parameter: Parameter): Promise<void>;
+
+    onRequestParameter(callback: (trialId: string) => Promise<void>): void;
```
ultmaster, line 32: One trial can still have multiple parameter?

ts/nni_manager/common/training_service_v3.ts
```diff
@@ -0,0 +1,38 @@
...
+    id: string;
+}
+
+export interface TrainingServiceV3 {
+    init(): Promise<void>;
+    start(): Promise<EnvironmentInfo[]>;
+    stop(): Promise<void>;
+
+    uploadDirectory(directoryName: string, path: string): Promise<void>;
+    createTrial(environmentId: string, trialCommand: string, directoryName: string): Promise<string | null>;
```
ultmaster, line 28: Would `createTrial` send a parameter with it?

ts/nni_manager/common/training_service_v3.ts
liuzhe-lz: Yes.
Each `get_next_parameter()` results in one `onRequestParameter()`, and then one `sendParameter()`.

ts/nni_manager/common/training_service_v3.ts
liuzhe-lz: By design, no.
If the trial does not invoke `get_next_parameter()` or `get_current_parameter()`, no parameter will be generated.

In current implementation, yes.
The compat layer caches a parameter on creating trial, and send it to first `onRequestParameter()`.

ts/nni_manager/common/training_service_v3.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export interface TrainingServiceV3 {
+    /**
+     *  Invoked during experiment initialization.
+     *
+     *  It should verify the config and raise error if the specified training resource is not available.
+     *
+     *  It should not start daemon on worker machine.
+     *  If another component of the experiment failed to initialize,
+     *  the process will abort without invoking any clean up function.
```
QuanluZhang, line 31: this implies that there should be clean-up but you decide not to do the clean-up, why?

QuanluZhang (approved):

ultmaster (approved):

ts/nni_manager/common/training_service_v3.ts
liuzhe-lz: This method is invoked during initializing stage.
Failing to initialize means NNI manager is in an unexpected condition, so nothing is guaranteed to work, and halting is the safest action.

PR #5069, opened 2022-08-15
Add Group Norm support for Pruning model

cin-xing:
### Description ###

Add Group Norm support for model pruning.

I am not sure if this algorithm works for GroupNorm 
--> Need to do experiments to recheck the algorithm.

There are two main ways to prune GroupNorm (one is pruning group (reduce the number of groups, keep the number of channels per group the same, finally the module will become `LayerNorm`), the other is pruning the channel per group (finally the module will become `InstanceNorm`).

This implementation is the second approach as I saw that for `Conv2d`, we keep the number of groups the same.

But personally, I prefer the first approach as LayerNorm is more widely accepted and outperforms `InstanceNorm` according to `GroupNorm` original paper.

p.s: Lacking the infrastructure to do these test :(

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### How to test ###
```python
from nni.compression.pytorch.pruning import FPGMPruner
from nni.compression.pytorch import ModelSpeedup
import torch
from torch import nn

class GroupNormModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.conv1 = nn.Conv2d(3, 128, 3)
        self.norm = nn.GroupNorm(32, 128)
        # self.norm = nn.BatchNorm2d(128)
        self.conv2 = nn.Conv2d(128, 128, 3, groups=32)

    def forward(self, x):
        x = self.conv1(x)
        x = self.norm(x)
        x = self.conv2(x)
        return x

config_list = [
    {
        'sparsity': 0.3,
        'op_types': ['Conv2d'],
    }
]

input_shape = (1, 3, 224, 224)
speed_up_config = dict(
    mode="dependency_aware",
    dummy_input=torch.rand(input_shape, device="cuda")
)
model = GroupNormModel().cuda().eval()
print("Before speed up")
print(model(speed_up_config["dummy_input"]).shape)
print(model)
print("---")

pruner = FPGMPruner(
    model,
    config_list,
    **speed_up_config,
)

masked_model, masks = pruner.compress()
pruner.show_pruned_weights()
pruner._unwrap_model()
ModelSpeedup(model, dummy_input=torch.rand(input_shape, device="cuda"), masks_file=masks).speedup_model()

print("After speed up")
print(model)
print(model(speed_up_config["dummy_input"]).shape)
```

### DIFF
--- a/nni/compression/pytorch/speedup/compress_modules.py
+++ b/nni/compression/pytorch/speedup/compress_modules.py
@@ -143,8 +143,8 @@ def replace_prelu(prelu, masks):
     weight_mask = weight_mask['weight']
     if weight_mask.size(0) == 1:
         return prelu
-    _, remained_in = convert_to_coarse_mask(in_mask, 1)
-    _, remained_out = convert_to_coarse_mask(output_mask, 1)
+    pruned_in, remained_in = convert_to_coarse_mask(in_mask, 1)
+    pruned_out, remained_out = convert_to_coarse_mask(output_mask, 1)
     n_remained_in = weight_mask.size(0) - pruned_in.size(0)
     n_remained_out = weight_mask.size(0) - pruned_out.size(0)
     remained_in, remained_out = remained_in.to(
@@ -332,8 +332,8 @@ def replace_groupnorm(norm: nn.GroupNorm, masks):
     in_mask = in_masks[0]
 
     # N, C, H, W
-    pruned_in, remained_in = convert_to_coarse_mask(in_mask, 1)
-    pruned_out, remained_out = convert_to_coarse_mask(output_mask, 1)
+    _, remained_in = convert_to_coarse_mask(in_mask, 1)
+    _, remained_out = convert_to_coarse_mask(output_mask, 1)
 
     assert len(remained_in.size()) == 1
     if remained_in.size(0) != remained_out.size(0):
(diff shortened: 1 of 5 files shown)


J-shang:
@cin-xing Thanks for your contribution!

"There are two main ways to prune GroupNorm (one is pruning group (reduce the number of groups, keep the number of channels per group the same, finally the module will become LayerNorm), the other is pruning the channel per group (finally the module will become InstanceNorm)."

Totally agree, these two ways are all important, can you consider implement them both?
FYI, the following is group conv replacement logic.
https://github.com/microsoft/nni/blob/cbc6273ae8de00a2c3e900987c7a42978d33ed11/nni/compression/pytorch/speedup/compress_modules.py#L434

cin-xing:
Hi @J-shang 
I have implemented both types of pruning GroupNorm: Constant number of groups and constant number of channels per group.
The implementation is still very rough, so I would love to have your opinions.

---
For more information, I found that the two achieve similar accuracies for my internal model.
My setting is to use FGBM Pruning with sparsity 20% and 30%. Then fine-tune the model after pruning (already replaced layers) for 3 epochs. 
I ran the experiment twice with the two stated settings above. The InstanceNorm only gets slightly higher accuracy (around 0.01%).

J-shang:
> Hi @J-shang I have implemented both types of pruning GroupNorm: Constant number of groups and constant number of channels per group. The implementation is still very rough, so I would love to have your opinions.
> 
> For more information, I found that the two achieve similar accuracies for my internal model. My setting is to use FGBM Pruning with sparsity 20% and 30%. Then fine-tune the model after pruning (already replaced layers) for 3 epochs. I ran the experiment twice with the two stated settings above. The InstanceNorm only gets slightly higher accuracy (around 0.01%).

I think we should consider merging the two-kind replacement, here is an example:

GroupNorm(num_groups=4, num_channels=12) with masks [0,0,1,1, | 1, 1, 0, 0, | 0, 0, 0, 0]
==>
GroupNorm(num_groups=2 num_channels=4)

cin-xing:
image
Just want to add this image here. A small problem with the original computation of Conv2d with Group.

It is much better to use PyTorch support directly.

Full benchmark code:
```python
# code block1
import torch
tensor = torch.arange(512)

# code block2
remained_in = tensor
in_start = 0
in_end = 32

#code block3
%timeit torch.logical_and(in_start <= remained_in, remained_in < in_end).sum().item()

# code block 4
%timeit list(filter(lambda x: in_start <= x and x < in_end, remained_in.tolist()))
```

nni/compression/pytorch/speedup/compress_modules.py
```diff
@@ -311,6 +312,90 @@ def replace_batchnorm2d(norm, masks):
...
+    for groupid in range(norm.num_groups):
+        in_start = groupid * ori_channel_step
+        in_end = in_start + ori_channel_step
+        masks = torch.logical_and(in_start <= remained_in, remained_in < in_end)
+        current_input_index = remained_in[masks]
+        if len(current_input_index) == 0:
+            continue
+
+        # remap the global index to the group index
+        current_input_index = current_input_index - in_start
```
J-shang, line 371: seems don't need this line

J-shang:
LGTM, could merge this pr after fix the comments

J-shang (approved):

cin-xing:
@J-shang could you help me merge this?
I don't have permission to merge this.

QuanluZhang (approved):

PR #5073, opened 2022-08-18
Use Random to generate experiment id

ultmaster:
### Description ###

Generate experiment id with ~shortuuid~ Random. This fixes the issue that when `random.seed` is called, experiment id is always the same.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/nni/experiment/management.py
+++ b/nni/experiment/management.py
@@ -7,7 +7,14 @@ import string
 
 
 def generate_experiment_id() -> str:
-    return ''.join(random.sample(string.ascii_lowercase + string.digits, 8))
+    try:
+        # This try..catch is for backward-compatibility,
+        # in case shortuuid is not installed for some reason.
+        import shortuuid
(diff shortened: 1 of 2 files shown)


J-shang (approved):

nni/experiment/management.py
```diff
@@ -2,12 +2,22 @@
...
 from pathlib import Path
+import logging
 import random
 import string
 
+_logger = logging.getLogger(__name__)
+
 
 def generate_experiment_id() -> str:
-    return ''.join(random.sample(string.ascii_lowercase + string.digits, 8))
```
liuzhe-lz, line 10: I don't think it worth introducing a dependency.
Replacing `random.sample()` with `Random().sample()` should be enough.

liuzhe-lz (approved):

PR #5092, opened 2022-08-29
Dedup for evolution search

ultmaster:
### Description ###

Add deduplication for evolutionary search.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


cruiseliu:
What will happen if the search space is exhausted?
HPO has a dedup util: https://github.com/microsoft/nni/blob/master/nni/common/hpo_utils/dedup.py

ultmaster:
> What will happen if the search space is exhausted? HPO has a dedup util: https://github.com/microsoft/nni/blob/master/nni/common/hpo_utils/dedup.py

It will give up the dedup, and fallback to the original "un-dedup" process.

I read the HPO dedup util. It's hard to reuse it because the format is not unified. Also, the generated config uses mutator as keys, and mutators are not JSON-serializable. (I'm working on both problems in v3.0 iteration.)

liuzhe-lz (approved):

nni/nas/strategy/evolution.py
```diff
@@ -43,6 +46,10 @@ class RegularizedEvolution(BaseStrategy):
         The number of individuals that should participate in each tournament. Default: 25.
     mutation_prob : float
         Probability that mutation happens in each dim. Default: 0.05
+    dedup : bool
+        Do not try the same configuration twice. When variational is true, deduplication is not supported. Default: true.
```
QuanluZhang, line 50: what is "variational"?

ultmaster: Mis-copied. Removed.

QuanluZhang (approved):

Issue #5097, opened 2022-08-30
torch.full can't use implicit scalars


SolomidHero:
**Describe the issue**:
For some functions like `torch.full` it is needed to have input from other tensors. For example I use it to create lengths array, when it is not provided

I need traceable usage:
```py3
torch.full(x.shape, x.shape[0], device=x.device, dtype=torch.int32) # desirable, not working
torch.full(x.shape, int(x.shape[0]), device=x.device, dtype=torch.int32) # not desirable, but working
```

**Environment**:
- NNI version: 2.9a2
- Python version: 3.9
- PyTorch version: checked 1.9, 1.12

**How to reproduce it?**:

```py3
import torch
import torch.nn as nn
from nni.compression.pytorch.pruning import L1NormPruner
from nni.compression.pytorch.speedup import ModelSpeedup


class M(nn.Module):
    def __init__(self, _in, _out):
        super().__init__()
        self.model = nn.Linear(_in, _out)

    def forward(self, x: torch.Tensor):
        add = torch.full(x.shape, x.shape[0], device=x.device, dtype=torch.int32)
        x = x + add
        x = self.model(x)
        return x

m = M(5, 7)
m_input = torch.randn(20, 5)

config_list = [{
    'sparsity_per_layer': 0.5,
    'op_types': ['Linear']
}]
pruner = L1NormPruner(m, config_list)
masked_m, masks = pruner.compress()
pruner._unwrap_model()

ModelSpeedup(m, m_input, masks, 20).speedup_model()
```

Error:
```pytb
[2022-08-30 09:58:02] start to speedup the model
no multi-dimension masks found.
[2022-08-30 09:58:02] infer module masks...
[2022-08-30 09:58:02] Update mask for .aten::size.1
[2022-08-30 09:58:02] Update mask for .aten::size.3
[2022-08-30 09:58:02] Update mask for .aten::size.5
[2022-08-30 09:58:02] Update mask for .aten::Int.2
[2022-08-30 09:58:02] Update mask for .aten::Int.4
[2022-08-30 09:58:02] Update mask for .aten::ScalarImplicit.6
[2022-08-30 09:58:02] ERROR: aten::ScalarImplicit is not Supported! Please report an issue at https://github.com/microsoft/nni. Thanks~
[2022-08-30 09:58:02] Update mask for .aten::full.7
[2022-08-30 09:58:02] WARNING: throw some args away when calling the function "full"
---------------------------------------------------------------------------
[...]

PR #5107, opened 2022-09-02
fix(speedup): refactor the execution logic of functions in speedup

Louis-J:
### Description ###

In speedup, we use a non-general trick to treat the prim ops. We analyze the relationship between the prim ops, then delete all the prim ops, finally connect the non-prim op directly. We also assume that all the intermediate variables are tensors, tuples of tensors, or lists of tensors. But it'll cause many problems in various models.

So, it's better to execute all the prim ops for real. In this pr, I remove the deletion of prim ops, and execute the prim recursively. And the assumption of variable types is also removed. Now it's closer to the original execution.

But it changed a lot and is untested now. More work is needed.

#5097

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


docs/source/compression/pruning.rst
```diff
@@ -105,4 +105,5 @@ In addition, for the convolutional layers that have more than one filter group,
 ``dependency-aware pruner`` will also try to prune the same number of the channels for each filter group.
 Overall, this pruner will prune the model according to the L1 norm of each filter and try to meet the topological constrains (channel dependency, etc) to improve the final speed gain after the speedup process. 
 
+Operations that will be recognized as having channel dependencies: add/sub/mul/div, addcmul/addcdiv, logical_and/or/xor
```
J-shang, line 108: `.`?

nni/common/graph_utils.py
```diff
@@ -725,8 +717,6 @@ def _build_graph(self):
 
         # associate module name with their trace graph nodes
         for node in graph.nodes():
-            if node.kind() == CONSTANT_KIND:
-                continue
```
J-shang, line 729: when we should skip the CONSTANT_KIND and when we should not? I found you did not remove all the `if node.kind() == CONSTANT_KIND:` in the code

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -175,12 +176,15 @@ def _prepare_dummy_input(self, node):
         # prepare the inputs and outputs mask for this node,
         # if there is already a mask in self.masks, then use
         # the original mask tensor, else create a new one.
-        inputs_name = node.inputs
+        if node.type == 'module':
+            inputs_name = node.inputs
+        else:
+            inputs_name = [val_node.debugName() for val_node in node.key_node.inputs()]
```
J-shang, line 182: what is key_node? maybe we need a doc to explain the important attrs in a node

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -297,16 +318,15 @@ def update_indirect_sparsity(self, node):
                 debug_name = auto_infer.input_debugname[in_id]
 
                 last_output = self.internal_result[debug_name]
-                # if isinstance(last_output, torch.Tensor):
-                # TODO what if last output is tuple/list of tensor
```
J-shang, line 301: so `what if last output is tuple/list of tensor`?

Louis-J:
totally refactored old speedup in https://github.com/Louis-J/nni/tree/refactor_speedup

Louis-J:
the speedup v2 now is runnable. so better to reassess that refactor the old speedup is suitable or not.

Lijiaoa:
https://github.com/microsoft/nni/issues/5143

PR #5108, opened 2022-09-05
fix(bugbash): bugs in compression and speedup

Louis-J:
### Description ###

1. fix the bug in compression that dependency_aware do not support functions like torch.div, torch.sub
2. fix the bug in speedup of support of aten::ones_like, zeros_like by adding a convertion from 'int layout' in TorchScript to 'torch.layout' in python.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [x] full test - compression

### Checklist ###
  - [x] test case
  - [x] doc


docs/source/compression/pruning.rst
```diff
@@ -105,4 +105,5 @@ In addition, for the convolutional layers that have more than one filter group,
 ``dependency-aware pruner`` will also try to prune the same number of the channels for each filter group.
 Overall, this pruner will prune the model according to the L1 norm of each filter and try to meet the topological constrains (channel dependency, etc) to improve the final speed gain after the speedup process. 
 
+Operations that will be recognized as having channel dependencies: add/sub/mul/div, addcmul/addcdiv, logical_and/or/xor
```
J-shang, line 108: .. Note:: Operations that will be recognized as having channel dependencies: add/sub/mul/div, addcmul/addcdiv, logical_and/or/xor.

J-shang (approved):

zheng-ningxin (approved):
LGTM

PR #5109, opened 2022-09-05
[nas] fix issue introduced by the trial recovery feature

QuanluZhang:
fix bugs in #4931


liuzhe-lz (approved):

nni/nas/execution/common/integration.py
```diff
@@ -239,4 +256,6 @@ def handle_import_data(self, data):
         pass
 
     def handle_add_customized_trial(self, data):
-        pass
+        previous_max_param_id = self.recover_parameter_id(data)
+        self.parameters_count = previous_max_param_id
+        self._advisor_initialized = True
```
ultmaster, line 261: Is it possible that `handld_add_customized_trial` is never called and `_advisor_initialized` is never set to true?

nni/nas/execution/common/integration_api.py
```diff
@@ -41,7 +45,11 @@ def send_trial(parameters: dict, placement_constraint=None) -> int:
     Send a new trial. Executed on tuner end.
     Return a ID that is the unique identifier for this trial.
     """
-    return get_advisor().send_trial(parameters, placement_constraint)
+    advisor = get_advisor()
+    while not advisor.initialized:
```
ultmaster, line 49: Suggest putting this into `RetiariiAdvisor.send_trial`.

nni/nas/execution/common/integration.py
QuanluZhang: good point! `handle_add_customized_trial` is called when experiment is resumed, but will not be called when experiment is created, this is a bug introduced by me...
I moved this flag to `handle_request_trial_jobs`, which means if trial is not requested, `send_trial` will be blocked. And "request trial" will always be sent by nnimanager

nni/nas/execution/common/integration_api.py
QuanluZhang: updated

ultmaster (approved):

Issue #5131, opened 2022-09-16
AttributeError in "Speedup Model with Mask" Tutorial Codes


hshwang88:
**Describe the issue**:
https://nni.readthedocs.io/en/stable/tutorials/pruning_speedup.html

```python
ModelSpeedup(model, torch.rand(10, 1, 28, 28).to(device), masks).speedup_model()
```

```
Traceback (most recent call last):
  File "e:\Projects\NNI\src\CompressionTuto.py", line 18, in <module>
    ModelSpeedup(model, torch.rand(10, 1, 28, 28).to(device), masks).speedup_model()
  File "d:\pt-env\lib\site-packages\nni\compression\pytorch\speedup\compressor.py", line 543, in speedup_model
    self.infer_modules_masks()
  File "d:\pt-env\lib\site-packages\nni\compression\pytorch\speedup\compressor.py", line 380, in infer_modules_masks
    self.update_direct_sparsity(curnode)
  File "d:\pt-env\lib\site-packages\nni\compression\pytorch\speedup\compressor.py", line 228, in update_direct_sparsity
    func = jit_to_python_function(node, self)
  File "d:\pt-env\lib\site-packages\nni\compression\pytorch\speedup\jit_translate.py", line 555, in jit_to_python_function
    return trans_func_dict[node.op_type](node, speedup)
  File "d:\pt-env\lib\site-packages\nni\compression\pytorch\speedup\jit_translate.py", line 490, in generate_aten_to_python
    positional_num, keyword_list, special_treat = parse_aten_schema(schema)
  File "d:\pt-env\lib\site-packages\nni\compression\pytorch\speedup\jit_translate.py", line 411, in parse_aten_schema
    if not arg.kwarg_only:
AttributeError: 'torch._C.Argument' object has no attribute 'kwarg_only'
```

**Environment**:
- NNI version: 2.9
- Training service: local
- Client OS: Win10
- Python version: 3.9.8
- PyTorch version: 1.8.1+cu111
- Is conda/virtualenv/venv used?: venv
- Is running in Docker?: No

**How to reproduce it?**:
```python
import torch
import time
from nni.compression.pytorch import ModelSpeedup
from scripts.compression_mnist_model import TorchModel, device

model = TorchModel().to(device)

conv1_mask = torch.ones_like(model.conv1.weight.data)
conv1_mask[0:3] = 0

masks = {'conv1': {'weight': conv1_mask}}

start = time.time()
[...]

PR #5138, opened 2022-09-23
fix(speedup): re-write aten schema parser to support pytorch versions < 1.9.0

Louis-J:
### Description ###

reason:
the official aten schema parser of python lost the 'kwarg_only' info in python versions <1.9.0.
and schema is not simple enough to use some regex to parser.

solution:
so I translated the schema parser code from pytorch 1.12 c++ code to python and adapted some deleted syntaxes in pytorch 1.8.
tested in pytorch versions 1.8, 1.9, 1.10, 1.11 and 1.12.

note:
`torch._C.Node` supports `schema` function since version 1.8. so speedup cannot use on pytorch 1.7 now.
#5131 can be solved.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [x] full test - compression

### Checklist ###
  - [x] test case
  - [ ] doc

### DIFF
--- a/nni/compression/pytorch/speedup/jit_translate.py
+++ b/nni/compression/pytorch/speedup/jit_translate.py
@@ -492,7 +492,7 @@ def parse_aten_schema(schema: str):
                 index += 1
                 return tk
         return default_value
- 
+
     def next_expect_pass_value(tk: str) -> str:
         nonlocal index
         if tk in spec_tokens:
@@ -551,7 +551,7 @@ def parse_aten_schema(schema: str):
                     return '*'
                 else:
                     return next_expect_pass_value('string')
- 
+
             value = '('.join(parse_list('|', None, parse_inner))
             value += next_if_pass_value('!', '')
             if next_if('-') and next_if('>', 1):
--- a/nni/compression/pytorch/speedup/jit_translate.py
+++ b/nni/compression/pytorch/speedup/jit_translate.py
@@ -399,7 +399,36 @@ schema_fix_dict = {
 def parse_aten_schema(schema: str):
     """
     Parse the schema, to positional_num and keyword_list, and detect if the argument should be specially treated.
-    Cannot use 'torch._C.parse_schema' because 'torch._C.Argument' has no 'kwarg_only' before pytorch v1.9.0
+    only available on pytorch >= v1.9.0
+    """
+    if schema in schema_fix_dict:
+        schema = schema_fix_dict[schema]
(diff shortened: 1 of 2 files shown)


liuzhe-lz:
I suggest to keep both version and use a flag to switch, if it's not too hard.
In future it's very likely that nobody can maintain this code. At that time dropping certain version's support will be a more practical choice.

liuzhe-lz (approved):

nni/compression/pytorch/speedup/jit_translate.py
```diff
@@ -486,7 +748,10 @@ def generate_aten_to_python(func: Callable, node: NodePyGroup, speedup: ModelSpe
     c_node = node.key_node
 
     schema = c_node.schema()
-    positional_num, keyword_list, special_treat = parse_aten_schema(schema)
+    if torch.__version__ < '1.9.0':
```
J-shang, line 751: image

I'm worried with this method of judging version.

nni/compression/pytorch/speedup/jit_translate.py
```diff
@@ -422,6 +424,266 @@ def parse_aten_schema(schema: str):
 
     return positional_num, keyword_list, special_treat
 
+@lru_cache
+def parse_aten_schema_version_1_8_x(schema: str):
+    """
+    Parse the schema, to positional_num and keyword_list, and detect if the argument should be specially treated.
+    Cannot use 'torch._C.parse_schema' because 'torch._C.Argument' has no 'kwarg_only' on pytorch v1.8.x
+    Using a lexer-parser like method to parse it.
+    Re-write from torch/csrc/jit/frontend/function_schema_parser.cpp
```
J-shang, line 433: if this function a mirror rewrite from this `cpp` file?

Louis-J: It's not totally equal to the original function in cpp. Actually, the original parser in that cpp file is too specific and too long (5k+ lines). I wrote a much simpler syntax analyzer, deleted some error check and specific type check to shorten the code.

nni/compression/pytorch/speedup/jit_translate.py
Louis-J: the type of `torch.__version__` is not a `str`. it's a version type that can correctly compare with str.

J-shang (approved):

PR #5141, opened 2022-09-26
fix(speedup): make the tensor contiguous before randomizing

Louis-J:
### Description ###

when randomizing, if the tensor is not contiguous, we cannot use torch.nn.init.uniform_ to randomize it.
so we should detect and make it contiguous.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/nni/compression/pytorch/speedup/infer_mask.py
+++ b/nni/compression/pytorch/speedup/infer_mask.py
@@ -135,6 +135,8 @@ class AutoMaskInference:
                     # this tensor, because our tensor scrambling is on the batch
                     # dimention. For example, if the tensor is a scalar(returned
                     # by the size operator), then we will skip this tensor
+                    if not tensor.is_contiguous():
+                        tensor = tensor.contiguous()
                     randomize_tensor(tensor, start, end)
             for para in self.weights:
                 randomize_tensor(self.weights[para].data, start, end)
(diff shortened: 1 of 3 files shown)


nni/compression/pytorch/speedup/infer_mask.py
```diff
@@ -127,14 +127,17 @@ def random_init(self, start=0.1, end=8.0):
...
-            for tensor in self.dummy_input:
+            for index, tensor in enumerate(self.dummy_input):
                 if isinstance(tensor, torch.Tensor) and len(tensor.size()) > self.batch_dim\
                     and tensor.size(self.batch_dim) == self.batch_size:
                     # if the input tensor only has one dimension, which means
                     # it doesn't have the batch dimension, then we don't randomize
                     # this tensor, because our tensor scrambling is on the batch
                     # dimention. For example, if the tensor is a scalar(returned
                     # by the size operator), then we will skip this tensor
+                    if not tensor.is_contiguous():
```
J-shang, line 138: can `tensor` not a `torch.Tensor`?

Louis-J: the tensor must be a torch.Tensor in this line

J-shang (approved):

QuanluZhang (approved):

Issue #3549, opened 2021-04-20
Broken link on WebUI


JunweiSUN:
![image](https://user-images.githubusercontent.com/30487595/115360329-1c118580-a1f2-11eb-93ab-37ead0feefd4.png)
Link here should be `https://github.com/Microsoft/nni/tree/v2.1` instead of `https://github.com/Microsoft/nni/tree/2.1.0`

Lijiaoa:
Yes. But I think this version should be `v2.1` rather than `v2.1.0`.

JunweiSUN:
> Yes. But I think this version should be `v2.1` rather than `v2.1.0`.

I think maybe the `version` in `package.json`  was set to `2.1.0` when releasing nni v2.1 which caused this issue. Maybe we should make the version of nni more consistent instead using `2.1.0`, `2.1`, `v2.1` together.

Lijiaoa:
hi @JunweiSUN Thanks for raising this bug. It had been fix in PR #5154.

PR #5154, opened 2022-10-09
fix issue 3549: Broken link on webui

Lijiaoa:
### Description ###
fix issue #3549

### How to test ###
test it with package


ts/webui/src/components/nav/Nav.tsx
```diff
@@ -63,7 +63,12 @@ class NavCon extends React.Component<NavProps, NavState> {
             method: 'GET'
         }).then(res => {
             if (res.status === 200) {
-                this.setState({ version: res.data });
+                let formatVersion = res.data;
+                // 2.0 will get 2.0.0 by node, so delete .0 to get real version
+                if (formatVersion.endsWith('.0')) {
```
ultmaster, line 68: What if `2.0.11`?

Count `.` in version and split on last `.`. Or use regex.

Lijiaoa: why? `2.0.11` is not end with '.0'. endsWith is enough to judge it, i think.

ultmaster (approved):

PR #5166, opened 2022-10-17
Let tuner auto reconnect to NNI manager

liuzhe-lz:
### Description ###

Dispatcher now tries to reconnect when it loses connection to NNI manager.
For one command it only retries once. If it fails twice for the same message then dispatcher crashes.
Because the reconnect is raised by client (dispatcher) side, dispatcher to manager messages will be resent on connection loss, but manager to dispatcher messages during the reconnect (~5s) will loss. I think it's okay because it only hurts tuner's accuracy.

Because connection error is now "expected", we add a new tuner command `ER` (error) to report real tuner error.
When the outermost function of dispatcher catches an exception, it tries to send the content with error command.
In NNI manager side, the shim layer of IPC channel is modified to report `ER` content to NNI manager and hide connection errors.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [x] test case
  - ~~doc~~

### How to test ###

I don't know...

### DIFF
--- a/nni/__main__.py
+++ b/nni/__main__.py
@@ -1,11 +1,12 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-import os
 import argparse
-import logging
-import json
 import base64
+import json
+import logging
+import os
+import traceback
 
 from .runtime.msg_dispatcher import MsgDispatcher
 from .runtime.msg_dispatcher_base import MsgDispatcherBase
@@ -65,6 +66,7 @@ def main():
         tuner._on_error()
         if assessor is not None:
             assessor._on_error()
+        dispatcher.report_error(traceback.format_exc())
         raise
 
 
--- a/nni/runtime/tuner_command_channel/channel.py
+++ b/nni/runtime/tuner_command_channel/channel.py
@@ -35,6 +35,7 @@ class TunerCommandChannel:
     """
 
     def __init__(self, url: str):
+        self._url = url
         self._channel = WebSocket(url)
 
     def connect(self) -> None:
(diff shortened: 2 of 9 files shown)


nni/__main__.py
```diff
@@ -65,6 +66,7 @@ def main():
         tuner._on_error()
         if assessor is not None:
             assessor._on_error()
+        dispatcher.report_error(traceback.format_exc())
```
ultmaster, line 69: If web socket already loses connection, will it end up here? What will `report_error` behave?

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -51,10 +52,19 @@ def disconnect(self) -> None:
...
+        except WebSocket.ConnectionClosed:
+            self._channel = WebSocket(self._url)
+            self._channel.send(command)
 
     def _receive(self) -> tuple[CommandType, str] | tuple[None, None]:
-        command = self._channel.receive()
+        try:
+            command = self._channel.receive()
+        except WebSocket.ConnectionClosed:
+            self._channel = WebSocket(self._url)
```
ultmaster, line 65: We'd better log this as a warning. Having more logs doesn't hurt.

Do you think retrying immediately and for only once is a good idea? What's the reason for this?

nni/runtime/tuner_command_channel/websocket.py
```diff
@@ -85,10 +88,10 @@ def receive(self) -> str | None:
             msg = _wait(self._ws.recv())
             _logger.debug(f'Received {msg}')
         except websockets.ConnectionClosed:  # type: ignore
-            _logger.debug('Connection closed by server.')
+            _logger.warning('Connection closed by server.')
```
ultmaster, line 91: Why is this only in receive? How about send?

QuanluZhang (approved):

nni/runtime/tuner_command_channel/channel.py
```diff
@@ -55,17 +61,45 @@ def _send(self, command_type: CommandType, data: str) -> None:
...
         if command is None:
-            raise RuntimeError('NNI manager closed connection')
+            command = self._retry_receive()
         command_type = CommandType(command[:2].encode())
         return command_type, command[2:]
+
+    def _retry_receive(self) -> str:
+        _logger.warning('Connection lost. Trying to reconnect...')
+        for interval in self._retry_intervals:
+            time.sleep(interval)
```
matluster, line 95: Suggest adding logs at every retrying.

ultmaster (approved):

PR #5189, opened 2022-10-27
Fix YAML encoding and WebSocket reconnect UT

liuzhe-lz:
### Description ###

Accidentally merged two PRs. Sorry for that.

Python part:
Fix the problem found by Lijiao that on Chinese system the YAML config files are loaded as GB encoding.

TS part:
The UT for WS reconnect is unstable. Fixed that.
Only one change outside UT: after reconnect the first ping is allowed to be skipped.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [x] test case
  - ~~doc~~

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -113,6 +113,7 @@ linkcheck_ignore = [
     r'https://www\.openml\.org/',                           # OpenML
     r'https://ml\.informatik\.uni-freiburg\.de/',
     r'https://docs\.nvidia\.com/deeplearning/',
+    r'https://cla\.opensource\.microsoft\.com',
 ]
 
 # Ignore all links located in release.rst
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -114,6 +114,7 @@ linkcheck_ignore = [
     r'https://ml\.informatik\.uni-freiburg\.de/',
     r'https://docs\.nvidia\.com/deeplearning/',
     r'https://cla\.opensource\.microsoft\.com',
+    r'https://www\.docker\.com/',
 ]
 
 # Ignore all links located in release.rst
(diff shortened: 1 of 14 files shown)


ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
```diff
@@ -142,6 +142,7 @@ class WebSocketChannelImpl implements WebSocketChannel {
         }
 
         this.serving = false;
+        this.waitingPong = false;
```
J-shang, line 145: what's this used for?

docs/source/conf.py
```diff
@@ -114,6 +114,7 @@
     r'https://ml\.informatik\.uni-freiburg\.de/',
     r'https://docs\.nvidia\.com/deeplearning/',
     r'https://cla\.opensource\.microsoft\.com',
+    r'https://www\.docker\.com/',
```
J-shang, line 117: maybe it's time to check and clean the permanently dead links in our docs😂

ts/nni_manager/core/tuner_command_channel/websocket_channel.ts
liuzhe-lz: There is a callback at 5 seconds intervals to handle ping.
It send a ping to client and set `waitingPong` to `true`.
In 5 seconds, if the client has send back a pong, the pong handler will set `waitingPong` to `false`.
5 seconds later, the next callback checks `waitingPong`'s value and if it's `true`, the client is considered lost responsive.

This line is executed at establishing of a new connection and effectively skips the first pong check.
This is because in UT the pong interval is reduced to a short 10ms; therefore the client might not be able to finish all on connect jobs and send first pong in time.
In real world this also somehow increases robustness. The drawback is if the Python client does fail on reconnection init, it will take an extra 5s to detect.

J-shang (approved):

ultmaster (approved):

J-shang:
should this merge to v2.10?

PR #5194, opened 2022-11-01
Mutable V3 (Stage 1) - Label namespace and utilities

ultmaster:
This is the first of 7 pull requests for the new mutable infrastructure for NAS and HPO.

This PR introduces a labeling utility, which automatically labels mutables in cases when labels are not manually provided.


nni/mutable/utils.py
```diff
@@ -0,0 +1,207 @@
...
+
+_last_uid = defaultdict(int)
+
+_logger = logging.getLogger(__name__)
+
+
+def uid(namespace: str = 'default') -> int:
+    """Global counter for unique id. Not thread-safe."""
+    _last_uid[namespace] += 1
+    return _last_uid[namespace]
```
QuanluZhang, line 24: does it support experiment resume?

matluster: Experiment resume is not related to this part.

nni/mutable/utils.py
```diff
@@ -0,0 +1,207 @@
...
+
+    :class:`LabelNamespace` is implemented based on :class:`ContextStack`.
+
+    Examples
+    --------
+    >>> with LabelNamespace('param'):
+    ...     label1 = auto_label()       # param_1
+    ...     label2 = auto_label()       # param_2
+    ...     with LabelNamespace('param'):
+    ...         label3 = auto_label()   # param_1_1
```
QuanluZhang, line 118: why is it `param_1_1` not `param_2_1`?

nni/mutable/utils.py
```diff
@@ -0,0 +1,207 @@
...
+    ...         label4 = auto_label()   # param_1_2
+    ...     with LabelNamespace('another'):
+    ...         label5 = auto_label()   # another_1
```
QuanluZhang, line 121: what if I wrote the code below?
```
...           with LabelNamespace('param'):
                   label7 = auto_label() # it is param_2_1? or param_1_3 or param_1_1?
```

nni/mutable/utils.py
ultmaster: I think the docstring is out-of-date. Just updated the docs and corresponding tests.

QuanluZhang (approved):

nni/mutable/utils.py
```diff
@@ -0,0 +1,226 @@
...
+    """Reset counter for a specific namespace."""
+    _last_uid[namespace] = 0
+
+
+class NoContextError(Exception):
+    """Exception raised when context is missing."""
+    pass
+
+
+class ContextStack:
```
J-shang, line 37: who will use this class? is it possible someone using it like:

```
with ContextStack(namespace, value):
    ContextStack.push(namespace, value2)
    ...
...
ContextStack.pop(namespace)
```

It is ok if only `LabelNamespace` using it.

nni/mutable/utils.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Examples
+    --------
+    >>> label1 = auto_label()               # global_1
+    >>> with LabelNamespace('param'):
+    ...     label2 = auto_label('param')    # param_1, because in the namespace "param"
+    >>> with LabelNamespace():
+    ...     label3 = auto_label()           # param_1, default key is used
+    >>> with LabelNamespace('another'):
+    ...     label4 = auto_label('another')  # another_1
```
J-shang, line 207: I don't understand why under `'another'` context, I should pass `'another'` to `auto_label`

ultmaster: Suppose you have:

```python
with LabelNamespace('abc'):
  with LabelNamespace('def'):
     auto_label()  # abc_1 or def_1?
```

Maybe I should auto-fetch the nearest one?

nni/mutable/utils.py
ultmaster: Haven't considered it carefully.

Right now it's mainly for internal usages. It's used in LabelNamespace and ModelSpace, to provide a global stack to access current active xxx. I'm not sure whether it's worth it to further polish it.

nni/mutable/utils.py
J-shang: I don't know the usage scenario very well, my intuition in this case is `abc_def_1`.

ultmaster: In current implementation:

This will generate an extra level.

```python
with LabelNamespace('abc'):
  with LabelNamespace('abc'):
    auto_label('abc')  # abc_1_1
    auto_label()  # global_1
```

The following is actually identical to `with LabelNamespace('abc'), LabelNamespace('def'):`. The nested-with here has no effect. 

```python
with LabelNamespace('abc'):
  with LabelNamespace('def'):
    auto_label('abc')  # abc_1
    auto_label('def')  # def_1
    auto_label()  # global_1
```

I think `auto_label()` can automatically fetch the nearest label namespace if you agree it's a good idea.

Nested-with can also be concatenated to generate labels like `abc_abc_1` or `abc_def_1`. Or maybe `abc/abc/1` and `abc/def/1`?

J-shang: agree, it seems more auto if it can fetch the nearest label namespace.

I have no idea if we should support the nested labels generation, if we do, I prefer `abc/abc/1`.

ultmaster: If then, I need to rewrite the whole label namespace. :)

ultmaster:
I almost rewrite the whole thing. Might need re-review.

nni/mutable/utils.py
```diff
@@ -0,0 +1,300 @@
...
+    >>> label1 = auto_label('bar')          # bar, because the scope is global
+    >>> label2 = auto_label()               # global/1, because label is not provided
+    >>> with label_scope('foo'):
+    ...     label3 = auto_label()           # foo/1, because in the scope "foo"
+    >>> with label_scope():
+    ...     label4 = auto_label()           # param/1, default key is used
+    >>> with label_scope('another'):
+    ...     label5 = auto_label()           # another/1
+    ...     label6 = auto_label('thing')    # another/thing
+    ...     label7 = auto_label()           # another/2
```
QuanluZhang, line 263: what if the following:
```
with label_scope('test'):
    label1 = auto_label('hello', 'world') #??
```

ultmaster: This is not valid. It must be

```python
label1 = auto_label('hello', label_scope('world'))
```

I think it's too advanced for regular users. I added a test to make sure it complains "ValueError: label_scope is not entered yet."

PR #5195, opened 2022-11-01
Mutable V3 (Stage 2) - Symbolic execution engine

ultmaster:
This PR migrates the symbolic execution engine, which is previously embedded in NAS, to mutable, and also makes it a general utility.

### DIFF
--- a/nni/mutable/symbol.py
+++ b/nni/mutable/symbol.py
@@ -420,6 +420,8 @@ class Symbol(SymbolicExpression):
 
     def evaluate(self, values: Iterator[Any] | dict[str, Any]) -> Any:
         if isinstance(values, dict):
+            if self.label is None:
+                raise ValueError('Cannot evaluate symbol without label when values is a dict.')
             if self.label not in values:
                 raise KeyError(f'{self.label} is not in {values}')
             value = values[self.label]
--- a/nni/mutable/symbol.py
+++ b/nni/mutable/symbol.py
@@ -1,0 +1,437 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+Infrastructure for symbolic execution.
+"""
+
+from __future__ import annotations
+
+__all__ = ['Symbol', 'SymbolicExpression']
+
+import math
+import operator
+from typing import Any, Iterable, Type, NoReturn, Callable, Iterator, overload
+
+# the code in ValueChoice can be generated with this codegen
+# this is not done online because I want to have type-hint supports
+# $ python -c "from nni.mutable.symbol import _symbol_expr_codegen; _symbol_expr_codegen(_internal=True)"
+def _symbol_expr_codegen(*, _internal: bool = False):
+    if not _internal:
+        raise RuntimeError("This method is set to be internal. Please don't use it directly.")
+    MAPPING = {
+        # unary
+        'neg': '-', 'pos': '+', 'invert': '~',
+        # binary
+        'add': '+', 'sub': '-', 'mul': '*', 'matmul': '@',
+        'truediv': '//', 'floordiv': '/', 'mod': '%',
+        'lshift': '<<', 'rshift': '>>',
+        'and': '&', 'xor': '^', 'or': '|',
+        # no reverse
+        'lt': '<', 'le': '<=', 'eq': '==',
+        'ne': '!=', 'ge': '>=', 'gt': '>',
+        # NOTE
+        # Currently we don't support operators like __contains__ (b in a),
+        # Might support them in future when we actually need them.
+    }
+
+    binary_template = """    def __{op}__(self, other: Any) -> Any:
+        return self.expr_cls(operator.{opt}, '{{}} {sym} {{}}', [self, other])"""
+
+    binary_r_template = """    def __r{op}__(self, other: Any) -> Any:
+        return self.expr_cls(operator.{opt}, '{{}} {sym} {{}}', [other, self])"""
+
+    unary_template = """    def __{op}__(self) -> Any:
+        return self.expr_cls(operator.{op}, '{sym}{{}}', [self])"""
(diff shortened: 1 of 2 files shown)


nni/mutable/symbol.py
```diff
@@ -0,0 +1,439 @@
...
+# $ python -c "from nni.mutable.symbol import _symbol_expr_codegen; _symbol_expr_codegen(_internal=True)"
+def _symbol_expr_codegen(*, _internal: bool = False):
+    if not _internal:
+        raise RuntimeError("This method is set to be internal. Please don't use it directly.")
+    MAPPING = {
+        # unary
+        'neg': '-', 'pos': '+', 'invert': '~',
+        # binary
+        'add': '+', 'sub': '-', 'mul': '*', 'matmul': '@',
+        'truediv': '//', 'floordiv': '/', 'mod': '%',
```
liuzhe-lz, line 27: truediv and floordiv are reversed.

nni/mutable/symbol.py
```diff
@@ -0,0 +1,439 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+Infrastructure for symbolic execution.
```
liuzhe-lz, line 5: I think it needs a brief description for "symbolic execution".
What it is used for and a minimal example about how to use it.

test/ut/sdk/mutable/test_symbol.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_symbol_repr():
+    x, y = Symbol('x'), Symbol('y')
+    expr = x * x + y * 2
+    assert str(expr)
+    (x * x) + (y * 2)
+    list(expr.leaf_symbols())
+    [Symbol('x'), Symbol('x'), Symbol('y')]
+    expr.evaluate({'x': 2, 'y': 3})
```
liuzhe-lz, line 11: Not asserting the result?

ultmaster: Left for lunch at that point. :(

nni/mutable/symbol.py
```diff
@@ -0,0 +1,472 @@
...
+    Operations on symbols (e.g., ``a + b``) will result in a new :class:`SymbolicExpression`.
+
+    Parameters
+    ----------
+    label
+        Each symbol is bound with a label, i.e., the variable name.
+    """
+
+    def __init__(self, label: str | None) -> None:
+        self.label = label
```
Louis-J, line 445: should a symbol without label exist? can we create a new symbol then a unique name is assigned to it automatically?

nni/mutable/symbol.py
```diff
@@ -0,0 +1,472 @@
...
+            if not isinstance(arg0, Iterable):
+                raise TypeError('Expect more than one items to compare min')
+            return SymbolicExpression.min(*list(arg0))
+        lst = list(arg0) if isinstance(arg0, Iterable) else [arg0] + list(args)
+        symbol_obj = first_symbolic_object(*lst)
+        if symbol_obj is not None:
+            return symbol_obj.expr_cls(min, 'min({})', lst)
+        return min(lst)
+
+    def __hash__(self):
```
Louis-J, line 259: why required?

nni/mutable/symbol.py
```diff
@@ -0,0 +1,472 @@
...
+
+        Useful for when you want to inspect when the symbols come from.
+        No deduplication even if the symbols has duplicates.
+        """
+        for arg in self.arguments:
+            if isinstance(arg, SymbolicExpression):
+                yield from arg.leaf_symbols()
+
+    @overload
+    def evaluate(self, values: Iterable[Any]) -> Any:
```
Louis-J, line 135: will the `Symbol('x') + Symbol('x')` got a 3 with input `[1, 2]`?

liuzhe-lz (approved):

nni/mutable/symbol.py
matluster: It used to be useful. But I think it's unnecessary now.

nni/mutable/symbol.py
matluster: Couldn't really recall. I guess it's because hash won't work because `__eq__` is overridden.

nni/mutable/symbol.py
matluster: Yes, but I think it doesn't really matter, because dict is more often-used now. Using a list of values, on the other hand, isn't really convenient...

Louis-J: It seems that there are many tests using a list input in #5200. Is it ok?

Louis-J: suggestion: try do next() in the end of the evaluate and assert getting the StopIteration

nni/mutable/symbol.py
Louis-J: suggestion: if hash isn't needed, '__hash__' can raise an error. '__eq__' will not return a bool so the hash cannot use.
but if hash is needed, we can determine whether to return a bool by detecting calling frame in '__hash__' and '__eq__'

ultmaster: I don't like calling frame because they are notoriously nasty and I don't think it's worth the effort here.

Left unimplemented can be an option. I just deleted `__hash__`. I guess I can add it back if some tests complain in future.

nni/mutable/symbol.py
ultmaster: > It seems that there are many tests using a list input in #5200. Is it ok?

Those tests are ported from a version. It always doesn't harm to have more tests.

> suggestion: try do next() in the end of the evaluate and assert getting the StopIteration

Might need to rewrite the whole `evaluate` to make this possible. Not sure whether worth the effort.

Louis-J (approved):

PR #5196, opened 2022-11-01
Mutable V3 (Stage 3) - Mutable base classes and exceptions

ultmaster:
This PR introduces mutable, which is now a fundamental building component to define a search space.

For the first time, I'm making `label` a well-defined concept. I created classes like `LabeledMutable` and `MutableSymbol`, which sits as the base class for most types of mutables.


mydmdm:
The design sounds reasonable for me, except for a few comments

- I'd like to see more details about how to construct and store the relationship between the variable and its dependencies. Though it's not feasible to demonstrate all the details in the abstract class, however, it would be helpful to declare the necessary data structure in it. For example, it make differences in the below example that `C` store the labels, or the reference to the two `Discrete` variables.   
```
A = Discrete([2, 3]) + Discrete([5, 7])
```

- The relationship between `MutableExpression`, `MutableSymbol` seems to be a little confusions. In my understanding, `Symbolic` is a special case of `Mutable` (with only one possible value). Both classes are similar to placeholder which will be filled later. Could they be combined together in the variable level and unified in the expression and later?

mydmdm (approved):
The design sounds reasonable for me, except for a few comments

- I'd like to see more details about how to construct and store the relationship between the variable and its dependencies. Though it's not feasible to demonstrate all the details in the abstract class, however, it would be helpful to declare the necessary data structure in it. For example, it make differences in the below example that `C` store the labels, or the reference to the two `Discrete` variables.   
```
A = Discrete([2, 3]) + Discrete([5, 7])
```

- The relationship between `MutableExpression`, `MutableSymbol` seems to be a little confusions. In my understanding, `Symbolic` is a special case of `Mutable` (with only one value). Both classes are similar to placeholder which will be filled later. Could they be combined together in the variable level and unified in the expression and later?

matluster:
> ```
> A = Discrete([2, 3]) + Discrete([5, 7])
> ```
>
> Whether `C` store the labels, or the reference to the two `Discrete` variables.

Both. More docs added.

>  Both `MutableExpression`, `MutableSymbol` are similar to placeholder which will be filled later.

`MutableSymbol` is a placeholder. `MutableExpression` isn't. Expression is a concrete class defined by an operator and several operands. One example is: `Discrete([2, 3]) + Discrete([5, 7])` will generate `MutableExpression(operator=operator.add, operands=[Discrete([2, 3]), Discrete([5, 7]))`.

nni/mutable/mutable.py
```diff
@@ -0,0 +1,482 @@
...
+            Use this to share the sampled value among mutables with the same label.
+        """
+        sample: Sample = {} if memo is None else memo
+        if random_state is None:
+            random_state = RandomState()
+        for mutable in self.simplify().values():
+            mutable.random(sample, random_state)
+        return self.freeze(sample)
+
+    def grid(self, memo: Sample | None = None, granularity: int | None = None) -> Iterable[Any]:
```
liuzhe-lz, line 346: So the caller of this function should iterate granularity in range(inf), and dedup with a set?

nni/mutable/mutable.py
```diff
@@ -0,0 +1,482 @@
...
+
+    The labeled mutables are by default created with one reproducible, auto-generated label.
+    But it can also contain multiple labels if the mutable is more complex.
+    The label can also be specified by users, or sub-classes.
+
+    Being a :class:`LabeledMutable` doesn't necessarily mean that it is a leaf mutable.
+    Some :class:`LabeledMutable` can be further simplified into multiple leaf mutables.
+    """
+
+    label: str
```
liuzhe-lz, line 398: Isn't this static?

nni/mutable/mutable.py
```diff
@@ -0,0 +1,482 @@
...
+        sample
+            See :meth:`freeze`.
+
+        Returns
+        -------
+        Optionally a :exc:`~nni.mutable.exception.SampleValidationError` if the sample is invalid.
+        """
+        raise NotImplementedError()
+
+    def leaf_mutables(self, is_leaf: Callable[[Mutable], bool]) -> Iterable[LabeledMutable]:
```
liuzhe-lz, line 170: Who provides the if_leaf callback?

nni/mutable/mutable.py
```diff
@@ -0,0 +1,482 @@
...
+        It's possible that :meth:`random` raises :exc:`~nni.mutable.exception.SampleValidationError`,
+        e.g., in cases when constraints are violated.
+
+        Parameters
+        ----------
+        memo
+            A dict of mutable labels and their random values.
+            Use this to share the sampled value among mutables with the same label.
+        """
+        sample: Sample = {} if memo is None else memo
```
liuzhe-lz, line 339: Should it raise if self is a leaf mutable?

nni/mutable/mutable.py
```diff
@@ -0,0 +1,482 @@
...
+            rv[mutable.label] = mutable
+        else:
+            if not mutable.equals(rv[mutable.label]):
+                raise ValueError(
+                    f'Mutables have the same label {mutable.label} but are different: {mutable} and {rv[mutable.label]}'
+                )
+    return rv
+
+
+class Mutable:
```
liuzhe-lz, line 74: Suggest more explicitly document which methods are to be used by "users", which are to be overridden by subclasses, and which are used by both.

nni/mutable/mutable.py
ultmaster: It's a instance variable. This is [Python syntax](https://peps.python.org/pep-0526/#class-and-instance-variable-annotations).

nni/mutable/mutable.py
ultmaster: Search algorithm. Algorithm should provide `is_leaf` based on the types of mutables they can handle.

`leaf_mutables` is not considered as first-level API, but a method to override. `simplify` is.

nni/mutable/mutable.py
```diff
@@ -0,0 +1,482 @@
...
+                return rv
+            except SampleValidationError:
+                if attempt == retries:
+                    raise ValueError(
+                        f'Cannot find a valid default sample after {retries} retries, for {self}. '
+                        'Please either set `default_value` manually, or loosen the constraints.')
+
+        raise RuntimeError('This should not happen.')
+
+    def random(self, memo: Sample | None = None, random_state: RandomState | None = None) -> Any:
```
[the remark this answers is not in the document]
ultmaster: Using the default value (None) should not hurt.

It's inspired by [Pytorch API](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.named_modules). I'm not saying that it's a good API though. But it seems necessary to the implementation.

nni/mutable/mutable.py
ultmaster: That's correct.

But I imagine most use cases of `grid` should be with a certain granularity level.

nni/mutable/mutable.py
ultmaster: I list the "must-override" list. There is no should-suggestion. Subclass can do whatever they want.

QuanluZhang (approved):

nni/mutable/mutable.py
liuzhe-lz: Wow, I though it was just type hint.

nni/mutable/mutable.py
liuzhe-lz: So in most cases algorithms call `simplify()` without argument, and if someone want do something special they call `leaf_mutables()` with a customized `is_leaf()`?

nni/mutable/mutable.py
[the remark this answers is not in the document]
liuzhe-lz: OK. That means it's acceptable for data scientists.

liuzhe-lz (approved):

nni/mutable/mutable.py
ultmaster: Correct.

PR #5197, opened 2022-11-01
Mutable V3 (Stage 4) - Categorical and Numerical

ultmaster:
Discrete and Continuous are both MutableSymbol, and they are the most useful two types of mutables in the first iteration of design.

This PR depends on stage 3.


nni/mutable/_notimplemented.py
```diff
@@ -0,0 +1,70 @@
...
+from .mutable import LabeledMutable, MutableSymbol, Discrete, Continuous
+
+
+class Permutation(MutableSymbol):
+    """Get a permutation of several values.
+    Not implemented. Kept as a placeholder.
+    """
+
+
+class RandomInteger(Discrete[int]):
```
liuzhe-lz, line 21: I want to change `randint` to default to ordered integers.
The old `randint` is nearly meaningless, as different tuners treat it differently.
`quniform` is not a perfect substitute and sometimes causes problems. For example `list_[quniform(0, 10, 1)]` does not work.
If someone wants unordered integers, they can simply use `choice(range(a, b))`.

nni/mutable/mutable.py
```diff
@@ -501,3 +508,608 @@ class MutableSymbol(LabeledMutable, Symbol, MutableExpression):
...
+class Discrete(MutableSymbol, Generic[Choice]):
+    """Choosing one from a list of values.
+
+    Parameters
+    ----------
+    values
+        The list of values to choose from.
+    distribution
+        The probability distribution of the values. Should be an array with the same length as ``values``.
+        The sum of the distribution should be 1.
```
liuzhe-lz, line 522: I think it's more convenient to use "weights".

nni/mutable/mutable.py
```diff
@@ -501,3 +508,608 @@ class MutableSymbol(LabeledMutable, Symbol, MutableExpression):
...
+        Distribution is respected if provided.
+
+        See Also
+        --------
+        Mutable.random
+        """
+        memo = {} if memo is None else memo
+        if random_state is None:
+            random_state = RandomState()
+        err = self.contains(memo)
```
liuzhe-lz, line 614: I just found that `if self.contains(sample):` evaluates to True when it does not contain sample...

nni/mutable/mutable.py
```diff
@@ -501,3 +508,608 @@ class MutableSymbol(LabeledMutable, Symbol, MutableExpression):
...
+        memo = {} if memo is None else memo
+        err = self.contains(memo)
+
+        if isinstance(err, SampleMissingError):
+            if all(dis == self.distribution[0] for dis in self.distribution):
+                # uniform distribution
+                values_perm = self.values
+            else:
+                # More heavily-distributed items are put upfront.
+                indices = sorted(list(range(len(self.values))), key=lambda i: self.distribution[i], reverse=True)
```
liuzhe-lz, line 636: sorted() takes an iterable, so `list` is unnecessary here. (too many parentheses)

nni/mutable/_notimplemented.py
ultmaster: I'm actually fine with both: either making randint a special case of Discrete, or making it an independent MutableSymbol.

The problem here is, however, what if the algorithm doesn't want to handle randint? In the design of `simplify`, there is a `is_leaf` callable, that supports to expand some mutables where you don't want to handle. In this case, you can return false when checking `is_leaf` for randint. And then what?

Option 1: it becomes a discrete.
Option 2: it becomes a quniform.
Option 3: it becomes a `NonNegativeRandomInteger` (see below). This is actually kept for those who doesn't even want to handle Discrete (i.e., don't care about the values, only indices matter.

If you think about this a little bit further, you will notice that `Continuous` are not that easy to handle. That's where `UnitUniform` comes from (and that's option 4).

My point here is that, the decision here is actually about the simplify-"fallback" of mutables. If the algorithm can handle this kind of mutable, it should never be a problem. If it can't, we need to have a backup plan to make their lives easier. And, everything will eventually fallback to `UnitUniform`.

Finally, let's vote for `RandomInteger -> Discrete -> NonNegativeRandomInteger -> UnitUniform`, or `RandomInteger -> NonNegativeRandomInteger -> UnitUniform` or `RandomInteger -> Continous -> UnitUniform`, or `RandomInteger -> UnitUniform`.

nni/mutable/mutable.py
ultmaster: Good point.

nni/mutable/_notimplemented.py
cruiseliu: I have found my original plan, it was `randint(low, high, categorical=False)`.

`UnitUniform` is bad in HPO because it does not support update search space. I know I used it in `ParameterSpec`. That's a mistake.

So,
  - RandomInteger (ordered) -> Continuous
  - RandomInteger (unordered) -> Discrete -> Continuous (one hot)

nni/mutable/mutable.py
cruiseliu: `check_contains()` should be better.
I will check the doc when I see a strange name like this.

nni/mutable/_notimplemented.py
ultmaster: Firstly I don't like utility functions here because it's notoriously hard to extend. When `simplify()` is a member function, all mutables can have the same interface (including the NAS-specific modules and model space). If `simplify()` is a util function, how should I do that?

Secondly Continuous is definitely not the final fallback layer because of its complexity right now. The algorithm needs to check all mu, sigma, low, high, quant, and optionally customized distribution to make sure that they have handled all the cases.
I understand algorithms might have preferences over what the fallback of Continuous should be (uniform or normal), but when `is_leaf` returns false for Continuous, the algorithm have already made the claim that Continuous is not supported. Continuous simplifies itself into uniform for the convenience of **most algorithms**. If you want to handle normal, you should've done it on the Continuous layer and never simplified it.
The position of `simpfily()` is that it makes writing algorithms easier. You can always choose NOT using it.

Adding `categorical=False` to randint might be a good idea, because it belongs to the semantics of mutable itself.

liuzhe-lz (approved):

nni/mutable/mutable.py
```diff
@@ -501,3 +524,608 @@ class MutableSymbol(LabeledMutable, Symbol, MutableExpression):
...
+        which can be obtained by calling ``scipy.stats.distribution_name(...)``.
+        If specified, ``low``, ``high``, ``mu``, ``sigma``, ``log_distributed`` will be ignored.
+    default
+        The default value. If not specified, the default value will be the median of distribution.
+    label
+        The label of the variable.
+
+    Examples
+    --------
+    To create a variable uniformly sampled from 0 to 1::
```
mydmdm, line 937: I don't think it's a good idea to let the distribution (e.g. normal, uniform) to be induced from whether `low, high` or `mu, sigma` is given. I prefer to let `Continuous` to be the base class (or with a default `uniform` random generator), and let `normal`, `lognormal` to have separate class.

nni/mutable/mutable.py
```diff
@@ -501,3 +524,608 @@ class MutableSymbol(LabeledMutable, Symbol, MutableExpression):
...
+
+        Continuous(low=0, high=1)
+
+    To create a variable normally sampled with mean 2 and std 3::
+
+        Continuous(mu=2, sigma=3)
+
+    To create a normally sampled variable with mean 0 and std 1, but
+    always in the range of [-1, 1] (note that it's not **truncated normal** though)::
```
mydmdm, line 946: Is it `normal` with constraints? What's the difference with **truncated normal**?

mydmdm (approved):
Most of the changes seems Ok to me. The main concern lies in the Continuous class. Instead of implying distributions from arguments, I prefer to have explicit child class for a specific distribution.

nni/mutable/mutable.py
```diff
@@ -501,3 +524,608 @@ class MutableSymbol(LabeledMutable, Symbol, MutableExpression):
 
     def __repr__(self) -> str:
         return f'{self.__class__.__name__}({self.extra_repr()})'
+
+
+class Discrete(MutableSymbol, Generic[Choice]):
+    """Choosing one from a list of values.
```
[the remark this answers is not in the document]
ultmaster: It's not an alias. Generic is used for types, so that we can have things like `Discrete[int]`.

Discrete is unordered.

nni/mutable/mutable.py
```diff
@@ -151,8 +158,10 @@ def freeze(self, sample: Sample) -> Any:
         """
         raise NotImplementedError()
 
-    def contains(self, sample: Sample) -> SampleValidationError | None:
+    def check_contains(self, sample: Sample) -> SampleValidationError | None:
```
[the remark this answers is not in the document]
ultmaster: `check_contains` is the implementation.
`contains` is used in `if` (return true if contains).
`validate` raises exception if not valid.

nni/mutable/mutable.py
ultmaster: uniform and normal are special cases of Continuous. They will be subclasses of Continuous.

The idea of Continuous is to support arbitrary distributions. Normal and uniform are only some shortcuts here.

nni/mutable/mutable.py
ultmaster: [Truncate normal](https://en.wikipedia.org/wiki/Truncated_normal_distribution) will scale ppf so that the total AUC is exactly 1. But what we did here is simply clip the values outside to the boundary.

Consider an extreme case where mu=0, sigma=100 and low and high is only [-0.1, 0.1]. Truncated normal will have an almost uniform distribution over [-0.1, 0.1]. But in our implementation, it's likely that the value will be either -0.1 or 0.1.

ultmaster:
Updated names to Categorical and Numerical per Yuqing's suggestion.

PR #5199, opened 2022-11-01
Mutable V3 (Stage 6) - Mutable Annotations

ultmaster:
Mutable Annotation is a component to provide extra semantics to search spaces. Typical examples here are constraints.


QuanluZhang (commented):
this pr is clear. one question, how should the algorithms such as RL respect the constraint? For algorithms such as tpe, GP, as long as the sampling respects the constraint, the algorithm would work.

QuanluZhang (approved):

matluster:
First of all, if users write a search space with constraints, the algo will see it, whether they like it or not. Algos like RL can also "creatively" manage to handle at least some kind of constraints if they want to do it. Now let's discuss what will happen if they choose to **ignore** it.

When algos use the *freeze* API to convert their samples (dict like `{"a": 1}`) to real trial payloads (objects like a sampled nn.Module, or a function, or a very complex list), the *freeze* of constraints will also be invoked and the constraint will be checked. If it doesn't work, a *ConstraintViolation* exception will be raised. In this case, the algo can either directly crash, or catch the exception and ignore, or give it a negative reward and learn not to sample it.
That being said, there is no way to **force** sending an invalid sample. Maybe we can design another *SoftConstraint* as another annotation, but that's beyond the scope of this PR.

liuzhe-lz (approved):

PR #5201, opened 2022-11-03
Fix RegularizedEvolution strategy hang issue

amznero:
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/nni/nas/experiment/pytorch.py
+++ b/nni/nas/experiment/pytorch.py
@@ -304,7 +304,7 @@ class RetiariiExperiment(Experiment):
                     raise RuntimeError(f'The experiment mode "{self._action}" is not supposed to invoke run() method.')
 
                 # non-block
-                self._run_strategy_thread = Thread(target=self._run_strategy, args=(base_model_ir, self.applied_mutators))
+                self._run_strategy_thread = Thread(target=self._run_strategy, args=(base_model_ir, self.applied_mutators), daemon=True)
                 self._run_strategy_thread.start()
                 # FIXME: move this logic to strategy with a new API provided by execution engine
                 self._wait_completion()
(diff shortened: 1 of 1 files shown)


amznero:
Related issue: #5202

nni/nas/experiment/pytorch.py
```diff
@@ -303,7 +303,9 @@ def run(self,
                 else:
                     raise RuntimeError(f'The experiment mode "{self._action}" is not supposed to invoke run() method.')
 
-                self._run_strategy(base_model_ir, self.applied_mutators)
+                # non-block
+                self._run_strategy_thread = Thread(target=self._run_strategy, args=(base_model_ir, self.applied_mutators), daemon=True)
+                self._run_strategy_thread.start()
```
amznero, line 308: some concerns here. 
Python cannot explicitly terminate a thread, so _run_strategy will still activate even Experiment.run() returned.

ultmaster: Starting strategy in the main thread is by design. We did this for debuggability and potential needs for interactive experiments.
Also, as you mentioned, there could be potential risks that prevent strategy from terminating in tricky cases.

Overall, I think the correct way to do this is: the strategy could have a way to check whether experiment "should" stop, and terminate itself when it "should". This is already in a draft design, and might come in v3.0 or v3.1.

amznero: Yes, I also think it would be better to let strategy decide when to terminate the search.

I'll close this pr. Look forward to your design :).

PR #5206, opened 2022-11-06
Bump node.js to 18

liuzhe-lz:
### Description ###

Bump node.js to 18.x and update some packages.

Drops `yarn` for NNI manager.
`npm` is good enough now. No need to add an extra dependency.

The target branch will be changed to master once v2.10 released and merged.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - ~~test case~~
  - ~~doc~~

### How to test ###

No need


setup_ts.py
```diff
@@ -24,8 +24,8 @@
 from zipfile import ZipFile
 
 
-node_version = 'v16.14.2'
-yarn_version = 'v1.22.10'
+node_version = 'v18.12.1'
+yarn_version = 'v1.22.19'
```
matluster, line 28: I thought there is no yarn any more.

J-shang: haven't written ts for a long time, now do we still have `yarn watch`?

ts/nni_manager/core/nniDataStore.ts
```diff
@@ -59,7 +59,7 @@ class NNIDataStore implements DataStore {
         try {
             await this.db.storeExperimentProfile(experimentProfile);
         } catch (err) {
-            throw NNIError.FromError(err, 'Datastore error: ');
+            throw NNIError.FromError(err as any, 'Datastore error: ');
```
J-shang, line 62: why we need add these `as any`

setup_ts.py
liuzhe-lz: > I thought there is no yarn any more.

yarn is still used in webui & jupyter extension.
It can be fully removed, but don't want to do it now.

liuzhe-lz: > haven't written ts for a long time, now do we still have `yarn watch`?

It will need a separate watcher like [npm-watch](https://www.npmjs.com/package/npm-watch). Have not added it because myself do not use it.

ts/nni_manager/core/nniDataStore.ts
liuzhe-lz: Because tsc has been updated and is complaining.
The code is somewhat wrong. The function only accepts Error and string, but there is no such guarantee for err's type. Don't want to fix it though.

J-shang (approved):

ultmaster (approved):

ts/nni_manager/common/training_service_v3.ts
```diff
@@ -0,0 +1,120 @@
...
+ *  To minimize the change, it is currently identical to HyperParameters.value
+ **/
+export type Parameter = string;
+
+export interface EnvironmentInfo {
+    id: string;
+    // TODO
+}
+
+export interface TrainingServiceV3 {
```
matluster, line 23: Hold on... where does this come from?

liuzhe-lz: Well this PR bases on v3 dev branch and was supposed to be merged after #5211 . I manually removed some files later but seems no that clean.
The merge schedule is as following:
 1. bump version (this)
 2. http command channel (#5216)
 3. gpu scheduler (#5219)
 4. trial keeper (#5229)
 5. local training service (next)  ⬅  this one depends on the old dev branch #5211

PR #5216, opened 2022-11-09
Training service util: HTTP command channel

liuzhe-lz:
### Description ###

A simple command channel based on HTTP and polling.

It will be used by the new trial keeper to interact with trials.
The HTTP protocol is chosen because:
 1. The channel client is user's trial code. So it is good to keep the client as simple as possible.
 2. There are not many commands and only one of them is issued by the server. For each trial there is only 1 request parameter, 1 send parameter (from server), N send intermediate metrics, and 1 send final metric. Therefore the performance does not matter, especially the server to client performance.

Note: Trial logs (stdout and stderr) are not sent by command channel.

The RestServer is modified to support register REST API handlers on the fly.
Previously the RestServer is created after all other components and register all handlers during its initialization.
Now RestServer is the first initialized component so it can response to "check status" requests at the earliest stage. And then each other component will install their handlers on ready.

I'm using node 18 APIs in new codes so it might depend on #5206

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [x] test case
  - ~~doc~~

### DIFF
--- a/nni/runtime/command_channel/http.py
+++ b/nni/runtime/command_channel/http.py
@@ -1,0 +1,28 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import requests
+
+from .base import Command, CommandChannel
+
+class HttpChannel(CommandChannel):
+    """
+    A simple command channel based on HTTP.
+
+    Check the server for details. (``ts/nni_manager/common/command_channel/http.ts``)
+    """
+
+    def __init__(self, url: str):
+        self._url: str = url
+
+    def send(self, command: Command) -> None:
+        requests.put(self._url, json=command)
+
+    def receive(self) -> Command:
+        while True:
+            r = requests.get(self._url)
+            if r.status_code == 200:
+                return r.json()
+            if r.status_code == 408:
+                continue
+            raise IOError(f'HTTP command channel received unexpected status code {r.status_code}')
--- a/nni/runtime/command_channel/http.py
+++ b/nni/runtime/command_channel/http.py
@@ -1,9 +1,13 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import logging
+
 import requests
 
 from .base import Command, CommandChannel
+
+_logger = logging.getLogger(__name__)
 
 class HttpChannel(CommandChannel):
     """
@@ -18,11 +22,14 @@ class HttpChannel(CommandChannel):
     def send(self, command: Command) -> None:
         requests.put(self._url, json=command)
 
-    def receive(self) -> Command:
+    def receive(self) -> Command | None:
         while True:
             r = requests.get(self._url)
             if r.status_code == 200:
                 return r.json()
             if r.status_code == 408:
                 continue
+            if r.status_code == 410:
+                return None
+            _logger.error('Bad status %s %s', r.status_code, r.text)
             raise IOError(f'HTTP command channel received unexpected status code {r.status_code}')
--- a/ts/nni_manager/common/command_channel/http.ts
+++ b/ts/nni_manager/common/command_channel/http.ts
@@ -95,6 +95,9 @@ export class HttpChannelServer implements CommandChannelServer {
     }
 }
 
+/**
+ *  A FIFO queue with asynchronous pop.
+ **/
 class CommandQueue {
     private commands: Command[] = [];
     private consumers: Deferred<Command | null>[] = [];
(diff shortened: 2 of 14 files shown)


ts/nni_manager/common/command_channel/http.ts
```diff
@@ -0,0 +1,119 @@
...
+    // the server can only send commands when the client requests, so it needs a queue
+    private outgoingQueues: Map<string, CommandQueue> = new Map();
+    private path: string;
+
+    constructor(name: string, urlPath: string) {
+        this.log = getLogger(`HttpChannelManager.${name}`);
+        this.path = urlPath;
+    }
+
+    public start(): void {
```
matluster, line 38: Where is stop?

nni/runtime/command_channel/http.py
```diff
@@ -0,0 +1,28 @@
...
+    def send(self, command: Command) -> None:
+        requests.put(self._url, json=command)
+
+    def receive(self) -> Command:
+        while True:
+            r = requests.get(self._url)
+            if r.status_code == 200:
+                return r.json()
+            if r.status_code == 408:
+                continue
```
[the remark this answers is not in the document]
cruiseliu: The delay is done at server side.

nni/runtime/command_channel/http.py
```diff
@@ -0,0 +1,28 @@
...
+            raise IOError(f'HTTP command channel received unexpected status code {r.status_code}')
```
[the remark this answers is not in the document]
cruiseliu: added

ts/nni_manager/common/command_channel/http.ts
cruiseliu: added

ts/nni_manager/common/command_channel/http.ts
```diff
@@ -0,0 +1,140 @@
...
+        this.serving = true;
+        const channelPath = globals.rest.urlJoin(this.path, ':channel');
+        globals.rest.registerSyncHandler('GET', channelPath, this.handleGet.bind(this));
+        globals.rest.registerSyncHandler('PUT', channelPath, this.handlePut.bind(this));
+    }
+
+    public shutdown(): void {
+        this.serving = false;
+        this.outgoingQueues.forEach(queue => { queue.clear(); });
```
ultmaster, line 51: It looks not symmetric to `start()`. For example, `start()` has registerHandler, but shudown doesn't have unregister.

ultmaster (approved):

ts/nni_manager/common/command_channel/http.ts
cruiseliu: That's why it's called shutdown rather than stop.
Removing handler is not an easy task for express. I have not seen needs for "stop" so far so it's not worthy the efforts.

ts/nni_manager/common/command_channel/interface.ts
```diff
@@ -0,0 +1,23 @@
...
+ *  Each channel is connected to a client.
+ *
+ *  Normally each client has a unique channel URL,
+ *  which can be got with `server.getChannelUrl(id)`.
+ *
+ *  The APIs might be changed to return `Promise<void>` in future.
+ **/
+export interface CommandChannelServer {
+    // constructor(name: string, urlPath: string)
+    start(): void;
```
QuanluZhang, line 19: missed `shutdown` here

ts/nni_manager/common/command_channel/http.ts
```diff
@@ -0,0 +1,140 @@
...
+        if (!this.outgoingQueues.has(channelId)) {
+            this.outgoingQueues.set(channelId, new CommandQueue());
+        }
+        return this.outgoingQueues.get(channelId)!;
+    }
+}
+
+class CommandQueue {
+    private commands: Command[] = [];
+    private consumers: Deferred<Command | null>[] = [];
```
QuanluZhang, line 100: why we need both `commands` and `consumers` here. what is `consumers` used for?

ts/nni_manager/rest_server/index.ts
```diff
@@ -64,7 +63,8 @@ export class RestServer {
         const app = express();
         expressWs(app, undefined, { wsOptions: { maxPayload: 4 * 1024 * 1024 * 1024 }});
 
-        app.use('/' + this.urlPrefix, rootRouter());
+        app.use('/' + this.urlPrefix, mainRouter());
+        app.use('/' + this.urlPrefix, fallbackRouter());
```
QuanluZhang, line 67: maybe a stupid question, one path or url can be registered with two handlers?

ts/nni_manager/common/command_channel/http.ts
liuzhe-lz: See added doc

ts/nni_manager/rest_server/index.ts
liuzhe-lz: [Yes](https://expressjs.com/en/guide/routing.html)

ts/nni_manager/common/command_channel/interface.ts
liuzhe-lz: added

QuanluZhang (approved):

Issue #5046, opened 2022-08-04
torch.cuda.is_available() changes to FALSE at the second trial


rockcor:
**Describe the issue**:
May like issue #4853 
**torch.cuda.is_available()** is FALSE at the second trial but TRUE at the first trial (by print). Also, it is TRUE when running the main.py.
May be something wrong in my **config.yml**? 
My project works well on old-version nni without keyword **use active GPU**. Is this the problem comes from?

**Environment**:
- NNI version: 2.8.0
- Training service (local|remote|pai|aml|etc):local
- Client OS:ubuntu18.04
- Server OS (for remote mode only):
- Python version: 3.8.0
- PyTorch/TensorFlow version: 1.12
- Is conda/virtualenv/venv used?: yes
- Is running in Docker?: no

**Configuration**:
 - Experiment config (remember to remove secrets!):experimentName: ***
searchSpaceFile: search_space.json
trialCommand: python3 main.py
trialCodeDirectory: .
trialGpuNumber: 1
trialConcurrency: 1
maxExperimentDuration: 2h
maxTrialNumber: 100
tuner:
  name: GridSearch
trainingService:
  platform: local
  useActiveGpu: True
  maxTrialNumberPerGpu: 1
 - Search space:{
    "split": {"_type": "choice", "_value": [4, 6]},
    "gen": {"_type": "choice", "_value": [24, 48]},
    "hidden_dim": {"_type": "choice", "_value": [32, 64]}
}

**Log message**:
 - nnimanager.log:[2022-08-04 17:12:50] INFO (main) Start NNI manager
[2022-08-04 17:12:50] INFO (NNIDataStore) Datastore initialization done
[2022-08-04 17:12:50] INFO (RestServer) Starting REST server at port 10089, URL prefix: "/"
[2022-08-04 17:12:51] WARNING (NNITensorboardManager) Tensorboard may not installed, if you want to use tensorboard, please check if tensorboard installed.
[2022-08-04 17:12:51] INFO (RestServer) REST server started.
[2022-08-04 17:12:51] INFO (NNIManager) Starting experiment: uenxijm9
[2022-08-04 17:12:51] INFO (NNIManager) Setup training service...
[2022-08-04 17:12:51] INFO (LocalTrainingService) Construct local machine training service.
[2022-08-04 17:12:51] INFO (NNIManager) Setup tuner...
[2022-08-04 17:12:51] INFO (NNIManager) Change NNIManager status from: INITIALIZED to: RUNNING
[...]

Issue #4830, opened 2022-05-01
please solve the bug on windows: WARNING (GPUScheduler) gpu_metrics file does not exist!


qqqck18:
NNI version: 2.7
Local OS: windows 10

when I use _nnictl create ..._ , it's always waiting, and nnimanager.log tell me _WARNING (GPUScheduler) gpu_metrics file does not exist!_

I notice that someone has mentioned before, the bug still exists.

Here's my solution:
#jianh is my usrname
1)cd C:\Users\jianh\AppData\Local\Temp\jianh\nni\script
2)edit _gpu_metrics_collector.ps1_ , change the first line from:
$env:METRIC_OUTPUT_DIR="C:\Users\jianh\AppData\Local\Temp/jianh/nni/script"
to
$env:METRIC_OUTPUT_DIR="C:\Users\jianh\AppData\Local\Temp\jianh\nni\script"
3) open a powershell, enter your environment containing nni, and execute _gpu_metrics_collector.ps1_ by:
.\gpu_metrics_collector.ps1
4) if you can find the file named "gpu_metrics" under C:\Users\jianh\AppData\Local\Temp\jianh\nni\script     done!!

It seems caused by the path separator '\' and '/' , and may lead to some unwanted escape character

Issue #4788, opened 2022-04-21
HPO: gpu_metrics file does not exist!


MinYounZhang:
**Describe the issue**:
I run`nnictl create --config exp.yaml --port 9090` and  cheak `nvidia-smi`, find its running without using GPU.
So  I run`nnictl create --config exp.yaml --port 9090 - f` and get error below:
![image](https://user-images.githubusercontent.com/19604387/164359111-4e2dd2ac-33ae-466f-b21d-774ff51977af.png)

I have followed tips in https://github.com/microsoft/nni/issues/4226
`METRIC_OUTPUT_DIR=${scriptFolder} python3 -m nni.tools.gpu_tool.gpu_metrics_collector` works under my path.

**Environment**:
- NNI version: 2.6
- Training service (local|remote|pai|aml|etc): local
- Client OS: Ubuntu 18.04
- Server OS (for remote mode only): \
- Python version:  py3.7
- PyTorch/TensorFlow version:  pytorch 1.10.2
- Is conda/virtualenv/venv used?: Y
- Is running in Docker?: N

**Configuration**:
 - Experiment config (remember to remove secrets!):
 - Search space:
![image](https://user-images.githubusercontent.com/19604387/164230399-10b95038-ce0d-4e3b-b6d5-1f118a78e76f.png)

**Log message**:
 - nnimanager.log:
 Dont exist
 - dispatcher.log:
[2022-04-20 20:12:10] INFO (nni.experiment/MainThread) Creating experiment, Experiment ID: 30pecn7s
[2022-04-20 20:12:10] INFO (nni.experiment/MainThread) Starting web server...
[2022-04-20 20:12:11] WARNING (nni.experiment/MainThread) Timeout, retry...
[2022-04-20 20:12:12] INFO (nni.experiment/MainThread) Setting up...
[2022-04-20 20:12:12] INFO (nni.experiment/MainThread) Web UI URLs: http://127.0.0.1:9090 http://172.17.7.99:9090 http://172.18.0.1:9090 http://172.19.0.1:9090 http://172.20.0.1:9090
[2022-04-20 20:12:12] INFO (nni.tools.nnictl.launcher/MainThread) To stop experiment run "nnictl stop 30pecn7s" or "nnictl stop --all"
[2022-04-20 20:12:12] INFO (nni.tools.nnictl.launcher/MainThread) Reference: https://nni.readthedocs.io/en/stable/Tutorial/Nnictl.html
[...]

PR #5219, opened 2022-11-10
Training service util: GPU scheduler (library part)

liuzhe-lz:
### Description ###

Current GPU scheduler is too buggy (#5046, #4830, #4788, etc). Time for a new one.

There will be a follow up PR for daemon part.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [x] test case
  - [x] doc

### DIFF
--- a/nni/tools/nni_manager_scripts/collect_gpu_info.py
+++ b/nni/tools/nni_manager_scripts/collect_gpu_info.py
@@ -8,7 +8,8 @@ Results are printed to stdout in JSON format.
 See `ts/nni_manager/common/gpu_scheduler/collect_info` for details.
 """
 
-# pylint: skip-file
+# pylint: disable=wildcard-import
+# pylint: disable=unused-wildcard-import
 
 from __future__ import annotations
 
@@ -30,7 +31,7 @@ def main() -> None:
         data = {'success': False}
     if errors:
         data['errors'] = sorted(errors)
-    print(json.dumps(data))
+    print(json.dumps(data), flush=True)
 
 def dict_factory(obj):
     ret = {}
--- a/nni/tools/nni_manager_scripts/collect_gpu_info.py
+++ b/nni/tools/nni_manager_scripts/collect_gpu_info.py
@@ -8,6 +8,8 @@ Results are printed to stdout in JSON format.
 See `ts/nni_manager/common/gpu_scheduler/collect_info` for details.
 """
 
+# pylint: skip-file
+
 from __future__ import annotations
 
 from dataclasses import asdict, dataclass
--- a/nni/tools/nni_manager_scripts/collect_gpu_info.py
+++ b/nni/tools/nni_manager_scripts/collect_gpu_info.py
@@ -37,6 +37,8 @@ def dict_factory(obj):
     for k, v in obj:
         if k.startswith('_'):
             continue
+        if v is None:
+            continue
         words = k.split('_')
         camel_k = words[0] + ''.join(word.title() for word in words[1:])
         ret[camel_k] = v
--- a/nni/tools/nni_manager_scripts/collect_gpu_info.py
+++ b/nni/tools/nni_manager_scripts/collect_gpu_info.py
@@ -1,6 +1,13 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+"""
+Collect GPU utilization metrics, and debug info if ``--detail`` is specified.
+Results are printed to stdout in JSON format.
+
+See `ts/nni_manager/common/gpu_scheduler/collect_info` for details.
+"""
+
 from __future__ import annotations
 
 from dataclasses import asdict, dataclass
--- a/ts/nni_manager/common/gpu_scheduler/scheduler.ts
+++ b/ts/nni_manager/common/gpu_scheduler/scheduler.ts
@@ -1,0 +1,175 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { getLogger } from 'common/log';
+import { GpuSystemInfo, collectGpuInfo } from './collect_info';
+
+const logger = getLogger('GpuScheduler');
+
+export interface ScheduleRestrictions {
+    onlyAcceptIndices?: number[];
+    rejectActiveGpus?: boolean;
+    rejectComputeActiveGpus?: boolean;
+}
+
+interface SchedulerGpuInfo {
+    index: number;
+    util: number;
+    coreUtil: number;
+    memUtil: number;
+    active: boolean;
+    computeActive: boolean;
+}
+
+interface SchedulerTrialInfo {
+    gpuIndex: number;
+    experimentId: string;
+    trialId: string;
+    util: number;
+}
+
+export class GpuScheduler {
+    private gpus: SchedulerGpuInfo[] = [];
+    private trials: SchedulerTrialInfo[] = [];
+
+    public async init(): Promise<void> {
+        const info = await collectGpuInfo(true);
+        if (info === null) {
+            throw new Error('GpuScheduler: Failed to collect GPU info');
+        }
+        if (info.gpuNumber === 0) {
+            throw new Error('GpuScheduler: No GPU found');
+        }
+
+        for (let i = 0; i < info.gpuNumber; i++) {
+            this.gpus.push({
+                index: i,
+                util: 0,
+                coreUtil: 0,
+                memUtil: 0,
+                active: false,
+                computeActive: false,
+            });
+        }
+
+        this.updateGpus(info);
+    }
+
+    public async update(forceUpdate?: boolean): Promise<void> {
+        const info = await collectGpuInfo(forceUpdate);
+        if (info === null) {
+            if (forceUpdate) {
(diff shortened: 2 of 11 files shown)


nni/tools/nni_manager_scripts/collect_gpu_info.py
```diff
@@ -0,0 +1,173 @@
...
+# Licensed under the MIT license.
+
+"""
+Collect GPU utilization metrics, and debug info if ``--detail`` is specified.
+Results are printed to stdout in JSON format.
+
+See `ts/nni_manager/common/gpu_scheduler/collect_info` for details.
+"""
+
+# pylint: skip-file
```
ultmaster, line 11: Why skip it?

nni/tools/nni_manager_scripts/collect_gpu_info.py
```diff
@@ -0,0 +1,173 @@
...
+def main() -> None:
+    info = collect('--detail' in sys.argv)
+    if info:
+        data = asdict(info, dict_factory=dict_factory)
+        data['success'] = True
+    else:
+        data = {'success': False}
+    if errors:
+        data['errors'] = sorted(errors)
+    print(json.dumps(data))
```
ultmaster, line 33: Flush it?

nni/tools/nni_manager_scripts/collect_gpu_info.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, info: Any, gpu_index: int, type_: Literal['compute', 'graphics'], detail: bool):
+        self.pid = info.pid
+        if detail:
+            self.name = nvmlSystemGetProcessName(self.pid)
+        self.gpu_index = gpu_index
+        self.type = type_
+        self.used_gpu_memory = info.usedGpuMemory
+
+if __name__ == '__main__':
+    main()
```
ultmaster, line 173: Missing EOL at EOF.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
```diff
@@ -0,0 +1,246 @@
...
+     *      This is useful for desktop systems with graphical interface.
+     **/
+    public async schedule(
+            experimentId: string,
+            trialId: string,
+            gpuNumber: number,
+            restrictions?: ScheduleRestrictions): Promise<number[] | null> {
+
+        if (gpuNumber >= this.gpus.length) {
+            throw new Error(`GpuScheduler: Only have ${this.gpus.length} GPUs, requesting ${gpuNumber}`);
```
ultmaster, line 126: Should it raise an error or simply reject the trial?

Will an error crash the whole experiment?

ts/nni_manager/common/gpu_scheduler/scheduler.ts
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        const gpus = this.sortGpus(restrictions ?? {});
+
+        if (gpuNumber < 1) {
```
ultmaster, line 131: I think allowing `gpuNumber` to be a float is a bad design that will bring us troubles in future.

Maybe requestedGpuMemory is a better idea.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     *  Update GPUs' utilization info.
+     *
+     *  If `force` is not true, it may use cached result.
+     *
+     *  This is normally unnecessary because it will implicitly update before scheduling.
+     **/
+    public async update(force?: boolean): Promise<void> {
```
ultmaster, line 76: Who will call this?

ts/nni_manager/common/gpu_scheduler/scheduler.ts
```diff
@@ -0,0 +1,246 @@
...
+            }
+            return a.index - b.index;
+        });
+    }
+
+    private releaseByFilter(filter: (trial: SchedulerTrialInfo) => boolean): void {
+        const trials = this.trials.filter(filter);
+        trials.forEach(trial => {
+            logger.debug(`Released ${trial.experimentId}/${trial.trialId}`);
+            this.gpus[trial.gpuIndex].util -= trial.util;
```
ultmaster, line 232: `this.gpus` might not be correct when someone is sharing the machine with others.

ts/nni_manager/test/common/gpu_scheduler.ts
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import assert from 'assert/strict';
```
ultmaster, line 4: Suggest adding a test on real GPU.

Maybe on full test HPO.

nni/tools/nni_manager_scripts/collect_gpu_info.py
liuzhe-lz: nvml* APIs are not pythonic. Wildcard import, unused import, not snake case, etc.

nni/tools/nni_manager_scripts/collect_gpu_info.py
liuzhe-lz: good idea

nni/tools/nni_manager_scripts/collect_gpu_info.py
liuzhe-lz: I believe my editor adds EOL (and hides EOL) automatically. If it's missing, there will be a red "⏎" line on github web.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
liuzhe-lz: Maybe an error message on web ui is a better option.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
liuzhe-lz: We have `trial_number_per_gpu` currently. It will be translated to `gpu_number: gpu_number / trial_number_per_gpu` somewhere. So it is at least useful internally inside NNI manager. We can reject float `gpu_number` at python side validation.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
liuzhe-lz: GPU scheduler will be a shared daemon process. For example when a new experiment connects to the scheduler it will request a full update in it's start().
And I plan to show machine info on web ui. The user will be able to request an update.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
liuzhe-lz: That's why it will be a daemon.

ts/nni_manager/test/common/gpu_scheduler.ts
liuzhe-lz: Er, maybe later as a part of integration test.
I'm scared of those environments :(

ts/nni_manager/common/gpu_scheduler/scheduler.ts
ultmaster: So it will crash the experiment?

ts/nni_manager/test/common/gpu_scheduler.ts
ultmaster: I don't agree it's integration test. Technically we don't have integration tests now. We only have full tests for HPO and training services. I don't think it's part of the training service. I think at least the GPU scheduler deserves a dedicated test rather than only be tested in an end-to-end experiment.

It's acceptable that it will be done in a separate PR.

ts/nni_manager/common/gpu_scheduler/scheduler.ts
liuzhe-lz: No. The trial dispatcher / scheduler will catch all exceptions.
Training services are assumed to be unreliable.

ts/nni_manager/test/common/gpu_scheduler.ts
liuzhe-lz: The GPU scheduler will become a separate daemon process. It will have its own entry point file alongside `main.ts`. So I will add GPU integration tests for the daemon, and use mocked unit tests for the library (the class).

ts/nni_manager/common/gpu_scheduler/scheduler.ts
ultmaster: According to my experience, GPUs are not that reliable. All exceptions should be caught and "GPU unavailable" should be reported when necessary, but it shouldn't crash the experiment because GPUs might "come back".

ultmaster (approved):

ts/nni_manager/common/gpu_scheduler/scheduler.ts
ultmaster: Seems ray is also doing so. I think it might work as well. https://docs.ray.io/en/latest/ray-core/tasks/using-ray-with-gpus.html#fractional-gpus

PR #5228, opened 2022-11-15
[Compression] remove pruning v1 & refactor directory

J-shang:
#### Note: 
* `sensitivity_analysis.py` is deleted because it depends on pruning v1, if it is necessary, will add it back on v2.
* `apply_compression.py` is deleted because it will expire after wrapper refactor.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


nni/algorithms/compression/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import logging
+
+_logger = logging.getLogger(__name__)
+_logger.warning('import path `nni.algorithms.compression` has be removed in v3.0, please import from `nni.compression`.')
```
matluster, line 7: has been

nni/algorithms/compression/pytorch/auto_compress/__init__.py
```diff
@@ -1,6 +0,0 @@
-# Copyright (c) Microsoft Corporation.
```
matluster, line 1: Removed?

nni/algorithms/compression/__init__.py
J-shang: good point, just want to give more information for different import levels, if raise error here, it will directly raise error in the upper level, for example:
```
>>> from nni.algorithms.compression.v2.pytorch.pruning import xx
  File "<stdin>", line 1, in <module>
  File "/home/xxx/nni/nni/algorithms/compression/__init__.py", line 4, in <module>
    raise RuntimeError('something')
RuntimeError: something
```
I can't tell user change `from nni.algorithms.compression.v2.pytorch.pruning import xx` to `from nni.compression.pytorch.pruning import xx` by raise error.

matluster: What I did in NAS is that I kept all the original files but deleted their contents. Added a shortcut to the new location.

J-shang: align with NAS

ultmaster (approved):

QuanluZhang (approved):

PR #5229, opened 2022-11-16
Training service util: Trial keeper (library part)

liuzhe-lz:
### Description ###

Unfinished tasks:
 1. Support graceful kill trial on Windows. Indefinitely delayed because not important.
 2. Trial keeper unit test. The trial keeper makes up 90% of the local training service, so I do not want to write a separate test.

This version does not include the daemon part or distributed trial support.

This PR will move GPU scheduler to inside trial keeper, after #5129 get merged.
The reason is that according to our discussion GPU scheduler will only be used in reuse mode, and reuse mode == have trial keeper.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,175 @@
...
+            } else if (command.type === 'metric') {
+                this.emitter.emit('metric', trialId, command.metric);
+            } else {
+                this.log.error('Received bad command:', trialId, command);
+            }
+        });
+    }
+
+    public async start(): Promise<void> {
+        await this.channels.start();
```
matluster, line 71: Users might need a way to setup custom setup in environment. For example, download some datasets, setup some account authentication. These only needs to be done once.

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,175 @@
...
+
+        if (this.gpuScheduler !== null) {
```
matluster, line 73: How does the scheulder work for pure-CPU workloads?

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,175 @@
...
+        promises = promises.concat(trials.map(trial => trial.kill()));
+
+        await Promise.all(promises);
+    }
+
+    public registerDirectory(name: string, path: string): void {
+        this.dirs.set(name, path);
+    }
+
+    public async createTrial(options: TrialKeeper.TrialOptions): Promise<string | null> {
```
matluster, line 97: I don't like the term trial here. Looks like we are abusing it.

What it actually does is: find a proper device placement and `createProcess`, and keep a channel to communicate with the process.

cruiseliu: The process has bound trial id and trial command channel. Do you think they should be renamed as well?
P.S. The trial command channel should not be renamed to "command channel", because this channel only accepts trial commands, while there are other channels accepting tuner commands, environment commands, GPU scheduler commands, etc.

ts/nni_manager/common/trial_keeper/keeper.ts
cruiseliu: Always accept. This is the v2 behavior.

ts/nni_manager/common/trial_keeper/keeper.ts
cruiseliu: Are you proposing a new feature? Like a `setup_command` config field similar to `trial_command`?

ts/nni_manager/common/trial_keeper/keeper.ts
matluster: My concern is that, in distributed trial, this only creates a worker, which is actually part of a trial.

ts/nni_manager/common/trial_keeper/keeper.ts
matluster: I remember remote training service used to have something like this.

ts/nni_manager/test/common/trial_keeper/process.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    while True:
+        pass
+except KeyboardInterrupt:
+    print('graceful exit', file=sys.stderr, flush=True)
+`
+    const stderr = (process.platform === 'win32' ? undefined : 'graceful exit');
+    await testKillHelper('test_grace', script, '', stderr);
+}
+
+async function testForceKill(): Promise<void> {
```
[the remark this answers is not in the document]
cruiseliu: I found they are not killed. Left a FIXME.

Try this script:
```
from subprocess import Popen
import os
import time

proc = Popen(['python', '-c', 'import time ; time.sleep(600)'])
print('parent:', os.getpid())
print('child:', proc.pid)
time.sleep(600)
```
On Linux, kill the parent process with whatever signal (SIGTERM, SIGKILL, SIGINT), the child process will not get killed. It's parent will become systemd.

ts/nni_manager/common/trial_keeper/keeper.ts
cruiseliu: Left a reminder. Will rename when add distributed trial support.

cruiseliu:
@matluster
Because my dev branch has made some progress since then, trial keeper now uses a placeholder "GPU scheduler client".
As a placeholder it is really simple inside so I decide to delay the split of dummy scheduler. The PR is already 700 lines large.

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
...
+        rejectComputeActive?: boolean;
+    }
+}
+
+export class TrialKeeper {
+    private envId: string;
+    private channels: HttpChannelServer;
+    private dirs: Map<string, string> = new Map();
+    private emitter: EventEmitter = new EventEmitter();
+    private gpuScheduler: GpuSchedulerClient;
```
matluster, line 44: I'm still feeling a little uncomfortable with `gpuScheduler`. Maybe `scheduler`?

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
...
+        this.dirs.set(name, path);
+    }
+
+    // FIXME: the method name will be changed when we support distributed trials
+    public async createTrial(options: TrialKeeper.TrialOptions): Promise<boolean> {
+        const trialId = options.id;
+
+        const gpuEnv = await this.gpuScheduler.schedule(trialId, options.gpuNumber, options.gpuRestrictions);
+        if (gpuEnv === null) {
+            this.log.info('No GPU available');
```
matluster, line 95: Maybe this should be "Scheduling failed because the required constraint ${sku} can't be satisfied."

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
...
+    private trials: Map<string, TrialProcess> = new Map();
+
+    constructor(environmentId: string, platform: string, enableGpuScheduler: boolean) {
+        this.envId = environmentId;
+        this.platform = platform;
+        this.log = getLogger(`TrialKeeper.${environmentId}`);
+
+        this.gpuScheduler = new GpuSchedulerClient(enableGpuScheduler);
+
+        this.channels = new HttpChannelServer(this.envId, `/env/${this.envId}`);
```
matluster, line 56: When the experiment keeps running for a long time, is it possible that I get a large number of registered (and useless) HTTP servers?

ts/nni_manager/common/trial_keeper/process.ts
```diff
@@ -0,0 +1,238 @@
...
+
+        this.proc = child_process.spawn(options.command, spawnOptions);
+
+        this.proc.on('spawn', () => { this.resolveStart('spawn'); });
+        this.proc.on('exit', (code, signal) => { this.resolveStop('exit', code, signal); });
+        this.proc.on('error', (err) => { this.handleError(err); });
+        this.proc.on('close', (code, signal) => { this.resolveStop('close', code, signal); });
+
+        await this.started.promise;
+        return Boolean(this.info.startSuccess);
```
matluster, line 93: Reminds me that there is no logging here in case of failure.

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
...
+import { TrialProcess, TrialProcessOptions } from './process';
+import { GpuSchedulerClient } from './gpu_scheduler_client';
+
+export declare namespace TrialKeeper {
+    export interface TrialOptions {
+        id: string;
+        command: string;
+        codeDirectoryName: string;
+        sequenceId?: number;
+        gpuNumber?: number;
```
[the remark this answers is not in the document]
liuzhe-lz: Our informal conclusion was that each training service can have its own sku definition so NNI don't need to do the translate.

ts/nni_manager/common/trial_keeper/keeper.ts
liuzhe-lz: There will also be a scheduler between NNI manager & training service, what should be its name?

ts/nni_manager/common/trial_keeper/keeper.ts
liuzhe-lz: Each trial keeper is running inside an environment (or inside a worker in distributed trial's term). In NNI manager process there will be only one trial keeper (if local training service is enabled).

ts/nni_manager/common/trial_keeper/process.ts
liuzhe-lz: It's logged when setting `startSuccess` to `false`.
At here we don't know why it fails.

ts/nni_manager/common/trial_keeper/keeper.ts
ultmaster: Maybe it's `taskScheduler`. The other one is `trialScheduler`.

As long as the interface is clear, I think name is the easy one.

ts/nni_manager/common/trial_keeper/keeper.ts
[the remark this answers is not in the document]
ultmaster: So does `trial_keeper` belong to the user part or the framework part?

ts/nni_manager/common/trial_keeper/keeper.ts
ultmaster: I'm not sure I've fully understood. We'll see when remote is ready.

ts/nni_manager/common/trial_keeper/gpu_scheduler_client.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ *  FIXME: This is a placeholder.
+ *
+ *  In near future (before 3.0 release) GPU scheduler will become a shared daemon so
```
QuanluZhang, line 7: How could the GPU scheduler be shared by multiple experiments?

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/**
+ *  The helper class to implement a "reuse" training service.
+ *
+ *  TrialKeeper has a very similar interface to TrainingSerivceV3.
+ *  In fact the local training service is a thin wrapper over TrialKeeper.
```
QuanluZhang, line 8: would be great to briefly summarize the duty of TrialKeeper here

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
...
+import { HttpChannelServer } from 'common/command_channel/http';
+import globals from 'common/globals';
+import { Logger, getLogger } from 'common/log';
+import { TrialProcess, TrialProcessOptions } from './process';
+import { GpuSchedulerClient } from './gpu_scheduler_client';
+
+export declare namespace TrialKeeper {
+    export interface TrialOptions {
+        id: string;
+        command: string;
```
QuanluZhang, line 25: any idea about distributed trial? what does `command` look like for distributed trial?

ts/nni_manager/common/trial_keeper/keeper.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    export interface GpuRestrictions {
+        onlyUseIndices?: number[];
+        rejectActive?: boolean;
+        rejectComputeActive?: boolean;
+    }
+}
+
+export class TrialKeeper {
+    private envId: string;
```
QuanluZhang, line 40: it seems TrialKeeper belongs to environment. so you have two schedulers, one is within environment, and one is out of environment?

ts/nni_manager/common/trial_keeper/keeper.ts
QuanluZhang: how about "trialScheduler" and "resourceAllocator"?

ts/nni_manager/common/trial_keeper/keeper.ts
QuanluZhang: what is this channel used for? only control plane? or also serves as data plane (i.e., metrics)?

liuzhe-lz: > what is this channel used for? only control plane? or also serves as data plane (i.e., metrics)?

Trial to manager:
- request parameter
- metrics
Manager to trial:
- parameter

QuanluZhang (approved):

ultmaster (approved):

PR #5233, opened 2022-11-18
Add legacy build for node.js v18

ultmaster:
Adding support for legacy glibc back.

It's becoming more and more difficult to find a machine to build those binaries, especially after ubuntu 18.04 are removed from pipelines in the coming December. I think we won't upgrade them for at least a few months unless necessary.

### DIFF
--- a/setup_ts.py
+++ b/setup_ts.py
@@ -76,7 +76,7 @@ def _get_node_downloader():
         _arch = platform.machine()
 
     if _arch.startswith('glibc'):
-        node_legacy_version = 'v18.12.1'  # We might not upgrade glibc version for legacy builds every time.
+        node_legacy_version = 'v18.12.1'  # We might not upgrade node version for legacy builds every time.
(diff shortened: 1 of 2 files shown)


setup_ts.py
```diff
@@ -27,13 +27,65 @@
...
+            _print('glibc version is too low. We will try to use the node version compiled with glibc 2.23, '
+                   'but it might not work.', color='yellow')
+            _print('Please check your glibc version by running `ldd --version`, and upgrade it if necessary.',
+                   color='yellow')
+            _arch = 'glibc-2.23'
+    else:
+        _arch = platform.machine()
+
+    if _arch.startswith('glibc'):
+        node_legacy_version = 'v18.12.1'  # We might not upgrade glibc version for legacy builds every time.
```
J-shang, line 79: upgrade node version?

ultmaster: Yes

J-shang (approved):

liuzhe-lz (approved):

PR #5242, opened 2022-11-23
fix ddp problem in ddp instead of compressor

Bonytu:
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### How to test ###
Use TaylorFOWeightPruner, ActivationMeanRankPruner and ActivationAPoZRankPrune to prune a ddp model to detect whether masks on different GPUs are the same after pruning.

### DIFF
--- a/nni/compression/pytorch/pruning/basic_pruner.py
+++ b/nni/compression/pytorch/pruning/basic_pruner.py
@@ -755,20 +755,11 @@ class ActivationPruner(EvaluatorBasedPruner):
                 _module = _module.module
             batch_dims, batch_num = get_output_batch_dims(output, _module)  # type: ignore
             activation = self._activation_trans(output, batch_dims)
-            batch_num = torch.tensor([batch_num]).to(activation.device)
-            if self.is_ddp_model:
-                import torch.distributed as dist
-                batch_num = batch_num.clone()
-                activation = activation.clone()
-
-                dist.all_reduce(batch_num, op=dist.ReduceOp.SUM)
-                dist.all_reduce(activation, op=dist.ReduceOp.SUM)
-
             if len(buffer) == 1:
                 buffer.append(torch.zeros_like(activation))
             if buffer[0] < self.training_steps:
                 buffer[1] += activation.to(buffer[1].device)  # type: ignore
-                buffer[0] += batch_num.item()
+                buffer[0] += batch_num
         return collect_activation
 
     def _activation_trans(self, output: Tensor, dim: int | list = 0) -> Tensor:
@@ -1032,12 +1023,6 @@ class TaylorFOWeightPruner(EvaluatorBasedPruner):
         buffer.append(0)
 
         def collect_taylor(grad: Tensor):
-            if self.is_ddp_model:
-                import torch.distributed as dist
-                grad = grad.clone()
-                size = dist.get_world_size()
-                dist.all_reduce(grad, op=dist.ReduceOp.SUM)
-                grad /= size
             if len(buffer) == 1:
                 buffer.append(torch.zeros_like(grad))
             if buffer[0] < self.training_steps:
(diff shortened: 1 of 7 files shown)


nni/compression/pytorch/pruning/basic_pruner.py
```diff
@@ -755,11 +755,20 @@ def collect_activation(_module: Module, _input: Tensor, output: Tensor):
                 _module = _module.module
             batch_dims, batch_num = get_output_batch_dims(output, _module)  # type: ignore
             activation = self._activation_trans(output, batch_dims)
+            batch_num = torch.tensor([batch_num]).to(activation.device)
+            if self.is_ddp_model:
+                import torch.distributed as dist
+                batch_num = batch_num.clone()
+                activation = activation.clone()
```
J-shang, line 762: do we need detach() here?

nni/compression/pytorch/speedup/compressor.py
```diff
@@ -553,3 +554,7 @@ def speedup_model(self):
         self.replace_compressed_modules()
         self.bound_model.train(training)
         _logger.info("speedup done")
+
+        if self.is_ddp_model:
+            self.bound_model = reset_ddp_model(self.bound_model, self.ddp_params)
```
J-shang, line 559: please add a Note to show the error met.

nni/compression/pytorch/utils/check_ddp.py
```diff
@@ -0,0 +1,36 @@
...
+        ddp_params["device_ids"] = attr_dicts.get("device_ids", None)
+        ddp_params["output_device"] = attr_dicts.get("output_device", None)
+        ddp_params["dim"] = attr_dicts.get("dim", 0)
+        ddp_params["broadcast_buffers"] = attr_dicts.get("broadcast_buffers", True)
+        ddp_params["process_group"] = attr_dicts.get("process_group", None)
+        ddp_params["bucket_cap_mb"] = attr_dicts.get("bucket_cap_mb", 25)
+        ddp_params["find_unused_parameters"] = attr_dicts.get("find_unused_parameters", False)
+        ddp_params["check_reduction"] = attr_dicts.get("check_reduction", False)
+
+        if "gradient_as_bucket_view" in attr_dicts:
```
J-shang, line 24: add a note to explain why these two attrs are special

nni/compression/pytorch/utils/evaluator.py
```diff
@@ -181,6 +182,13 @@ def unbind_model(self):
         """
         raise NotImplementedError
 
+    def rewrap_ddp_model(self, model):
```
J-shang, line 185: maybe `rewrap_if_ddp_model`?

nni/compression/pytorch/pruning/basic_pruner.py
Bonytu: I think there is no need to use .detach(), the operation here is to average the grads on multiple GPUs, without changing the graph structure of the variables.

nni/compression/pytorch/pruning/basic_pruner.py
```diff
@@ -755,11 +755,20 @@ def collect_activation(_module: Module, _input: Tensor, output: Tensor):
...
+
+                dist.all_reduce(batch_num, op=dist.ReduceOp.SUM)
+                dist.all_reduce(activation, op=dist.ReduceOp.SUM)
```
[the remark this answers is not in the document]
Bonytu: yes, I will share the time cost. I think hook operations are used for each step, so if we all_reduce this after training/inference done, then we only collect infos about the last step

nni/compression/pytorch/pruning/basic_pruner.py
J-shang: if we don't not detach `activation`, will it be recorded as a data that for backward needs and never released?

J-shang (approved):

QuanluZhang (approved):

PR #5243, opened 2022-11-23
Local Training Service V3

liuzhe-lz:
#### Test Options ####
  - [ ] fast test
  - [x] full test - HPO
  - [x] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


nni/runtime/platform/v3.py
```diff
@@ -0,0 +1,35 @@
...
+
+_channel = HttpChannel(os.environ['NNI_TRIAL_COMMAND_CHANNEL'])
+
+def get_next_parameter():
+    req = {'type': 'request_parameter'}
+    _channel.send(req)
+
+    res = _channel.receive()
+    if res is None:
+        logging.getLogger(__name__).error('Command channel is closed')
```
matluster, line 19: Print the channel url in log.

ts/nni_manager/training_service/local_v3/local.ts
```diff
@@ -0,0 +1,123 @@
...
+     **/
+    public async uploadDirectory(directoryName: string, path: string): Promise<void> {
+        this.log.info(`Register directory ${directoryName} = ${path}`);
+        this.trialKeeper.registerDirectory(directoryName, path);
+    }
+
+    public async createTrial(_envId: string, trialCommand: string, directoryName: string, sequenceId?: number):
+            Promise<string | null> {
+
+        const trialId = uuid();
```
matluster, line 52: So, a trial is launched without a parameter, and it will get a parameter when it is launched and it "requests" a parameter.
Don't you think it's a a bit complicated? Sometimes (in amulet I imagine), it would be most convenient to send the parameter at job submit. In this way, I don't have to implement a two-way communication, but only metrics collection.

ts/nni_manager/training_service/local_v3/local.ts
```diff
@@ -0,0 +1,123 @@
...
+            codeDirectoryName: directoryName,
+            sequenceId,
+            gpuNumber,
+            gpuRestrictions: {
+                onlyUseIndices: this.config.gpuIndices,
+                rejectActive: !this.config.useActiveGpu,
+            },
+        };
+
+        const success = await this.trialKeeper.createTrial(opts);
```
matluster, line 71: Do JS-guys prefer `.done().error()` or something like this (`if (success)`)?

ts/nni_manager/training_service/local_v3/local.ts
liuzhe-lz: But currently by design a trial can request multiple parameters (maybe not documented). That's why we could drop multi-phase mode.
If a trial can only have one parameter, it's a regression.
If a trial can still request more parameters, then pre-allocating parameter is merely a performance optimize.

ts/nni_manager/training_service/local_v3/local.ts
liuzhe-lz: In non-local training services this is an RPC, so exception is not a choice. (promise rejection === async exception)

ts/nni_manager/training_service/local_v3/local.ts
ultmaster: IIUC, I don't know the trial parameter here when it submits. The parameter is only provided after calling `createTrial`.
This brings a headache to create a "trial placeholder" at the beginning, who watches out for new parameters and does nothing until parameters have been received.
Maybe you believe trial keeper is a good idea to handle all these details, but IMHO, this brings an extra level of uncertainty given the chaotic situations of clusters here at Microsoft. As the cluster executes the trial keeper instead of the bare trial command, I feel rather skeptical whether the trial keeper can work transparently when the environment variables, command syntax, distributed training (e.g., MPI), signals get more complicated.

Therefore, I'm asking here is: whether it's possible to invoke `requestParameter` inside `createTrial`. In that case, the training service itself is the one who is asking for new parameters. It can directly submit a trial **with** a parameter.

ultmaster (approved):

ts/nni_manager/training_service/local_v3/local.ts
QuanluZhang: what is the conclusion here?

ts/nni_manager/training_service/local_v3/local.ts
```diff
@@ -0,0 +1,123 @@
...
+        });
+    }
+
+    public onMetric(callback: (trialId: string, metric: Metric) => Promise<void>): void {
+        this.trialKeeper.onReceiveCommand('metric', (trialId, command) => {
+            callback(trialId, (command as any)['metric']);
+        });
+    }
+
+    public onEnvironmentUpdate(_callback: (environments: EnvironmentInfo[]) => Promise<void>): void {
```
QuanluZhang, line 112: environment does not update even in hybrid mode?

QuanluZhang (approved):

ts/nni_manager/training_service/v3/compat.ts
```diff
@@ -56,21 +70,29 @@ export class V3asV1 implements TrainingService {
     public async submitTrialJob(form: TrialJobApplicationForm): Promise<TrialJobDetail> {
```
QuanluZhang, line 70: The logic of `submitTrialJob` does not keep all the features of the previous version. For example, in the previous version, `form` could be `id` in it, if there is, `trialId` should respect that `id`.

ts/nni_manager/training_service/local_v3/local.ts
```diff
@@ -0,0 +1,123 @@
...
+            return null;
+        }
+    }
+
+    public async stopTrial(trialId: string): Promise<void> {
+        this.log.info('Stop trial', trialId);
+        await this.trialKeeper.stopTrial(trialId);
+    }
+
+    public async sendParameter(trialId: string, parameter: Parameter): Promise<void> {
```
QuanluZhang, line 86: this function is deprecated, right?

ts/nni_manager/training_service/local_v3/local.ts
```diff
@@ -0,0 +1,123 @@
...
+    private log: Logger;
+    private trialKeeper: TrialKeeper;
+
+    constructor(trainingServiceId: string, config: TrainingServiceConfig) {
+        this.log = getLogger(`LocalV3.${trainingServiceId}`);
+        this.log.debug('Training sevice config:', config);
+
+        this.config = config as LocalConfig;
+        this.env = { id: `${trainingServiceId}-env` };
+        this.trialKeeper = new TrialKeeper(this.env.id, 'local', Boolean(config.trialGpuNumber));
```
QuanluZhang, line 21: `trialKeeper` is only for local training service, right? because it is in typescript.

PR #5247, opened 2022-11-25
Mutable equal, frozen context, new labels

ultmaster:
This PR:

* Add `_mutable_equal` to compare two mutables without generating MutableExpression (useful in tests).
* Add `frozen_context` that freezes mutables upon first initialization and fixed module creation (in NAS).
* Improve the implementation of `auto_label()` to make it more intuitive and idempotent.

This should be merged before #5200.


nni/mutable/frozen.py
```diff
@@ -0,0 +1,226 @@
...
+        The sample in frozen context.
+        If no sample is found, return none.
+        """
+        try:
+            ContextStack.top(_FROZEN_CONTEXT_KEY)
+            sample: Sample = {}
+            for ctx in ContextStack.stack(_FROZEN_CONTEXT_KEY):
+                if not isinstance(ctx, dict):
+                    raise TypeError(f'Expect architecture to be a dict, found: {ctx}')
+                sample.update(ctx)
```
J-shang, line 141: just curious, key in `ctx` will not conflict?

ultmaster: It will. The top of the stack will overwrite the bottom.

J-shang (approved):

liuzhe-lz (approved):

PR #5281, opened 2022-12-13
Refactor nas.nn (Stage 3) - Cell and Repeat

ultmaster:
### Description ###

Adapt Cell and Repeat to the new framework.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


nni/nas/nn/pytorch/repeat.py
```diff
@@ -61,80 +67,154 @@ class Repeat(Mutable):
...
+        return None
 
     def forward(self, x):
-        for block in self.blocks:
-            x = block(x)
+        for i, block in enumerate(self.blocks):
+            if i < self._dry_run_depth:
+                x = block(x)
+            # Make JIT happy
+            # else:
```
liuzhe-lz, line 160: ?

ultmaster: Should have an `else` here. But we can't because otherwise pytorch.jit will complain.

This is again for the (useless) graph converter.

QuanluZhang (approved):

liuzhe-lz (approved):

PR #5292, opened 2022-12-23
[Compression] v2.5 stage 1: add setting & test

J-shang:
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


nni/contrib/compression/base/setting.py
```diff
@@ -0,0 +1,222 @@
...
+            'quant_dtype': None,
+            'quant_scheme': None,
+            'granularity': 'default',
+            'apply_method': 'clamp_round',
+        }
+    }
+
+    # TODO: add more default module type
+    registry = {
+        'Linear': default_setting,
```
Bonytu, line 186: Is it necessary to provide an interface for the user to register the settings of the submodel? For example, the user want to prune multi-head attention layer,  there is no interface to register the settings of the multi-head attention layer  in this pipeline.

J-shang: the interface is at line 26

Bonytu (approved):

nni/contrib/compression/utils/evaluator.py
```diff
@@ -723,17 +734,28 @@ def unbind_model(self):
...
-
-        def patched_criterion(*args, **kwargs):
-            loss = old_criterion(*args, **kwargs)
-            return patch(loss)
-
-        self._criterion = patched_criterion
+    def patch_loss(self, patch: Callable[[Tensor, Any], Tensor]):
+        old_training_step = self._training_step
+
+        def patched_training_step(*args, **kwargs):
```
matluster, line 740: I'm not sure whether this works, because Pytorch-lightning sometimes use inspect.parameters to inspect the arguments of `training_step`. If the signature has `batch_idx`, it will give you `batch_idx`. If the signature doesn't, it won't.

ultmaster (approved):

nni/contrib/compression/utils/evaluator.py
J-shang: could you give a link for this logic, I didn't find that

PR #5293, opened 2022-12-23
[Compression] v2.5 stage 2: add target space and config

J-shang:
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


nni/contrib/compression/base/config.py
```diff
@@ -0,0 +1,157 @@
...
+                target_names.append(OUTPUT_PREFIX)
+                quant_bit = quant_bits if isinstance(quant_bits, int) else quant_bits['output']
+                target_settings[OUTPUT_PREFIX] = {'quant_dtype': f'int{quant_bit}'}
+            config['target_names'] = target_names
+            config['target_settings'] = target_settings
+    return config_list
+
+
+def select_modules_by_config(model: torch.nn.Module, config: Dict[str, Any]) -> Tuple[Dict[str, torch.nn.Module], Dict[str, Any]]:
+    """
```
matluster, line 105: Better to add more explanation on how this "selection" works.

nni/contrib/compression/base/config.py
```diff
@@ -0,0 +1,157 @@
...
+    Parameters
+    ----------
+    model
+        The modules selected from.
+    config
+        The config contains keys ['op_names', 'op_types', 'op_names_re', 'exclude_op_names', 'exclude_op_types', 'exclude_op_names_re']
```
matluster, line 111: What's the meaning of each key?

nni/contrib/compression/base/target_space.py
```diff
@@ -0,0 +1,372 @@
...
+from nni.compression.pytorch.utils.scaling import Scaling
+
+
+class TargetType(Enum):
+    INPUT = 'input'
+    OUTPUT = 'output'
+    PARAMETER = 'parameter'
+
+
+class TargetSpace:
```
matluster, line 23: What's the purpose of a TargetSpace?

nni/contrib/compression/base/config.py
```diff
@@ -0,0 +1,157 @@
...
+        target_names = []
+        target_settings = {}
+        if 'quant_types' in config:
+            quant_types, quant_bits = config.pop('quant_types'), config.pop('quant_bits')
+            if 'input' in quant_types:
+                target_names.append(INPUT_PREFIX)
+                quant_bit = quant_bits if isinstance(quant_bits, int) else quant_bits['input']
+                target_settings[INPUT_PREFIX] = {'quant_dtype': f'int{quant_bit}'}
+            if 'weight' in quant_types:
+                target_names.extend(['weight', 'bias'])
```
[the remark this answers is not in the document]
J-shang: remove 'bias'

nni/contrib/compression/base/config.py
J-shang: add comments

```
    This is a helper function for selecting the modules in model specified in config.

    There are six optional keys in config to specify the module,
    note that the module name should be the same as the name obtained from ``model.named_modules()``,
    module type is the ``__name__`` of the module class:
        - ``op_names``: a module name list, the modules with these names will be selected.
        - ``op_types``: a module type name list, the modules satisfied these types will be selected.
        - ``op_names_re``: a regular expression list, the modules satisfied the regular expressions will be selected.
        - ``exclude_op_names``: a module name list, the modules with these names will be excluded.
        - ``exclude_op_types``: a module type name list, the modules satisfied these types will be excluded.
        - ``exclude_op_names_re``: a regular expression list, the modules satisfied the regular expressions will be excluded.

    A module is selected if it satisfies all the following conditions:
        1. If ``op_names`` or ``op_names_re`` is not empty, the module name should in ``op_names``
           or satisfied one of regular expressions in ``op_names_re``.
        2. If ``op_types`` is not empty, the module type name should in ``op_types``.
        3. If ``exclude_op_names`` or ``exclude_op_names_re`` is not empty, the module name should not in ``exclude_op_names``
           and should not satisfied each of regular expressions in ``exclude_op_names_re``.
        4. If ``exclude_op_types`` is not empty, the module type name should not in ``exclude_op_types``.

```

nni/contrib/compression/base/config.py
J-shang: add comments

nni/contrib/compression/base/target_space.py
J-shang: add class doc str
```
    """
    TargetSpace is a container to handle all compression information related to a specific compression target,
    note that it has the ability to register information in the wrapper,
    this because in most comopression cases need to register tensor (mask, scale, zero point) as buffer/parameter in the model.

    Parameters
    ----------
    wrapper
        The wrapper of a module.
    target_name
        If the target is a parameter, the name is the attr name in the module, i.e., 'weight', 'bias'.
        If the target is the module input/output, the name should have prefix '_input_'/'_output_',
        then follow a number or name, i.e., '_input_0', '_output_hidden_state',
        number is the position of this input/output in all inputs/outputs,
        name is the input argument name or outputs dict key.
    target_type
        'input' or 'output' or 'parameter'.
    setting
        The compression setting, i.e., {'sparse_ratio': 0.1}.
    """
```

nni/contrib/compression/base/target_space.py
```diff
@@ -0,0 +1,393 @@
...
+    """
+    TargetSpace is a container to handle all compression information related to a specific compression target,
+    note that it has the ability to register information in the wrapper,
+    this because in most comopression cases need to register tensor (mask, scale, zero point) as buffer/parameter in the model.
```
ultmaster, line 27: compression

ultmaster (approved):

Bonytu (approved):

PR #5315, opened 2023-01-11
fix(speedup): fix the error on jit translation of 'aten::to.dtype_layout' in pytorch <= 1.10

Louis-J:
### Description ###

the jit translation of 'aten::to.dtype_layout' in pytorch 1.10 will cause some error.
fix it by adding a check.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [x] full test - compression

### Checklist ###
  - [x] test case
  - [ ] doc

### DIFF
--- a/nni/compression/pytorch/speedup/jit_translate.py
+++ b/nni/compression/pytorch/speedup/jit_translate.py
@@ -767,7 +767,7 @@ def hook_to_dtype_layout(positional, keyword, undetermined, undetermined_special
 
     def ret_func(*args):
         the_self = args[0]
-        if the_self.layout != to_layout
+        if the_self.layout != to_layout:
             the_self = layout_names_to_trans_dict[to_layout](the_self)
         return real_to(the_self, *args[1:])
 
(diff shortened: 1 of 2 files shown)


nni/compression/pytorch/speedup/jit_translate.py
```diff
@@ -763,7 +767,8 @@ def hook_to_dtype_layout(positional, keyword, undetermined, undetermined_special
 
     def ret_func(*args):
         the_self = args[0]
-        the_self = layout_names_to_trans_dict[to_layout](the_self)
+        if the_self.layout != to_layout:
```
J-shang, line 770: is this because pytorch 1.10 do not have `to_dense`? what about pytorch 1.9? better to add a comment here.

J-shang (approved):

nni/compression/pytorch/speedup/jit_translate.py
Louis-J: it's not support to do transform from torch.strided to torch.strided by to_dense on cuda backend. actually it's not necessary to do transform when source equals to aim.

Louis-J: and it's support to do transform from torch.sparse_coo to torch.strided by to_dense on both cpu or cuda backend.

PR #5336, opened 2023-02-05
Mutable shortcut and enhancements

ultmaster:
This PR brings the following enhancements to mutable APIs.

- Shortcuts to create mutables via functions like `nni.choice`, `nni.uniform` and `nni.randint`.
- Add a strict flag to control whether to raise error when context is missing in `ensure_frozen`.
- Add `case` and `switch_case` to MutableExpression, in preparation for hardware-aware NAS.
- `as_legacy_dict` to convert the new mutables into legacy search space JSON.
- Improved representation string for mutables.
- Fix the case of numpy.array in mutable_equal.
- Miscellenous bug fixes and test enhancements.

It should be merged prior to #5279.

### DIFF
--- a/nni/mutable/mutable.py
+++ b/nni/mutable/mutable.py
@@ -660,8 +660,8 @@ class Categorical(MutableSymbol, Generic[Choice]):
     ) -> None:
         values = list(values)
         assert values, 'Categorical values must not be empty.'
-        self.label = auto_label(label)
-        self.values = values
+        self.label: str = auto_label(label)
+        self.values: list[Choice] = values
         self.weights = weights if weights is not None else [1 / len(values)] * len(values)
 
         if default is not MISSING:
--- a/nni/mutable/mutable.py
+++ b/nni/mutable/mutable.py
@@ -101,6 +101,11 @@ def _mutable_equal(mutable1: Any, mutable2: Any) -> bool:
             else:
                 return False
         return True
+    if isinstance(mutable1, np.ndarray):
+        if not isinstance(mutable2, np.ndarray):
+            return False
+        return np.array_equal(mutable1, mutable2)
+
     return mutable1 == mutable2
 
 
@@ -342,6 +347,14 @@ class Mutable:
         """Return a string representation of the extra information."""
         return ''
 
+    def as_legacy_dict(self) -> dict:
+        """Convert the mutable into the legacy dict representation.
+
+        For example, ``{"_type": "choice", "_value": [1, 2, 3]}`` is the legacy dict representation of
+        ``nni.mutable.Categorical([1, 2, 3])``.
+        """
+        raise NotImplementedError(f'as_legacy_dict is not implemented for this type of mutable: {type(self)}.')
+
     def equals(self, other: Any) -> bool:
         """Compare two mutables.
 
@@ -482,7 +495,7 @@ class Mutable:
         # Used in ``nni.trace``.
         # Calling ``ensure_frozen()`` by default.
         from .frozen import ensure_frozen
-        return ensure_frozen(self)
+        return ensure_frozen(self, strict=False)
 
 
 class LabeledMutable(Mutable):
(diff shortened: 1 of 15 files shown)


nni/mutable/mutable.py
```diff
@@ -678,6 +699,12 @@ def freeze(self, sample: Sample) -> Any:
     def __len__(self):
         return len(self.values)
 
+    def as_legacy_dict(self) -> dict:
+        return {
+            '_type': 'choice',
+            '_value': self.values,
```
liuzhe-lz, line 705: Suggest to add type hint for fields. It took me some time to find out `self.values`' type.

nni/mutable/shortcut.py
```diff
@@ -0,0 +1,235 @@
...
+    #     # In case PyTorch is not installed.
+    #     pass
+
+    return Categorical(choices, label=label)
+
+
+def randint(label: str, lower: int, upper: int) -> Categorical[int]:
+    """Choosing a random integer between lower (inclusive) and upper (exclusive).
+
+    Currently it is translated to a :func:`choice`.
```
liuzhe-lz, line 74: I want to make it default to quniform. This shortcut is not used in NAS I guess?
To achieve this the mutable need to aware that the result should be converted to int.

nni/mutable/shortcut.py
```diff
@@ -0,0 +1,235 @@
...
+    --------
+    >>> nni.qnormal('x', 0., 1., 0.1)
+    Numerical(-inf, inf, mu=0.0, sigma=1.0, q=0.1, label='x')
+    >>> nni.qnormal('x', 0., 1., 0.1).random()
+    -0.1
+    """
+    return Numerical(mu=mu, sigma=sigma, quantize=quantize, label=label)
+
+
+def lognormal(label: str, mu: float, sigma: float) -> Numerical:
```
liuzhe-lz, line 205: I suggest to remove `lognormal` and `qlognormal`. They are misleading and really unnecessary.

nni/mutable/symbol.py
```diff
@@ -228,6 +229,76 @@ def condition(pred: Any, true: Any, false: Any) -> SymbolicExpression | Any:
...
+        .. code-block:: cpp
+
+            switch (branch) {  // c-style switch
+                case 0: return 17;
+                case 1: return 31;
+            }
+
+        Equivalent to::
+
+            SymbolicExpression.switch_case(branch, {0: 17, 1: 31})
```
liuzhe-lz, line 279: So fancy...

nni/mutable/mutable.py
ultmaster: Are you suggesting `TypedDict`?

I don't want to waste the effort because it's **legacy**_dict.

nni/mutable/shortcut.py
ultmaster: It's not straightforward for `nni.randint().random()` to return a float, rather than an integer.

Also see quniform: `nni.quniform(low - 0.5, high + 0.5, 1).int()`. I think it's somewhat complicated and requires `simplify()` to get the variable.

nni/mutable/shortcut.py
ultmaster: Agreed. Let's make sure @QuanluZhang agrees with us.

nni/mutable/symbol.py
ultmaster: I think it's actually simpler than nested `if-else`.

nni/mutable/shortcut.py
ultmaster: And it's currently never used in NAS.

nni/mutable/mutable.py
liuzhe-lz: I mean `self.values: list[Any] = values`

nni/mutable/symbol.py
liuzhe-lz: I mean, `case` + `switch_case` is a little bit too fancy to me.
I have just checked the Pythonic way, `match` statement, but unluckily it's even more fancy...  :(

ultmaster: One of them is needed in symbolic shape inference. The other one is used in nn-meter profiler.

Tensorflow have both [switch_case](https://www.tensorflow.org/api_docs/python/tf/switch_case) and [case](https://www.tensorflow.org/api_docs/python/tf/case).

nni/mutable/mutable.py
ultmaster: My IDE inferred these types automatically. But I can surely add them on.

nni/mutable/shortcut.py
QuanluZhang: we can follow our discussion.

QuanluZhang (approved):

ultmaster:
Moved several debatable items to `_notimplemented.py`.

liuzhe-lz (approved):

Issue #5338, opened 2023-02-07
Add a parameter in generate_scenario for the file 'scenario.txt' instead of harcoding


miguelgfierro:
**What would you like to be added**:
It would be great in the function https://github.com/microsoft/nni/blob/e85f029bd4e4b1bdf3e679893fb6447e4d6b2c79/nni/algorithms/hpo/smac_tuner/convert_ss_to_scenario.py#L192 to add a parameter for `scenario.txt` instead of hardcoding it.

**Why is this needed**:
I'm using SMAC tuner in a Databricks compute where I can't write to `.`, I can only write to `/tmp/`. If I could parametrize the file, then I would be able to use SMAC.

**Without this feature, how does current nni work**：
SMAC won't work in Databricks and probably on Synapse (I haven't tried, but it is my guess)

**Components that may involve changes**:

**Brief description of your proposal if any**:
Difficult fix:
Change the signature of [generate_scenario](https://github.com/microsoft/nni/blob/e85f029bd4e4b1bdf3e679893fb6447e4d6b2c79/nni/algorithms/hpo/smac_tuner/convert_ss_to_scenario.py#L114), and then also the SMAC arguments.

Easy fix:
Instead of hardcoding `scenario.txt` hardcode `/tmp/scenario.txt`. I'll send this PR in case you want it

PR #5339, opened 2023-02-07
Fix to #5338 in generate_scenario function for SMAC tunner

miguelgfierro:
### Description ###
This is a simple fix for #5338

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc

### DIFF
--- a/nni/algorithms/hpo/smac_tuner/convert_ss_to_scenario.py
+++ b/nni/algorithms/hpo/smac_tuner/convert_ss_to_scenario.py
@@ -189,7 +189,7 @@ def generate_scenario(ss_content):
         The scenario-object (smac.scenario.scenario.Scenario) is used to configure SMAC and can be constructed
         either by providing an actual scenario-object, or by specifing the options in a scenario file
     """
-    with open('scenario.txt', 'w') as sce_fd:
+    with open('/tmp/scenario.txt', 'w') as sce_fd:
         sce_fd.write('deterministic = 0\n')
         # sce_fd.write('output_dir = \n')
         sce_fd.write('paramfile = param_config_space.pcs\n')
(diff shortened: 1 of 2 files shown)


liuzhe-lz (approved):

nni/algorithms/hpo/smac_tuner/convert_ss_to_scenario.py
```diff
@@ -189,7 +189,7 @@ def generate_scenario(ss_content):
         The scenario-object (smac.scenario.scenario.Scenario) is used to configure SMAC and can be constructed
         either by providing an actual scenario-object, or by specifing the options in a scenario file
     """
-    with open('scenario.txt', 'w') as sce_fd:
+    with open('/tmp/scenario.txt', 'w') as sce_fd:
```
QuanluZhang, line 192: This change leads to the situation that smac cannot find 'scenario.txt', because for simplicity nni uses default parameters of the original smac which reads this file from the working directory. You can refer to this line for details: https://github.com/QuanluZhang/SMAC3/blob/master/smac/utils/io/cmd_reader.py#L36

miguelgfierro: yes, I thought that this would be an easy fix, but it seems it involves a good amount of refactoring. Is there interest in fixing this? if yes, it would be interesting to know the steps so I can see if I have bandwidth. Otherwise, happy to close the PR

QuanluZhang: Thanks for your contribution. The way is updating the package smac4nni to change the path accordingly. But this may induce version mismatch between smac4nni and nni. Thus, it would be better to keep it as is for now.

PR #5367, opened 2023-02-16
NAS strategy (stage 2) - utils and bruteforce strategy

ultmaster:


nni/nas/strategy/bruteforce.py
```diff
@@ -1,138 +1,232 @@
...
-class GridSearch(BaseStrategy):
+class GridSearch(Strategy):
     """
     Traverse the search space and try all the possible combinations one by one.
 
     Parameters
     ----------
-    shuffle : bool
-        Shuffle the order in a candidate list, so that they are tried in a random order. Default: true.
+    shuffle
```
liuzhe-lz, line 30: Does it present in v2.10? Seems it's incompatible with HPO grid search's design

nni/nas/strategy/bruteforce.py
```diff
@@ -1,138 +1,232 @@
...
+        Shuffle the order in a candidate list, so that they are tried in a random order.
+        Currently, the implementation is a pseudo-random shuffle, which only shuffles the order of every 100 candidates.
+    dedup
```
liuzhe-lz, line 33: In what case one should turn off dedup?

ultmaster: For example when deduplication causes performance issues?

It won't happen in the wild I guess, but might happen in some benchmarks.

nni/nas/strategy/bruteforce.py
ultmaster: Yes.

I think shuffled-grid-search is a nice feature and I prefer enabling it by default.

QuanluZhang (approved):

liuzhe-lz (approved):

PR #5372, opened 2023-02-16
NAS oneshot (stage 2) - Supernet modules

ultmaster:


QuanluZhang (approved):
it would be better to provide simple tutorials about how to use these operators, sampling, and differentiables.

ultmaster:
> it would be better to provide simple tutorials about how to use these operators, sampling, and differentiables.

Sure. Will do.

nni/nas/oneshot/pytorch/supermodule/sampling.py
```diff
@@ -420,10 +376,16 @@ def mutate(cls, module, name, memo, mutate_kwargs):
...
-                op_factory = {  # create a factory
-                    name: lambda _, __, ___: copy.deepcopy(op_candidates_lc[name])
-                    for name in op_candidates_lc.names
-                }
+                candidates = op_candidates_lc.candidates
+                def _copy(_, __, ___, op):
+                    return copy.deepcopy(op)
+
+                if isinstance(candidates, list):
+                    op_factory = [functools.partial(_copy, op=op) for op in candidates]
```
liuzhe-lz, line 384: Is this equal to `lambda *args: copy.deepcopy(op)`?
The mental overhead is massive...

liuzhe-lz: Oh `op` is the iterator so you need to shallow copy `op` each time before constructing the lambda...

liuzhe-lz (approved):

nni/nas/oneshot/pytorch/supermodule/sampling.py
matluster: I suppose there is a better way, but I don't think it's worth the effort.

PR #5378, opened 2023-02-16
NAS experiment (stage 3) - main body

ultmaster:


nni/nas/experiment/experiment.py
```diff
@@ -0,0 +1,462 @@
...
+        """Stop a NAS experiment.
+        """
+        _logger.info('Stopping experiment, please wait...')
+        atexit.unregister(self.stop)
+
+        self.save_checkpoint()
+
+        if self._nni_manager_required():
+            _logger.debug('Stopping NNI manager...')
+            self._stop_nni_manager()
```
liuzhe-lz, line 281: I have just found out the real reason that it returns before actually stop.
The stopping command is an HTTP request, and HTTP requests must be responded in several seconds before it times out.
So the only solution is to add a loop waiting for subprocess to exit.

nni/nas/experiment/experiment.py
```diff
@@ -0,0 +1,462 @@
...
+        """Save the whole experiment state.
+
+        It will dump the config first (as a JSON) and then states of components like strategy and engine.
+        It calls :meth:`state_dict` to get the states.
+        """
+        try:
+            if not self._checkpoint_directory.exists():
+                self._checkpoint_directory.mkdir(parents=True, exist_ok=True)
+
+            with (self._checkpoint_directory / 'config.json').open('w') as f:
```
liuzhe-lz, line 421: encoding=utf8

liuzhe-lz (approved):

nni/nas/experiment/experiment.py
```diff
@@ -0,0 +1,462 @@
...
+            _logger.debug('Stopping NNI manager...')
+            self._stop_nni_manager()
+            # NNI manager should be stopped before engine.
+            # Training service engine need the terminate signal so that the listener thread can be stopped.
+        else:
+            _logger.debug('Updating experiment manifest...')
+            self._stop_without_nni_manager()
+
+        # NOTE: Engine is designed to be disposable.
+        # It should never restart because one experiment can't run twice.
```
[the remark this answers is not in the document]
matluster: Both strategy and engine can't run twice, for simplicity. To "resume" a strategy / engine, create a new one and load the old one's state_dict.

J-shang (approved):

PR #5414, opened 2023-03-01
Kwargs shape prop

junesnow0806:
### Description ###
Add overrided interpreter and shape_prop to support Shape Inference when the dummy input for a model forward is in the keyword arguments form. And use a bert-base-cased model from huggingface as an example to show how to use the new shape_prop.


nni/common/concrete_trace_utils/passes/kwargs_shape_prop.py
```diff
@@ -0,0 +1,93 @@
+import torch
+import traceback
+from torch.fx.node import Node, map_aggregate
+from typing import Optional, Union, NamedTuple, Tuple, Any, Dict
+from ..kwargs_interpreter import KwargsInterpreter
+
+class TensorMetadata(NamedTuple):
```
super-dainiu, line 7: Can reuse the original one from shape prop actually. I believe they are the same.

junesnow0806: Sure. Updated in the latest commit.

super-dainiu (approved):

PR #5460, opened 2023-03-20
Count module prams utils

junesnow0806:
### Description ###
Based on KwargsInterpreter, this branch implements a fx graph pass utils for counting flops and module param size.


J-shang:
close this pr for another pr have added this feature

PR #5462, opened 2023-03-20
[Common] let **kwargs be possible during intermediate tracing

super-dainiu:
### Description ###
cc @QuanluZhang can try if this helps.
```python
from nni.common.concrete_trace_utils import concrete_trace
import torch

class KwargsModel(torch.nn.Module):
    def __init__(self, model):
        super().__init__()
        self.model = model
    
    def forward(self, kwargs):
        return self.model(**kwargs)
    
class Model(torch.nn.Module):
    def forward(self, x, y):
        return x + y
    
model = KwargsModel(Model())
kwargs = {'x': torch.tensor(1), 'y': torch.tensor(2)}
concrete_trace(model, (kwargs, ))
```
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


nni/common/concrete_trace_utils/concrete_proxy.py
```diff
@@ -58,6 +58,7 @@ class ConcreteProxy(Proxy):
     op_call_ex = dis.opmap['CALL_FUNCTION_EX']
     op_not = dis.opmap['UNARY_NOT']
     op_unpack_sequence = dis.opmap['UNPACK_SEQUENCE']
+    op_dict_merge = dis.opmap['DICT_MERGE']
```
super-dainiu, line 61: dont know why **kwargs call DICT_MERGE first before UNPACK_SEQUENCE...
but this hack should work for simple cases

super-dainiu: likewise, if i call `dict(**kwargs)``, python calls DICT_MERGE first also

J-shang (approved):

PR #5481, opened 2023-03-27
cpu/gpu mixed trace

junesnow0806:
### Description ###
Modified nni.common.concrete_trace_utils.concrete_tracer.py to support cpu/gpu-mixed trace.
Model and input building is on cpu. `call_function` and `call_module` nodes will be moved to gpu when call `run_target` and moved back to cpu after running.


super-dainiu (approved):

nni/common/concrete_trace_utils/concrete_tracer.py
```diff
@@ -1461,10 +1512,50 @@ def f(x, y):
         else:
             assert node_a.op == node_b.op and target_a == target_b
 
+    # apply dead code elimination according to the switch param `dce`
+    # note that we perform dce before instantiating GraphModule
+    # since there maybe some problems if we use GraphModule.recompile()
+    dont_delete = [operator.setitem, builtins.next]
+    if dce:
```
super-dainiu, line 1519: is this the same as
https://github.com/pytorch/pytorch/blob/e9a7115605ee5b6ae38ea5716a4abea5aa415333/torch/fx/graph.py#L1375

J-shang:
close this pr for another pr have added this feature.

PR #5492, opened 2023-03-29
[Compression] Add support for torch 2.0

Bonytu:
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


examples/tutorials/pruning_bert_glue.py
```diff
@@ -159,11 +159,21 @@ def preprocess_function(examples):
...
+else:
+    from torch.optim.lr_scheduler import _LRScheduler
+    SCHEDULER = _LRScheduler
+
 
 def training(model: torch.nn.Module,
              optimizer: torch.optim.Optimizer,
              criterion: Callable[[torch.Tensor, torch.Tensor], torch.Tensor],
-             lr_scheduler: torch.optim.lr_scheduler._LRScheduler = None,
+             lr_scheduler: SCHEDULER = None, # type: ignore
```
J-shang, line 176: after modifying the tutorial, re-build doc is needed (`make en`).
and before `# type: ignore` prefer double space.

examples/tutorials/pruning_bert_glue.py
```diff
@@ -159,11 +159,21 @@ def preprocess_function(examples):
...
 
+from nni.common.version import torch_version_is_2
+
+
+if torch_version_is_2():
+    from torch.optim.lr_scheduler import LRScheduler
+    SCHEDULER = LRScheduler
+else:
+    from torch.optim.lr_scheduler import _LRScheduler
+    SCHEDULER = _LRScheduler
```
J-shang, line 170: what about put this `if ... else ...` in some where and we just `from xxx import SCHEDULER` here, then we don't need to copy paste it everywhere.

examples/tutorials/pruning_bert_glue.py
J-shang: and why need type ignore here?

Bonytu: good suggestion, I will modify it.

nni/contrib/compression/utils/evaluator.py
```diff
@@ -656,7 +657,7 @@ def get_dummy_input(self) -> Any:
 
 _OPTIMIZERS = Union[Optimizer, List[Optimizer]]
 _TRAINING_STEP = Callable[..., Union[Tensor, Tuple[Tensor], Dict[str, Tensor]]]
-_SCHEDULERS = Union[None, _LRScheduler, List[_LRScheduler]]
+_SCHEDULERS = Union[None, SCHEDULER, List[SCHEDULER]]  # type: ignore
```
J-shang, line 660: what's the type error?

J-shang (approved):

QuanluZhang (approved):

PR #5513, opened 2023-04-10
[Testing] limit mmdet to 2.x

super-dainiu:
### Description ###
Temporarily, we shall refine mmdet to 2.x before we need extra support.

#### Test Options ####
  - [x] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [x] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


super-dainiu:
Errors happen in the azure pipeline only. I cannot reproduce these errors locally...

super-dainiu:
no need for this

PR #5564, opened 2023-05-15
[Compression] avoid duplicated replacement for a single target submodule

super-dainiu:
### Description ###
In model speedup, we might replace one module twice.

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


J-shang:
please add the logic of comprehensively considering all masks of the same target of different nodes in module replacement.

super-dainiu:
> please add the logic of comprehensively considering all masks of the same target of different nodes in module replacement.

I think this will be solved if we add another dependency check for this.

J-shang (approved):

Bonytu (approved):

PR #5569, opened 2023-05-18
[compression] pruning dependency v2

super-dainiu:
#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


J-shang:
please design a representative model (should have all dependencies we try to deal with) and test dependency on it, add this into unit test.

super-dainiu:
I see

nni/compression/pytorch/speedup/v2/dependency.py
```diff
@@ -0,0 +1,444 @@
...
+    for key, value in dependency.items():
+        if key not in visited:
+            tmp = set([key]) | value
+            visited.update(tmp)
+            if len(tmp) > 1:
+                d_sets.append(tmp)
+    return d_sets
+
+
+def build_weight_sharing_dependency(graph_module: torch.fx.GraphModule) -> List[List[torch.fx.Node]]:
```
zheng-ningxin, line 230: Could you please explain more about the weight-sharing dependency? What scenarios is it for? why it has such dependencies?

nni/compression/pytorch/speedup/v2/dependency.py
```diff
@@ -0,0 +1,444 @@
...
+            return meta[0][0]
+        return None
+
+    in_shape = get_shape(find_parent(node))
+    out_shape = get_shape(node)
+    if not in_shape or not out_shape or len(in_shape) <= 1 or len(out_shape) <= 1:
+        return True
+    in_channel = in_shape[1]
+    out_channel = out_shape[1]
+    return in_channel != out_channel
```
J-shang, line 105: Is it possible that the number of channels is the same but it is a break operation?

nni/compression/pytorch/speedup/v2/dependency.py
```diff
@@ -0,0 +1,444 @@
...
+
+            elif node.op == 'call_function':
+                if node.target in CALL_FUNCTION_REDUCE:
+                        # refer issue 4540 for more details. Multiplication actually
+                        # will not introduce the channel dependency, cause the misaligned
+                        # channels can propagate to each other. However, when one of the input
+                        # tensor is from skip connection(residual), the channel propagation
+                        # may be failed(the input is also used by another layer and cannot be
+                        # pruned), in this case, we need to fix the conflict maunally.
+                        d_set = set(find_adjacent_layers(node, graph_module, target_types, 'parent'))
```
J-shang, line 331: indentation

nni/compression/pytorch/speedup/v2/dependency.py
super-dainiu: this sounds like
```python
class Example(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.conv = nn.Conv2d(3, 3, 3, padding=1)
        self.conv2 = nn.Conv2d(3, 3, 3, padding=1)
        
    def forward(self, x):
        x = self.conv(x) + x
        x = self.conv2(x) + x
        x = self.conv(x) + x
        return x
```
where `self.conv` is repeatedly used in forward. in `torch.fx` we might prune ``self.conv`` twice.

nni/compression/pytorch/speedup/v2/dependency.py
super-dainiu: it's really hard for me to find an example. maybe I add a TODO here?

J-shang (approved):

zheng-ningxin (approved):
LGTM

PR #5575, opened 2023-05-23
[bug fix] fix prefix experiment nav highlight issue

Lijiaoa:
### Description ###

start an experiment by using v3.0rc `nnictl create --config examples/trials/mnist-tfv2/config.yml --url_prefix xxx`, and will find that nav bar had the wrong highlight with different page.

**related component**
image

**PR merged**
prefix exp had the correct actions with nav bar

#### Test Options ####
no need

### Checklist ###
 no need

### How to test ###
run experiment


ultmaster (approved):

liuzhe-lz:
I think there should be a global util for URL without prefix.

liuzhe-lz (approved):

PR #5576, opened 2023-05-23
Update pyrightconfig.json

Jvr2022:
Changes made:

Removed the "reportMissingImports" and "reportPrivateImportUsage" settings since they were set to false by default, and explicitly setting them is unnecessary.

Added two plugins: "mypy.plugins.django_plugin" and "mypy.plugins.sqlalchemy_plugin". These plugins extend mypy's capabilities to work with Django and SQLAlchemy code, respectively. You can add more plugins if needed.

Added the "mypy_path" setting to specify the root directory for the code. Change "src" to the actual root directory if it's different.

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,24 +1,28 @@
 {
-    "ignore": [
-        "nni/algorithms/compression/pytorch",
-        "nni/algorithms/compression/tensorflow",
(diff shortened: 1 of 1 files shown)


matluster:
The real problem is that NNI is a large code base and enabling those rules brings us effort to make existing code comply with the new rules. If you feel happy about it, go ahead to modify those code, and we will review. Otherwise, enabling new rules is pointless.

PR #5598, opened 2023-06-06
fix tensorboard bug

yjjinjie:
### Description ###
when we use the tensorboard,we find this tensorboard bug
https://github.com/tensorflow/tensorboard/issues/1713

#### Test Options ####
  - [ ] fast test
  - [ ] full test - HPO
  - [ ] full test - NAS
  - [ ] full test - compression

### Checklist ###
  - [ ] test case
  - [ ] doc


ultmaster:
I've looked at the original issue, and the workaround looks reasonable. However, a better idea might be to support customized tensorboard command.

ultmaster (approved):

liuzhe-lz (approved):

yjjinjie:
@microsoft-github-policy-service agree

PR #5606, opened 2023-06-11
adapt pytorch lighting 2.0 AKA lightning

qmpzzpmq:
adapt pytorch lightning 2.0 since it already rename as lightning


qmpzzpmq:
@microsoft-github-policy-service agree

nni/compression/pytorch/utils/evaluator.py
```diff
@@ -21,6 +21,13 @@
 else:
     LIGHTNING_INSTALLED = True
 
+try:
+    import lightning as L
+except ImportError:
+    LIGHTNING2_INSTALLED = False
+else:
+    LIGHTNING2_INSTALLED = True
+
```
J-shang, line 30: Suggest merge these two lightning, then you don't need to modify too much code:
```
try:
    import lightning as pl
    from lightning.callbacks import Callback
except ImportError:
    try:
        import pytorch_lightning as pl
        from pytorch_lightning.callbacks import Callback
    except ImportError:
        LIGHTNING_INSTALLED = False
    else:
        LIGHTNING_INSTALLED = True
else:
    LIGHTNING_INSTALLED = True
```

Bonytu (approved):

J-shang:
hello @qmpzzpmq , please try to fix the lint error. You could using `python -m pyright nni` to reproduce the error.

image

qmpzzpmq:
@J-shang I tried to fix it, but unfortunately I failed on install node.js during `python setup.py develop`. FYI, I am in China.

qmpzzpmq:
@J-shang Hi, I run the command as you mentioned. but it looks all error is not about my PR. Please check

PR #5780, opened 2024-05-08
Training Services all URL fixed

Onkarpatil7:
The URL's of mentioned training services in support section and Compression section are fixed which were showing **error 404 not found**. From this reference  
https://nni.readthedocs.io/en/latest/experiment/training_service/overview.html


Onkarpatil7:
Why Training services files are in _removed folder?

